PR #5, opened 2024-04-25
Use Rye instead of Poetry

Kludex:


.github/workflows/main.yml
```diff
@@ -16,9 +16,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.8'
-      - run: |
-            pip install poetry
-            poetry install -E test
-      - uses: pre-commit/action@v3.0.1
-        with:
-          extra_args: --all-files --verbose
+      - uses: eifinger/setup-rye@v2
```
samuelcolvin, line 19: Surely simplest to just run pre-commit in CI like other projects.

pyproject.toml
```diff
@@ -118,21 +85,23 @@ test = [
...
 ]
 
-[tool.poetry.scripts]
-logfire = "logfire.cli:main"
+[tool.rye.scripts]
+type-check = "pyright"
 
-[tool.poetry.plugins."pydantic"]
-logfire-plugin = "logfire.integrations.pydantic:plugin"
+[build-system]
```
samuelcolvin: I think this should go right at the top of the file.

pyproject.toml
```diff
@@ -1,89 +1,56 @@
...
-starlette = ["opentelemetry-instrumentation-starlette"]
-sqlalchemy = ["opentelemetry-instrumentation-sqlalchemy"]
-psycopg2 = ["opentelemetry-instrumentation-psycopg2"]
-pymongo = ["opentelemetry-instrumentation-pymongo"]
-redis = ["opentelemetry-instrumentation-redis"]
-requests = ["opentelemetry-instrumentation-requests"]
-test = [
+[tool.rye]
+managed = true
+dev-dependencies = [
```
samuelcolvin, line 57: I couldn't work out how to have multiple dev depenency groups, e.g. for docs, but I guess if install is fast enough, it shouldn't matter.

.github/workflows/main.yml
Kludex: It was not simple, I'll add pre-commit here later.

PR #12, opened 2024-04-26
improve readme and related faff

samuelcolvin:
I couldn't get rye to respect `UV_EXTRA_INDEX_URL` despite numerous different attempts, so I ended up using `uv` directly for docs build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
       - name: Ensure requests to production domains fail


Makefile
Kludex, line 1: You can configure rye scripts for it...

Kludex: I mean that a Makefile is not needed.

.github/workflows/main.yml
```diff
@@ -18,13 +18,23 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
+          python-version: "3.8"
+      - uses: eifinger/setup-rye@v2
+      - run: rye sync
```
Kludex: You need to add:

```suggestion
      - run: rye sync
        env:
          UV_EXTRA_INDEX_URL: ${{ secrets.UV_EXTRA_INDEX_URL }}
```

Makefile
samuelcolvin: I’d strongly prefer to keep a Makefile, it makes it much easier for new contributors to get started

And can include use hints like how to install rye

.github/workflows/main.yml
samuelcolvin: why for linting?

Kludex: oh, my bad, i thought this was the docs part

PR #30, opened 2024-04-28
`logfire.instrument_psycopg()` function

alexmojaki:
Part of https://linear.app/pydantic/issue/PYD-874/extra-instrument-methods

### DIFF
--- a/docs/extra/tweaks.css
+++ b/docs/extra/tweaks.css
@@ -1,7 +1,17 @@
+/* https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#custom-colors */
+:root {
+  --md-primary-fg-color: #1D0214;
+  --md-primary-fg-color--light: #1D0214;
+  --md-primary-fg-color--dark: #1D0214;
+}
+
 /* Revert hue value to that of pre mkdocs-material v9.4.0 */
 [data-md-color-scheme="slate"] {
   --md-hue: 230;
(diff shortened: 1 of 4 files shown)


logfire/_internal/main.py
```diff
@@ -831,6 +831,29 @@ def instrument_openai(
...
+        Args:
+            conn_or_module: Can be:
+
+                - The `psycopg` (version 3) or `psycopg2` module.
+                - The string `'psycopg'` or `'psycopg2'` to instrument the module.
+                - `None` (the default) to instrument whichever module(s) are installed.
+                - A `psycopg` or `psycopg2` connection.
+
+            **kwargs: Additional keyword arguments to pass to the OpenTelemetry `instrument` methods,
+                particularly `enable_commenter` and `commenter_options`.
```
Kludex, line 851: Does this renders nicely on the API docs?

logfire/_internal/integrations/psycopg.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import annotations
+
+import contextlib
+import importlib
+from importlib.util import find_spec
+from types import ModuleType
+from typing import TYPE_CHECKING, Any
+
+from packaging.requirements import Requirement
```
Kludex, line 9: Do we have this as dependency?

alexmojaki: Good point, added.

PR #37, opened 2024-04-29
Handle cyclic references in JSON encoding and schema

alexmojaki:
Fixes bug seen in https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1714397069462369

### DIFF
--- a/logfire/_internal/json_schema.py
+++ b/logfire/_internal/json_schema.py
@@ -36,7 +36,7 @@ __all__ = 'create_json_schema', 'attributes_json_schema_properties', 'attributes
 
 
 @lru_cache
-def type_to_schema() -> dict[type[Any], JsonDict | Callable[[Any], JsonDict]]:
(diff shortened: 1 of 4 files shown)


logfire/_internal/json_schema.py
```diff
@@ -106,16 +107,21 @@ def create_json_schema(obj: Any) -> JsonDict:
     obj_type = obj.__class__
     if obj_type in {str, int, bool, float}:
         return {}
-    elif obj_type in {list, tuple, set, frozenset, deque}:
-        return _array_schema(obj)
+
+    if id(obj) in seen:
+        return {}
```
dmontagu, line 112:
```suggestion
        return {'x-circular-reference': True}
```
this way we can render something more nicely in the UI for circular references, rather than having it just look like a normal string

alexmojaki: Leaving this for now because it could make the overall schema significantly larger if the data consists only of JSON values which don't need a schema. This can be added if we actually decide to spend time on rendering in the UI.

PR #47, opened 2024-04-30
Change announcement on the headers

Kludex:
Screenshot 2024-04-30 at 16 24 55

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -1,19 +1,16 @@
 {% extends "base.html" %}
 
 {% block announce %}
-  🔔 Pydantic Logfire is in <strong>closed beta</strong>. 🛠️
-  <br>
-  We are excited for you to try the platform, come join our
-  <a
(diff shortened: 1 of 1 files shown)


docs/overrides/main.html
```diff
@@ -1,19 +1,16 @@
...
 
 {% block extrahead %}
-  <meta name="robots" content="noindex, nofollow" />
+<meta name="robots" content="noindex, nofollow" />
+{% endblock %}
+
+{% block content %}
+{{ super() }}
+<script src="/flarelytics/client.js"></script>
+{% include 'mkdocs_run_deps.html' ignore missing %}
```
Kludex, line 15: I've just copied this from https://github.com/pydantic/pydantic/blob/62822045515364f2ec5ea2a9070da7d13bdd6012/docs/theme/main.html#L9, but don't know what this is about...

samuelcolvin: it'll do nothing since that template doesn't exist yet, it'll be useful if/when we make examples runable.

PR #51, opened 2024-04-30
Update pyproject.toml to include logfire in sdist build target

syniex:
Currently the source distribution just copy the entire repository.
This change will make it include only the source folder.

image


samuelcolvin:
thanks so much, I think the convention is to include more in the sdist than just the source, e.g. see [what pydantic includes](https://github.com/pydantic/pydantic/blob/62822045515364f2ec5ea2a9070da7d13bdd6012/pyproject.toml#L71-L81).

syniex:
update the configuration to include more then only the source code.
I am not sure why the docs test fails

Kludex:
Thanks for the help here @syniex :)

Kludex:
We still don't have an sdist...

Issue #64, opened 2024-05-01
Index error for Azure OpenAI streaming


gabrielchua:
### Description

I'm using `AzureOpenAI` from the `openai` SDK and getting this error.

Is it due to there being an empty chunk?

```
Traceback (most recent call last):
  File "/Users/XXXX/.pyenv/versions/3.11.5/envs/XXXX/lib/python3.11/site-packages/logfire/_internal/integrations/openai.py", line 137, in __stream__
    chunk_content = content_from_stream(chunk)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/XXXX/.pyenv/versions/3.11.5/envs/XXXX/lib/python3.11/site-packages/logfire/_internal/integrations/openai.py", line 183, in <lambda>
    content_from_stream=lambda chunk: chunk.choices[0].delta.content,
                                      ~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

### Python, Logfire & OS Versions, related packages

```TOML
logfire="0.28.0"
platform="macOS-13.5.2-arm64-arm-64bit"
python="3.11.5 (main, Dec 23 2023, 11:01:02) [Clang 14.0.3 (clang-1403.0.22.14.1)]"
[related_packages]
requests="2.31.0"
pydantic="2.6.1"
fastapi="0.109.2"
openai="1.12.0"
protobuf="4.25.2"
rich="13.7.0"
opentelemetry-api="1.24.0"
opentelemetry-exporter-otlp-proto-common="1.24.0"
opentelemetry-exporter-otlp-proto-grpc="1.22.0"
opentelemetry-exporter-otlp-proto-http="1.24.0"
opentelemetry-instrumentation="0.43b0"
opentelemetry-instrumentation-asgi="0.43b0"
opentelemetry-instrumentation-fastapi="0.43b0"
opentelemetry-proto="1.24.0"
opentelemetry-sdk="1.24.0"
opentelemetry-semantic-conventions="0.45b0"
opentelemetry-util-http="0.43b0"
```

hramezani:
Thanks @gabrielchua for reporting this. Yes, it can be related to an empty chunk. we will prepare a fix.

Kludex:
This is fixed on version 0.28.2.

Thanks for the help @gabrielchua :)

PR #69, opened 2024-05-01
Fix OpenAI streaming empty chunk error

hramezani:
Fixes https://github.com/pydantic/logfire/issues/64


Kludex (approved):

Kludex:
Hold on

hramezani:
> Hold on...
> 
> But the `chunk` was available on the issue, what was not available was the `choices` list, isn't it the case?

Yes, let me handle `choices` as well

hramezani:
@Kludex I handled the empty `choices` and added a test for it

Kludex (approved):

PR #101, opened 2024-05-03
Replace `insert_assertion` from devtools by `inline_snapshot.snapshot`

Kludex:
@alexmojaki Can you help me to fix the tests that are failing? They are failing because we customized those assertions.

If possible, guiding me on what is the supposed workflow instead of doing it yourself.

- [ ] Regenerate the lockfile.


Kludex:
I'll open another PR to solve this.

PR #102, opened 2024-05-03
Add `dynamic_context` on coverage

Kludex:
I use this frequently, and I always need to google the "test_function" name...


alexmojaki (approved):
I can't object to adding something commented out lol

PR #103, opened 2024-05-03
docs: add example query for HTTP requests duration per percentile

frankie567:
As suggested by @Kludex on [Slack](https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1714725281924919), here is a simple doc contribution with a (slightly) advanced query, which I very useful to visualize HTTP requests duration.

Feel free to move it around (or ask me to): not sure if it's the right place in the docs structure 🙂


Kludex:
Thanks @frankie567 !

Issue #62, opened 2024-05-01
OpenAI SDK traces will fail when `.with_raw_response` is used


tonybaloney:
### Description

I have this code:

```python
chat_completion_response = await openai_client.chat.completions.with_raw_response.create(
            messages=query_messages,  # type: ignore
            # Azure OpenAI takes the deployment name as the model name
            model=self.chatgpt_deployment if self.chatgpt_deployment else self.chatgpt_model,
            temperature=0.0,  # Minimize creativity for search query generation
            max_tokens=100,  # Setting too low risks malformed JSON, setting too high may affect performance
            n=1,
            tools=tools,
            tool_choice="auto",
        )
self.meter_ratelimit_remaining_tokens.set(
            int(chat_completion_response.headers.get("x-ratelimit-remaining-tokens", 0))
        )
self.meter_ratelimit_remaining_requests.set(
            int(chat_completion_response.headers.get("x-ratelimit-remaining-requests", 0))
        )
chat_completion = chat_completion_response.parse()
```

That causes a crash when instrumentation is enabled using logfire:

```default
ERROR:root:Exception while generating response stream: 'LegacyAPIResponse' object has no attribute 'choices'
Traceback (most recent call last):
  File "/Users/anthonyshaw/projects/azure-search-openai-demo/app/backend/app.py", line 181, in format_as_ndjson
    async for event in r:
  File "/Users/anthonyshaw/projects/azure-search-openai-demo/app/backend/approaches/chatapproach.py", line 152, in run_with_streaming
    extra_info, chat_coroutine = await self.run_until_final_call(
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anthonyshaw/projects/azure-search-openai-demo/app/backend/approaches/chatreadretrieveread.py", line 140, in run_until_final_call
    chat_completion_response = await self.openai_client.chat.completions.with_raw_response.create(
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anthonyshaw/projects/azure-search-openai-demo/.venv/lib/python3.11/site-packages/openai/_legacy_response.py", line 349, in wrapped
    return cast(LegacyAPIResponse[R], await func(*args, **kwargs))
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #54, opened 2024-04-30
Feature Request: Handle Tool Call Assistant Messages For OpenAI (and other providers)


willbakst:
Hi! I noticed when using tool calls (function calling) that I get an empty assistant message in the UI (because there's no content, just tool calls). See below:

![April 30 Screenshot from Pydantic Logfire](https://github.com/pydantic/logfire/assets/99370834/502c3d63-ea78-4284-a12c-b6acf49fe5ca)

Would be awesome if this rendered something like a pretty JSON object so it's not empty and I don't have to scroll down to the original response to find the data.

Kludex:
@willbakst Do you have a code snippet for me to reproduce this?

willbakst:
I've confirmed this is working with OpenAI (as well as other providers through Mirascope).

Here is a Mirascope snippet that works with Logfire and the new function calling UI:

```python
from typing import Literal, Type

import logfire
from pydantic import BaseModel

from mirascope.logfire import with_logfire
from mirascope.openai import OpenAIExtractor

logfire.configure()


class TaskDetails(BaseModel):
    description: str
    due_date: str
    priority: Literal["low", "normal", "high"]


@with_logfire
class TaskExtractor(OpenAIExtractor[TaskDetails]):
    extract_schema: Type[TaskDetails] = TaskDetails
    prompt_template = """
    Extract the task details from the following task:
    {task}
    """

    task: str


task = "Submit quarterly report by next Friday. Task is high priority."
task_details = TaskExtractor(
    task=task
).extract()  # this will be logged automatically with logfire
assert isinstance(task_details, TaskDetails)
print(task_details)
```

PR #104, opened 2024-05-03
Improve the OpenAI integration

Kludex:
- Closes https://github.com/pydantic/logfire/issues/62

### DIFF
--- a/logfire/_internal/integrations/openai.py
+++ b/logfire/_internal/integrations/openai.py
@@ -1,21 +1,48 @@
 from __future__ import annotations
 
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Any, AsyncIterator, Callable, ContextManager, Iterator, NamedTuple
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    AsyncIterator,
+    Callable,
+    ContextManager,
+    Generic,
+    Iterator,
+    NamedTuple,
+    TypeVar,
+    cast,
+)
 
 import openai
+from openai._legacy_response import LegacyAPIResponse
+from openai.types.chat.chat_completion import ChatCompletion
+from openai.types.chat.chat_completion_chunk import ChatCompletionChunk
+from openai.types.completion import Completion
+from openai.types.create_embedding_response import CreateEmbeddingResponse
+from openai.types.images_response import ImagesResponse
 from opentelemetry import context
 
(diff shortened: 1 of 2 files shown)


logfire/_internal/integrations/openai.py
```diff
@@ -59,30 +86,30 @@ def instrument_openai_sync(logfire_openai: Logfire, openai_client: openai.OpenAI
     # WARNING: this method is vey similar to `instrument_openai_async` below, any changes here should be reflected there
     openai_client._original_request_method = original_request_method = openai_client._request  # type: ignore
 
-    def instrumented_openai_request(**kwargs: Any) -> Any:
+    def instrumented_openai_request(**kwargs: Unpack[OpenAIRequest[ResponseT, _StreamT]]) -> ResponseT | _StreamT:
         if context.get_value('suppress_instrumentation'):
             return original_request_method(**kwargs)
 
-        options: FinalRequestOptions | None = kwargs.get('options')
+        options = kwargs.get('options')
```
Kludex: `options` is mandatory, it will never be `None`.

logfire/_internal/integrations/openai.py
```diff
@@ -1,22 +1,49 @@
...
-    from openai.types.completion import Completion
-    from openai.types.create_embedding_response import CreateEmbeddingResponse
-    from openai.types.images_response import ImagesResponse
-    from typing_extensions import LiteralString
+    from openai._types import ResponseT
+    from typing_extensions import LiteralString, TypedDict, Unpack
 
     from ..main import Logfire, LogfireSpan
 
+    _AsyncStreamT = TypeVar('_AsyncStreamT', bound=AsyncStream[Any])
```
samuelcolvin, line 35: can you add a comment explaining roughly what these do and why they're needed - just one comment for all of them is probably enough.

logfire/_internal/integrations/openai.py
samuelcolvin:
```suggestion
        options = kwargs['options']
```
?

samuelcolvin (approved):
otherwise LGTM

alexmojaki:
Does this take https://github.com/pydantic/logfire/issues/54 into account?

Kludex:
> Does this take #54 into account?

Kludex:
> Does this take #54 into account?

No. Unrelated.

PR #119, opened 2024-05-04
Enable pyright

Kludex:


alexmojaki:
Looks like there's no CI job that runs pre-commit?

Kludex:
Yeah, I'm on it.

Issue #126, opened 2024-05-05
Offer a command or argument that returns only the packages listed by `logfire inspect`


shaunhegarty:
### Description

I ran `logfire inspect`, and it spits out a a lot of text plus an overly formatted version of the pip install command. Copying it gives me numerous newlines (where it wraps) which I have to clean up to run the command. Admittedly this could just be some terminal setting that is inconveniencing me. 

```
pip install opentelemetry-instrumentation-httpx opentelemetry-instrumentation-psycopg2 opentelemetry-instrumentation-urllib opentelemetry-instrumentation-sqlite3 opentelemetry-instrumentation-fastapi          
                                                                      opentelemetry-instrumentation-requests opentelemetry-instrumentation-sqlalchemy
```

It would have been slicker if I could just return the list so that I could run `pip install $(logfire <some command>)`

Kludex:
I've open a PR to address this. :)

alexmojaki:
Hmmm, I worry that the `inspect` command (particularly that pip install command it prints) may be creating a misunderstanding, and this might reinforce that. Installing these packages will not do anything on its own. Logfire doesn't automatically detect and use them at runtime. You still have to manually add some code for each one. @shaunhegarty I'm curious what your expectation/impression was here. Were you prepared to still manually use each installed package? Am I overthinking this?

I think it'd be better for the inspect command to:

1. Make it clear what the user needs to do in addition to installing packages.
2. Make it clear that the `OpenTelemetry instrumentation package` column contains links.
3. Add a column with links to logfire docs where possible.

> Admittedly this could just be some terminal setting that is inconveniencing me.

No, it's `rich` wrapping the text by default. I'll fix that.

shaunhegarty:
@alexmojaki @Kludex thanks for making that change!

> I'm curious what your expectation/impression was here. 

The inspect offered a command to install packages for things I may want instrumented. Since it was offered, I tried to copy it and then I had to clean it up. Looks like that shouldn't be an issue now at least :). If otel package names weren't so cumbersome or the command wasn't offered in the first place, I probably wouldn't have thought to raise anything.

PR #128, opened 2024-05-06
Add `--only-requirements` argument for `logfire inspect`

Kludex:
- Closes https://github.com/pydantic/logfire/issues/126


Kludex:
Merging https://github.com/pydantic/logfire/pull/130 instead of this.

PR #137, opened 2024-05-06
Use `stack_info` instead of `stack_offset`

Kludex:


logfire/_internal/stack_info.py
```diff
@@ -65,3 +70,35 @@ def get_caller_stack_info(stack_offset: int = 3) -> StackInfo:
...
+
+def get_user_stack_offset() -> int:
+    """Get the stack offset of the user code.
+
+    We want to skip the internal code, and third party code, and get the user code stack info.
+
+    Returns:
+        The stack offset of the user code.
+    """
+    stack_offset = 2
```
Kludex: @alexmojaki Do you know why the number 2 here works? I didn't check the stack closely...

logfire/_internal/stack_info.py
```diff
@@ -65,3 +70,35 @@ def get_caller_stack_info(stack_offset: int = 3) -> StackInfo:
...
+    """Check if the filename is a user filename.
+
+    Args:
+        filename: The filename to check.
+
+    Returns:
+        True if the filename is a user filename, False otherwise.
+    """
+    filename = get_filepath(filename)
+    return not (filename.startswith('logfire') or 'site-packages' in filename)
```
Kludex: Does this conditional make sense?

logfire/_internal/stack_info.py
alexmojaki: It doesn't seem like it *would* work as a default, and the pragma suggests that it hasn't been tested.

logfire/_internal/stack_info.py
alexmojaki: I would prefer something like this:

```python
from pathlib import Path

import opentelemetry.sdk.trace

import logfire

site_packages_dir = str(Path(opentelemetry.sdk.trace.__file__).parent.parent.parent.parent.absolute())
logfire_dir = str(Path(logfire.__file__).parent.absolute())
prefixes = (site_packages_dir, logfire_dir)


def is_ignored(path):
    return str(Path(path).absolute()).startswith(prefixes)
```

Plus some generous caching - caching `get_code_object_info` was a big win because pathlib is slow.

logfire/_internal/stack_info.py
Kludex: The pragma on the exception you mean?

alexmojaki: Yes, since there's no exceptions happening (and I guess there generally wouldn't be) I think the final `stack_offset` is always greater than 2. Not sure if a lower starting point is needed.

Kludex: What should I do here?

alexmojaki: Nothing, I just realized I misunderstood the meaning of this number. I think this is fine.

logfire/_internal/stack_info.py
```diff
@@ -65,3 +73,35 @@ def get_caller_stack_info(stack_offset: int = 3) -> StackInfo:
...
+def is_user_filename(filename: str) -> bool:
+    """Check if the filename is a user filename.
+
+    Args:
+        filename: The filename to check.
+
+    Returns:
+        True if the filename is a user filename, False otherwise.
+    """
+    return not filename.startswith(PREFIXES)
```
alexmojaki:
```suggestion
    return not Path(filename).absolute().startswith(PREFIXES)
```

I'm not sure if `frame.f_code.co_filename` is always absolute.

logfire/_internal/stack_info.py
```diff
@@ -65,3 +73,35 @@ def get_caller_stack_info(stack_offset: int = 3) -> StackInfo:
...
+
+@lru_cache(maxsize=2048)
+def is_user_filename(filename: str) -> bool:
+    """Check if the filename is a user filename.
+
+    Args:
+        filename: The filename to check.
+
+    Returns:
+        True if the filename is a user filename, False otherwise.
```
alexmojaki: This is a bit redundant lol

alexmojaki (approved):

alexmojaki:
Can you try using this with structlog?

logfire/_internal/stack_info.py
Kludex: I think it is? https://github.com/python/cpython/issues/64642

(I only read the first message, and the one before closing the issue)

alexmojaki: It looks like that was released in 3.9. And since this is cached I don't think it hurts to be safe.

Kludex:
> Can you try using this with structlog?

Yeah, it doesn't work. That number 2 doesn't look correct. I'll not merge this.

logfire/_internal/stack_info.py
```diff
@@ -65,3 +73,35 @@ def get_caller_stack_info(stack_offset: int = 3) -> StackInfo:
...
+    """
+    return not Path(filename).absolute().startswith(PREFIXES)
```
Kludex:
```suggestion
    return not str(Path(filename).absolute()).startswith(PREFIXES)
```

Kludex:
I've fixed this to also work with structlog. I've explained the + and - I've added. I think we need to refactor the logic we have around those `stack_offset` anyway

tests/otel_integrations/test_fastapi.py
```diff
@@ -825,8 +825,8 @@ def test_scrubbing(client: TestClient, exporter: TestExporter) -> None:
                     'logfire.level_num': 5,
                     'logfire.msg_template': 'FastAPI arguments',
                     'logfire.msg': 'FastAPI arguments',
-                    'code.filepath': 'fastapi.py',
-                    'code.function': 'solve_dependencies',
+                    'code.filepath': 'events.py',
+                    'code.function': '_run',
```
Kludex: This is still wrong.

tests/otel_integrations/test_openai.py
```diff
@@ -231,8 +231,8 @@ def test_sync_chat_completions(instrumented_client: openai.Client, exporter: Tes
                 'start_time': 1000000000,
                 'end_time': 2000000000,
                 'attributes': {
-                    'code.filepath': 'openai.py',
-                    'code.function': 'instrumented_openai_request',
+                    'code.filepath': 'test_openai.py',
+                    'code.function': 'test_sync_chat_completions',
```
Kludex, line 235: This is fixed.

tests/test_pydantic_plugin.py
```diff
@@ -436,8 +436,8 @@ class MyModel(BaseModel, plugin_settings={'logfire': {'record': 'all'}}):
                     'logfire.span_type': 'span',
                     'logfire.msg_template': 'Pydantic {schema_name} {validation_method}',
                     'logfire.msg': 'Pydantic MyModel validate_python failed',
-                    'code.filepath': 'pydantic.py',
-                    'code.function': '_on_enter',
+                    'code.filepath': 'test_pydantic_plugin.py',
+                    'code.function': 'test_pydantic_plugin_python_error',
```
Kludex, line 440: This is also fixed.

tests/test_structlog.py
```diff
@@ -44,9 +44,9 @@ def test_structlog(exporter: TestExporter, logger: Logger) -> None:
                     'logfire.level_num': 9,
                     'logfire.msg_template': 'This is now being logged.',
                     'logfire.msg': 'This is now being logged.',
-                    'code.filepath': 'python.py',
-                    'code.function': 'pytest_pyfunc_call',
-                    'code.lineno': 195,
+                    'code.filepath': 'test_structlog.py',
+                    'code.function': 'test_structlog',
+                    'code.lineno': 33,
```
Kludex, line 49: This is also fixed.

Kludex (commented):
The `logfire.log` still has the `stack_info` parameter, but it's not used. Should we remove it or deprecate it?

alexmojaki:
I've made the following changes, hope it's OK:

1. Removed the `stack_info` param from `logfire_format`. When it needs a `stacklevel` it can just find the user frame again. This makes things a bit cleaner and also means that `instrument` doesn't have to recompute `stack_info` just for formatting, it already has that info in its attributes.
2. Also excluded stdlib frames. This means that the fastapi arguments logs now have no stack info at all, which I think is fine. We can make maybe them point at the endpoint function in the future, but there was no way they could get stack info from calling frames.
3. Excluded list/set/dict comprehension frames, partly just to make tests consistent across versions.
4. Remove the `stack_offset` param of `Logfire.log` as you suggested. I think it's best to just rip it out rather than deprecate, and it'd be good to hear from users who were using it (which is unlikely). Besides, the `_stack_offset` param of `span` was also removed, and that was also a public API.

alexmojaki:
Merged so that I can immediately resolve conflicts in https://github.com/pydantic/logfire/pull/151.

PR #139, opened 2024-05-06
Get attribute values from frame variables

alexmojaki:
This makes this work:

```python
import logfire

foo = 1
logfire.info('foo is {foo}')
```

The ultimate goal is allow passing an f-string, this is the first step.


alexmojaki:
https://pydantic.slack.com/archives/C05AF4A4WRM/p1715092161303329 is making me think that this is NOT a first step towards how the f-string magic will work. So unless we want this as a feature in addition to f-string magic, I might throw a lot of this away.

alexmojaki:
Canceling this feature and closing in favour of https://github.com/pydantic/logfire/pull/151

PR #146, opened 2024-05-07
Create `logfire-noop`

### LINKED ISSUE #121: `logfire-noop`
### Description

My idea is that we release `logfire-noop`, this package has no dependencies and allows third-party libraries to integrate with logfire while still giving their users complete choice over whether to actually use Logfire.

The idea is that `logfire-noop` would contain two modules: `logfire_noop` and `logfire_if_installed` (name TBC)

`logfire_noop` exports types matching logfire but that do nothing, or do the minimum required for code to run, e.g. `logfire.span()` needs to return a context manager.

`logfire_if_installed` behaves like this:
* when `logfire` is installed, it just the contents of `logfire`
* when logfire is not installed, it just the contents of `logfire_noop` 

Third party libraries would wuite code like this:

```py
import logfire_if_installed

...

def my_library_method(...):
    ...
    with logfire.span('doing a thing {sniffle=}', sniffle=rofl):
        ...
```

Kludex:
- Closes #121

### Checklist

- [ ] Test should compare the docstrings.
- [ ] Test should compare the signatures.
- [ ] Test should compare the `__all__`.


Kludex:
We had a discussion about this... This may not be the way we want to go.

The proposed was to make `logfire` depend on `logfire_noop`.

PR #151, opened 2024-05-07
f-string magic

alexmojaki:
Leaving docs to another PR because https://docs.pydantic.dev/logfire/guides/onboarding_checklist/add_manual_tracing/ needs significant reworking anyway - it already uses f-strings where it shouldn't have and doesn't mention 'manual' formatting.


logfire/_internal/config.py
```diff
@@ -156,6 +156,7 @@ def configure(
     fast_shutdown: bool = False,
     scrubbing_patterns: Sequence[str] | None = None,
     scrubbing_callback: ScrubCallback | None = None,
+    fstring_magic: bool | None = None,
```
Kludex: I think we said we'd change it for `fstring_inspect` or something? I don't recall the conversation properly.

logfire/_internal/formatter.py
```diff
@@ -1,19 +1,27 @@
...
+import logfire
 from logfire._internal.stack_info import get_user_frame_and_stacklevel
 
-__all__ = 'chunks_formatter', 'LiteralChunk', 'ArgChunk', 'logfire_format'
-
 from .constants import MESSAGE_FORMATTED_VALUE_LENGTH_LIMIT
 from .scrubbing import Scrubber
 from .utils import truncate_string
 
+__all__ = 'chunks_formatter', 'LiteralChunk', 'ArgChunk', 'logfire_format'
```
Kludex: Do we need this? 👀

pyproject.toml
```diff
@@ -47,6 +47,7 @@ dependencies = [
     "protobuf >= 4.23.4",
     "typing-extensions >= 4.1.0",
     "tomli >= 2.0.1; python_version < '3.11'",
+    "executing>=2.0.1",
```
Kludex, line 50: `python_version >= 3.9`? Or... optional if not using the feature?

logfire/_internal/main.py
```diff
@@ -532,7 +545,26 @@ def log(
...
         merged_attributes = {**stack_info, **attributes}
         if (msg := attributes.pop(ATTRIBUTES_MESSAGE_KEY, None)) is None:
-            msg = logfire_format(msg_template, merged_attributes, self._config.scrubber)
+            fstring_frame = None
+            if self._config.inspect_arguments:
+                fstring_frame = inspect.currentframe()
+                if fstring_frame.f_back.f_code.co_filename == Logfire.log.__code__.co_filename:  # type: ignore
+                    # fstring_frame.f_back should be the user's frame.
+                    # The user called logfire.info or a similar method rather than calling logfire.log directly.
+                    fstring_frame = fstring_frame.f_back  # type: ignore
```
Kludex, line 554: Can't we use some utility function we have already?

logfire/_internal/formatter.py
```diff
@@ -113,29 +318,132 @@ def chunks(
...
+    in the source positions of AST nodes inside f-strings.
+    """
+    # ast.unparse is not available in Python 3.8, which is why inspect_arguments is forbidden in 3.8.
+    source_unparsed = ast.unparse(node)
+    source_segment = ast.get_source_segment(ex_source.text, node) or ''
+    try:
+        # Verify that the source segment is correct by checking that the AST is equivalent to what we have.
+        source_segment_unparsed = ast.unparse(ast.parse(source_segment, mode='eval'))
+    except Exception:
+        source_segment_unparsed = ''
```
Kludex: This is `AttributeError` only?

Kludex (approved):

logfire/_internal/main.py
alexmojaki: Here 'user' includes us. We need the place where the actual logfire method was called, even if it's inside an integration.

logfire/_internal/formatter.py
alexmojaki: No, added a comment

logfire/_internal/config.py
alexmojaki: Changed to `inspect_arguments`

pyproject.toml
alexmojaki: Started doing this but decided not to. If we accidentally eagerly import `executing` somewhere we won't notice this breaking 3.8 because it'll be installed in CI.

Kludex (approved):

Issue #145, opened 2024-05-07
Generate metric points from pydantic class changes?


inimaz:
### Question

Hello team! thanks for the great tool.

I haven't found it in the docs and maybe it is a use case you might be interested in covering? I have a `pydantic` class and I want to create a gauge/counter metric point out of it every time its value gets updated.

In the docs I see you have an integration detecting the validation results, so maybe there is a way already?

Thanks a lot.

Kludex:
We have that already! 

The name of the metric is "pydantic.validations". We are recording this information:

https://github.com/pydantic/logfire/blob/a5026b1fd57e7a99daecaaa8c7613c97a01a7d32/logfire/integrations/pydantic.py#L258-L260

What's your use case? Do you want to create some dashboard? We are interested. 👀

inimaz:
Good to know! But that counter counts the number of times it does the validation. Sorry for the misunderstanding. Instead I would like to monitor the value of the updated field.
For example if I have this:
```
from pydantic import BaseModel

class Temperature(BaseModel):
    value: int = 22
```
Every time someone does `Temperature(value=27)` ==> metric point with value 27.
Right now the way to do this is to create the metric yourself using `otel` and then pass this metric to `logfire` I guess?

alexmojaki:
Here's what you can do:

```python
import pydantic

import logfire

gauge = logfire.DEFAULT_LOGFIRE_INSTANCE._config.meter.create_gauge('temperature')


class Temperature(pydantic.BaseModel, validate_assignment=True):
    value: int

    @pydantic.field_validator('value')
    def validate_value(cls, value):
        gauge.set(value)
        return value


t = Temperature(value=1)
t.value = 2
```

`create_gauge` is a new OTEL method, we should add a `logfire.metric_gauge` method.

alexmojaki:
BTW you can now write `gauge = logfire.metric_gauge('temperature')`.

PR #153, opened 2024-05-08
Add `logfire.metric_gauge()`

Kludex:
- Closes https://github.com/pydantic/logfire/issues/145

## Checklist
- [x] Add tests.
- [ ] Add documentation.
   - [ ] Improve the explanation on "why" to use each metric, and give examples.

### DIFF
--- a/logfire/_internal/metrics.py
+++ b/logfire/_internal/metrics.py
@@ -25,8 +25,10 @@ from opentelemetry.util.types import Attributes
 try:
     # This only exists in opentelemetry-sdk>=1.23.0
     from opentelemetry.metrics import _Gauge as Gauge
+
+    GAUGE_IMPORTED = True
 except ImportError:  # pragma: no cover
-    Gauge = None
+    GAUGE_IMPORTED = False  # type: ignore
 
 # All the cpu_times fields provided by psutil (used by system_metrics) across all platforms,
 # except for 'guest' and 'guest_nice' which are included in 'user' and 'nice' in Linux (see psutil._cpu_tot_time).
@@ -212,7 +214,7 @@ class _ProxyMeter(Meter):
         name: str,
         unit: str = '',
         description: str = '',
-    ):  # pragma: no cover
+    ) -> Gauge:  # pragma: no cover
         with self._lock:
             proxy = _ProxyGauge(self._meter.create_gauge(name, unit, description), name, unit, description)
             self._instruments.add(proxy)
@@ -353,7 +355,7 @@ class _ProxyUpDownCounter(_ProxyInstrument[UpDownCounter], UpDownCounter):
         return meter.create_up_down_counter(self._name, self._unit, self._description)
 
 
-if Gauge is not None:  # pragma: no branch
+if not GAUGE_IMPORTED:  # pragma: no branch
 
     class _ProxyGauge(_ProxyInstrument[Gauge], Gauge):
         def set(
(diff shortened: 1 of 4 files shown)


logfire/_internal/metrics.py
```diff
@@ -25,8 +25,10 @@
 try:
     # This only exists in opentelemetry-sdk>=1.23.0
     from opentelemetry.metrics import _Gauge as Gauge
+
+    GAUGE_IMPORTED = True
 except ImportError:  # pragma: no cover
-    Gauge = None
+    GAUGE_IMPORTED = False  # type: ignore
```
alexmojaki: I don't see why this is better than `Gauge = None`, and it requires 2 extra type-ignores.

Kludex: It actually avoids the type ignore on the type hint of the `metric_gauge` method - but I just had an idea on how to avoid one more.

I'll eat and come back to this.

logfire/_internal/metrics.py
```diff
@@ -206,13 +208,19 @@ def create_histogram(
...
             return proxy
 
-    # TODO(Marcelo): We should test this method.
     def create_gauge(
         self,
         name: str,
         unit: str = '',
         description: str = '',
-    ):  # pragma: no cover
+    ) -> _Gauge:  # pragma: no cover
```
Kludex: This is the problem.

logfire/_internal/metrics.py
```diff
@@ -353,7 +361,7 @@ def _create_real_instrument(self, meter: Meter) -> UpDownCounter:
         return meter.create_up_down_counter(self._name, self._unit, self._description)
 
 
-if Gauge is not None:  # pragma: no branch
+if Gauge is not None:  # pragma: no cover
```
alexmojaki: Surely this is covered?

logfire/_internal/metrics.py
```diff
@@ -24,7 +24,9 @@
 
 try:
     # This only exists in opentelemetry-sdk>=1.23.0
-    from opentelemetry.metrics import _Gauge as Gauge
+    from opentelemetry.metrics import _Gauge
+
+    Gauge = _Gauge
```
alexmojaki, line 29: interesting, pyright treats this differently to `import _Gauge as Gauge`?

tests/test_metrics.py
```diff
@@ -4,12 +4,15 @@
...
 from typing import Any, cast
 
+import pytest
 from dirty_equals._numeric import IsInt
 from inline_snapshot import snapshot
 from opentelemetry.metrics import CallbackOptions, Observation
 from opentelemetry.sdk.metrics.export import AggregationTemporality, InMemoryMetricReader, MetricsData
 
 import logfire
+import logfire._internal
```
alexmojaki:
```suggestion
```

logfire/_internal/metrics.py
Kludex: I did this because we need a return type for `create_gauge` in this file, and without this line, Gauge can either be `_Gauge` or `None`. Since it can be either, then the type of `create_gauge` is wrong.

That's why the use of another variable - the type checker will know that `Gauge` is `_Gauge`.

logfire/_internal/metrics.py
Kludex: My bad, I was just trying to readd what was previously there.

Kludex:
```suggestion
if Gauge is not None:  # pragma: no branch
```

alexmojaki (approved):

adriangb:
Thanks I needed this the other day 😆

PR #154, opened 2024-05-08
Set `send_to_logfire` to `False` when running under Pytest

Kludex:


logfire/_internal/config_params.py
```diff
@@ -47,10 +47,23 @@ class ConfigParam:
...
+    """A default value that is computed at runtime.
+
+    A good example is when we want to check if we are running under pytest and set a default value based on that.
+    """
+
+    callback: Callable[[], Any]
+
+
+_send_to_logfire_default = _DefaultCallback(lambda: 'PYTEST_CURRENT_TEST' not in os.environ)
+"""When running under pytest, don't send spans to Logfire by default."""
```
Kludex, line 61: If this can be avoided, suggestions are welcome.

alexmojaki (approved):

PR #155, opened 2024-05-08
Enable tests on multiple Python versions

Kludex:


.github/workflows/main.yml
```diff
@@ -68,8 +68,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - uses: eifinger/setup-rye@v2.6
+      - name: Pin python-version ${{ matrix.python-version }}
+        run: rye pin ${{ matrix.python-version }}
       - run: rye config --set-bool behavior.use-uv=true
-      - run: rye sync --no-lock
+      - run: rye sync
```
Kludex, line 75: This ignores the lockfile we have, because it generates a new one.

pyproject.toml
```diff
@@ -116,7 +116,7 @@ dev-dependencies = [
     "loguru",
     "ruff",
     "pyright>=1.1.360",
-    "pre-commit>=3.7.0",
+    "pre-commit>=3.5.0",
```
Kludex, line 119: pre-commit 3.6.0 doesn't support Python 3.8.0.

.github/workflows/main.yml
alexmojaki: please add a comment to this line. maybe even remove the existing lockfiles first to be sure.

tests/test_source_code_extraction.py
```diff
@@ -111,7 +111,7 @@ def test_source_code_extraction_method(exporter: TestExporter) -> None:
                     'end_time': 2000000000,
                     'attributes': {
                         'code.filepath': 'tests/test_source_code_extraction.py',
-                        'code.lineno': 17,
+                        'code.lineno': 19,
                         'code.function': 'method',
```
alexmojaki: It looks like this is the only part that differs between the versions, can the test reflect that more clearly instead of these big branches?

.github/workflows/main.yml
alexmojaki: Surely this uses the lockfile in Python 3.12? It doesn't have to be in this PR but I still think we should be testing without a lockfile as much as possible.

alexmojaki (approved):

PR #160, opened 2024-05-10
Raise an exception when Pydantic plugin is enabled on Pydantic <2.5.0

bossenti:
see https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1714923558748049


alexmojaki:
I'm not sure what the minimum supported version is, but it's definitely less than 2.7.0.

Kludex:
I think it's 2.6.2 - need to check.

bossenti:
I can only say that I haven't worked with `2.4.2` but with `2.7.0`.
Best would be Samuels solution and adding a console statement, I just filed this PR so there is at least some information about it.
But I'm fine to close it if you don't think it's appropriate.

Kludex:
2.5.0 is the minimum version.

logfire/_internal/config.py
```diff
@@ -383,6 +383,14 @@ def _load_configuration(
             # This is particularly for deserializing from a dict as in executors.py
             pydantic_plugin = PydanticPlugin(**pydantic_plugin)  # type: ignore
         self.pydantic_plugin = pydantic_plugin or param_manager.pydantic_plugin
+        if self.pydantic_plugin.record != 'off':
+            import pydantic
+
+            # setuptools is a dependency of opentelemetry-instrumentation, we use it instead of adding "packaging" as a dependency.
+            from setuptools._vendor.packaging.version import Version
```
alexmojaki: I'd like it if we had a util function which tried to import things from `packaging` and then fell back to `setuptools._vendor.packaging`, so that we use the proper module where possible. `instrument_psycopg` could then also use that and we could remove the `packaging` dep from the extra.

logfire/integrations/pydantic.py
```diff
@@ -292,61 +291,68 @@ def get_schema_name(schema: CoreSchema) -> str:
 class LogfirePydanticPlugin:
     """Implements a new API for pydantic plugins.
 
-    Patches pydantic to accept this new API shape.
+    Patches Pydantic to accept this new API shape.
 
     Set the `LOGFIRE_DISABLE_PYDANTIC_PLUGIN` environment variable to `true` to disable the plugin.
```
alexmojaki:
```suggestion
    Set the `LOGFIRE_DISABLE_PYDANTIC_PLUGIN` or `PYDANTIC_DISABLE_PLUGINS` environment variable to `true` to disable the plugin.
```

Kludex: The reason I didn't do this, is because I'd have to explain that `PYDANTIC_DISABLE_PLUGINS` disables all the plugins. I don't think it's necessary to do it here.

logfire/integrations/pydantic.py
```diff
@@ -292,61 +291,69 @@ def get_schema_name(schema: CoreSchema) -> str:
...
-        logfire_settings = plugin_settings.get('logfire')
-        if logfire_settings and 'record' in logfire_settings:
-            record = logfire_settings['record']
-        else:
-            record = _pydantic_plugin_config().record
+    if get_version(pydantic.__version__) < get_version('2.5.0'):  # pragma: no cover
+
+        def new_schema_validator(  # type: ignore[reportRedeclaration]
+            self, schema: CoreSchema, config: CoreConfig | None, plugin_settings: dict[str, Any]
+        ) -> tuple[_ValidateWrapper, ...] | tuple[None, ...]:
```
alexmojaki:
```suggestion
    if get_version(pydantic.__version__) < get_version('2.5.0') or os.environ.get('LOGFIRE_PYDANTIC_RECORD') == 'off':  # pragma: no cover

        def new_schema_validator(  # type: ignore[reportRedeclaration]
            self, *_: Any, *__: Any
        ) -> tuple[_ValidateWrapper, ...] | tuple[None, ...]:
```

Best to make `LOGFIRE_PYDANTIC_RECORD=off` a complete no-op in case of bugs such as https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1714528611959999.

logfire/integrations/pydantic.py
```diff
@@ -292,61 +291,69 @@ def get_schema_name(schema: CoreSchema) -> str:
...
-
-        logfire_settings = plugin_settings.get('logfire')
-        if logfire_settings and 'record' in logfire_settings:
-            record = logfire_settings['record']
-        else:
-            record = _pydantic_plugin_config().record
+    if get_version(pydantic.__version__) < get_version('2.5.0') or os.environ.get('LOGFIRE_PYDANTIC_RECORD') == 'off':  # pragma: no cover
+
+        def new_schema_validator(  # type: ignore[reportRedeclaration]
+            self, *_: Any, *__: Any
```
alexmojaki:
```suggestion
            self, *_: Any, **__: Any
```

alexmojaki (approved):

Issue #161, opened 2024-05-11
Error/Warning: Loguru Success Level Integration throws warning


bllchmbrs:
### Description

Hey! Loving logfire so far... have a question.

I am getting this error (or rather, it's a warning).

```
UserWarning: Invalid log level name: 'level 25' warnings.warn(f'Invalid log level name: {level!r}')
```

While I am using loguru and logfire together.

I do not believe I am doing anything special with loguru at all - I am using all of the defaults and I don't completely understand how /why this is happening.

After looking at [the loguru docs](https://loguru.readthedocs.io/en/stable/api/logger.html#levels), it seems that level 25 is "success" in Loguru, but isn't recognized by logfire - this seems strange.

I believe this error triggers on `logger.success` calls.

### Edit 1:

here's the repro:

```
import logfire
from loguru import logger

logger.configure(handlers=[logfire.loguru_handler()])
logger.success('Hello, {name}!', name='World')
```

### Edit 2 

Actually, this is worse than I thought. It throws an exception in the Logfire dashboard when I run `loguru.success` basically the opposite of what I want. below is how a "success" shows up.
<img width="167" alt="image" src="https://github.com/pydantic/logfire/assets/1642503/434c25d3-c538-4337-be89-a3c8a698d447">

alexmojaki:
This has been released, so if you upgrade the SDK then `logger.success` shouldn't cause issues.

PR #162, opened 2024-05-12
Handle intermediate logging/loguru levels

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/161


alexmojaki:
> Should we mention this in the docs as a quick note?

If you want :stuck_out_tongue_closed_eyes:

Issue #175, opened 2024-05-14
`logfire projects use` should give you a list of projects to choose from


samuelcolvin:
### Description

Currently the command just shows

```
usage: logfire projects use [-h] [--org ORG] [--data-dir DATA_DIR] project_name
logfire projects use: error: the following arguments are required: project_name
```

I'm not sure if this was ever right, but it should definitely be fixed

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="0.31.0"
platform="macOS-14.4.1-arm64-arm-64bit"
python="3.12.1 (main, Jan 10 2024, 15:15:22) [Clang 15.0.0 (clang-1500.0.40.1)]"
[related_packages]
requests="2.31.0"
protobuf="4.25.3"
rich="13.7.1"
opentelemetry-api="1.24.0"
opentelemetry-exporter-otlp-proto-common="1.24.0"
opentelemetry-exporter-otlp-proto-http="1.24.0"
opentelemetry-instrumentation="0.45b0"
opentelemetry-proto="1.24.0"
opentelemetry-sdk="1.24.0"
opentelemetry-semantic-conventions="0.45b0"
```

PR #177, opened 2024-05-14
Do not require project name on `logfire projects use` command

Kludex:
- Closes https://github.com/pydantic/logfire/issues/175


alexmojaki:
Is it not possible to have an optional positional argument? I think `projects use myproject` should still work.

Kludex:
> Is it not possible to have an optional positional argument? I think `projects use myproject` should still work.

I thought it was not... I've replaced it 👍

samuelcolvin (approved):
I've tried it and it works as expected 👍 .

PR #181, opened 2024-05-15
Anthropic instrumentation

### LINKED ISSUE #110: Support of instrumenting Anthropic models
> Originally posted by @salahaz  as a discussion #88

Is there going to be in the near future support of instrumenting other AI models such Anthropic?

willbakst:
Closes #110 

Some notes:
- Tried my best to refactor with OpenAI. This ultimately led to a decision to have a slightly worse internal DX due to lack of great typing for certain things. Wasn't sure how to resolve this without a dependency on any particular package in the new `llm_providers.py` file. If there are any suggestions for how to improve this, I'm all ears.
- I made some decisions with Anthropic to get it working with the UI, but I'm not certain this is the best behavior. I left a detailed comment in the code. Might be best to update the UI instead of doing what I did.


logfire/_internal/integrations/llm_providers/types.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Callable, NamedTuple
+
+if TYPE_CHECKING:
+    from typing_extensions import LiteralString
+
+
+__all__ = ('EndpointConfig',)
```
alexmojaki:
```suggestion
from typing import Any, Callable, NamedTuple

from typing_extensions import LiteralString
```

alexmojaki: (I don't know why the original openai.py had `__all__` or imported from typing_extensions inside `if TYPE_CHECKING`)

logfire/_internal/integrations/llm_providers/llm_provider.py
```diff
@@ -0,0 +1,162 @@
...
+    logfire: Logfire,
+    client: Any,
+    suppress_otel: bool,
+    scope_suffix: str,
+    get_endpoint_config_fn: Callable[[Any], EndpointConfig],
+    on_response_fn: Callable[[Any, LogfireSpan], Any],
+    is_async_client_fn: Callable[[Any], bool],
+) -> ContextManager[None]:
+    """Instruments the provided `client` with `logfire`."""
+    logfire_llm = logfire.with_settings(custom_scope_suffix=scope_suffix.lower(), tags=['llm'])
```
alexmojaki: I'm not sure if we need tags in addition to the scope at all, but if we do I think it should be `LLM`.

logfire/_internal/integrations/llm_providers/llm_provider.py
```diff
@@ -0,0 +1,162 @@
...
+    client._is_instrumented_by_logfire = True
+    client._original_request_method = original_request_method = client._request
+
+    is_async = is_async_client_fn(client)
+
+    def _instrumentation_setup(**kwargs: Any) -> Any:
+        if context.get_value('suppress_instrumentation'):
+            return None, None, kwargs
+
+        options = kwargs['options']
```
alexmojaki, line 39: Might as well make this a normal keyword argument.

logfire/_internal/integrations/llm_providers/llm_provider.py
```diff
@@ -0,0 +1,162 @@
...
+
+            class LogfireInstrumentedAsyncStream(stream_cls):
+                async def __stream__(self) -> AsyncIterator[Any]:
+                    with record_streaming(logfire_llm, span_data, content_from_stream) as record_chunk:
+                        async for chunk in super().__stream__():  # type: ignore
+                            record_chunk(chunk)
+                            yield chunk
+
+            if is_async:
+                kwargs['stream_cls'] = LogfireInstrumentedAsyncStream
```
alexmojaki, line 65: Pleae define the class here, defining classes isn't free.

logfire/_internal/integrations/llm_providers/llm_provider.py
```diff
@@ -0,0 +1,162 @@
...
+            else:
+                kwargs['stream_cls'] = LogfireInstrumentedStream
+
+        return message_template, span_data, kwargs
+
+    def instrumented_llm_request_sync(**kwargs: Any) -> Any:
```
alexmojaki, line 79: I think this would work, but the fact that I'm not sure suggests it's maybe too complex to be worth it:

```python
    @contextmanager
    def instrumented_llm_request(**kwargs: Any) -> Any:
        message_template, span_data, kwargs = _instrumentation_setup(**kwargs)
        if message_template is None:
            yield lambda x: x
            return

        stream = kwargs['stream']
        with logfire_llm.span(message_template, **span_data) as span:
            with maybe_suppress_instrumentation(suppress_otel):
                if stream:
                    yield lambda x: x
                    return
                else:
                    yield lambda x: on_response_fn(x, span)
                    return

    def instrumented_llm_request_sync(**kwargs: Any) -> Any:
        with instrumented_llm_request(**kwargs) as request_fn:
            return request_fn(original_request_method(**kwargs))

    async def instrumented_llm_request_async(**kwargs: Any) -> Any:
        with instrumented_llm_request(**kwargs) as request_fn:
            return request_fn(await original_request_method(**kwargs))
```

logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -0,0 +1,83 @@
...
+    url = options.url
+    json_data = options.json_data
+    if not isinstance(json_data, dict):
+        raise ValueError('Expected `options.json_data` to be a dictionary')
+
+    if url == '/v1/messages' or url == '/v1/messages?beta=tools':
+        # Note: this enables the UI to display the system message; however, it also shows
+        # the system message in the messages array, which isn't actually what's being sent.
+        # Likely better to enable the UI separately so the request data actually matches
+        # what is sent but the system message still shows up pretty in the UI.
```
alexmojaki: @dmontagu any suggestions here?

logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -0,0 +1,83 @@
...
+    if isinstance(chunk, ContentBlockStartEvent):
+        return chunk.content_block.text
+    if isinstance(chunk, ContentBlockDeltaEvent):
+        return chunk.delta.text
+    return None
+
+
+def on_response(response: ResponseT, span: LogfireSpan) -> ResponseT:
+    """Updates the span based on the type of response."""
+    if isinstance(response, (Message, ToolsBetaMessage)):  # pragma: no cover
```
alexmojaki:
```suggestion
    if isinstance(response, (Message, ToolsBetaMessage)):  # pragma: no branch
```

is this what you meant?

tests/otel_integrations/test_anthropic.py
```diff
@@ -0,0 +1,534 @@
...
+                        type='message_delta',
+                        usage=MessageDeltaUsage(output_tokens=55),
+                    ),
+                    MessageStopEvent(type='message_stop'),
+                ]
+                return httpx.Response(
+                    200, text=''.join(f'event: {chunk.type}\ndata: {chunk.model_dump_json()}\n\n' for chunk in chunks)
+                )
+        else:
+            if json_body['system'] == 'tool response':
```
alexmojaki: `elif`

tests/otel_integrations/test_anthropic.py
```diff
@@ -0,0 +1,534 @@
...
+                        type='message_start',
+                    ),
+                    ContentBlockStartEvent(
+                        content_block=TextBlock(text='', type='text'), index=0, type='content_block_start'
+                    ),
+                    ContentBlockDeltaEvent(
+                        delta=TextDelta(text='The answer', type='text_delta'), index=0, type='content_block_delta'
+                    ),
+                    ContentBlockDeltaEvent(
+                        delta=TextDelta(text=' is nine', type='text_delta'), index=0, type='content_block_delta'
```
alexmojaki: Please mimic https://github.com/pydantic/logfire/pull/173/files

tests/otel_integrations/test_anthropic.py
```diff
@@ -0,0 +1,534 @@
...
+                            )
+                        ],
+                        model='claude-3-haiku-20240307',
+                        role='assistant',
+                        type='message',
+                        usage=Usage(input_tokens=2, output_tokens=3),
+                    ).model_dump(mode='json'),
+                )
+    else:
+        assert request.url == 'https://api.anthropic.com/v1/complete', f'Unexpected URL: {request.url}'
```
alexmojaki: Please put this case first with an early return so the rest can be dedented

logfire/_internal/integrations/llm_providers/llm_provider.py
willbakst: We need the `llm` tag for things to show up nicely in the UI. Unfortunately `LLM` isn't caught. Options seem like:
1. Update UI to notice `LLM` tag as well (or just not have case sensitivity?).
2. Update UI to enable the UI if the `custom_scope_suffix` is Anthropic, in which case we can remove the `llm` tag from the SDK internals (but leave in the UI for external integrations such as Mirascope).

logfire/_internal/integrations/llm_providers/llm_provider.py
willbakst: So the `kwargs` passed in here is being passed in from the below `instrumented_llm_request_sync` (or async) calls, which is passing the `kwargs` it gets directly from the client call (so I would either need to pull the options out there or here). I figured pulling the options out once here made more sense than doing it in both the `sync` and `async` functions. Thoughts?

logfire/_internal/integrations/llm_providers/llm_provider.py
willbakst: Good call!

logfire/_internal/integrations/llm_providers/llm_provider.py
alexmojaki: Can't you still just call `_instrumentation_setup(**kwargs)` below while having `def _instrumentation_setup(options: Any, **kwargs: Any)`?

willbakst: If I do that, then I will have to add `options` back into `kwargs` before I return `kwargs` from setup because I need them for the call.

logfire/_internal/integrations/llm_providers/llm_provider.py
alexmojaki: I think this depends on the resolution of https://github.com/pydantic/logfire/pull/181/files/14193d540735896728724e5943459d83d08e01fe#r1603349085, because it seems like openai and anthropic will each have their own code paths in the frontend.

logfire/_internal/integrations/llm_providers/llm_provider.py
willbakst: For some reason this drops the second span containing the final streamed results, which breaks the streaming tests. Any ideas why this might be the case?

tests/otel_integrations/test_anthropic.py
willbakst: Done.

tests/otel_integrations/test_anthropic.py
willbakst: Done.

tests/otel_integrations/test_anthropic.py
willbakst: Done.

logfire/_internal/integrations/llm_providers/anthropic.py
willbakst: Update: makes most sense to just enable the UI to handle this properly so that the `request_data` remains accurate

logfire/_internal/integrations/llm_providers/llm_provider.py
willbakst: Couldn't figure out why this was failing for stream. I think it's clean enough (at least for now) to resolve this and we can maybe address this in any future integration PRs (e.g. Groq)?

logfire/_internal/integrations/llm_providers/llm_provider.py
willbakst: For now looks like best solution would be to support both `llm` and `LLM` so that internally only `LLM` is used but integrations (like Mirascope) won't break.

alexmojaki (approved):

alexmojaki:
Thank you!!!

PR #183, opened 2024-05-15
Vendor Version logic from packaging

Kludex:


alexmojaki:
I don't feel good about this. If we don't want to rely on other packages for version parsing then I'd rather find a different solution. For example:

```python
version = '2.5.3'
parts = tuple(map(int, version.split('.')[:2]))
print(parts < (2, 6))
print(parts < (2, 5))
```

Parsing a full version requires accounting for betas and stuff (which might be easy enough for known packages like pydantic) but in this case we don't even need the full version, just the first two parts.

Issue #182, opened 2024-05-15
TypeError on _exit_span() with Pydantic 1


jordan-dimov:
### Description

```
  File "/app/.venv/lib/python3.10/site-packages/logfire/_internal/integrations/fastapi.py", line 108, in patched_run_endpoint_function
    return await instrumentation.run_endpoint_function(
  File "/app/.venv/lib/python3.10/site-packages/logfire/_internal/integrations/fastapi.py", line 223, in run_endpoint_function
    with self.logfire_instance.span(
  File "/app/.venv/lib/python3.10/site-packages/logfire/_internal/main.py", line 1236, in __exit__
    _exit_span(self._span, exc_value)
  File "/app/.venv/lib/python3.10/site-packages/logfire/_internal/main.py", line 1350, in _exit_span
    _record_exception(span, exception, escaped=True)
  File "/app/.venv/lib/python3.10/site-packages/logfire/_internal/main.py", line 1378, in _record_exception
    err_json = exception.json(include_url=False)
  File "pydantic/error_wrappers.py", line 67, in pydantic.error_wrappers.ValidationError.json
TypeError: json() got an unexpected keyword argument 'include_url'
```

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="0.32.0"
platform="Linux-6.5.0-35-generic-x86_64-with-glibc2.38"
python="3.10.13 (main, Apr  8 2024, 12:55:10) [GCC 13.2.0]"
[related_packages]
requests="2.31.0"
pydantic="1.10.15"
fastapi="0.111.0"
openai="1.26.0"
protobuf="4.25.3"
rich="13.7.1"
tomli="2.0.1"
opentelemetry-api="1.24.0"
opentelemetry-exporter-otlp-proto-common="1.24.0"
opentelemetry-exporter-otlp-proto-http="1.24.0"
opentelemetry-instrumentation="0.45b0"
opentelemetry-instrumentation-aiohttp-client="0.45b0"
opentelemetry-instrumentation-asgi="0.45b0"
opentelemetry-instrumentation-asyncpg="0.45b0"
opentelemetry-instrumentation-fastapi="0.45b0"
opentelemetry-instrumentation-httpx="0.45b0"
opentelemetry-instrumentation-starlette="0.45b0"
opentelemetry-proto="1.24.0"
opentelemetry-sdk="1.24.0"
opentelemetry-semantic-conventions="0.45b0"
opentelemetry-util-http="0.45b0"
```

[...]

PR #184, opened 2024-05-15
Don't use `include_url` with Pydantic's V1 `ValidationError`

Kludex:
- Closes https://github.com/pydantic/logfire/issues/182

### DIFF
--- a/logfire/_internal/main.py
+++ b/logfire/_internal/main.py
@@ -1375,7 +1375,11 @@ def _record_exception(
     attributes = {**(attributes or {})}
     if ValidationError is not None and isinstance(exception, ValidationError):
         # insert a more detailed breakdown of pydantic errors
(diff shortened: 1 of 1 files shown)


logfire/_internal/main.py
```diff
@@ -1375,7 +1375,11 @@ def _record_exception(
     attributes = {**(attributes or {})}
     if ValidationError is not None and isinstance(exception, ValidationError):
         # insert a more detailed breakdown of pydantic errors
-        err_json = exception.json(include_url=False)
+        try:
+            err_json = exception.json(include_url=False)
+        except TypeError:  # pragma: no cover
+            # pydantic v1
```
Kludex, line 1381: We have 3 places already with this "pydantic v1" comment, FYI.

alexmojaki (approved):
btw i don't even think we use this data anywhere

Issue #135, opened 2024-05-06
Ignore stdlib logs from urllib3


alexmojaki:
https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1713885605381719

We can specifically ignore the message `https://logfire-api.pydantic.dev:443 "POST /v1/traces HTTP/1.1" 200 2` (and same for `/v1/metrics`) so that the user doesn't have to silence all of urllib3 to avoid this noise.

epicwhale:
+1 👍

epicwhale:
fyi: when I put this into my fastapi project, I realised the noise is VERY HIGH - almost takes over the entire real estate.. 
![image](https://github.com/pydantic/logfire/assets/670655/b7325d9e-4be8-4b32-a91f-a9d6e61163b6)

adriangb:
Shouldn't the urllib3 instrumentation respect the contextvars to not infinite loop instrumentation? Are we maybe not propagating them correctly?

alexmojaki:
This is urllib3 using stdlib logging, i.e. this fix would be in the stdlib logging instrumentation. The urllib3 instrumentation is not involved.

But you do bring up a good point. Looking more closely, we use the string `suppress_instrumentation`, whereas OTEL mostly uses `_SUPPRESS_INSTRUMENTATION_KEY = create_key("suppress_instrumentation")` which is different.

adriangb:
OTEL currently has a haphazard mix of suppress instrumentation keys. We should probably just set all of them.

alexmojaki:
Well in this case it's about reading the correct key. OTEL conveniently sets it in the right places in BatchSpanProcessor and SimpleSpanProcessor.

alexmojaki:
Fixed by https://github.com/pydantic/logfire/pull/197

PR #197, opened 2024-05-20
Support all OTEL keys for instrument suppression

jlondonobo:
Add support for all OTEL keys for instrument suppression.

This PR adds support for the following keys:

- `"suppress_instrumentation"`
- `_SUPPRESS_INSTRUMENTATION_KEY`


logfire/_internal/utils.py
```diff
@@ -184,3 +190,23 @@ def get_version(version: str) -> Version:
...
+
+def is_instrumentation_suppressed() -> bool:
+    return bool(
+        context.get_value(SUPPRESS_INSTRUMENTATION_CONTEXT_KEY) or context.get_value(OTEL_SUPPRESS_INSTRUMENTATION_KEY)
+    )
+
+
+@contextmanager
+def suppress_instrumentation():
+    """Context manager to suppress instrumentation from the LLM client."""
```
adriangb: Why not use (and fix / expand if necessary) https://github.com/open-telemetry/opentelemetry-python-contrib/blob/f8758c6902ed725864ff677f739829aa6bce2078/opentelemetry-instrumentation/src/opentelemetry/instrumentation/utils.py#L188-L192?

alexmojaki: Some parts of OTEL still use the plain `'suppress_instrumentation'` key, so we need to check/set that too, as you said in https://github.com/pydantic/logfire/issues/135#issuecomment-2098318558.

We could maybe upstream a fix to OTEL to consistently use one key, but we need this to also work for older OTEL versions.

adriangb: We should fix as much of OTEL as we can to make sure it both sets and respects all keys. But yes the unfortunate reality is that both us and OTEL will have to set and check for any keys ever used.

image

alexmojaki:
In https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1714658053546979 a user essentially requested the `suppress_instrumentation` function, so I've made it public here.

This PR fixes https://github.com/pydantic/logfire/issues/135 in the sense that those particular urllib3 logs don't get sent to logfire. They are now logged by the fallback handler of LogfireLoggingHandler which logs to stderr by default. After this, I want to add a snippet like the following to the docs showing how to filter those logs if the noise in the console is still annoying:

```python
import logging
import time
from logging import DEBUG, basicConfig, getLogger

from logfire.integrations.logging import LogfireLoggingHandler

logfire_handler = LogfireLoggingHandler()

# Don't write to stderr the urllib3 debug logs created while exporting traces
urllib3_filter = logging.Filter('urllib3')
logfire_handler.fallback.addFilter(lambda record: not urllib3_filter.filter(record))

basicConfig(handlers=[logfire_handler], level=DEBUG)

logger = getLogger(__name__)

for _ in range(1000):
    logger.info('log')
    time.sleep(0.6)
```

alexmojaki:
Thanks @jlondonobo for your help, your changes were pretty much exactly what was needed!

logfire/_internal/utils.py
adriangb: @alexmojaki before you merge this could you address this comment? I'd like to understand why we can't just use the OTEL function or if we can't because it is missing some functionality I'd like to know that there is a plan to upstream it and then replace our function with theirs or something like that.

alexmojaki: > But yes the unfortunate reality is that both us and OTEL will have to set and check for any keys ever used.

I thought we were already in agreement. Similarly, the OTEL util function is somewhat recent, I'd like to not reduce compatibility with older OTEL versions. We especially can't rely on a future version.

alexmojaki: Opened https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2543

adriangb: Great so to make sure I understand we agree that OTEL needs to make some changes that will make future versions work correctly because we do not want to bump our minimum version we are going to do the fixes by exposing our own utility functions.

Will our utility functions be able to to just point to the open telemetry ones in the future?

alexmojaki: > Will our utility functions be able to to just point to the open telemetry ones in the future?

Sure, although I think it'll be some time before we decide we don't need a fallback for older OTEL versions.

PR #198, opened 2024-05-20
Add `logfire.instrument_httpx()`

tlpinney:
Mostly copied from asyncpg.


alexmojaki:
Thanks again @tlpinney !

Issue #193, opened 2024-05-19
`logfire.instrument_*()` methods


alexmojaki:
e.g. `logfire.instrument_sqlalchemy()` instead of `from opentelemetry.instrumentation.sqlalchemy import SQLAlchemyInstrumentor; SQLAlchemyInstrumentor().instrument()`. Same for all integrations documented in https://docs.pydantic.dev/logfire/integrations/#opentelemetry-integrations.

alexmojaki:
Remaining methods to add:

- [ ] `logfire.instrument_flask()`
- [ ] `logfire.instrument_starlette()`
- [ ] `logfire.instrument_aiohttp()`
- [ ] `logfire.instrument_sqlalchemy()`
- [ ] `logfire.instrument_pymongo()`
- [ ] `logfire.instrument_redis()`
- [ ] `logfire.instrument_celery()`
- [ ] `logfire.instrument_asgi_middleware()`
- [ ] `logfire.instrument_wsgi_middleware()`
- [ ] `logfire.instrument_logging_handler()`
- [ ] `logfire.instrument_loguru_handler()`
- [ ] `logfire.instrument_structlog_processor()`

I'm not sure about the names of the last 5. It is important that they're methods of logfire objects rather than classes imported from logfire. It would be nice for everything to start with `instrument_` but `instrument_logging_handler` sounds like it's instrumenting a logging handler, rather than returning a handler that 'instruments logging'.

@Kludex I'm going to start working on the first 6, please do the other half.

alexmojaki:
@Kludex I think we should have `logfire.instrument_loguru()` which calls `logger.add(**logfire.loguru_handler())`. That makes the naming less weird and should be easier for users, avoiding problems like https://github.com/pydantic/logfire/issues/228. `logfire.loguru_handler()` can be kept for compatibility but not recommended.

Kludex:
Is this done? @alexmojaki

alexmojaki:
No because I did half and I asked you to do the other half.

Kludex:
About the ASGI and WSGI, I'm not sure if it makes sense to create those methods.

Maybe wrapping them, and providing a `logfire.InstrumentASGIMiddleware` and `logfire.InstrumentWSGIMiddleware` makes more sense. wdyt @alexmojaki ?

alexmojaki:
They need to be methods so that you can use them with different logfire objects that have different configs.

[...]

PR #207, opened 2024-05-22
Add 6 instrument methods

alexmojaki:
Part of https://github.com/pydantic/logfire/issues/193


alexmojaki:
Need to remember to add `self._warn_if_not_initialized_for_instrumentation()` to these methods after merging https://github.com/pydantic/logfire/pull/219/

logfire/_internal/integrations/starlette.py
```diff
@@ -0,0 +1,12 @@
+from typing import Any
+
+from opentelemetry.instrumentation.starlette import StarletteInstrumentor
+from starlette.applications import Starlette
+
+
+def instrument_starlette(app: Starlette, **kwargs: Any):
```
Kludex, line 7: As a user, I would have to go to the OTel documentation to see which ones are the kwargs. I think we should correctly type hint instead of using the kwargs.

If the argument for doing it like this is because they may break our API, we can just do the `if TYPE_CHECKING` trick. It just provides a far better experience.

Kludex (approved):

logfire/_internal/integrations/starlette.py
alexmojaki: Yes, we should do that eventually.

PR #214, opened 2024-05-27
Retry connection errors

alexmojaki:
After some investigation, I've found that neither OTEL nor `requests` does any retrying by default when there's a connection or timeout error while making a request.

OTEL does retry on certain status codes in a way that looks sensible so I'm leaving that case as is, although it's not the kind of problem we usually see.

It's possible to configure retries with `requests` but OTEL doesn't do so and it didn't seem easier than just coding the retries myself.

### DIFF
--- a/logfire/_internal/exporters/otlp.py
+++ b/logfire/_internal/exporters/otlp.py
@@ -53,7 +53,8 @@ class OTLPExporterHttpSession(Session):
                     continue
                 raise
             return response
-        raise RuntimeError('Unreachable code')
+
+        raise RuntimeError('Unreachable code')  # for pyright
 
     def _check_body_size(self, size: int) -> None:
         if size > self.max_body_size:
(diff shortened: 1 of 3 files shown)


logfire/_internal/exporters/otlp.py
```diff
@@ -38,7 +41,20 @@ def gen() -> Iterable[bytes]:
                         yield chunk
 
                 request.body = gen()  # type: ignore
-        return super().send(request, **kwargs)
+
+        max_attempts = 7
```
adriangb, line 45: Should this be configurable in any way?

adriangb (approved):

logfire/_internal/exporters/otlp.py
alexmojaki: I don't think so. Making it configurable would be painful with little benefit, and OTEL's retrying of status codes isn't configurable.

Issue #194, opened 2024-05-19
Don't configure automatically


alexmojaki:
As announced in https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1716143149301069, the log/span/instrument methods shouldn't implicitly initialize config. This requires a few steps:

1. Make sure that all docs explicitly call `logfire.configure()` before anything else.
2. https://github.com/pydantic/logfire/issues/193 so that `instument_*` methods can emit a warning if called before configuring. Currently the problem is that users may call e.g. `SQLAlchemyInstrumentor().instrument()` and forget to configure and we can't do anything about that.
3. Emit warnings when calling methods before configure.
4. Stop emitting spans (but keep the warning) when calling methods before configure.

bllchmbrs:
Please!

PR #219, opened 2024-05-28
Don't configure automatically

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/194

Step 2 there isn't done but I realize now that it's not really a blocker, and there's strong demand for this change.

The implementation here is simple but not ideal. In particular `logfire.span` etc will still do a lot of processing which will slow things down and maybe even raise errors, it's just that the actual OTEL span will be a no-op. But making `logfire.span` a proper no-op is nontrivial since it still needs to return a context manager with methods, and that work overlaps with https://github.com/pydantic/logfire/pull/146. Optimizing this can be done later, and the same strategy can also be applied to checking if the span/log will be sampled out or if instrumentation is currently suppressed.

`logfire.instrument_<module>` will also apply instrumentation including patching, but (1) that's necessary to allow calling `configure` later and having it all just work and (2) that's how OTEL works anyway.

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -74,6 +74,7 @@ from .exporters.remove_pending import RemovePendingSpansExporter
 from .integrations.executors import instrument_executors
 from .metrics import ProxyMeterProvider, configure_metrics
 from .scrubbing import Scrubber, ScrubCallback
+from .stack_info import get_user_frame_and_stacklevel
 from .tracer import PendingSpanProcessor, ProxyTracerProvider
 from .utils import UnexpectedResponse, ensure_data_dir_exists, get_version, read_toml_file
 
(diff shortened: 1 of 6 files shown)


logfire/_internal/config.py
```diff
@@ -1228,3 +1238,7 @@ def sanitize_project_name(name: str) -> str:
 
 def default_project_name():
     return sanitize_project_name(os.path.basename(os.getcwd()))
+
+
+class LogfireNotConfiguredWarning(UserWarning):
```
samuelcolvin, line 1238: should this be public?

samuelcolvin (approved):
Generally looks good, I just want to confirm that none of the behaviour actually changes with this PR?

alexmojaki:
> I just want to confirm that none of the behaviour actually changes with this PR?

No, the behaviour does actually change. We can have a deprecation period (1 week?) if you want but I'm not sure it's worth it.

logfire/_internal/config.py
alexmojaki: Currently we have a few warnings but none are public. I think that if we decide to make them public we should make a new `logfire.warnings` module for them in a new PR.

PR #231, opened 2024-06-03
Improve error raised when opentelemetry.instrumentation.django is not installed

deepakdinesh1123:
I was looking through the logfire integrations with frameworks and libraries and found that a `ModuleNotFoundError` was not being raised in the django integration so I have added it here.

### DIFF
--- a/logfire/_internal/integrations/django.py
+++ b/logfire/_internal/integrations/django.py
@@ -1,6 +1,13 @@
 from typing import Any
 
-from opentelemetry.instrumentation.django import DjangoInstrumentor
+try:
+    from opentelemetry.instrumentation.django import DjangoInstrumentor
+except ModuleNotFoundError:
+    raise RuntimeError(
+        'The `logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.\n'
+        'You can install this with:\n'
+        "    pip install 'logfire[django]'"
+    )
 
 
 def instrument_django(**kwargs: Any):
--- a/logfire/_internal/integrations/django.py
+++ b/logfire/_internal/integrations/django.py
@@ -4,7 +4,7 @@ try:
     from opentelemetry.instrumentation.django import DjangoInstrumentor
 except ModuleNotFoundError:
     raise RuntimeError(
-        'The `logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.\n'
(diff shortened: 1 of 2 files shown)


logfire/_internal/integrations/django.py
```diff
@@ -1,6 +1,13 @@
 from typing import Any
 
-from opentelemetry.instrumentation.django import DjangoInstrumentor
+try:
+    from opentelemetry.instrumentation.django import DjangoInstrumentor
+except ModuleNotFoundError:
+    raise RuntimeError(
+        'The `logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.\n'
```
alexmojaki:
```suggestion
        'logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.\n'
```

tests/otel_integrations/test_django.py
```diff
@@ -115,3 +122,14 @@ def test_no_matching_route(client: Client, exporter: TestExporter):
...
         ]
     )
+
+
+def test_missing_opentelemetry_dependency() -> None:
+    with mock.patch.dict('sys.modules', {'opentelemetry.instrumentation.django': None}):
+        with pytest.raises(RuntimeError) as exc_info:
+            importlib.reload(logfire._internal.integrations.django)
+        assert str(exc_info.value) == snapshot("""\
+The `logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.
```
alexmojaki:
```suggestion
`logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.
```

alexmojaki:
Thanks! We should probably make something general like this for the other instrument_ methods.

logfire/_internal/integrations/django.py
```diff
@@ -1,6 +1,13 @@
 from typing import Any
 
-from opentelemetry.instrumentation.django import DjangoInstrumentor
+try:
+    from opentelemetry.instrumentation.django import DjangoInstrumentor
+except ModuleNotFoundError:
+    raise RuntimeError(
+        'logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.\n'
```
alexmojaki:
```suggestion
        '`logfire.instrument_django()` requires the `opentelemetry-instrumentation-django` package.\n'
```

Issue #99, opened 2024-05-02
Ability to configure additional span processors (e.g. to export to another sink)


adriangb:
### Description

https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1714686467955179

> feature request! being able to add another span processor while still including the default logfire one. what I’m doing right now is hacking it by not providing the second processor up front, then doing:

```python
from logfire._internal.config import GLOBAL_CONFIG
from logfire._internal.exporters.processor_wrapper import SpanProcessorWrapper

span_processor = SpanProcessorWrapper(SentrySpanProcessor(), GLOBAL_CONFIG.scrubber)
GLOBAL_CONFIG.processors = [span_processor]
```

PR #233, opened 2024-06-03
Keep the default span processor when `processors` is passed to `logfire.configure()`

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/99

This is a breaking change, but since we'll probably do this eventually anyway it's best to do it now. I can't see a reason for the current behaviour.


adriangb:
I assume if you want to get rid of the default processor you're supposed to pass `send_to_logfire=False`?

adriangb (approved):
OK, as long as that’s documented properly. Since you’re making a break I’d suggest changing the name of the parameter to `additional_span_processors` or similar.

alexmojaki:
OK I've renamed, and `processors` will now raise an error telling you to use `additional_span_processors`.

adriangb (approved):

adriangb:
Looks great!

PR #236, opened 2024-06-04
Raise error if use `logfire clean` on home directory

Kludex:
https://linear.app/pydantic/issue/PYD-875/logfire-clean-documentation-is-wrong

### DIFF
--- a/logfire/_internal/cli.py
+++ b/logfire/_internal/cli.py
@@ -77,7 +77,6 @@ def parse_whoami(args: argparse.Namespace) -> None:
 
 def parse_clean(args: argparse.Namespace) -> None:
     """Remove the contents of the Logfire data directory."""
-    # Check if running directory is HOME_LOGFIRE.
     if Path.cwd() == Path.home():
         sys.stderr.write('The clean command cannot be run in the home directory. ')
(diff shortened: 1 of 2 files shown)


logfire/_internal/cli.py
```diff
@@ -77,6 +77,12 @@ def parse_whoami(args: argparse.Namespace) -> None:
 
 def parse_clean(args: argparse.Namespace) -> None:
     """Remove the contents of the Logfire data directory."""
+    # Check if running directory is HOME_LOGFIRE.
```
Kludex:
```suggestion
```

logfire/_internal/cli.py
```diff
@@ -77,6 +77,11 @@ def parse_whoami(args: argparse.Namespace) -> None:
 
 def parse_clean(args: argparse.Namespace) -> None:
     """Remove the contents of the Logfire data directory."""
+    if Path.cwd() == Path.home():
+        sys.stderr.write('The clean command cannot be run in the home directory. ')
+        sys.stderr.write('Please run the command from a different directory.\n')
```
alexmojaki, line 82: This doesn't make sense to me. How should the user clean out the files if they happen to be in the home dir? Can't we just delete the specific files?

PR #241, opened 2024-06-04
Log request (e.g. connection) errors when exporting spans/metrics more quietly

alexmojaki:


alexmojaki:
I'm changing how this works in the system for retrying from disk.

alexmojaki:
Closing in favour of https://github.com/pydantic/logfire/pull/252 (which actually ended up containing most of this)

Issue #237, opened 2024-06-04
Add changelog 


Kludex:
### Description

We should a changelog - maybe the same logic as pydantic.

PR #245, opened 2024-06-05
Adding changelog and CI check for changes on version bump

sydney-runkle:
Fix https://github.com/pydantic/logfire/issues/237


Kludex:
Can we add the changelog to the docs, like this: https://github.com/encode/uvicorn/blob/master/docs/release-notes.md ?

Or just have the changelog on the docs itself?

sydney-runkle:
Please review

.github/workflows/check_changelog.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          python -m pip install toml
+          for file in pyproject.toml HISTORY.md; do
+              base_name=$(basename $file .${file##*.})
+              if git diff --name-only HEAD^ | grep -q "^$file$"; then
+                echo "${base_name}_changed=true" >> $GITHUB_ENV
+              else
+                echo "${base_name}_changed=false" >> $GITHUB_ENV
+              fi
+          done
```
Kludex: You don't need to install toml on python 3.12 to read TOML files, you can use https://docs.python.org/3/library/tomllib.html, which is available from 3.11.

Kludex (changes requested):

Kludex:
Feel free to merge it after removing the toml dependency 👍

PR #246, opened 2024-06-05
Update dependency versions, ensure that CI runs tests with latest working versions

alexmojaki:


Kludex (approved):

adriangb:
Alas I'm afraid one huge piece of pain we're going to encounter for every integration that we own is that there's no "correct" version to test CI against. We probably have to run at least against the oldest and newest supported versions, right?

alexmojaki:
Testing against older versions would be good too. Testing against the newest version seems like the most important thing.

PR #249, opened 2024-06-06
Update Anthropic to use tools that are no longer in beta

willbakst:
Anthropic released tools not in beta, so this PR upgrades to the most recent version that has GA tools.


requirements-dev.lock
```diff
@@ -94,7 +94,11 @@ gitdb==4.0.11
 gitpython==3.1.43
 googleapis-common-protos==1.63.1
     # via opentelemetry-exporter-otlp-proto-http
+<<<<<<< HEAD
+griffe==0.44.0
+=======
 griffe==0.45.2
+>>>>>>> main
```
alexmojaki:
```suggestion
griffe==0.45.2
```

alexmojaki (approved):
Thanks!

PR #252, opened 2024-06-07
Improved handling of request errors when exporting

alexmojaki:
Rather than relying on a fallback file which the user must manually pass to `logfire backfill` and likely can't even access, write to temporary files and automatically retry exporting on a separate thread.

Also prevent request errors from bubbling up to OTEL where it will log a big ugly traceback.


logfire/_internal/exporters/fallback.py
```diff
@@ -18,9 +19,15 @@ def __init__(
...
-            raise
+        except Exception as e:
+            if isinstance(e, requests.exceptions.RequestException):
+                # Silence the exception so that OTEL doesn't log a huge traceback.
+                # Rely on OTLPExporterHttpSession to log this kind of error periodically.
+                return SpanExportResult.FAILURE
+            else:
+                # Only write to fallback file if this isn't already being retried by OTLPExporterHttpSession.
+                self.fallback.export(spans)
+                raise
```
adriangb, line 30: This code looks good but I don't think it should live in FallbackExporter. It would be better as `SilencingProcessor` or similar.

adriangb (approved):
Overall this looks good.

I do feel like we're cramming stuff into `Session` that has no business being there. I'd be interested in us writing a new `Processor` to replace `BatchSpanExporter that handles these concerns and does it more cleanly and efficiently than distributing logic over `Session`, `FallbackExporter`, etc.

logfire/_internal/exporters/fallback.py
alexmojaki: I disagree because we're specifically preventing the exception from resulting in writing to a second fallback file. The extra exporter would be coupled to FallbackExporter and need to be inside.

PR #261, opened 2024-06-12
[PYD-617] Remove default for modules parameter of `install_auto_tracing`

alexmojaki:
Currently by default:

- Modules under the root of the current module are traced if no modules are specified
- An error is raised if modules to be traced have already been imported

This means that providing no arguments will always raise an error about wanting to trace the current module which has obviously already been imported. Either `modules` or `check_imported_modules` must be specified. This is a bit weird.

This PR makes things less weird by requiring the user to specify `modules`. This seems like a case where explicit really is better than implicit.

### DIFF
--- a/logfire/_internal/auto_trace/__init__.py
+++ b/logfire/_internal/auto_trace/__init__.py
@@ -23,10 +23,10 @@ def install_auto_tracing(
 
     See `Logfire.install_auto_tracing` for more information.
     """
-    if isinstance(modules, Sequence):  # pragma: no branch
+    if isinstance(modules, Sequence):
         modules = modules_func_from_sequence(modules)
 
-    if not callable(modules):  # pragma: no cover
+    if not callable(modules):
         raise TypeError('modules must be a list of strings or a callable')
 
     if check_imported_modules not in ('error', 'warn', 'ignore'):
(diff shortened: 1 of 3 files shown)


logfire/_internal/auto_trace/__init__.py
```diff
@@ -24,13 +23,7 @@ def install_auto_tracing(
...
-        frame = inspect.stack()[2]
-        module = inspect.getmodule(frame[0])
-        if module is None:  # pragma: no cover
-            raise KeyError('module not found')
-        modules = modules_func_from_sequence([module.__name__.split('.')[0]])
-    elif isinstance(modules, Sequence):  # pragma: no branch
+    if isinstance(modules, Sequence):  # pragma: no branch
         modules = modules_func_from_sequence(modules)
 
     if not callable(modules):  # pragma: no cover
```
davidhewitt: Maybe `if` / `elif` here? This seems like a prime opportunity to add tests for callable and also wrong-type inputs.

```suggestion
    if isinstance(modules, Sequence):
        modules = modules_func_from_sequence(modules)

    elif not callable(modules):
```

davidhewitt (approved):
LGTM with a suggestion to add more tests for related cases :)

logfire/_internal/auto_trace/__init__.py
alexmojaki: Added a test for the wrong type. Testing a callable feels a bit redundant. Not using elif to make it clear that the sequence gets converted to a callable, so `modules` is always eventually a callable.

PR #263, opened 2024-06-14
Tail sampling

alexmojaki:
There's some details which I'm unsure about such as defaults, but I'd like to get this through so we can test it out for ourselves. It could be quite useful in our backend, especially in places where we're using random trace sampling. Documentation and maybe some tweaking (including env var configuration) can come after we've used it.

Usage in a nutshell:

```python
import logfire

logfire.configure(
    tail_sampling=logfire.TailSamplingOptions(
        # These are the defaults of TailSamplingOptions, tail_sampling is None by default.
        level='notice',  # include traces with at least one span/log at this level or higher
        duration=1.0,  # include traces with at least this duration
    ),
    # Also include 10% of traces randomly from the beginning, before checking the other conditions.
    trace_sample_rate=0.1,
)
```

Improper usage will eat up memory as the spans are buffered.


logfire/_internal/exporters/processor_wrapper.py
```diff
@@ -47,13 +48,7 @@ def on_end(self, span: ReadableSpan) -> None:
         _tweak_http_spans(span_dict)
         self.scrubber.scrub_span(span_dict)
         span = ReadableSpan(**span_dict)
-        self.processor.on_end(span)
-
-    def shutdown(self) -> None:
-        self.processor.shutdown()
-
-    def force_flush(self, timeout_millis: int = 30000) -> bool:
```
samuelcolvin, line 55: is this the method we were using to get logfire to work with AWS lambda? If so, I guess we need it.

logfire/_internal/config.py
```diff
@@ -332,6 +336,12 @@ class _LogfireConfigData:
     scrubbing_callback: ScrubCallback | None
     """A function that is called for each match found by the scrubber."""
 
+    inspect_arguments: bool
+    """Whether to enable f-string magic"""
+
+    tail_sampling: TailSamplingOptions | None
+    """Tail sampling options"""
```
adriangb, line 343: Could just be a typeddict?

logfire/_internal/config.py
```diff
@@ -592,8 +613,15 @@ def _initialize(self) -> ProxyTracerProvider:
             # Both recommend generating a UUID.
             resource = Resource({ResourceAttributes.SERVICE_INSTANCE_ID: uuid4().hex}).merge(resource)
 
+            # Avoid using the usual sampler if we're using tail-based sampling.
+            # The TailSamplingProcessor will handle the random sampling part as well.
+            sampler = (
+                ParentBasedTraceIdRatio(self.trace_sample_rate)
+                if self.trace_sample_rate < 1 and self.tail_sampling is None
+                else None
+            )
```
adriangb, line 622: How do we document / tell users they can't combine these? Is there a world where I want to head sample down to 10% (to reduce overhead in the SDK) and then tail sample down to 1%? There's still advantages to head sampling.

logfire/_internal/exporters/tail_sampling.py
```diff
@@ -0,0 +1,145 @@
...
+
+    started: list[tuple[Span, context.Context | None]]
+    ended: list[ReadableSpan]
+
+    @cached_property
+    def first_span(self) -> Span:
+        return self.started[0][0]
+
+
+class TailSamplingProcessor(WrapperSpanProcessor):
```
adriangb, line 47: I'm really starting to feel like we should stop adding layers / wrapping processors like this. I would prefer to have a single processor that handles everything (sampling, batching, retries, etc.). I feel like it could be optimized more and would be easier to understand. It can still have pluggable bits just more explicit and less abstract.

logfire/_internal/exporters/tail_sampling.py
```diff
@@ -0,0 +1,145 @@
...
+    """Passes spans to the wrapped processor if any span in a trace meets the sampling criteria."""
+
+    def __init__(self, processor: SpanProcessor, options: TailSamplingOptions, random_rate: float) -> None:
+        super().__init__(processor)
+        self.duration: float = (
+            float('inf') if options.duration is None else options.duration * ONE_SECOND_IN_NANOSECONDS
+        )
```
adriangb, line 54: I'd like these durations to have the unit in their name. So `self.duration` -> `self.duration_ns` and `TailsamplingOptions.duration` -> `TailSamplingOptions.duration_sec` or something like that.

logfire/_internal/exporters/processor_wrapper.py
alexmojaki: It's moved to the base class.

adriangb (commented):
Welp I guess you merged this already but please address comments

logfire/_internal/config.py
alexmojaki: I tried that, I've also wanted `ConsoleOptions` and `PydanticPlugin` to be typed dicts, but they seemed worse. They can't define defaults in the class, and passing in a plain dict is actually not that user friendly.

adriangb: Makes sense

logfire/_internal/config.py
alexmojaki: They can combine them, see the PR body or the new `test_random_sampling`. 

> Is there a world where I want to head sample down to 10% (to reduce overhead in the SDK) and then tail sample down to 1%?

I don't know what this means. Tail sample down to a percentage?

It sounds like you want to be able to discard most spans up front randomly regardless of whether tail-sampling would include them, so that a span only gets through if it's 'notable' AND 'lucky'.

adriangb: Yes exactly

logfire/_internal/exporters/tail_sampling.py
alexmojaki: This layer wraps all processors, including the console and user-defined processors.

adriangb: I get that. But I feel like we should just make `LogfireSpanProcessor` which does all of those things in one place. In particular we avoid double buffering.

PR #268, opened 2024-06-18
Add `logfire-api`

Kludex:
This PR creates the `logfire-api` package.

We'll have to remove rye - or it needs to work with editable mode, otherwise developers need to remember to remove the `.venv` and run `rye sync --no-lock` every time they perform changes.

The approach here is simple - the `logfire-api/logfire_api/__init__.py` contains all the runtime logic, and the type hints are generated from the `make generate-stubs` command - we have a lot of `pyi` files.

There are two tests:
1. One that makes sure both `logfire` and `logfire_api` have the same objects in the `__init__.py`.
2. Another that makes sure we have the `__init__.pyi` type compliant.


logfire-api/src/logfire_api/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+            sample_rate: float | None = None,
+            tags: Sequence[str] = (),
+            console_log: bool = True,
+            otel_scope: str = 'logfire',
+        ) -> None: ...
+
+else:
+
+    class Logfire:
+        def __init__(self, *args: Any, **kwargs: Any) -> None:
```
samuelcolvin:
```suggestion
        def __new__(self, *args: Any, **kwargs: Any) -> real_logfire.Logfire | Logfire:
```

logfire-api/src/logfire_api/__init__.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Sequence
+
+try:
+    import logfire
+
+    logfire_installed = True
+except ImportError:
+    logfire_installed = False
```
samuelcolvin:
```suggestion
try:
    import logfire as real_logfire
except ImportError:
    real_logfire = None
```

logfire-api/src/logfire_api/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    class Logfire:
+        def __init__(
+            self,
+            *,
+            config: LogfireConfig = GLOBAL_CONFIG,
+            sample_rate: float | None = None,
+            tags: Sequence[str] = (),
+            console_log: bool = True,
+            otel_scope: str = 'logfire',
+        ) -> None: ...
```
samuelcolvin: we need all the methods here.

logfire-api/pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[project]
+name = "logfire-api"
+version = "0.1.0"
+description = "Add your description here"
+authors = [
+    { name = "Marcelo Trylesinski", email = "marcelotryle@gmail.com" }
```
samuelcolvin: best to use pydantic.dev email if you're okay with that, also maybe add me or `engineeering@pydantic.dev` in case people want to get in touch.

logfire-api/pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[project]
+name = "logfire-api"
+version = "0.1.0"
+description = "Add your description here"
```
samuelcolvin:
```suggestion
description = "Shim for the logfire SDK Python API which does nothing unless logfire is installed"
```
Or something?

logfire-api/README.md
```diff
@@ -0,0 +1,3 @@
+# logfire-api
+
+Describe your project here.
```
samuelcolvin:
```suggestion
Shim for the logfire SDK Python API which does nothing unless logfire is installed.

This package is designed to be used by packages that want to provide opt-in integration with [Logfire](https://github.com/pydantic/logfire).

The package provides a clone of the Python API exposed by the `logfire` package which does nothing if the `logfire` package is not installed, but makes real calls when it is.
```

logfire-api/src/logfire_api/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+else:
+
+    class Logfire:
+        def __init__(self, *args: Any, **kwargs: Any) -> None:
+            if logfire_installed:
+                return logfire.Logfire(*args, **kwargs)
```
samuelcolvin: You need to add methods like `span()` which return something functionally consistent with `logfire.span()`.

And also.

```py
def __getattr__(self, attr):
    return MagicMock()
```

We perhaps we shoud also implement all the know methods with a signature of `(*args, **kwargs) -> None`, then leave this as a fallback.

logfire-api/src/logfire_api/__init__.py
Kludex: This makes the type checker think that `real_logfire` is `Logfire | None`.

The current logic doesn't.

logfire-api/src/logfire_api/__init__.py
adriangb: Returning a MagicMock outside of tests seems bad IMO. Can we follow the opentelemetry-api way of doing this?

Kludex:
@alexmojaki do you know why those tests are failing? 🤔 There are different tests failing depending on the Python version...

.github/workflows/main.yml
```diff
@@ -144,5 +144,14 @@ jobs:
...
         uses: pypa/gh-action-pypi-publish@release/v1
+
+      - name: Build logfire-api
+        run: rye build
+        working-directory: logfire-api/
+
+      - name: Publish logfire-api to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: logfire-api/
```
Kludex, line 160: I've created the publisher... Not sure if this works.

Screenshot 2024-06-21 at 17 55 49

alexmojaki:
I don't know why. How new/consistent are they?

samuelcolvin (approved):
This looks awesome.

Thank you!

adriangb:
It would be nice to get a description of the overall approach here. From looking at the files it seems like the idea is to generate stubs / API files from _all_ of logfire. IMO this is unnecessary. The use case for a `logfire-api` is that libraries can instrument with logfire without depending on requests, etc. I think all that's needed is `logfire.span(...)`, `logfire.info(...)` and maybe the things those return. It can just be something like this:

```python
try:
    import logfire
except ImportError:
   logfire = None

def info(...) -> None:
    if logfire:
       logfire.info(...)
 
def span(...) -> ?:
    if logfire:
        return logfire.span(...)
    return ?
```

I leave `?` as "not sure what we should return for `LogfireSpan`. Maybe we implement a `LogfireSpanApi` and have `LogfireSpan` inherit from it? I also think we should audit what methods/attributes are on `LogfireSpan`. E.g. if it gives you access to the underlaying `opentelemetry.sdk.Span` it maybe should be switched to `opentelemetry.Span`.

Kludex:
> It seems like the idea is to generate stubs / API files from all of logfire. IMO this is unnecessary.

Since it's automated, I don't see the issue.

> I think all that's needed is logfire.span(...), logfire.info(...) and maybe the things those return.

This was already discussed. Also, it was the first question before I started implementing it.

> It can just be something like this:

If it were few objects, yes... But for every object on `logfire/__init__.py`, the current logic is easy to maintain because you don't need to add the conditionals on every `logfire_api` object.

adriangb:
My point is that libraries instrumenting with logfire don't need e.g. `logfire.instrument_fastapi()`.

alexmojaki:
> My point is that libraries instrumenting with logfire don't need e.g. `logfire.instrument_fastapi()`.

See https://github.com/search?q=repo%3AMirascope%2Fmirascope%20instrument_openai&type=code

samuelcolvin:
Let's go ahead with this, I don't really see the point of putting effort into limiting the type hint coverage. 

Let's go ahead with this.

@adriangb we have to make some decisions on calls without the whole team present, if you don't join calls you have to accept some decisions are made without you.

tests/test_logfire_api.py
```diff
@@ -0,0 +1,220 @@
...
+
+    assert hasattr(logfire_api, 'VERSION')
+    logfire__all__.remove('VERSION')
+
+    assert hasattr(logfire_api, 'LevelName')
+    logfire__all__.remove('LevelName')
+
+    assert hasattr(logfire_api, 'LogfireSpan')
+    with logfire_api.span('test span'):
+        ...
```
alexmojaki:
```suggestion
    with logfire_api.span('test span') as span:
        assert isinstance(span, logfire_api.LogfireSpan)
        span.set_attribute('foo', 'bar')
```

logfire-api/logfire_api/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+except ImportError:
+    if not TYPE_CHECKING:  # pragma: no branch
+        LevelName = Literal['trace', 'debug', 'info', 'notice', 'warn', 'warning', 'error', 'fatal']
+        VERSION = '0.0.0'
+        METRICS_PREFERRED_TEMPORALITY = {}
+
+        def configure(*args, **kwargs): ...
+
+        class LogfireSpan:
```
alexmojaki, line 22: Please mimic this: https://github.com/pydantic/logfire/blob/770780d4e0f5610bd8cbd596029d0cb1e23f25ae/logfire/_internal/main.py#L1695-L1714

logfire-api/logfire_api/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+            def install_auto_tracing(self, *args, **kwargs) -> None: ...
+
+            def instrument(self, *args, **kwargs):
+                def decorator(func):
+                    return func
+
+                return decorator
+
+            def instrument_fastapi(self, *args, **kwargs) -> None: ...
```
alexmojaki: This, instrument_openai, and maybe others should be a context manager. Might as well make them all return nullcontext.

pyproject.toml
```diff
@@ -96,6 +96,7 @@ dev-dependencies = [
     "requests",
     "sqlalchemy",
     "pandas",
+    "numpy<2.0.0",
```
alexmojaki: This is temporary, right?

Kludex: Yeah, I just removed this change on my last commit 👍

Kludex:
@alexmojaki Anything else, or can I merge?

alexmojaki (approved):
Thanks!

Issue #247, opened 2024-06-06
`logfire.configure()` makes a blocking HTTP request


alexmojaki:
`logfire.configure(token='...')` makes a request to `/v1/info`. This should happen on a separate thread. See e.g. https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1717655660357799?thread_ts=1717642232.650849&cid=C06EDRBSAH3 complaining about startup time (although this may not be the reason for that).

PR #274, opened 2024-06-21
Check token in separate thread

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/247


adriangb (approved):
Generally looks good to me. Seems like we never join the thread right? Do have a lock on configure()? What happens if I do;

configure()
while True:
  info()

?

alexmojaki:
```python
from logfire import configure, info

configure(token='...')
info('hi')
```

```
14:01:46.559 hi
/Users/alex/work/logfire/logfire/_internal/config.py:849: UserWarning: Invalid Logfire token.
  warnings.warn('Invalid Logfire token.')
Failed to export batch code: 401, reason: {"detail":"Invalid token"}
/Users/alex/work/logfire/logfire/_internal/exporters/file.py:58: WritingFallbackWarning: Failed to export spans, writing to fallback file: /Users/alex/Library/Application Support/JetBrains/PyCharm2024.1/scratches/.logfire/logfire_spans.bin
  warnings.warn(
Failed to export batch code: 401, reason: {"detail":"Invalid token"}
```

adriangb:
I mean what if it's a valid token. Does that `info()` call (1) warn if `configure()` hasn't actually finished, (2) wait for `configure()` to finish or (3) kick off another call to `configure()`?

alexmojaki:
The code inside `check_token` does nothing critical, it just prints the dashboard URL if the token is valid or warns (previously errored) if not. We already have the token and can start using it immediately to export without checking it.

Issue #248, opened 2024-06-06
Unnecessary `Redacted due to 'auth'`


mattzcarey:
### Description

I have a FastAPI endpoint which I am hitting to get some authentication types from some documentation. The endpoint is /auth/classify. Neither the input or the output from that endpoint contains secrets yet the body is still redacted:

```json
input": {
  "auth_types" :[
    "[Redacted due to 'auth']"
  ]
}
```

Is there anyway to switch this off as it would be nice to see these in the logs. To be clear this isnt a header or anything. This is just a body with the key auth and a value of string[] (it's actually an enum of values like 'BASIC', 'BEARER' etc.)

I guess I can manually emit a span but that kinda defeats the point of auto instrumentation. 

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="0.40.0"
platform="macOS-14.5-arm64-arm-64bit"
python="3.11.6 (main, May  8 2024, 18:55:10) [Clang 15.0.0 (clang-1500.3.9.4)]"
[related_packages]
requests="2.32.3"
pydantic="2.7.2"
fastapi="0.111.0"
openai="1.30.5"
protobuf="4.25.3"
rich="13.7.1"
executing="2.0.1"
opentelemetry-api="1.25.0"
opentelemetry-exporter-otlp-proto-common="1.25.0"
opentelemetry-exporter-otlp-proto-http="1.25.0"
opentelemetry-instrumentation="0.46b0"
opentelemetry-instrumentation-aiohttp-client="0.46b0"
opentelemetry-instrumentation-asgi="0.46b0"
opentelemetry-instrumentation-dbapi="0.46b0"
opentelemetry-instrumentation-fastapi="0.46b0"
opentelemetry-instrumentation-httpx="0.46b0"
opentelemetry-instrumentation-jinja2="0.46b0"
opentelemetry-instrumentation-requests="0.46b0"
opentelemetry-instrumentation-sqlite3="0.46b0"
opentelemetry-instrumentation-urllib="0.46b0"
opentelemetry-instrumentation-urllib3="0.46b0"
opentelemetry-proto="1.25.0"
opentelemetry-sdk="1.25.0"
opentelemetry-semantic-conventions="0.46b0"
opentelemetry-util-http="0.46b0"
```

alexmojaki:
See https://docs.pydantic.dev/logfire/guides/advanced/scrubbing/#scrubbing-less-with-a-callback for how to whitelist values.

[...]

PR #278, opened 2024-06-25
Add attribute noting details of scrubbed values

alexmojaki:
For https://github.com/pydantic/logfire/issues/248

The idea is to use this to generate code for the user to whitelist values from scrubbing. Here's some example code which I imagine being ported to the frontend:

```python
attributes = {
    'logfire.scrubbed': [
        {'path': ['attributes', 'usage'], 'matched_substring': 'token'},
        {'path': ['attributes', 'prompt'], 'matched_substring': 'secret'},
    ],
}

print(f"""
import logfire

def scrubbing_callback(match: logfire.ScrubMatch):
{"\n".join(f"""\
    if match.path == {tuple(match['path'])} and match.pattern_match.group(0) == {repr(match['matched_substring'])}:
        return match.value
"""
           for match in attributes['logfire.scrubbed'])}

logfire.configure(scrubbing=logfire.ScrubbingOptions(callback=scrubbing_callback))
""")
```

This generates:

```python
import logfire

def scrubbing_callback(match: logfire.ScrubMatch):
    if match.path == ('attributes', 'usage') and match.pattern_match.group(0) == 'token':
        return match.value

    if match.path == ('attributes', 'prompt') and match.pattern_match.group(0) == 'secret':
        return match.value


logfire.configure(scrubbing=logfire.ScrubbingOptions(callback=scrubbing_callback))
```


samuelcolvin (approved):
The code here looks good.

What do we need to change in the backend?

The PR body seems kind of confusing, and/or related to something slightly different. Am I being dumb or should or be updated.

alexmojaki:
> The PR body seems kind of confusing, and/or related to something slightly different. Am I being dumb or should or be updated.

It's related. The first bit of code is a sample of attributes that this PR adds to a span:

```python
attributes = {
    'logfire.scrubbed': [
        {'path': ['attributes', 'usage'], 'matched_substring': 'token'},
        {'path': ['attributes', 'prompt'], 'matched_substring': 'secret'},
    ],
}
```

The rest of that code block is an example of how the attributes can be used to generate sample Python code which is displayed to the user for them to copy and paste to whitelist values from being scrubbed. That's the ultimate goal here.

That code generation logic could also be baked into the SDK, but I think it makes more sense to do that part in the frontend.

The last code block is the output of the first, i.e. the sample code which would be generated and shown to the user.

The backend doesn't really need anything. But since the new attribute is stored as a string and it's not mentioned in the JSON schema, I think the backend should explicitly check for this attribute and parse the JSON just in case someone wants to query it or something.

PR #282, opened 2024-06-26
Use exponential histogram buckets

alexmojaki:


samuelcolvin:
@dmontagu or @adriangb please take a look.

It looks fine to me, but I don't pretend to fully understand it.

dmontagu (approved):

PR #283, opened 2024-06-27
Add `scrubbing: ScrubbingOptions | False` parameter to `logfire.configure`, replacing `scrubbing_patterns` and `scrubbing_callback`

alexmojaki:
Copying from https://pydantic.slack.com/archives/C05AF4A4WRM/p1719229546293829:

> Currently scrubbing is configured like this:
> 
> ```python
> logfire.configure(
>     scrubbing_callback=...,
>     scrubbing_patterns=...,
> )
> ```
> 
> I want to change it to this:
> 
> ```python
> logfire.configure(
>     scrubbing=logfire.ScrubbingOptions(
>         callback=...,
>         extra_patterns=...,
>     )
> )
> ```
> 
> 1. Combining related options into an object is what we’ve done elsewhere and it was a mistake not to do it here
> 2. People think that `scrubbing_patterns=[]` will disable all patterns, so `extra_patterns` should clarify that
> 3. I want to allow `scrubbing=False` to disable scrubbing entirely (similar to `console=False`), since many people want that and it would be more performant than `scrubbing_callback=lambda match: match.value`


adriangb (approved):
Looks good to me!

PR #303, opened 2024-07-05
Version 0.46.1

Kludex:
I made a mistake on https://github.com/pydantic/logfire/actions/runs/9807998370/job/27082996905

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -157,4 +157,4 @@ jobs:
       - name: Publish logfire-api to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -157,4 +157,4 @@ jobs:
       - name: Publish logfire-api to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          packages-dir: logfire-api/
+          packages-dir: logfire-api/dist
```
Kludex, line 160: I think this was the only problem, but I can't predict if there will be another...

alexmojaki (approved):

PR #305, opened 2024-07-05
Single dist folder on release

Kludex:
This simplifies the CI 👍

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -145,19 +145,15 @@ jobs:
 
       - uses: eifinger/setup-rye@v2
       - run: rye config --set-bool behavior.use-uv=true
-      - run: rye build
 
-      - name: Publish logfire to PyPI
+      - name: Build logfire
+        run: rye build
+
+      - name: Build logfire-api
+        run: rye build --pyproject logfire-api/pyproject.toml --out dist
+
+      - name: Publish artifacts to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           skip-existing: true
-
-      - name: Build logfire-api
-        run: rye build
-        working-directory: logfire-api/
-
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -145,19 +145,15 @@ jobs:
...
-        run: rye build
-        working-directory: logfire-api/
+        run: rye build --pyproject logfire-api/pyproject.toml --out dist
 
-      - name: Publish logfire-api to PyPI
+      - name: Publish artifacts to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          packages-dir: logfire-api/dist
+          skip-existing: true
```
Kludex:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -145,19 +145,12 @@ jobs:
...
+        run: rye build
 
       - name: Build logfire-api
-        run: rye build
-        working-directory: logfire-api/
+        run: rye build --pyproject logfire-api/pyproject.toml --out dist
 
-      - name: Publish logfire-api to PyPI
+      - name: Publish artifacts to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
```
alexmojaki, line 156: I'm not sure this action is helping us. It was tricky for you to debug when it went wrong. It requires creating a GH release to test and iterate. And it's hard to understand what exactly it's doing - it's somehow magically publishing two projects to PyPI that are both in `dist`?

I would much prefer to have shell commands here that I can see and understand and easily run on my local machine when I need to test changes or debug problems.

alexmojaki: Since the current setup is already working and is unlikely to change I guess it's not worth the effort to replace the action with shell commands. But I do think this PR adds to the problem. I'd rather keep it verbose and explicit.

Issue #313, opened 2024-07-13
Anthropic Instrumentation errors


bllchmbrs:
### Description

Received the following error on a `instrument_anthopic` call.
```

File ~/an-env/lib/python3.11/site-packages/logfire/_internal/main.py:1012, in Logfire.instrument_anthropic(self, anthropic_client, suppress_other_instrumentation)
    962 """Instrument an Anthropic client so that spans are automatically created for each request.
    963 
    964 The following methods are instrumented for both the sync and the async clients:
   (...)
   1008         Use of this context manager is optional.
   1009 """
   1010 import anthropic
-> 1012 from .integrations.llm_providers.anthropic import get_endpoint_config, is_async_client, on_response
   1013 from .integrations.llm_providers.llm_provider import instrument_llm_provider
   1015 self._warn_if_not_initialized_for_instrumentation()

File ~/an-env/lib/python3.11/site-packages/logfire/_internal/integrations/llm_providers/anthropic.py:6
      3 from typing import TYPE_CHECKING, Any
      5 import anthropic
----> 6 from anthropic.types import Message, RawContentBlockDeltaEvent, RawContentBlockStartEvent, TextBlock, TextDelta
      8 from .types import EndpointConfig
     10 if TYPE_CHECKING:

ImportError: cannot import name 'RawContentBlockDeltaEvent' from 'anthropic.types' (/Users/someuser/an-env/lib/python3.11/site-packages/anthropic/types/__init__.py)

### Python, Logfire & OS Versions, related packages (not required)
```

```TOML
logfire="0.46.1"
platform="macOS-10.16-x86_64-i386-64bit"
python="3.11.5 (main, Sep 11 2023, 08:19:27) [Clang 14.0.6 ]"
anthropic==0.25.9
```

Kludex:
What's the antrophic version?

bllchmbrs:
Added above `anthropic==0.25.9` (from `pip freeze | grep anthro`)

PR #316, opened 2024-07-15
Handle older versions of anthropic

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/313

Ideally we'd also test on the older version in CI, but that seems like less of a priority.

### DIFF
--- a/tests/otel_integrations/test_anthropic.py
+++ b/tests/otel_integrations/test_anthropic.py
@@ -1,10 +1,11 @@
 from __future__ import annotations as _annotations
 
 import json
-from typing import AsyncIterator, Iterator
+from typing import Any, AsyncIterator, Iterator
 
 import anthropic
 import httpx
+import pydantic
 import pytest
 from anthropic._models import FinalRequestOptions
 from anthropic.types import (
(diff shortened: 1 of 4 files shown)


tests/otel_integrations/test_anthropic.py
```diff
@@ -385,7 +377,7 @@ def test_sync_messages_stream(instrumented_client: anthropic.Anthropic, exporter
                     'logfire.span_type': 'log',
                     'logfire.tags': ('LLM',),
                     'duration': 1.0,
-                    'response_data': '{"combined_chunk_content":"The answer is secret","chunk_count":3}',
+                    'response_data': '{"combined_chunk_content":"The answer is secret","chunk_count":2}',
```
alexmojaki, line 380: The more liberal checking in `content_from_messages` started counting an additional chunk with empty text. So I started excluding all chunks with empty text, and it turns out there was already one being counted.

alexmojaki:
cc @willbakst

willbakst (approved):
LGTM

Kludex (approved):

Issue #239, opened 2024-06-04
`collect_headers=True` kwarg to `instrument_<webframe>()`


samuelcolvin:
### Description

So it's easy to enable collecting HTTP headers for any of the supported webframeworks.

This might just involve manually setting the env vars.

PR #318, opened 2024-07-16
capture_headers

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/239

### DIFF
--- a/logfire/_internal/integrations/django.py
+++ b/logfire/_internal/integrations/django.py
@@ -1,5 +1,7 @@
 from typing import Any
 
+from logfire._internal.utils import maybe_capture_server_headers
+
 try:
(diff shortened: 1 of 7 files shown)


docs/integrations/use_cases/web_frameworks.md
Kludex, line 1: I don't think this page should exist anymore.

alexmojaki: It still applies to ASGI and WSGI until we have all the instrument methods.

alexmojaki: It also lets users capture specific headers instead of all of them.

Kludex (approved):

PR #320, opened 2024-07-17
Add type hints to instrument methods

Kludex:
No runtime behavior changed.

### DIFF
--- a/logfire/_internal/integrations/flask.py
+++ b/logfire/_internal/integrations/flask.py
@@ -1,10 +1,34 @@
-from typing import Any
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 from flask.app import Flask
 from opentelemetry.instrumentation.flask import FlaskInstrumentor
 
+if TYPE_CHECKING:
+    from wsgiref.types import WSGIEnvironment
 
-def instrument_flask(app: Flask, **kwargs: Any):
+    from opentelemetry.metrics import MeterProvider
+    from opentelemetry.trace import Span, TracerProvider
+    from typing_extensions import Protocol, TypedDict, Unpack
+
+    class RequestHook(Protocol):
+        def __call__(self, span: Span, environment: WSGIEnvironment) -> None: ...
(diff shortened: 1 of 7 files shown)


logfire/_internal/integrations/psycopg.py
```diff
@@ -11,13 +11,25 @@
...
 
+    class CommenterOptions(TypedDict, total=False):
+        db_driver: bool
+        db_framework: bool
+        opentelemetry_values: bool
+
+    class PsycopgInstrumentKwargs(TypedDict, total=False):
+        enable_commenter: bool
+        commenter_options: CommenterOptions
+        skip_dep_check: bool
```
alexmojaki, line 26: We already pass this option.

logfire/_internal/integrations/flask.py
```diff
@@ -1,10 +1,34 @@
...
+    class RequestHook(Protocol):
+        def __call__(self, span: Span, environment: WSGIEnvironment) -> None: ...
+
+    class ResponseHook(Protocol):
+        def __call__(self, span: Span, status: str, response_headers: list[tuple[str, str]]) -> None: ...
+
+    class FlaskInstrumentKwargs(TypedDict, total=False):
+        request_hook: RequestHook | None
+        response_hook: RequestHook | None
+        tracer_provider: TracerProvider | None
```
alexmojaki, line 24: None of these should have `tracer_provider` or `meter_provider`, that would bypass `logfire.configure()`.

Kludex: Those are the parameters the `intrument()` methods accept. You don't want to show them because we already have the on `logfire.configure()`, and ignore the fact that they can accept those?

I can remove them. I was complying with what they accept.

logfire/_internal/integrations/psycopg.py
Kludex: Ah, you mean that it will be ignored on the user's side.

I'll remove it.

PR #322, opened 2024-07-18
Add `instrument_celery` method

Kludex:


Kludex:
I don't understand how I broke the tests... 🤔

docs/integrations/celery.md
```diff
@@ -27,14 +27,13 @@ Below we have a minimal example using Celery. You can run it with `celery -A tas
...
 from celery.signals import worker_process_init
-from opentelemetry.instrumentation.celery import CeleryInstrumentor
 
 
 logfire.configure()
 
 @worker_process_init.connect(weak=False)
 def init_celery_tracing(*args, **kwargs):
-    CeleryInstrumentor().instrument()
+    logfire.instrument_celery()
```
alexmojaki, line 37: Also need to update the start and end of this file like this: https://github.com/pydantic/logfire/pull/207/files#diff-1be28819fdb1a99ca760744cfa586fd5a755f990b140867e3bf67825c8dbdc75

tests/otel_integrations/test_celery.py
```diff
@@ -0,0 +1,100 @@
...
+    with start_worker(celery_app, perform_ping_check=False) as worker:  # type: ignore
+        yield worker
+
+
+def test_instrument_celery(celery_app: Celery, exporter: TestExporter) -> None:
+    logfire.instrument_celery()
+
+    # Send and wait for the task to be executed
+    result = celery_app.send_task('tasks.say_hello')  # type: ignore
+    result.get(timeout=10)  # type: ignore
```
alexmojaki: Please make `say_hello` return something and check it here.

.github/workflows/main.yml
```diff
@@ -87,9 +87,52 @@ jobs:
           name: coverage-${{ matrix.python-version }}
           path: coverage
 
+  test-integration:
```
alexmojaki: i really don't like that so much has been copied here. can we just have one workflow, with separate steps for unit and integration tests?

Kludex: I think having a different job is a way to "motivate" us to write more tests like those, since it's going to eventually slow us down.

Kludex: Your complaint seems more about the amount of duplicated code, rather than the idea of having a separate flow. Which can be solved by creating a common GH action.

tests/otel_integrations/test_celery.py
```diff
@@ -0,0 +1,106 @@
...
+    try:
+        yield app
+    finally:
+        CeleryInstrumentor().uninstrument()  # type: ignore
+
+
+@pytest.fixture(autouse=True)
+def celery_worker(celery_app: Celery) -> Iterator[WorkController]:
+    logger = logging.getLogger()
+    with start_worker(celery_app, perform_ping_check=False, loglevel=logger.level) as worker:  # type: ignore
```
Kludex, line 54: The `loglevel` has a default value to `ERROR`, and it modifies the root logger.

This code just sets the current level.

alexmojaki: This comment should be in the code

alexmojaki (approved):

PR #323, opened 2024-07-18
Update dependencies, handle change in importlib

alexmojaki:
A couple of tests started failing with the newest dependencies. The behaviour change comes from importlib but this hasn't changed in the lockfiles so I'm guessing it has to do with setuptools.

### DIFF
--- a/logfire/_internal/cli.py
+++ b/logfire/_internal/cli.py
@@ -388,12 +388,14 @@ def parse_info(_args: argparse.Namespace) -> None:
     related_packages: list[tuple[int, str, str]] = []
 
     for dist in importlib_metadata.distributions():
-        name = dist.metadata['Name']
+        metadata = dist.metadata
+        name = metadata.get('Name', '')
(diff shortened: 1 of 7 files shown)


logfire/_internal/cli.py
```diff
@@ -388,12 +388,14 @@ def parse_info(_args: argparse.Namespace) -> None:
     related_packages: list[tuple[int, str, str]] = []
 
     for dist in importlib_metadata.distributions():
-        name = dist.metadata['Name']
+        metadata = dist.metadata
+        name = metadata.get('Name', '')
```
alexmojaki, line 392: `dist.metadata` was an 'empty' metadata object in some cases, so `dist.metadata['Name']` was giving a deprecation warning about missing keys which the tests translated to an error, and the warning said that it would eventually be a KeyError. So the fix is to not assume that `Name` and `Version` exist.

Kludex (approved):

Issue #195, opened 2024-05-20
Allow `logfire.configure()` to return multiple separate non-global logfire configurations


alexmojaki:
I'm thinking something like `my_logfire = logfire.configure(..., global=False)`. With the default `global=True` then `configure` will return the default logfire instance. Otherwise it returns a new instance with a new config object, containing a tracer/meter provider that aren't set as the global providers in OTEL.

All the OTEL instrumentation libraries should accept a tracer/meter provider argument where appropriate instead of setting the global providers, so the methods created in https://github.com/pydantic/logfire/issues/193 should pass providers from the logfire config.

PR #324, opened 2024-07-19
Add ASGI & WSGI instrument methods

Kludex:
Is this what you had in mind? @alexmojaki 

I don't like it.


alexmojaki:
Pretty much. What would you prefer? Just a different name?

Kludex:
I don't like that every `instrument_` method is returns `None`, but those 2 here need to return the app itself.

alexmojaki:
I agree with that. The difference is more obvious in OTEL because it's not e.g. `ASGIInstrumentor().instrument(app)` like other things. We could maybe have `instrumented_asgi` but people might not notice the difference. If people copy the code from the docs then they'll probably get it right, but they might not do that.

If it was `name logfire.asgi_middleware(app)` or something else which didn't start with `instrument_`, but the rest was the same, how would you feel?

Kludex:
> If it was name logfire.asgi_middleware(app) or something else which didn't start with instrument_, but the rest was the same, how would you feel?

I think I'll prefer more if it was a class itself e.g. `logfire.ASGIMiddleware()`. Because then it's clear for those familiar with ASGI that it returns an ASGI application.

alexmojaki:
But an ASGI application is just a callable, often a function.

Kludex:
It's a callable, but I don't think it's usually a function. In any case, I think it's irrelevant if an app itself is usually a function or a class, but ASGI middlewsres in the ecosystem are usually presented as classes.

alexmojaki:
Currently you can do this:

```python
import logfire

logfire_instance = logfire.with_tags('foo')

logfire.instrument_bar()
# or
logfire_instance.instrument_bar()
```

Right now that's not very useful, but we may add other `logfire.with_*` methods. More importantly I want to eventually implement https://github.com/pydantic/logfire/issues/195 which would allow `logfire_instance = logfire.configure(..., global=False)`.

I want the `instrument` methods to be instances of the Logfire class so that they all support `logfire_instance.instrument_bar()`. If `logfire.ASGIMiddleware` is a class then `logfire_instance.ASGIMiddleware` will be an error, and the user will have to write `app = logfire.ASGIMiddleware(app, logfire_instance=logfire_instance)`. That feels a bit sad. BUT maybe it's fine if we're already giving up on a name like `logfire.instrument_asgi` anyway. WDYT?

Kludex:
> I want the instrument methods to be instances of the Logfire class so that they all support logfire_instance.instrument_bar(). If logfire.ASGIMiddleware is a class then logfire_instance.ASGIMiddleware will be an error, and the user will have to write app = logfire.ASGIMiddleware(app, logfire_instance=logfire_instance). That feels a bit sad. BUT maybe it's fine if we're already giving up on a name like logfire.instrument_asgi anyway. WDYT?

Ah, I see. Then I think we can just have it as functions.

alexmojaki:
Do you mean methods of the `Logfire` class?

Kludex:
> Do you mean methods of the `Logfire` class?

Yes.

sydney-runkle:
@Kludex, ping me when this is ready for review, I'd love to take a look.

Kludex:
I'll work on this today.

docs/integrations/asgi.md
```diff
@@ -1,23 +1,20 @@
 # ASGI
 
 If the [ASGI][asgi] framework doesn't have a dedicated OpenTelemetry package, you can use the
```
alexmojaki: I know it's not changed by this PR, but this sentence referring to 'the ASGI framework' sounds weird.

docs/integrations/asgi.md
```diff
@@ -34,15 +31,17 @@ async def app(scope, receive, send):
     )
     await send({"type": "http.response.body", "body": b"Hello, world!"})
 
-app = OpenTelemetryMiddleware(app)
+app = logfire.instrument_asgi(app)
```
alexmojaki, line 34: Maybe we should emphasise that they need to use the return value, unlike other instrument methods.

logfire/_internal/main.py
```diff
@@ -1241,6 +1246,30 @@ def instrument_starlette(
             **kwargs,
         )
 
+    def instrument_asgi(self, app: ASGIApp, **kwargs: Unpack[ASGIInstrumentKwargs]) -> ASGIApp:
+        """Instrument `app` so that spans are automatically created for each request.
+
+        Uses the
+        [OpenTelemetry ASGI Instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/asgi/asgi.html)
+        library, specifically `ASGIInstrumentor().instrument_app()`, to which it passes `**kwargs`.
```
alexmojaki: ASGIInstrumentor doesn't exist

logfire/_internal/main.py
```diff
@@ -1241,6 +1246,30 @@ def instrument_starlette(
             **kwargs,
         )
 
+    def instrument_asgi(self, app: ASGIApp, **kwargs: Unpack[ASGIInstrumentKwargs]) -> ASGIApp:
+        """Instrument `app` so that spans are automatically created for each request.
```
alexmojaki, line 1261:
```suggestion
        """Instrument `app` so that spans are automatically created for each request.
        
        Returns an instrumented app, so use like this:
        
            app = logfire.instrument_asgi(app)
```

logfire/_internal/main.py
```diff
@@ -1241,6 +1246,30 @@ def instrument_starlette(
...
+
+        self._warn_if_not_initialized_for_instrumentation()
+        return instrument_asgi(self, app, **kwargs)
+
+    def instrument_wsgi(self, app: WSGIApplication, **kwargs: Unpack[WSGIInstrumentKwargs]) -> WSGIApplication:
+        """Instrument `app` so that spans are automatically created for each request.
+
+        Uses the
+        [OpenTelemetry WSGI Instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/wsgi/wsgi.html)
+        library, specifically `WSGIInstrumentor().instrument_app()`, to which it passes `**kwargs`.
```
alexmojaki: WSGIInstrumentor doesn't exist

docs/integrations/wsgi.md
```diff
@@ -18,14 +16,14 @@ Below we have a minimal example using the standard library [`wsgiref`][wsgiref].
...
-from opentelemetry.instrumentation.wsgi import OpenTelemetryMiddleware
+import logfire
 
 
 def app(env, start_response):
     start_response('200 OK', [('Content-Type','text/html')])
     return [b"Hello World"]
 
-app = OpenTelemetryMiddleware(app)
+app = logfire.instrument_wsgi(app)
```
alexmojaki, line 28: please run this example and make sure it works

logfire/_internal/integrations/asgi.py
```diff
@@ -48,3 +73,26 @@ def start_span(self, name: str, context: Context | None = None, *args: Any, **kw
...
     # is roughly equivalent to `with use_span(start_span(...)):`
     start_as_current_span = SDKTracer.start_as_current_span
+
+
+def instrument_asgi(
+    logfire_instance: Logfire,
+    app: ASGIApp,
+    *,
+    record_send_receive: bool = False,
+    capture_headers: bool = False,
```
alexmojaki, line 83: These need documenting in the Logfire methods' docstrings, same for wsgi

docs/integrations/asgi.md
Kludex: Please suggest something.

alexmojaki:
```suggestion
If the [ASGI][asgi] web framework you're using doesn't have a dedicated integration, you can use the
```

alexmojaki: did you mean to commit this?

Kludex: I think I applied and then I force pushed other changes 🤔

logfire/_internal/main.py
```diff
@@ -1217,16 +1225,19 @@ def instrument_starlette(
...
-        If enabled, they will be set to debug level, meaning they will usually still be hidden in the UI.
-
         Uses the
         [OpenTelemetry Starlette Instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/starlette/starlette.html)
         library, specifically `StarletteInstrumentor.instrument_app()`, to which it passes `**kwargs`.
+
+        Args:
+            app: The Starlette app to instrument.
+            capture_headers: Set to `True` to capture all request and response headers.
+            record_send_receive: Set `record_send_receive` to `True` to allow the OpenTelemetry ASGI to create send/receive spans.
```
```suggestion
            record_send_receive: Set `record_send_receive` to `True` to allow the OpenTelemetry ASGI middleware to create send/receive spans.
```

docs/integrations/asgi.md
```diff
@@ -34,15 +31,17 @@ async def app(scope, receive, send):
...
     uvicorn.run(app)
 ```
 
-You can read more about the OpenTelemetry ASGI middleware [here][opentelemetry-asgi].
+The keyword arguments of [`logfire.instrument_asgi()`][logfire.Logfire.instrument_asgi] are passed to the
+[`OpenTelemetryMiddleware`][opentelemetry.instrumentation.asgi.OpenTelemetryMiddleware] class
+of the OpenTelemetry ASGI Instrumentation package.
 
 ## Excluding URLs from instrumentation
 <!-- note that this section is duplicated for different frameworks but with slightly different links -->
```
alexmojaki, line 47: The note here should refer to instrument_asgi, not OpenTelemetryMiddleware.

In fact we should probably just implement the env var ourselves.

docs/integrations/wsgi.md
```diff
@@ -1,15 +1,13 @@
 # WSGI
 
 If the [WSGI][wsgi] framework doesn't have a dedicated OpenTelemetry package, you can use the
```
alexmojaki: same change as in asgi

logfire/_internal/main.py
alexmojaki: same for the other integrations

Kludex:
@alexmojaki something else here?

alexmojaki:
Thanks!!! 🙏

PR #329, opened 2024-07-23
Handle message formatting errors

alexmojaki:
Addresses https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1721433723530879

There were several places where we explicitly or implicitly raised errors which could lead to similar complaints from users. Many of these errors could be caused in the same way: by passing a preformatted string which gets unintentionally interpreted as a formatting template. I've converted all these known cases to warnings.

Previously we would warn about missing fields but continue trying to format the string. This had the small benefit of formatting other fields that weren't present, so the user gets something partially formatted. This helped in the specific case where they know what they're doing but made a typo. Keeping this behaviour would mean that something like `logfire.info("log {'1': {'2': {'3': {'4': 5}}}}")`would produce three warnings about missing fields `'1'`, `'2'`, and `'3'`, and a fourth warning about recursion depth, all from the same line. So instead I chose to just fail quickly when a field is missing. This just means that the message is just the format string, but the user can still find any correct fields in the attributes, so it's not a big loss. Also the better way to prevent these kinds of typos is to use an f-string, and the warning points to that page in the docs.

### DIFF
--- a/logfire/_internal/formatter.py
+++ b/logfire/_internal/formatter.py
@@ -275,15 +275,17 @@ class ChunksFormatter(Formatter):
                 except IndexError:
                     raise KnownFormattingError('Numeric field names are not allowed.')
                 except KeyError as exc1:
-                    missing = str(exc1)
-                    if missing == repr(field_name):
+                    if str(exc1) == repr(field_name):
                         raise KnownFormattingError(f'The field {{{field_name}}} is not defined.') from exc1
 
                     try:
-                        # fall back to getting a key with the dots in the name
+                        # field_name is something like 'a.b' or 'a[b]'
+                        # Evaluating that expression failed, so now just try getting the whole thing from kwargs.
+                        # In particular, OTEL attributes with dots in their names are normal and handled here.
                         obj = kwargs[field_name]
                     except KeyError as exc2:
-                        raise KnownFormattingError(f'The fields {str(exc1)} and {str(exc2)} are not defined.') from exc2
+                        # e.g. neither 'a' nor 'a.b' is defined
+                        raise KnownFormattingError(f'The fields {exc1} and {exc2} are not defined.') from exc2
                 except Exception as exc:
                     raise KnownFormattingError(f'Error getting field {{{field_name}}}: {exc}') from exc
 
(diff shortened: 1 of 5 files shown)


logfire/_internal/formatter.py
```diff
@@ -450,3 +451,29 @@ def warn_inspect_arguments(msg: str, stacklevel: int):
...
+
+class KnownFormattingError(Exception):
+    """An error raised when there's something wrong with a format string or the field values.
+
+    In other words this should correspond to errors that would be raised when using `str.format`,
+    and generally indicate a user error, most likely that they weren't trying to pass a template string at all.
+    """
+
+
+class FormattingFailedWarning(Warning):
```
dmontagu:
```suggestion
class FormattingFailedWarning(UserWarning):
```

I think if we view a `FormattingFailedWarning` as a result of a mistake in user code, it would be marginally better for it to inherit from `UserWarning`. If not, the current implementation is fine.

It seems to me in practice, 9 times out of 10 that a `KnownFormattingError` is raised, it will be due to a "mistake" in user code, which is why I would suggest using `UserWarning` as the base class here but if I'm missing something or you just disagree with that change we can keep it as is.

logfire/_internal/formatter.py
```diff
@@ -450,3 +451,29 @@ def warn_inspect_arguments(msg: str, stacklevel: int):
...
+    pass
+
+
+def warn_formatting(msg: str):
+    _frame, stacklevel = get_user_frame_and_stacklevel()
+    warnings.warn(
+        f'\n'
+        f'    Ensure you are either:\n'
+        '      (1) passing an f-string directly, with inspect_arguments enabled and working, or\n'
+        '      (2) passing a literal `str.format`-style template, not a preformatted string.\n'
```
dmontagu, line 474: You are also allowed to pass non-string-literals right? (By that I mean, a string-valued variable.) The way this message is phrased it sounds like that isn't allowed

dmontagu (approved):
Other than my note about the string literal message, LGTM. Let me know if you want input on how to tweak that message. Alternatively, feel free to merge without changes if you think it's good.

logfire/_internal/formatter.py
alexmojaki: You're allowed, if you know what you're doing. On the one hand we used to have a `LiteralString` type hint for the argument, and it only got removed to allow passing f-strings. On the other hand the linked docs do suggest a reasonable way to pass a variable string. We could say 'low cardinality' but that's a bit heavy.

dmontagu: That makes sense, given that I'm good with just leaving the message as is. 👍

PR #340, opened 2024-07-26
Test emitting both new and old semantic convention attribute names

alexmojaki:
OTEL is migrating the HTTP semantic conventions. A good explanation of what this means is here: https://www.honeycomb.io/blog/opentelemetry-http-attributes

The latest OTEL release implements this transition for several big instrumentations, search for "semantic" in https://github.com/open-telemetry/opentelemetry-python-contrib/releases/tag/v0.47b0

Because of https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2745, the Django duration metrics don't have an `http.target` attribute, but if you opt into the new semconv then they have `http.route`.

Switching over entirely to the new semconv breaks some of our tweaks in the SDK and probably in the frontend, as well as existing queries in dashboards etc. So for customers who want the improved Django metrics, we should recommend setting the env var `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup`. That mostly just means that spans get a bunch of extra attributes, most of which have the same values as existing attributes but different names.

This PR just tests what happens when doing that and updates the snapshots. Fortunately no lines are removed, so this looks safe to recommend.


tests/otel_integrations/test_django.py
```diff
@@ -79,6 +81,36 @@ def test_good_route(client: Client, exporter: TestExporter, metrics_reader: InMe
...
+                'description': 'Duration of HTTP server requests.',
+                'unit': 's',
+                'data': {
+                    'data_points': [
+                        {
+                            'attributes': {
+                                'http.request.method': 'GET',
+                                'url.scheme': 'http',
+                                'network.protocol.version': '1.1',
+                                'http.route': 'django_test_app/<int:item_id>/',
```
alexmojaki, line 95: This is what this is all about. Note that it's a whole new metric name. The old duration metrics are still above, unchanged, without anything similar to `http.route`.

Kludex (approved):

PR #325, opened 2024-07-20
Log more info about internal errors

dmontagu:
In https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1721433723530879 you can see an internal error where we don't produce enough information to determine what line of user code produced the issue. (We only get stack frames underneath the `try: except:` so can't see what user code led to the error.)  I think it would be better if we log a bit more frames up the stack when an internal error is produced to give us a better handle on debugging.

In particular, in this case, it seems to be an issue with magic formattting — seeing the source code could be very insightful.

PR #346, opened 2024-07-29
Log more info about internal errors

alexmojaki:
Replaces https://github.com/pydantic/logfire/pull/325


logfire/_internal/utils.py
```diff
@@ -252,17 +256,83 @@ def log_internal_error():
...
+
+def _handle_internal_errors():
     try:
         yield
     except Exception:
         log_internal_error()
 
 
+_HANDLE_INTERNAL_ERRORS_CODE = _handle_internal_errors.__code__
+handle_internal_errors = contextmanager(_handle_internal_errors)
```
Kludex: If you do this, I think the stub generator will work properly, and not generate the `Incomplete` - not sure tho. Can we try?
```suggestion
handle_internal_errors: ContextManager[None, None, None] = contextmanager(_handle_internal_errors)
```

logfire-api/logfire_api/_internal/utils.pyi
```diff
@@ -1,5 +1,6 @@
 from _typeshed import Incomplete
 from collections.abc import Generator
+from logfire._internal.stack_info import is_user_code as is_user_code
```
Kludex, line 3: Why is the `as` needed?

logfire/_internal/utils.py
```diff
@@ -252,7 +256,70 @@ def log_internal_error():
         raise
 
     with suppress_instrumentation():  # prevent infinite recursion from the logging integration
-        logger.exception('Internal error in Logfire')
+        logger.exception('Internal error in Logfire', exc_info=_internal_error_exc_info())
+
+
+def _internal_error_exc_info() -> Any:
```
Kludex:
```suggestion
ExcInfoType = tuple[Type[BaseException] | None, BaseException | None, types.TracebackType | None]

def _internal_error_exc_info() -> ExcInfoType:
```

Kludex (approved):

logfire-api/logfire_api/_internal/utils.pyi
alexmojaki: That's just what stubgen does, you can see it in lots of other places in the pyi files

PR #347, opened 2024-07-30
Switch UpDownCounters to cumulative aggregation temporality

alexmojaki:
In https://github.com/pydantic/platform/pull/1456#discussion_r1506686563 I couldn't understand the recommendation in https://docs.datadoghq.com/opentelemetry/guide/otlp_delta_temporality/ and chose to use delta temporality for all instruments. This was a mistake.

If we want to measure e.g. the total number of requests, that's a monotonic sum, a Counter is an appropriate instrument, and I can add up all delta measurements between two times to get the number of requests handled in a time period.

But if we're measuring e.g. active requests, that's non-monotonic and we use an UpDownCounter. Some of the measurements will have negative values, indicating that the server stopped processing a request. If we use delta temporality and add up values in a time period, we get the change in the number of active requests being handled, e.g. at time T2 the server was handling 5 more requests than it was handling at time T1. That's not very useful. Usually we just want to know how many requests were being handled at a particular time. With delta temporality, we'd have to add up all the values starting from the beginning up to that point. Cumulative temporality gives that to us for free.

I still think aggregation temporality is meaningless for ObservableGauge, but now I think I should just go with DataDog's recommendation.


alexmojaki:
Interesting code I just stumbled across in the OTEL SDK:

```python
    def _get_temporality(
        self, preferred_temporality: Dict[type, AggregationTemporality]
    ) -> Dict[type, AggregationTemporality]:

        otel_exporter_otlp_metrics_temporality_preference = (
            environ.get(
                OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE,
                "CUMULATIVE",
            )
            .upper()
            .strip()
        )

        if otel_exporter_otlp_metrics_temporality_preference == "DELTA":
            instrument_class_temporality = {
                Counter: AggregationTemporality.DELTA,
                UpDownCounter: AggregationTemporality.CUMULATIVE,
                Histogram: AggregationTemporality.DELTA,
                ObservableCounter: AggregationTemporality.DELTA,
                ObservableUpDownCounter: AggregationTemporality.CUMULATIVE,
                ObservableGauge: AggregationTemporality.CUMULATIVE,
            }

        elif otel_exporter_otlp_metrics_temporality_preference == "LOWMEMORY":
            instrument_class_temporality = {
                Counter: AggregationTemporality.DELTA,
                UpDownCounter: AggregationTemporality.CUMULATIVE,
                Histogram: AggregationTemporality.DELTA,
                ObservableCounter: AggregationTemporality.CUMULATIVE,
                ObservableUpDownCounter: AggregationTemporality.CUMULATIVE,
                ObservableGauge: AggregationTemporality.CUMULATIVE,
            }

        else:
            if otel_exporter_otlp_metrics_temporality_preference != (
                "CUMULATIVE"
            ):
                _logger.warning(
                    "Unrecognized OTEL_EXPORTER_METRICS_TEMPORALITY_PREFERENCE"
                    " value found: "
                    f"{otel_exporter_otlp_metrics_temporality_preference}, "
                    "using CUMULATIVE"
                )
            instrument_class_temporality = {
                Counter: AggregationTemporality.CUMULATIVE,
                UpDownCounter: AggregationTemporality.CUMULATIVE,
                Histogram: AggregationTemporality.CUMULATIVE,
                ObservableCounter: AggregationTemporality.CUMULATIVE,
                ObservableUpDownCounter: AggregationTemporality.CUMULATIVE,
                ObservableGauge: AggregationTemporality.CUMULATIVE,
            }

        instrument_class_temporality.update(preferred_temporality or {})

        return instrument_class_temporality
```

So effectively this is like setting the env var `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` to `DELTA`

Kludex (approved):

Issue #298, opened 2024-07-02
Do not require auth if using custom exporters


astrojuanlu:
### Description

I tried Logfire for the first time and the UI looks so beautiful ✨ 

I then tried to set up a custom exporter using Jaeger as follows:

```
export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces
```

and the following `docker-compose.yaml`:

```yaml
services:
  jaeger:
    image: jaegertracing/all-in-one:1.58
    ports:
      - "4318:4318"
      - "16686:16686"
```

As expected, the logs were sent to my local Jaeger instance, and not to the Logfire UI. However, it still required me to do `logfire auth` and set up an empty project.

I then tried `export LOGFIRE_SEND_TO_LOGFIRE=0` and it did not ask me to auth, but also it didn't send the traces to Jaeger.

Am I missing something? And if indeed this is not yet possible, would you folks consider enabling that use case?

alexmojaki:
You're not missing anything, this is a mistake on our part.

S4mpl3r:
@alexmojaki Is it possible for you guys to put up a guide for using logfire sdk with a local backend solution?
I've been trying the beta and am very impressed, but I'd like to be able to send my data to my own servers in production.

astrojuanlu:
@S4mpl3r it's not a guide, but in https://github.com/pydantic/logfire/issues/298#issue-2386213778 I wrote the steps I followed

S4mpl3r:
@astrojuanlu Yeah I've seen your comment and I'll be trying that out, but an 'official' guide of sorts will be worth the effort as I'm sure a lot of other developers will be looking for such a solution.

alexmojaki:
@astrojuanlu @S4mpl3r @joaoflaviosantos @mastizada @gustavosett I've started work on a fix for this in https://github.com/pydantic/logfire/pull/350. As I've explained there it's tricky to design user-friendly behavior, would love to hear everyone's perspectives.

alexmojaki:
OK, if you upgrade to the newest release (0.50.0) you should be able to do this sanely. Just set `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318/v1/traces`. I think leave out the `OTEL_EXPORTER_OTLP_ENDPOINT` or it will try to also send metrics to Jaeger and log errors about that.

[...]

Issue #265, opened 2024-06-17
Irregular failing to connect to Logfire API errors


yovelcohen:
### Description

I have a bunch of python microservices, currently running on azure container apps.
they write logs to two major projects in Logfire.
For some reason, even though I'm using the same configuration code and the exact same write token, some of the services can't connect to logfire, with a weird open telemetry error msg:
```
/usr/local/lib/python3.11/site-packages/logfire/_internal/config.py:842: UserWarning: Logfire API is unreachable, you may have trouble sending data. Error: HTTPConnectionPool(host='k8se-otel.k8se-apps.svc', port=4317): Max retries exceeded with url: /v1/info (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x71713aae7e90>: Failed to resolve 'k8se-otel.k8se-apps.svc' ([Errno -2] Name does not resolve)"))
  warnings.warn(f'Logfire API is unreachable, you may have trouble sending data. Error: {e}')
```

I haven't been able to find consistency as to why some services can connect and some connect (double-checked the logfire configuration in each)

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="0.42.0"
platform="Linux-5.15.153.1-1.cm2-x86_64-with"
python="3.11.7 (main, Jan 27 2024, 07:04:48) [GCC 13.2.1 20231014]"
[related_packages]
requests="2.32.3"
pydantic="2.7.4"
fastapi="0.111.0"
protobuf="4.25.3"
rich="13.7.1"
executing="2.0.1"
opentelemetry-api="1.25.0"
opentelemetry-exporter-otlp-proto-common="1.25.0"
opentelemetry-exporter-otlp-proto-http="1.25.0"
opentelemetry-instrumentation="0.46b0"
opentelemetry-proto="1.25.0"
opentelemetry-sdk="1.25.0"
opentelemetry-semantic-conventions="0.46b0"
```

alexmojaki:
`k8se-otel.k8se-apps.svc` is not the Logfire API. If you didn't configure that yourself, then I think something in Azure is setting the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. You can override it with `base_url='https://logfire-api.pydantic.dev'` in `logfire.configure()`.

yovelcohen:
@alexmojaki thanks, that was indeed the issue, I'm just a bit curios as to why would this happens on some container apps and not others, but thanks :)

alexmojaki:
[...]

PR #350, opened 2024-07-31
Allow sending to custom OTEL_EXPORTER_OTLP_ENDPOINT without logfire auth

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/298. Still thinking through some details, could use feedback. Here's how this currently works:

1. If you want to export over HTTP, you need `send_to_logfire=True` (the default) even if you're sending to something that isn't logfire. If you want to simultaneously export to both logfire and something custom, you can't just use the env vars, you have to manually set up the exporters and processors. Nothing below applies when `send_to_logfire=False`.
2. If you want to send both traces and metrics somewhere custom, set the env var `OTEL_EXPORTER_OTLP_ENDPOINT`. Traces will then be sent to `{OTEL_EXPORTER_OTLP_ENDPOINT}/v1/traces`, similar for metrics.
3. Otherwise you can set `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and/or `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` if you need to be more specific.
4. If `OTEL_EXPORTER_OTLP_ENDPOINT` and/or `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is set, we assume that you are sending traces to something that isn't logfire, and set up basic exporters with minimal configuration. If neither is set, we assume that you're sending to a logfire instance hosted at `base_url`, and refine the exporters with extra logic. Similar for metrics.
5. Unless you're sending both traces and metrics to somewhere that isn't logfire, we require authentication. So for example if you only set `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, we assume that you're still going to send metrics to logfire and expect to find a token somewhere. We don't wait until you start creating metrics data.

Problems:

1. The name `send_to_logfire` is misleading
2. `base_url` is now only for a logfire URL. For users this means it's only relevant when self-hosting logfire, which isn't possible yet. If we want to reuse it for general OTLP endpoints we need a way to indicate whether or not `base_url` represents a logfire URL. We could maybe do this by sending a special request to the backend, but I feel iffy about this.
3. OTEL endpoints can only be configured via env vars, not `logfire.configure()` or `pyproject.toml`. Adding them as regular config options could be a fair amount of API bloat.
4. This is quite a bit of logic that's hard to document in a way that's both clear and concise.

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -23,6 +23,7 @@ from opentelemetry.exporter.otlp.proto.http.metric_exporter import OTLPMetricExp
 from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
 from opentelemetry.sdk.environment_variables import (
     OTEL_BSP_SCHEDULE_DELAY,
+    OTEL_EXPORTER_OTLP_ENDPOINT,
     OTEL_EXPORTER_OTLP_METRICS_ENDPOINT,
     OTEL_EXPORTER_OTLP_TRACES_ENDPOINT,
     OTEL_RESOURCE_ATTRIBUTES,
@@ -375,9 +376,6 @@ class _LogfireConfigData:
         param_manager = ParamManager.create(config_dir)
 
         self.base_url = param_manager.load_param('base_url', base_url)
-        self.metrics_endpoint = os.getenv(OTEL_EXPORTER_OTLP_METRICS_ENDPOINT) or urljoin(self.base_url, '/v1/metrics')
-        self.traces_endpoint = os.getenv(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT) or urljoin(self.base_url, '/v1/traces')
-
         self.send_to_logfire = param_manager.load_param('send_to_logfire', send_to_logfire)
         self.token = param_manager.load_param('token', token)
         self.project_name = param_manager.load_param('project_name', project_name)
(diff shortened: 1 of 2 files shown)


astrojuanlu:
> 4. If `OTEL_EXPORTER_OTLP_ENDPOINT` and/or `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is set, we assume that you are sending traces to something that isn't logfire, and set up basic exporters with minimal configuration. If neither is set, we assume that you're sending to a logfire instance hosted at base_url, and refine the exporters with extra logic. Similar for metrics.
> 5. Unless you're sending both traces and metrics to somewhere that isn't logfire, we require authentication. So for example if you only set `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, we assume that you're still going to send metrics to logfire and expect to find a token somewhere. We don't wait until you start creating metrics data.

This seems good enough to me 👍🏼 and I appreciate that Logfire is willing to go forward with this tough product decision 🙏🏼 

> The name `send_to_logfire` is misleading

Yes it is. Something that could be changed in the future maybe? (Before you folks get too famous and there's no way back 😄)

S4mpl3r:
Truly appreciate the willingness for going the extra mile to implement this feature!

alexmojaki:
I'm actually now leaning strongly in the opposite direction of this PR and more towards my second comment. I think that the logfire SDK should avoid looking at `OTEL_EXPORTER_OTLP_ENDPOINT` or `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`. More generally I think it would be best to draw a hard line between sending to logfire and sending elsewhere. The latter should require `logfire.configure(additional_metric_readers=[...], additional_span_processors=[...])`. We can write guides in documentation explaining how to wire that up, but doing just a little bit to make that more convenient _sometimes_ by checking specific env vars doesn’t seem like a good strategy. It makes things confusing and hard to document, and can be surprising in cases like https://github.com/pydantic/logfire/issues/265.

@astrojuanlu @S4mpl3r do you have any strong objections to using the snippet in https://github.com/pydantic/logfire/pull/350#issuecomment-2260956211?

S4mpl3r:
@alexmojaki tbh I don't really mind it either way. I just want to be able to send logs to a custom server, and as long as it's possible and is not a real burden to do, I'm fine (a guide in the official docs would be awesome).

samuelcolvin:
I think we should make it very easy to send data to another service - it's really important for the way people perceive our SDK(s) in particular and the while company in general, that we clearly support sending data to another store, and clearly make that easy.

Having spoken to hundreds of people at conferences, the "you can just set `base_url` and send to any otel store" is an incredibly reassuring thing for potential users to hear.

For that reason, I think we should definitely support `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`.

My suggests for how other stuff works:
* `send_to_logfire` alone decides whether data is sent to logfire, we introduce a new `logfire_url` (medium term I think it should replace `base_url`) which you can use for custom logfire instances (e.g. our meta instance, and the team running logfire locally)
* `OTEL_EXPORTER_OTLP_ENDPOINT` etc. decide whether data is sent to another URL (as well as logfire if `send_to_logfire=True`)
* we could also support kwargs like `otel_exporter_otlp_endpoint` but maybe that's just confusing?

alexmojaki:
> we introduce a new `logfire_url` (medium term I think it should replace `base_url`) which you can use for custom logfire instances

Is the goal here just to make it explicitly clear that it only applies to logfire instances, in a way that the name `base_url` doesn't?

I'm still a little worried about https://github.com/pydantic/logfire/issues/265. If Azure or whatever automatically sets `OTEL_EXPORTER_OTLP_ENDPOINT` without the user knowing, it will export to both logfire and that endpoint. Then it'll log noisy tracebacks when the latter fails, and I'm not sure if we should do anything about that. Currently, as I suggested in that issue, you can explicitly set `base_url` to override the `OTEL_EXPORTER_OTLP_ENDPOINT`. In the new plan, either we add some kind of param to check/ignore OTEL env vars, or we tell the user to directly set `OTEL_EXPORTER_OTLP_ENDPOINT` to an empty string and make sure that this happens after the other place where the variable was set but before calling `logfire.configure()`. I guess that's not so bad.

alexmojaki:
Replacing with https://github.com/pydantic/logfire/pull/351

PR #351, opened 2024-08-02
Separate sending to logfire from using standard OTEL env vars

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/298, based on https://github.com/pydantic/logfire/pull/350#issuecomment-2262810630

Notable changes:

- `OTEL_EXPORTER_OTLP_ENDPOINT` is no longer just an alternative to `LOGFIRE_BASE_URL`. Setting `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, and/or `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` will set up appropriate exporters *in addition* to sending to logfire, which must be turned off separately if desired. These are basic exporters relying on OTEL defaults. In particular they don't use our custom retrying logic.
- Setting `OTEL_TRACES_EXPORTER` to an unsupported value will silently disable the extra OTEL exporter rather than raising an error. The idea is that it's a standard variable which (1) may be set by something other than the user without them knowing, and (2) may be consumed by something other than logfire.
- `OTEL_METRICS_EXPORTER` is similarly checked, it wasn't used at all before.
- Pending spans are only produced for logfire exporters (sending to the backend, the console, or the test exporters)


logfire/_internal/config.py
```diff
@@ -203,8 +205,6 @@ def configure(
         metric_readers: Legacy argument, use `additional_metric_readers` instead.
         additional_metric_readers: Sequence of metric readers to be used in addition to the default reader
             which exports metrics to Logfire's API.
-            Ensure that `preferred_temporality=logfire.METRICS_PREFERRED_TEMPORALITY`
-            is passed to the constructor of metric readers/exporters that accept the `preferred_temporality` argument.
```
alexmojaki, line 207: Thinking about exporting to other backends, I realize that this is somewhat opinionated, and we shouldn't recommend it for all alternative metric readers.

docs/guides/advanced/alternative_backends.md
```diff
@@ -0,0 +1,56 @@
+# Alternative backends
+
+**Logfire** uses the OpenTelemetry standard. This means that you can configure the SDK to export to any backend that supports OpenTelemetry.
+
+The easiest way is to set the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable to a URL that points to your backend.
+This will be used as a base, and the SDK will append `/v1/traces` and `/v1/metrics` to the URL to send traces and metrics, respectively.
+
+Alternatively, you can use the `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` environment variables to specify the URLs for traces and metrics separately. These URLs should include the full path, including `/v1/traces` and `/v1/metrics`.
+
+The data will be encoded using Protobuf (not JSON) and sent over HTTP (not gRPC), so make sure that your backend supports this.
```
Kludex: I would add an `info` admonition or something to catch attention here.

Kludex (approved):

PR #353, opened 2024-08-05
Remove extra deps that were only for tests now skipped in 3.8

alexmojaki:
Since we skip test_mysql in Python 3.8, we don't need these extra dependencies that were only for the combination of `testcontainers`, mysql, and 3.8.

Also updated the requirements locks while I'm at it.

### DIFF
--- a/requirements-dev.lock
+++ b/requirements-dev.lock
@@ -9,33 +9,34 @@
 #   generate-hashes: false
 
 -e file:.
-aiohttp==3.9.5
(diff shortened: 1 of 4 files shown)


requirements-dev.lock
```diff
@@ -9,33 +9,34 @@
 #   generate-hashes: false
 
 -e file:.
-aiohttp==3.9.5
+aiohappyeyeballs==2.3.4
```
alexmojaki, line 12: wtf

PR #359, opened 2024-08-05
Remove `TypeAlias` from code source

Kludex:
The `TypeAlias` that we have are correctly used, but `stubgen` doesn't work properly with them. It creates `Incomplete` types if you use `TypeAlias`, which on the user's side you have `exc_info: Unkown`, and for users with strict typing configuration it's bothersome.


Kludex:
I didn't check the why, nor if they have issues mentioning this, but I don't want to spend more time on this when the fix is simple.

alexmojaki (approved):

PR #361, opened 2024-08-05
Turn `ParamSpec` non-private

Kludex:


logfire-api/logfire_api/_internal/instrument.pyi
```diff
@@ -3,9 +3,13 @@ from .ast_utils import BaseTransformer as BaseTransformer, LogfireArgs as Logfir
 from .main import Logfire as Logfire
 from dataclasses import dataclass
 from types import CodeType
-from typing import Callable
+from typing import Callable, TypeVar
+from typing_extensions import ParamSpec
 
-def instrument(logfire: Logfire, args: LogfireArgs) -> Callable[[Callable[_PARAMS, _RETURN]], Callable[_PARAMS, _RETURN]]: ...
+P = ParamSpec('P')
+R = TypeVar('R')
```
Kludex, line 10: Those are not there because `_PARAMS` and `_RETURN` are private, and we are not generating the stubs with private objects.

Issue #367, opened 2024-08-06
logfire 0.50 requires pytest


baggiponte:
### Description

When I launch a service instrumented with logfire I get this error:

```
dice-player-1  | Traceback (most recent call last):
dice-player-1  |   File "/usr/local/bin/uvicorn", line 8, in <module>
dice-player-1  |     sys.exit(main())
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
dice-player-1  |     return self.main(*args, **kwargs)
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1078, in main
dice-player-1  |     rv = self.invoke(ctx)
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
dice-player-1  |     return ctx.invoke(self.callback, **ctx.params)
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/click/core.py", line 783, in invoke
dice-player-1  |     return __callback(*args, **kwargs)
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/uvicorn/main.py", line 410, in main
dice-player-1  |     run(
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/uvicorn/main.py", line 577, in run
dice-player-1  |     server.run()
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/uvicorn/server.py", line 65, in run
dice-player-1  |     return asyncio.run(self.serve(sockets=sockets))
dice-player-1  |   File "/usr/local/lib/python3.10/asyncio/runners.py", line 44, in run
dice-player-1  |     return loop.run_until_complete(main)
dice-player-1  |   File "/usr/local/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
dice-player-1  |     return future.result()
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/uvicorn/server.py", line 69, in serve
dice-player-1  |     await self._serve(sockets)
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/uvicorn/server.py", line 76, in _serve
dice-player-1  |     config.load()
dice-player-1  |   File "/usr/local/lib/python3.10/site-packages/uvicorn/config.py", line 434, in load
dice-player-1  |     self.loaded_app = import_from_string(self.app)
[...]

PR #368, opened 2024-08-06
Move TestExporter to avoid requiring pytest

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/367

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -52,7 +52,6 @@ from typing_extensions import Self
 from logfire.exceptions import LogfireConfigError
 from logfire.version import VERSION
 
-from ..testing import TestExporter
 from .auth import DEFAULT_FILE, DefaultFile, is_logged_in
 from .collect_system_info import collect_package_info
 from .config_params import ParamManager, PydanticPluginRecordValues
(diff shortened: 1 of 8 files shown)


logfire/_internal/config.py
```diff
@@ -75,6 +74,7 @@
 from .exporters.quiet_metrics import QuietMetricExporter
 from .exporters.remove_pending import RemovePendingSpansExporter
 from .exporters.tail_sampling import TailSamplingOptions, TailSamplingProcessor
+from .exporters.test import TestExporter
```
alexmojaki, line 77: @Kludex it's not critical but I don't know why there isn't a stub for `exporters.test` when I run `rye run generate-stubs`. There's other stubs in the `exporters` folder.

Kludex: I think it's because the `testing` module is not in the `__init__.py`.

alexmojaki: I'm talking about `logfire._internal.exporters.test`, not `logfire.testing`

Kludex (approved):

logfire/_internal/config.py
Kludex: Ah, because we don't include private objects.

Kludex: We'd need to add `--include-private` on the `stubgen` command for it to appear, I think...

alexmojaki: But all the other internal exporter files are there

PR #371, opened 2024-08-08
Disable ASGI send/receive spans by default

alexmojaki:
A user reported that generating many `http receive` spans significantly slowed down requests: https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1722626116726899

In addition to these spans adding overhead in user code, they will eventually have a significant cost, since we will be charging based on the number of spans. This is out of proportion with the value they provide, especially since they're hidden by default in the UI.

So by default we're now preventing them from being created. The user can still opt in with `record_send_receive=True` in `instrument_fastapi/starlette`, in which case they will still have the debug log level.

We *might* want to eventually move the logic in `processor_wrapper.py` that sets the log level and span name to the new `TweakAsgiSpansTracer`. This will improve performance in general by avoiding checking whether we need to tweak every span on both start and end. But this will mean that the tweaking only works when going through `instrument_*` methods. At the very least this requires creating `instrument_asgi`.


tests/otel_integrations/test_fastapi.py
```diff
@@ -1508,3 +1512,360 @@ def test_scrubbing(client: TestClient, exporter: TestExporter) -> None:
             },
         ]
     )
+
+
+def make_request_hook_spans(record_send_receive: bool):
+    # Instrument an app with request hooks to make sure that they work.
+    # We make a new app here instead of using the fixtures because the same app can't be instrumented twice.
+    # Then make a request to it to generate spans.
+    # The tests then check that the spans are in the right place both with and without send/receive spans.
```
alexmojaki, line 1521: This is what the spans look like:

Screenshot 2024-08-08 at 13 45 09

PR #373, opened 2024-08-08
Add `logfire.instrument_system_metrics()`

alexmojaki:
## Breaking change!

With this, you must call `logfire.instrument_system_metrics()` to enable basic system metrics, previously system metrics were always sent if you had the appropriate package installed.

---

## Background:

In preparation for starting to charge we looked our usage from our larger beta users and realised a lot of companies were sending us a lot of system metrics which we are guessing they didn't care about, but would be charged a lot for.

This was because the standard [`opentelemetry-instrumentation-system-metrics`](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/system_metrics/system_metrics.html) logic was collecting around 400 metrics per minute per host — stuff like this:

| metric_name                    | count |
|--------------------------------|-------|
| system.disk.operations         | 74    |
| system.disk.time               | 74    |
| system.disk.io                 | 74    |
| system.cpu.time                | 64    |
| system.cpu.utilization         | 64    |
| system.memory.utilization      | 10    |
| system.memory.usage            | 10    |
| system.network.io              | 6     |
| system.network.dropped_packets | 6     |
| system.network.errors          | 6     |

(This is from looking at a machine running on render)

Most people don't need 219 datapoints on disk usage every minute!

## Solution

The solution (described in detail in the docs changes in this PR) are:
1. Make system metrics opt-in, `logfire` will now only send them if you call `logfire.instrument_system_metrics()`
2. Send a curated set of the most useful system metrics, reducing the metric volume by around 98% in common cases

Feedback on whether we should adjust the set of metrics sent in "basic" mode very welcome.


docs/integrations/system_metrics.md
```diff
@@ -1,28 +1,68 @@
...
+logfire.configure()
+
+logfire.instrument_system_metrics()
+```
+
+Then in your project, click on 'Dashboards' in the top bar, click 'New Dashboard', and select 'Basic System Metrics' from the dropdown.
+
+## Configuration
+
+By default, `instrument_system_metrics` collects only the metrics it needs to display the 'Basic System Metrics' dashboard. You can choose exactly which metrics to collect and how much data to collect about each metric. The default is equivalent to this:
```
alexmojaki: The current 'Basic System Metrics' dashboard template won't work with this new set of reduced metrics. I will create a PR with a new template. That PR will need to be ready to merge before these changes can be merged, and it should be deployed approximately when this is released.

When users upgrade the SDK, system metrics will stop being collected, if they were being collected at all. Since their collection was the implicit default, we can't warn them about this in code, and must rely on the changelog and announcements in slack. If the only thing that users do is add `logfire.instrument_system_metrics()` then existing basic system metrics dashboards will still not work, so we need to make it clear that they also need to switch to the new template.

The old template is still useful though, because it works with OTEL conventions, so system metrics collected by means other than the logfire SDK can be used with the old template. So we should keep and document both. We need to figure out how to name them to clarify the difference, e.g. 'Basic System Metrics (Logfire)' and 'Basic System Metrics (OpenTelemetry)'.

docs/integrations/system_metrics.md
```diff
@@ -1,28 +1,68 @@
...
+
+```py
+logfire.instrument_system_metrics({
+    'system.cpu.simple_utilization': None,
```
alexmojaki: This is a new metric I've implemented. It's similar to `system.cpu.utilization` but without any attributes, meaning just one data point per export instead of one data point per CPU core per CPU state. It aggregates nicely over the 'busy' states and the CPU cores.

In the future we might want to add similar metrics which have a single attribute, e.g. `simple_utilization_by_cpu` and `simple_utilization_by_mode`, to get more detail without multiplying the number of rows. Not a priority right now.

I'm a little hesitant to create new metric names under the official `system.cpu.` namespace, but it seems probably OK. My impression from https://opentelemetry.io/docs/specs/semconv/general/metrics/ and https://opentelemetry.io/docs/specs/semconv/system/system-metrics/ was that this was better than adding a `logfire.` prefix.

docs/integrations/system_metrics.md
```diff
@@ -1,28 +1,68 @@
...
+The [`logfire.instrument_system_metrics()`][logfire.Logfire.instrument_system_metrics] method can be used to collect system metrics with **Logfire**, such as CPU and memory usage.
 
-To enable metrics, you need to install the `logfire[system-metrics]` extra:
+## Installation
+
+Install `logfire` with the `system-metrics` extra:
 
 {{ install_logfire(extras=['system-metrics']) }}
 
-### Available Metrics
```
alexmojaki, line 7: This section was not very informative, and is even less useful now that the full config dict is documented. In the future I'd like to add a section documenting each metric in more detail, but it's not a priority.

docs/integrations/system_metrics.md
alexmojaki: PR here: https://github.com/pydantic/platform/pull/3838

logfire-api/logfire_api/_internal/integrations/system_metrics.pyi
```diff
@@ -0,0 +1,16 @@
+from _typeshed import Incomplete
+from logfire import Logfire as Logfire
+from typing import Iterable
+from typing_extensions import LiteralString
+
+MetricName: Incomplete
```
alexmojaki: @Kludex why is this Incomplete?

Kludex: Good question. I usually try stuff, and look if there are issues on mypy repository...

If you add the `type[Literal["potato"]]` explicitly, does it work?

alexmojaki: sort of :/

docs/integrations/system_metrics.md
samuelcolvin: Can you add a pop over explaining what `None` means or linking to the description below.

logfire/_internal/config.py
```diff
@@ -192,8 +192,7 @@ def configure(
             `LOGFIRE_CONFIG_DIR` environment variable, otherwise defaults to the current working directory.
         data_dir: Directory to store credentials, and logs. If `None` uses the `LOGFIRE_CREDENTIALS_DIR` environment variable, otherwise defaults to `'.logfire'`.
         base_url: Root URL for the Logfire API. If `None` uses the `LOGFIRE_BASE_URL` environment variable, otherwise defaults to https://logfire-api.pydantic.dev.
-        collect_system_metrics: Whether to collect system metrics like CPU and memory usage. If `None` uses the `LOGFIRE_COLLECT_SYSTEM_METRICS` environment variable,
-            otherwise defaults to `True`.
+        collect_system_metrics: Legacy argument, use `logfire.instrument_system_metrics()` instead.
```
samuelcolvin: can we make `logfire.instrument_system_metrics()` a link?

logfire/_internal/integrations/system_metrics.py
```diff
@@ -0,0 +1,189 @@
...
+    from logfire import Logfire
+
+try:
+    import psutil
+    from opentelemetry.instrumentation.system_metrics import (
+        _DEFAULT_CONFIG,  # type: ignore
+        SystemMetricsInstrumentor,
+    )
+except ModuleNotFoundError as e:  # pragma: no cover
+    raise RuntimeError(
```
samuelcolvin, line 21:
```suggestion
    raise ImportError(
```
?

docs/integrations/system_metrics.md
```diff
@@ -1,28 +1,68 @@
...
+
+```py
+logfire.instrument_system_metrics({
+    'system.cpu.simple_utilization': None,
+    'system.memory.utilization': ['available'],
+    'system.swap.utilization': ['used'],
+})
+```
+
+To collect lots of detailed data about all available metrics, use `logfire.instrument_system_metrics(base='full')`. This is equivalent to:
```
samuelcolvin: Maybe good to have a warning here saying roughly how many metrics this will generate, and that that might be expensive if run full time, especially on multiple machines.

samuelcolvin (commented):
Sorry to be so slow, otherwise looks great.

logfire/_internal/integrations/system_metrics.py
```diff
@@ -0,0 +1,189 @@
...
+            # So we can't use `# type: ignore` here, because on Linux it's not needed.
+            if not TYPE_CHECKING:  # pragma: no branch
+                cpu_num = process.cpu_num()
+                # `cpu_num` can be -1 on some platforms according to psutil.
+                if cpu_num >= 0:  # pragma: no branch
+                    percents.append(psutil.cpu_percent(percpu=True)[cpu_num])
+
+        # Return the highest of the three values.
+        # This means interpreting the value is not straightforward,
+        # but any unusual activity will show a notable spike, regardless of the infra setup.
```
alexmojaki: @samuelcolvin @dmontagu I'm having second thoughts about this. Maybe I should split this into 3 metrics:

- `system.cpu.simple_utilization`: just the average across all cores
- `process.runtime.cpu.utilization`: just the usage of this process - this is already an existing standard metric, it's part of the full config
- `process.core.cpu.utilization` (made up naming): just the usage of the core this process is using, if available.

Then the dashboard query would take the max of the three values at each recorded timestamp.

Pro: more detailed and interpretable data, less opinionated

Cons:

- The meaning of the defaults that the user must copy from the docs if they want to export a subset of those metrics goes from 3 metrics to 5:

```diff
logfire.instrument_system_metrics({
    'system.cpu.simple_utilization': None,
+  'process.runtime.cpu.utilization': None,
+  'process.core.cpu.utilization': None,
    'system.memory.utilization': ['available'],
    'system.swap.utilization': ['used'],
})
```

- More complex queries, particularly if users want to analyze CPU usage themselves.
- More data points. Probably not worth worrying about since it's still just 5 series, but it's 66% more than before.

logfire/_internal/integrations/system_metrics.py
alexmojaki: This is just following an existing pattern in the codebase:

https://github.com/pydantic/logfire/blob/fe127144cfe89e221e92268317383b1cfa7cfd79/logfire/_internal/integrations/fastapi.py#L27

https://github.com/pydantic/logfire/blob/fe127144cfe89e221e92268317383b1cfa7cfd79/logfire/_internal/integrations/django.py#L8

@Kludex any reason you chose `RuntimeError` for fastapi?

logfire/_internal/integrations/system_metrics.py
samuelcolvin: My weak preference would be to record:
* `system.cpu.simple_utilization`
*  and `process.runtime.cpu.utilization`

I think those are the things people will want in almost all cases.

alexmojaki: The problem with `process.runtime.cpu.utilization` is that if you have 5 processes running at full throttle on a single CPU, each will record 20% usage.

I suppose in that case `system.cpu.simple_utilization` will probably be high, I'm just not sure how reliable that is. I think some have mentioned that in some containerized environments on shared instances the CPU usage stats will include the usage of cores that the user doesn't actually have permission to use.

docs/integrations/system_metrics.md
```diff
@@ -1,28 +1,80 @@
...
+})
+```
+
+1. `None` simply means that there are no fields to configure for this metric.
+
+To collect lots of detailed data about all available metrics, use `logfire.instrument_system_metrics(base='full')`.
+
+!!! warning
+    The amount of data collected by `base='full'` can be expensive, especially if you have many servers,
+    and this is easy to forget about. Be sure to monitor your usage and costs.
```
dmontagu:
```suggestion
    and this is easy to forget about. If you enable this, be sure to monitor your usage and costs.
```
I mean, I don't think it's a bad idea to monitor the usage more generally, but imo this phrasing better expresses the intent

docs/integrations/system_metrics.md
```diff
@@ -1,28 +1,80 @@
...
+
+    The most expensive metrics are `system.cpu.utilization/time` which collect data for each core and each mode,
+    and `system.disk.*` which collect data for each disk device. This can result in hundreds of data points per minute.
+    The exact number depends on your machine.
```
dmontagu:
```suggestion
    and `system.disk.*` which collect data for each disk device. The exact number depends on the machine hardware,
    but this can result in hundreds of data points per minute from each instrumented host.
```
I think useful to make it clear that the number of hosts is another multiplier here.

docs/integrations/system_metrics.md
```diff
@@ -1,28 +1,80 @@
...
+    'process.runtime.thread_count': None,
+    'process.runtime.cpu.utilization': None,
+    'process.runtime.context_switches': ['involuntary', 'voluntary'],
+    'process.open_file_descriptor.count': None,
+})
+```
+
+Each key here is a metric name. The values have different meanings for different metrics. For example, for `system.cpu.utilization`, the value is a list of CPU modes. So there will be a separate row for each CPU core saying what percentage of time it spent idle, another row for the time spent waiting for IO, etc. There are no fields to configure for `system.thread_count`, so the value is `None`.
+
+The first dict argument is merged with the base. For example, if you want to collect disk read operations (but not writes) you can write:
```
dmontagu:
```suggestion
For convenient customizability, the first dict argument is merged with the base. For example, if you want to collect disk read operations (but not writes) you can write:
```

logfire/_internal/integrations/system_metrics.py
dmontagu: For what it's worth, I think we should be summing CPU usage in most scenarios rather than averaging, and rely on the user being aware of the capacity of the machine.

The number of CPUs also seems to me to be a reasonable thing to include in resource attributes, if anywhere.

alexmojaki: OK, I'm going to split it up into `system.cpu.simple_utilization` and `process.runtime.cpu.utilization` and drop the third value. However, `process.runtime.cpu.utilization` in the config doesn't actually mean that OTEL will export a metric called that, it'll name the metric something like `process.runtime.cpython.cpu.utilization` and it depends on the python implementation. The reason for this is in https://opentelemetry.io/docs/specs/semconv/general/metrics/#general-metric-semantic-conventions:

> Semantic ambiguity SHOULD be avoided. Use prefixed metric names in cases where similar metrics have significantly different implementations across the breadth of all existing metrics. For example, every garbage collected runtime has slightly different strategies and measures. Using a single set of metric names for GC, not divided by the runtime, could create dissimilar comparisons and confusion for end users. (For example, prefer `process.runtime.java.gc*` over `process.runtime.gc.*`.) Measures of many operating system metrics are similarly ambiguous.

I think OTEL made a bit of a mistake here. The way CPU usage is measured depends on the language, and garbage collection depends on the Python implementation, but all Python implementations measure CPU usage the same way via psutil. I'm thinking I should add a metric called `process.runtime.python.cpu.utilization` (that's `python`, not `cpython`) so that the name is stable. Then the dashboard query can just check `metric_name = 'process.runtime.python.cpu.utilization'`. I worry that something like `metric_name like 'process.runtime.%.cpu.utilization'` might not be reliably efficient, especially with variadic partitioning on metric name in FF.

PR #379, opened 2024-08-14
Add py.typed file to logfire-api

jackmpcollins:
Add a `py.typed` file to `logfire-api` so type hints are correctly resolved.


logfire-api/logfire_api/__init__.pyi
jackmpcollins, line 1: > Does the `rye generate-stubs` command erases it?

It doesn't erase the py.typed file, but it did make changes here.

jackmpcollins: Reverted now as it appears to make tests fail.

alexmojaki (approved):
Thanks!

PR #385, opened 2024-08-19
Add docs page warning about generators

alexmojaki:
Step 1 of https://pydantic.slack.com/archives/C05AF4A4WRM/p1715430326710039


dmontagu (approved):

dmontagu:
for what it's worth, I think it's more likely that someone would discover they need to read this starting from the error message. Not sure if there's a good way for us to make it easier to go from error message -> answer as opposed to answer -> error message (other than google, which maybe is enough).

PR #386, opened 2024-08-19
Don't auto-trace generators

alexmojaki:


davidhewitt (commented):
This reminds me a lot of rust `async` functions (which are very similar to generators). The way that the rust tracing library handles this is that it adds the span to the current context only while doing work.

Something like this:

```python
async def generate_items():
    with logfire.span('Generating items', end_on_exit=False) as span:
        for i in range(3):
            span.__exit__(None, None, None)
            yield i
            span.__enter__()  # BUG: it looks like __enter__ sets end_on_exit=True
    span.end()
```

I think that translation doesn't exactly work as-is, but maybe there's a way to come up with a similar API?

alexmojaki:
I would quite like to implement something like this someday, particularly to measure the time spent blocking in an async function (https://linear.app/pydantic/issue/PYD-595/measure-time-spent-blockingwaiting-in-coroutinesgenerators-using) but today is not that day.

davidhewitt (approved):

PR #407, opened 2024-08-29
Tail sampling

alexmojaki:
Adds a new parameter `sampling: SamplingOptions` to `logfire.configure()` which can configure both head and tail sampling.

The `trace_sample_rate` parameter has been deprecated, the new equivalent is `sampling=logfire.SamplingOptions(head=...)`. However the `trace_sample_rate` key in `pyproject.toml` and the `LOGFIRE_TRACE_SAMPLE_RATE` env var will still work without warnings. Not sure what to do about that and how much it's worth thinking about at the moment.


docs/guides/advanced/sampling.md
```diff
@@ -1,38 +1,313 @@
...
+stored and analyzed. Sampling is a trade-off between cost and completeness of data.
 
-To configure sampling for the SDK:
+_Head sampling_ means the decision to sample is made at the beginning of a trace. This is simpler and more common.
 
-- Set the [`trace_sample_rate`][logfire.configure(trace_sample_rate)] option of [`logfire.configure()`][logfire.configure] to a number between 0 and 1, or
-- Set the `LOGFIRE_TRACE_SAMPLE_RATE` environment variable, or
-- Set the `trace_sample_rate` config file option.
+_Tail
+sampling_ means the decision to sample is delayed, possibly until the end of a trace. This means there is more
```
Kludex:
```suggestion
_Tail sampling_ means the decision to sample is delayed, possibly until the end of a trace. This means there is more
```

Kludex (approved):

PR #409, opened 2024-08-30
Add `CodeSource` parameter

Kludex:


logfire/_internal/constants.py
```diff
@@ -145,6 +145,15 @@ def log_level_attributes(level: LevelName | int) -> dict[str, otel_types.Attribu
 RESOURCE_ATTRIBUTES_PACKAGE_VERSIONS = 'logfire.package_versions'
 """Versions of installed packages, serialized as list of json objects with keys 'name' and 'version'."""
 
+RESOURCE_ATTRIBUTES_CODE_GIT_REF = 'logfire.code.git_ref'
```
alexmojaki, line 148: In general we should try following semantic conventions where they exist. With a bit of effort I found https://opentelemetry.io/docs/specs/semconv/attributes-registry/vcs/. This would be `vcs.repository.ref.revision`.

Just noting that, not saying we should go with this overall approach or spend time on this now.

Kludex: Ah, I didn't find it. Nice. Thanks.

Kludex: They are experimental. They are not included in the `opentelemetry.semconv` package yet. What should I do?

Kludex:
If I use [opentelemetry.io/docs/specs/semconv/attributes-registry/vcs](https://opentelemetry.io/docs/specs/semconv/attributes-registry/vcs/), what do I do with the "root_path"?

Kludex:
I'll open another PR.

alexmojaki:
I'd like to discuss the approach a bit more when we're under less pressure from GA and stuff.

PR #414, opened 2024-09-05
Use `birthday`, not `age` on original example

sydney-runkle:
A more intuitive question for the intro example


sydney-runkle:
I've broken everything 😭

PR #419, opened 2024-09-11
Update screenshots on docs intro page

ChristopherGS:
Updated screenshots to show new UI and removed Beta button


docs/index.md
```diff
@@ -225,9 +225,13 @@ and it'll end up as structured data in our platform ready to be queried.
 For example, using data from the `User` model above, we could list users from the USA:
 
 ```sql
-SELECT attributes->'result'->>'name' as name, age(attributes->'result'->>'dob') as age
+SELECT
```
ChristopherGS: when I ran this query, I got this error:
image

So I updated with a query that does work

.gitignore
```diff
@@ -9,3 +9,6 @@ __pycache__
 
 # stubgen
 out
+
+# intellij
+.idea
```
Kludex: I prefer for ppl to use their own global `.gitignore` on IDE specific files.

```suggestion
```

Kludex (approved):
Screenshots looks fine.

docs/index.md
alexmojaki: I personally significantly prefer the original example. I suggest:

```sql
SELECT attributes->'result'->>'name' as name, extract(year from (attributes->'result'->>'dob')::date) as "birth year"
FROM records
WHERE attributes->'result'->>'country_code' = 'USA'
```

to demonstrate some fancy-looking SQL. Calculating the age would be better, but apparently datafusion doesn't support extracting years from durations.

.gitignore
alexmojaki: What's the harm? Does .gitignore need to be clean?

alexmojaki:
It's not important, but I think future screenshots should probably have a profile pic in them.

docs/images/index/logfire-screenshot-search-query.png
alexmojaki, line 1: The mixture of with and without service names doesn't look very good on the left, nor does `unknown_service` on the right (although other people probably wouldn't notice that bit). It'd also be nice if the service names weren't truncated to `example-s...` and `example-a...`. How about just `api` and `worker`?

.gitignore
Kludex: I wouldn't accept this on the projects I maintain cause it opens space for ppl to keep adding stuff. I prefer correctness. The gitignore for a project should only contain what is generated from the project itself but shouldn't be committed.

docs/index.md
ChristopherGS: updated

.gitignore
ChristopherGS: removed

docs/images/index/logfire-screenshot-search-query.png
ChristopherGS: fixed

PR #424, opened 2024-09-15
Add uv installation instructions to docs

montasaurus:
Adds installation commands for the [uv](https://docs.astral.sh/uv/) package manager.


Kludex (approved):
Thanks! :)

PR #428, opened 2024-09-17
Deprecate `project_name` in `logfire.configure()`, remove old kwargs from signature 

alexmojaki:
Deprecated kwargs are now moved into `**kwargs: DeprecatedKwargs` and undocumented to leave the remaining docs cleaner.

Internal discussion: https://pydantic.slack.com/archives/C05AF4A4WRM/p1725302678666539

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -14,7 +14,7 @@ from dataclasses import dataclass, field
 from functools import cached_property
 from pathlib import Path
 from threading import RLock, Thread
-from typing import TYPE_CHECKING, Any, Callable, Literal, Sequence, cast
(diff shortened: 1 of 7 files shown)


logfire/_internal/config.py
```diff
@@ -144,11 +144,14 @@ class PydanticPlugin:
     """Exclude specific modules from instrumentation."""
 
 
-def configure(
+class DeprecatedKwargs(TypedDict):
+    pass
```
Kludex, line 149: By "undocumented" you mean you don't have type hints? and that's why you didn't add them here? 🤔

alexmojaki: yes, added a comment

Kludex (approved):

PR #432, opened 2024-09-18
Move `base_url`, `id_generator`, and `ns_timestamp_generator` parameters of `logfire.configure()` into `advanced: AdvancedOptions`

alexmojaki:
Internal discussion: https://pydantic.slack.com/archives/C05AF4A4WRM/p1725302678666539


Kludex:
Can we not move `base_url`? I use that all the time... I understand the goal is exactly to move things only logfire developers use, but for me it's a huge improvement of my developer experience...

alexmojaki:
@Kludex `base_url` will now only emit a warning, not an error.

Kludex (approved):

PR #444, opened 2024-09-25
Add `metrics` parameter to `logfire.configure()`

alexmojaki:
Internal discussion: https://pydantic.slack.com/archives/C05AF4A4WRM/p1727260278734989

This allows users to pass `metrics=False` to disable sending metrics entirely, since OTEL instrumentations often produce metrics where users only want/need the spans.

Alternatively users can pass `logfire.MetricsOptions(...)`. For now the only option is `additional_readers` which replaces the existing top-level `additional_metric_readers` parameter of `configure`. In the future I want to add other options, including a way to filter specific metrics instead of just disabling them all, and a way to create metrics from spans in the SDK.


alexmojaki:
https://github.com/pydantic/logfire/pull/263#discussion_r1649080146

It seems that PyCharm is handling a plain dict being passed to a TypedDict param more cleverly than I remember in terms of stuff like autocomplete. Maybe things have improved and this should be reconsidered.

PR #446, opened 2024-09-25
Remove default `min_duration` for `install_auto_tracing`

alexmojaki:
Internal discussion: https://pydantic.slack.com/archives/C05AF4A4WRM/p1727175487605829


Kludex (approved):

Kludex:
I think 0.1 is a good default, it's the same number `PYTHONASYNCIODEBUG` uses. 🤷‍♂️

PR #451, opened 2024-09-26
Add `CodeSource` parameter to `configure`

Kludex:
I'll add documentation after this is supported by the frontend.


logfire/_internal/config.py
```diff
@@ -174,6 +177,25 @@ class MetricsOptions:
...
+    """The repository URL for the code e.g. https://github.com/pydantic/logfire"""
+
+    revision: str
+    """The git revision of the code e.g. branch name, commit hash, tag name etc."""
+
+    root_path: str
+    """The root path for the source code in the repository.
+
+    Example:
+        If the `code.filename` is `/path/to/project/src/logfire/main.py` and the `root_path` is `src/`, the URL
```
alexmojaki, line 213: This is a long and not very readable example.

It's `code.filepath`, not `filename`.

`code.filepath` is not necessarily something users will know about. It's relative to the current working directory, so it probably shouldn't start with `/`. And users are likely to not know what the current working directory is or will be when running the app, so setting this correctly may be tricky.

logfire/_internal/config.py
```diff
@@ -174,6 +177,25 @@ class MetricsOptions:
...
+class CodeSource:
+    """Settings for the source code of the project."""
+
+    repository: str
+    """The repository URL for the code e.g. https://github.com/pydantic/logfire"""
+
+    revision: str
+    """The git revision of the code e.g. branch name, commit hash, tag name etc."""
+
+    root_path: str
```
alexmojaki, line 209: Should this default to `''`?

logfire/_internal/config.py
```diff
@@ -174,6 +177,25 @@ class MetricsOptions:
     """Sequence of metric readers to be used in addition to the default which exports metrics to Logfire's API."""
 
 
+@dataclass
+class CodeSource:
```
alexmojaki, line 196: Should this be `CodeSourceOptions`? I genuinely don't know.

Kludex: It just seems cleaner the way it is, even if the others have the `Options`.

logfire/_internal/config.py
Kludex: Should I default to `.`?

logfire/_internal/config.py
Kludex: Suggestions?

Kludex: I'm planning to document this feature btw... 

This PR is just the first step. After the feature is on the frontend, I'll add to the docs.

Kludex:
@alexmojaki Your comments seem not that blocking, can you check them and tell me what would make you happy? I want to push this forward.

Kludex:
We can even put the feature under experimental or something?

alexmojaki:
I think just document the dataclass and parameter as not ready for public use and then it's fine to go ahead.

Or just don't change the SDK for now if it's only going to add resource attributes for the sake of developing the frontend, those can already be set in environment variables.

alexmojaki (approved):

logfire/_internal/config.py
alexmojaki: > Logfire spans automatically have a `code.filepath` attribute indicating where they were created. That path is relative to the current working directory of the Python process. The `root_path` setting is combined with the `code.filepath` attribute to construct a link, so `root_path` should be a path starting from the root of the repository that corresponds to the current working directory of the Python process.

> For example, suppose that your repository contains `a/b/c/main.py`, the folder `a/b/` is copied into the `/docker/root/` folder of your docker container, and within the container the command `python ./b/c/main.py` is run from within the `/docker/root/a/` directory. Then `code.filepath` will be `b/c/main.py` for spans created in that file, and the `root_path` should be set to `a` so that the final link is `a/b/c/main.py`.

PR #458, opened 2024-09-27
Upgrade `DeprecationWarning`s to `UserWarning`s

alexmojaki:
`DeprecationWarning` is hidden by default. This will mean using deprecated args will start making more noise. Only to be merged when making the v1.0 release.


Kludex:
Well... They are deprecation warnings...

PR #463, opened 2024-10-02
Improve getting started docs images

ChristopherGS:
Fixes images to show just the code from the snippet above it


alexmojaki:
"Arguments (as Python)" look better than "as JSON" because they're rendered with Python syntax. Click the "as JSON" to switch mode.

sydney-runkle (approved):
Nice, thanks!

ChristopherGS:
> "Arguments (as Python)" look better than "as JSON" because they're rendered with Python syntax. Click the "as JSON" to switch mode.

@alexmojaki the new images display `Arguments (as Python)` as your comment suggests. OK to merge?

alexmojaki:
Ah, sorry, I got confused.

I'd like it if the details panel on the right looked less squashed, but it's not essential.

ChristopherGS:
> Ah, sorry, I got confused.
> 
> I'd like it if the details panel on the right looked less squashed, but it's not essential.

I know what you mean. The reason it's like that is to show more of the log/span text so the docs reader groks the link with the code snippet. Not an ideal trade off but if you don't view it as a blocker I'll merge. Happy to revisit

PR #480, opened 2024-10-07
Move from Rye to UV

Kludex:


alexmojaki:
Unsubscribed, please re-request review once CI passes

logfire-api/logfire_api/__init__.py
```diff
@@ -81,7 +81,7 @@ def with_settings(self, *args, **kwargs) -> Logfire:
 
             def force_flush(self, *args, **kwargs) -> None: ...
 
-            def log_slow_async_callbacks(self, *args, **kwargs) -> None:  # pragma: no branch
+            def log_slow_async_callbacks(self, *args, **kwargs) -> None:  # pragma: no cover
```
alexmojaki, line 84: ??

Makefile
```diff
@@ -1,59 +1,64 @@
...
 .PHONY: testcov  # Run tests and generate a coverage report
 testcov: test
 	@echo "building coverage html"
-	@rye run coverage html --show-contexts
+	@uv run coverage html --show-contexts
 
 .PHONY: docs  # Build the documentation
 docs:
-	rye run docs
+	mkdocs build
```
alexmojaki: should this use uv run?

Makefile
```diff
@@ -1,59 +1,64 @@
...
 generate-stubs:
-	rye run generate-stubs
+	stubgen -p logfire --include-docstrings --no-analysis
+	rsync -a out/logfire/ logfire-api/logfire_api/
 
 .PHONY: testcov  # Run tests and generate a coverage report
 testcov: test
 	@echo "building coverage html"
-	@rye run coverage html --show-contexts
+	@uv run coverage html --show-contexts
```
alexmojaki: what's with the `@`?

Makefile
```diff
@@ -1,59 +1,64 @@
...
 .PHONY: lint  # Lint the code
 lint:
-	rye lint
-	rye format --check
+	uv run ruff check
+	uv run ruff format --check --diff
+
+.PHONY: typecheck  # Typecheck the code
+typecheck:
+	uv run pyright $(sources)
```
alexmojaki: what does `$(sources)` do?

.pre-commit-config.yaml
```diff
@@ -25,8 +25,8 @@ repos:
         types: [python]
       - id: pyright
         name: pyright
-        entry: rye
+        entry: uv
```
alexmojaki: what about `make` here?

.github/workflows/main.yml
```diff
@@ -93,24 +103,34 @@ jobs:
...
           merge-multiple: true
           path: coverage
-      - run: pip install coverage[toml]
-      - run: ls -lh coverage
-      - run: coverage combine coverage
-      - run: coverage xml
+
+      - run: uv pip install coverage[toml]
+      - run: uv run coverage combine coverage
+      - run: uv run coverage xml
```
alexmojaki, line 127: Is there not a way to 'activate' uv so that we don't have to remember to always use `uv run`? What about scripts that internally just use `python` or whatever other command?

.github/workflows/main.yml
```diff
@@ -134,27 +154,30 @@ jobs:
...
         with:
           version_file_path: pyproject.toml
 
-      - uses: eifinger/setup-rye@v2
-      - run: rye config --set-bool behavior.use-uv=true
-      - run: rye build
+      - run: uv build
 
       - name: Publish logfire to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
```
alexmojaki, line 175: Does this not likely use the `python` command or whatever? Does it work without `uv run`? Is there a way to dry run to test it?

.github/workflows/main.yml
```diff
@@ -57,26 +68,25 @@ jobs:
...
-          pip install uv
-          uv pip install "logfire-api @ file://logfire-api"
-      - run: rye show
+          version: "0.4.12"
+          enable-cache: true
+
+      - name: Set up Python ${{ matrix.python-version }}
+        run: uv python install ${{ matrix.python-version }}
+
+      - run: uv sync --frozen
```
alexmojaki: This means we use the lockfile. We want to be always using the latest versions of things in CI tests, that's important.

alexmojaki: Why do other `uv sync` lines here have `--python 3.12` but not this one?

.github/workflows/main.yml
```diff
@@ -93,24 +103,34 @@ jobs:
...
         with:
-          python-version: "3.12"
+          version: "0.4.12"
+          enable-cache: true
+
+      - name: Set up Python 3.12
+        run: uv python install 3.12
+
+      - name: Create venv
+        run: uv venv --python 3.12
```
alexmojaki, line 117: Why do we create a venv here but not elsewhere? Are we using it? Can it be activated to avoid needing `uv run`?

logfire-api/logfire_api/__init__.py
Kludex: I don't understand how "no branch" was working before... 🤔

Makefile
Kludex:
```suggestion
	uv run mkdocs build
```

.github/workflows/main.yml
Kludex: No. Also, the recommendation is to do this: https://github.com/Kludex/python-multipart/blob/5303590e767c8f6125729912923cddedb1cc0946/.github/workflows/publish.yml#L40-L59

Makefile
Kludex: It's just to not show the command

Makefile
Kludex: it was actually not needed

.pre-commit-config.yaml
Kludex: not related to the pr, but applied

.github/workflows/main.yml
Kludex: Because everywhere else we run `uv sync`.

.github/workflows/main.yml
```diff
@@ -16,34 +16,45 @@ jobs:
...
-      - run: rye config --set-bool behavior.use-uv=true
-      - run: rye sync --no-lock
+          version: "0.4.12"
+          enable-cache: true
+
+      - name: Set up Python 3.12
+        run: uv python install 3.12
+
+      - name: Install dependencies
+        run: uv sync --python 3.12 --frozen
```
alexmojaki, line 30: feels like linting should also be using latest dependencies

.github/workflows/main.yml
```diff
@@ -57,26 +68,25 @@ jobs:
...
-          python-version: ${{ matrix.python-version }}
-      - uses: eifinger/setup-rye@v2.6
-      - name: Pin python-version ${{ matrix.python-version }}
-        run: rye pin ${{ matrix.python-version }}
-      - run: rye config --set-bool behavior.use-uv=true
-      # Update all dependencies to the latest version possible
-      - run: rye sync --update-all
-      - run: |
-          pip install uv
-          uv pip install "logfire-api @ file://logfire-api"
```
sydney-runkle, line 78: Do we not need this anymore?

.github/workflows/main.yml
```diff
@@ -16,34 +16,45 @@ jobs:
...
-      - uses: actions/setup-python@v5
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
         with:
-          python-version: "3.8"
-      - uses: eifinger/setup-rye@v2.6
-      - run: rye config --set-bool behavior.use-uv=true
-      - run: rye sync --no-lock
+          version: "0.4.12"
```
sydney-runkle, line 23: Can we just use `latest`?

sydney-runkle (approved):
LGTM overall, a few questions...

.github/workflows/main.yml
Kludex: We can. I don't want to not pin the dependencies.

sydney-runkle: Ok

.github/workflows/main.yml
samuelcolvin: Apparently the modern take is that activating environments is an anti-pattern we should now avoid.

I suspect you can active `uv`, but commands like this should not assume you have.

.github/workflows/main.yml
Kludex: Nop, I've added logfire-api as a workspace on the pyproject.

.github/workflows/main.yml
Kludex: I'm maintaining the same behavior here.

alexmojaki: OK, it doesn't have to change here, but I'm questioning it.

Makefile
```diff
@@ -1,59 +1,63 @@
...
 
 .PHONY: test  # Run the tests
 test:
-	rye run coverage run -m pytest
+	uv run coverage run -m pytest
 
 .PHONY: generate-stubs  # Generate stubs for logfire-api
 generate-stubs:
-	rye run generate-stubs
+	stubgen -p logfire --include-docstrings --no-analysis
```
alexmojaki: uv run?

.github/workflows/main.yml
Kludex: I don't mind those, just the tests.

Makefile
Kludex: yep

alexmojaki (approved):

Issue #469, opened 2024-10-04
Update `__init__.pyi` after running `rye run generate-stubs`


alexmojaki:
When I run `rye run generate-stubs`, I immediately have to run `test_override_init_pyi` afterwards to fix `__init__.pyi`, during which the test fails. Then I run it again to make sure it passes. This should be automated within the rye command.

PR #482, opened 2024-10-08
Automate rerun of test for stubs update in `Makefile`

sydney-runkle:
Closes https://github.com/pydantic/logfire/issues/469

Also, updated the stubs :)


Makefile
```diff
@@ -36,6 +36,8 @@ test:
 generate-stubs:
 	uv run stubgen -p logfire --include-docstrings --no-analysis
 	rsync -a out/logfire/ logfire-api/logfire_api/
+	uv run pytest ./tests/test_logfire_api.py::test_override_init_pyi || true
```
sydney-runkle, line 41: `|| true` so that we ignore the test failure on the first pass, it should report as skipped on the second

Makefile
```diff
@@ -36,6 +36,8 @@ test:
...
+	uv run pytest ./tests/test_logfire_api.py::test_override_init_pyi
```
sydney-runkle, line 42: Lmk if you want me to check for `"skipped"` or something here

Makefile
alexmojaki: this should be a comment in the code

Makefile
```diff
@@ -36,6 +36,8 @@ test:
 generate-stubs:
 	uv run stubgen -p logfire --include-docstrings --no-analysis
 	rsync -a out/logfire/ logfire-api/logfire_api/
```
alexmojaki, line 38:
```suggestion
	rsync -a out/logfire/ logfire-api/logfire_api/
	rm -rf out
```

Sometimes things in `out` from other branches stay there and have caused me confusion.

alexmojaki:
Thanks!

Issue #474, opened 2024-10-04
Document excluding `/` as an example of `excluded_urls`


alexmojaki:
https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1722462772309099

PR #487, opened 2024-10-09
Add `excluded_urls` example for `/`

sydney-runkle:
Closes https://github.com/pydantic/logfire/issues/474


docs/integrations/fastapi.md
```diff
@@ -102,6 +102,40 @@ To avoid tracing certain URLs, you can specify a string of comma-separated regex
 - The environment variable `OTEL_PYTHON_FASTAPI_EXCLUDED_URLS`.
 - The environment variable `OTEL_PYTHON_EXCLUDED_URLS` (which will also apply to other instrumentation).
 
+!!! example
+    If you'd like to trace all URLs except the base `/` URL, you can use the following regex: `^https?://[^/]+/$`
```
dmontagu: Are all URLs guaranteed to end with a `/`? Like, I'd think you'd want `^https?://[^/]+` as your regex if it only needs to match a prefix, or `^https?://[^/]+.*$` if it needs to match the whole thing.

docs/integrations/fastapi.md
```diff
@@ -102,6 +102,40 @@ To avoid tracing certain URLs, you can specify a string of comma-separated regex
...
+    logfire.instrument_fastapi(app, excluded_urls='^https?://[^/]+/$')
+
+    if __name__ == '__main__':
+        import uvicorn
+
+        uvicorn.run(app)
+    ```
+
+    If you visit http://127.0.0.1:8000/, that matches the above regex, so no span will be sent to logfire.
+    If you visit http://127.0.0.1:8000/hello/ (or any other endpoing that's not `/`, for that matter), a span will be sent to logfire.
```
dmontagu: (I would expect `http://127.0.0.1:8000/hello` does not match this regex)

docs/integrations/fastapi.md
sydney-runkle: Yes good call, I've changed to `^https?://[^/]+.*$`

docs/integrations/fastapi.md
```diff
@@ -102,6 +102,40 @@ To avoid tracing certain URLs, you can specify a string of comma-separated regex
 - The environment variable `OTEL_PYTHON_FASTAPI_EXCLUDED_URLS`.
 - The environment variable `OTEL_PYTHON_EXCLUDED_URLS` (which will also apply to other instrumentation).
 
+!!! example
+    If you'd like to trace all URLs except the base `/` URL, you can use the following regex: `^https?://[^/]+.*$`
```
dmontagu:
```suggestion
    If you'd like to trace all URLs except the base `/` URL, you can use the following regex for `excluded_urls`: `^https?://[^/]+.*$`
```

docs/integrations/fastapi.md
```diff
@@ -102,6 +102,40 @@ To avoid tracing certain URLs, you can specify a string of comma-separated regex
...
+    If you'd like to trace all URLs except the base `/` URL, you can use the following regex for `excluded_urls`: `^https?://[^/]+/$`
+
+    Breaking it down:
+
+    * `^` matches the start of the string
+    * `https?` matches `http` or `https`
+    * `://` matches `://`
+    * `[^/]+` matches one or more characters that are not `/`
+    * `/` matches `/`
+    * `$` matches the end of the string
```
dmontagu:
```suggestion
    * `^` matches the start of the string
    * `https?` matches `http` or `https`
    * `://` matches `://`
    * `[^/]+` matches one or more characters that are not `/` (this will be the host part of the URL)
    * `/` matches `/`
    * `$` matches the end of the string
    
    So this regex will only match routes that have no path after the host.
```

docs/integrations/fastapi.md
```diff
@@ -102,6 +102,40 @@ To avoid tracing certain URLs, you can specify a string of comma-separated regex
...
+    logfire.instrument_fastapi(app, excluded_urls='^https?://[^/]+/$')
+
+    if __name__ == '__main__':
+        import uvicorn
+
+        uvicorn.run(app)
+    ```
+
+    If you visit http://127.0.0.1:8000/, that matches the above regex, so no span will be sent to logfire.
+    If you visit http://127.0.0.1:8000/hello/ (or any other endpoing that's not `/`, for that matter), a span will be sent to logfire.
```
dmontagu:
```suggestion
    If you visit http://127.0.0.1:8000/hello/ (or any other endpoing that's not `/`, for that matter), a span will be sent to logfire.
    
    Note that under the hood, the `opentelemetry` library is using `re.search` (not `re.match` or `re.fullmatch`) to check for a match between the route and the `excluded_urls` regex, which is why we need to include the `^` at the start and `$` at the end of the regex.
```

dmontagu (approved):

docs/integrations/fastapi.md
```diff
@@ -102,6 +102,44 @@ To avoid tracing certain URLs, you can specify a string of comma-separated regex
...
```
dmontagu:
```suggestion
    If you visit http://127.0.0.1:8000/, that matches the above regex, so no tracing data will be recorded or sent to Logfire.
    If you visit http://127.0.0.1:8000/hello/ (or any other endpoing that's not `/`, for that matter), a trace will be started and sent to Logfire.
```
@alexmojaki is this right though? In particular, what happens if a user manually starts spans inside their endpoint, and isn't relying on the middleware. Does that misbehave because it just opens a trace below the middleware? I think what a user would _want_ to happen is that instrumentation is suppressed inside an excluded URL but I'm not sure if that's how it works now, and if it isn't, we should probably explain that..

alexmojaki: Yes, we should explain that it currently only prevents producing spans that come directly from the instrumentation.

alexmojaki:
Maybe this should be a section under https://logfire.pydantic.dev/docs/integrations/use-cases/web-frameworks/ which is linked to from specific integrations the way that capturing headers is, since this applies to other integrations as well, not just fastapi.

docs/integrations/fastapi.md
sydney-runkle: @alexmojaki, so do you think we should 1) apply the above suggestion and 2) add further clarification?

sydney-runkle:
Didn't get to this one today, will apply these suggestions tomorrow! Thanks for the feedback, folks!

docs/integrations/fastapi.md
alexmojaki: I think @dmontagu's suggestion was meant to imply that it suppresses all tracing within, and we want to clarify that it doesn't, so I think no on the suggestion.

sydney-runkle: Gotcha. I've reverted the suggestion and added an additional note.

sydney-runkle:
@alexmojaki, should be ready for review again 👍

docs/integrations/asgi.md
```diff
@@ -44,6 +44,11 @@ if __name__ == "__main__":
 
 You can read more about the OpenTelemetry ASGI middleware [here][opentelemetry-asgi].
 
+## Excluding URLs from instrumentation
+<!-- note that this section is duplicated for different frameworks but with slightly different links -->
+
+- [Quick guide](use-cases/web-frameworks.md#excluding-urls-from-instrumentation)
```
alexmojaki, line 50: Please add a note that `OpenTelemetryMiddleware` does accept an `excluded_urls` parameter (it's barely visible in the OTEL docs) but not an environment variable.

An `instrument_asgi` method would make this a bit more sensible, @Kludex what's happening with 
 https://github.com/pydantic/logfire/pull/324?

docs/integrations/use-cases/web-frameworks.md
```diff
@@ -79,3 +79,62 @@ Notice how we filtered on records that have the `http.method` attributes set. It
...
 
 ![Requests duration per percentile as Time Series chart](../../images/integrations/use-cases/web-frameworks/logfire-screenshot-chart-percentiles.png)
+
+
+## Excluding URLs from instrumentation
+
+If you want to exclude certain URLs from tracing, you can either use logfire's instrumentation methods or OpenTelemetry configuration.
+You can specify said URLs using a string of comma-separated regexes which will be matched against the full request URL.
+
+### Using logfire
```
alexmojaki:
```suggestion
### Using Logfire
```

docs/integrations/use-cases/web-frameworks.md
```diff
@@ -79,3 +79,62 @@ Notice how we filtered on records that have the `http.method` attributes set. It
...
+Some methods (e.g. `logfire.instrument_fastapi()`) allow you to pass the argument `excluded_urls` as a string of comma-separated regexes.
+
+### Using OpenTelemetry
+
+You can set one of two environment variables to exclude URLs from tracing:
+
+- `OTEL_PYTHON_EXCLUDED_URLS`, which will also apply to all instrumentations for which excluded URLs apply).
+- `OTEL_PYTHON_FASTAPI_EXCLUDED_URLS`, for example, which will only apply to FastAPI instrumentation. You can replace `FASTAPI` with the name of the framework you're using.
+
+!!! example
```
alexmojaki: I don't think putting this all in a big box looks good

docs/integrations/use-cases/web-frameworks.md
```diff
@@ -79,3 +79,62 @@ Notice how we filtered on records that have the `http.method` attributes set. It
...
+
+    If you visit http://127.0.0.1:8000/, that matches the above regex, so no spans will be sent to Logfire.
+    If you visit http://127.0.0.1:8000/hello/ (or any other endpoint that's not `/`, for that matter), a trace will be started and sent to Logfire.
+
+    !!! note
+        Under the hood, the `opentelemetry` library is using `re.search` (not `re.match` or `re.fullmatch`) to check for a match between the route and the `excluded_urls` regex, which is why we need to include the `^` at the start and `$` at the end of the regex.
+
+!!! note
+    Specifying excluded URLs for a given instrumentation suppresses tracing for any automatic instrumentation that applies to said framework.
+    It does not suppress tracing for manual instrumentation.
```
alexmojaki:
```suggestion
    Specifying excluded URLs for a given instrumentation only prevents that specific instrumentation from creating spans/metrics, it doesn't suppress other instrumentation within the excluded endpoints.
```

We also probably need to explain how to prevent inner traces using custom head sampling, but not in this PR.

docs/integrations/wsgi.md
```diff
@@ -36,6 +36,11 @@ with make_server("", 8000, app) as httpd:
 
 You can read more about the OpenTelemetry WSGI middleware [here][opentelemetry-wsgi].
 
+## Excluding URLs from instrumentation
+<!-- note that this section is duplicated for different frameworks but with slightly different links -->
+
+- [Quick guide](use-cases/web-frameworks.md#excluding-urls-from-instrumentation)
```
alexmojaki: I don't see any sign of being able to exclude URLs using the WSGI instrumentation.

alexmojaki (changes requested):

docs/integrations/use-cases/web-frameworks.md
```diff
@@ -81,3 +81,59 @@ You can use this query in a Time Series chart in a dashboard:
 ![Requests duration per percentile as Time Series chart](../../images/integrations/use-cases/web-frameworks/logfire-screenshot-chart-percentiles.png)
 
 See the [DataFusion documentation](https://datafusion.apache.org/user-guide/sql/aggregate_functions_new.html#approx-percentile-cont) for more information on the `approx_percentile_cont` function.
+
+## Excluding URLs from instrumentation
+
+If you want to exclude certain URLs from tracing, you can either use logfire's instrumentation methods or OpenTelemetry configuration.
```
alexmojaki:
```suggestion
If you want to exclude certain URLs from tracing, you can either use Logfire's instrumentation methods or OpenTelemetry configuration.
```

alexmojaki (approved):
Thanks!

Issue #441, opened 2024-09-25
`LOGFIRE_SEND_TO_LOGFIRE` env var doesn't accept `if-token-present` as a value


samuelcolvin:
### Description

It would be very useful if it did, also more consistent with `configure()` kwargs

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="0.53.0"
platform="macOS-14.6.1-arm64-arm-64bit"
python="3.12.2 (main, Feb 25 2024, 03:55:42) [Clang 17.0.6 ]"
[related_packages]
requests="2.32.3"
pydantic="2.9.2"
fastapi="0.115.0"
openai="1.47.1"
protobuf="4.25.5"
rich="13.8.1"
executing="2.1.0"
opentelemetry-api="1.27.0"
opentelemetry-exporter-otlp-proto-common="1.27.0"
opentelemetry-exporter-otlp-proto-http="1.27.0"
opentelemetry-instrumentation="0.48b0"
opentelemetry-instrumentation-asgi="0.48b0"
opentelemetry-instrumentation-asyncpg="0.48b0"
opentelemetry-instrumentation-fastapi="0.48b0"
opentelemetry-instrumentation-httpx="0.48b0"
opentelemetry-instrumentation-system-metrics="0.48b0"
opentelemetry-proto="1.27.0"
opentelemetry-sdk="1.27.0"
opentelemetry-semantic-conventions="0.48b0"
opentelemetry-util-http="0.48b0"
```

PR #488, opened 2024-10-09
Support `'if-token-present'` for env var `'LOGFIRE_SEND_TO_LOGFIRE'`

sydney-runkle:
Fixes https://github.com/pydantic/logfire/issues/441

Working on adding a test to fix coverage at the moment.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,7 @@ We'd love anyone interested to contribute to the Logfire SDK and documentation.
 
 1. Fork and clone the repository
 2. [Install uv](https://docs.astral.sh/uv/getting-started/installation/)
+3. [Install pre-commit](https://pre-commit.com/#install)
 3. Run `make install` to install dependencies
 4. Run `make test` to run unit tests
 5. Run `make format` to format code
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,10 +7,10 @@ We'd love anyone interested to contribute to the Logfire SDK and documentation.
 1. Fork and clone the repository
 2. [Install uv](https://docs.astral.sh/uv/getting-started/installation/)
 3. [Install pre-commit](https://pre-commit.com/#install)
-3. Run `make install` to install dependencies
-4. Run `make test` to run unit tests
-5. Run `make format` to format code
-6. Run `make lint` to lint code
-7. run `make docs` to build docs and `make docs-serve` to serve docs locally
+4. Run `make install` to install dependencies
+5. Run `make test` to run unit tests
+6. Run `make format` to format code
+7. Run `make lint` to lint code
+8. run `make docs` to build docs and `make docs-serve` to serve docs locally
 
 You're now set up to start contributing!
--- a/logfire/_internal/config_params.py
+++ b/logfire/_internal/config_params.py
@@ -189,7 +189,7 @@ class ParamManager:
                     return self._cast(value, name, arg)
                 except LogfireConfigError:
                     pass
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -6,6 +6,7 @@ We'd love anyone interested to contribute to the Logfire SDK and documentation.
 
 1. Fork and clone the repository
 2. [Install uv](https://docs.astral.sh/uv/getting-started/installation/)
+3. [Install pre-commit](https://pre-commit.com/#install)
 3. Run `make install` to install dependencies
```
Kludex: etc
```suggestion
4. Run `make install` to install dependencies
```

logfire/_internal/config_params.py
```diff
@@ -183,6 +183,13 @@ def _cast(self, value: Any, name: str, tp: type[T]) -> T | None:
             return value
         if get_origin(tp) is Literal:
             return _check_literal(value, name, tp)
+        if get_origin(tp) is Union:
+            for arg in get_args(tp):
+                try:
+                    return self._cast(value, name, arg)
+                except LogfireConfigError:
+                    pass
+            raise LogfireConfigError(f'Expected {name} to be one of {get_args(tp)}, got {value!r}')
```
sydney-runkle: Seems like this is a great place to use `pydantic` - why don't we?

sydney-runkle: Reducing deps perhaps?

Kludex: > Reducing deps perhaps?

Yes. Pydantic is not a dependency of logfire.

tests/test_configure.py
```diff
@@ -454,6 +454,13 @@ def test_read_config_from_environment_variables() -> None:
         ):
             fresh_pydantic_plugin()
 
+    with patch.dict(os.environ, {'LOGFIRE_SEND_TO_LOGFIRE': 'not-valid'}):
+        with pytest.raises(
```
alexmojaki: This is a good place to use https://15r10nk.github.io/inline-snapshot/extra/#inline_snapshot.extra.raises, mostly just to avoid escaping the regex.

tests/test_configure.py
```diff
@@ -1254,6 +1261,20 @@ def test_send_to_logfire_if_token_present_empty() -> None:
...
+    os.environ['LOGFIRE_SEND_TO_LOGFIRE'] = 'if-token-present'
+    try:
+        with ExitStack() as stack:
+            stack.enter_context(mock.patch('logfire._internal.config.Confirm.ask', side_effect=RuntimeError))
+            requests_mocker = stack.enter_context(requests_mock.Mocker())
+            configure(console=False)
+            assert len(requests_mocker.request_history) == 0
+    finally:
+        del os.environ['LOGFIRE_TOKEN']
+        del os.environ['LOGFIRE_SEND_TO_LOGFIRE']
```
alexmojaki:
```suggestion
    with patch.dict(
        os.environ,
        {'LOGFIRE_TOKEN': '', 'LOGFIRE_SEND_TO_LOGFIRE': 'if-token-present'},
    ), mock.patch(
        'logfire._internal.config.Confirm.ask',
        side_effect=RuntimeError,
    ), requests_mock.Mocker() as requests_mocker:
        configure(console=False)
    assert len(requests_mocker.request_history) == 0
```

(i'm not sure that this is better than the exit stack, i started writing it before i realised that i couldn't make it pretty)

tests/test_configure.py
sydney-runkle: Cool done!

sydney-runkle:
@alexmojaki, anything else you'd like here?

alexmojaki:
Thanks!

PR #489, opened 2024-10-09
Suppress instrumentation in excluded URLs in FastAPI

dmontagu:
Here's a proof of concept demonstrating one way this could work — on this branch if you run the following demo app:
```python
import logfire
from fastapi import FastAPI

app = FastAPI()

logfire.configure(token='your_token')
logfire.instrument_fastapi(app, excluded_urls='^https?://[^/]+/$')


@app.get("/")
async def root(name: str):
    with logfire.span("root"):
        return {"message": f"root {name}"}


@app.get("/hello")
async def hello(name: str):
    return {"message": f"hello {name}"}


if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app)
```
Requests to the `/hello` endpoint end up as traces in logfire, but requests to `/` are completely suppressed; on main, you'll get a span with name `'root'` showing up as a trace root.

I guess it would probably make sense to expose this as an opt-in choice rather than an always-on the way this PR does it, but I wanted to share the implementation in case it's useful rather than basically pixel-twiddle it in case there are differences of opinion about how that should work.

I'll note that there are other ways to achieve this with a context var, and there may be cleaner ways to store the "current app" (which is all you need to look up its instrumentation in the registry, and therefore determine whether the URL is excluded), but this approach of storing the current _request_ in a contextvar and looking up the app from that was straightforward to me and I felt it was worthy enough to share as a proof of concept.

@alexmojaki I know you have thoughts about getting upstream agreement with OTel about this, happy to move forward with that, though I think it might make sense to try to expose this in some opt-in way in the interim (or if they reject that idea).

### DIFF
--- a/logfire/_internal/integrations/fastapi.py
+++ b/logfire/_internal/integrations/fastapi.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import dataclasses
 import inspect
 from contextlib import contextmanager
+from contextvars import ContextVar
 from functools import lru_cache
 from typing import Any, Awaitable, Callable, ContextManager, Iterable, cast
 from weakref import WeakKeyDictionary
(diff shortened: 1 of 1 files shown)


alexmojaki:
Not going with this approach because it doesn't handle spans created in middleware.

Issue #477, opened 2024-10-04
Is there a way to add tags to a span after it is created?


ohmeow:
### Question

I have additional tags I want to that come from the result of a function running inside a span, but I'm not able to add them via `span.tags += ["my_tag"]` with the current API.

Is there some other way to do this?

Thanks!

alexmojaki:
```python
import logfire

logfire.configure()

with logfire.span('my_span', _tags=['foo', 'bar']) as span:
    span._span.set_attribute('logfire.tags', span.tags + ('baz', 'qux'))
```

We can certainly make this easier.

AdolfoVillalobos:
Hi! @alexmojaki 

I would love to work on this.  I thought about two approaches.

I would appreciate guidance on the best approach to the project vision and design principles.

## 1. create `add_tags`method

The implementation would be something like:

```python
class LogfireSpan(ReadableSpan):
    # ... 

    def add_tags(self, *tags: str) -> None:
        """Add tags to the span."""
        with handle_internal_errors():
            if self._span is None:
                self._otlp_attributes.setdefault(ATTRIBUTES_TAGS_KEY, []).extend(tags)
            else:
                existing_tags = self._span.attributes.get(ATTRIBUTES_TAGS_KEY, [])
                new_tags = uniquify_sequence(existing_tags + list(tags))
                self._span.set_attribute(ATTRIBUTES_TAGS_KEY, new_tags)

    # ... 

```

Then calls would be:
```python
with logfire.span("my_span") as span:
    span.add_tags("tag1", "tag2", "tag3")
```

## 2. add a setter to the `tags`property

This would give us even more control over the tags:

```python
class LogfireSpan(ReadableSpan):
    # ... 

    @property
    def tags(self) -> Sequence[str]:
        return self._get_attribute(ATTRIBUTES_TAGS_KEY, [])

    @tags.setter
    @handle_internal_errors()
    def tags(self, new_tags: Sequence[str]) -> None:
        """Set or add tags to the span."""
        existing_tags = self.tags
        updated_tags = uniquify_sequence(existing_tags + list(new_tags))
        if self._span is None:
            self._otlp_attributes[ATTRIBUTES_TAGS_KEY] = updated_tags
        else:
            self._span.set_attribute(ATTRIBUTES_TAGS_KEY, updated_tags)

    # ... 
```

Then we would have a much more flexible approach, as we would leverage the features from python's list type:

```python
with logfire.span("my_span") as span:
[...]

PR #497, opened 2024-10-14
feat: add tag as tuples through setter

AdolfoVillalobos:
This PR addresses https://github.com/pydantic/logfire/issues/477.

From now on, the property LogfireSpan.tags will return a tuple and can be modified using common tuple operations.

Trying to use list methods will raise errors from now on.

This is a first attempt. I'm happy to improve and receive feedback.


logfire/_internal/main.py
```diff
@@ -1698,7 +1698,17 @@ def message_template(self) -> str | None:  # pragma: no cover
 
     @property
     def tags(self) -> Sequence[str]:  # pragma: no cover
```
```suggestion
    def tags(self) -> tuple[str, ...]:
```

Please test this case.

logfire/_internal/main.py
```diff
@@ -1698,7 +1698,17 @@ def message_template(self) -> str | None:  # pragma: no cover
...
-        return self._get_attribute(ATTRIBUTES_TAGS_KEY, [])
+        return self._get_attribute(ATTRIBUTES_TAGS_KEY, ())
+
+    @tags.setter
+    @handle_internal_errors()
+    def tags(self, new_tags: Sequence[str]) -> None:
+        """Set or add tags to the span."""
+        updated_tags = uniquify_sequence(list(new_tags))
```
```suggestion
        updated_tags: tuple = uniquify_sequence(list(new_tags))
```

alexmojaki: Why the conversion to a list first?

alexmojaki: Also please test setting a non-list

logfire/_internal/main.py
```diff
@@ -1698,7 +1698,17 @@ def message_template(self) -> str | None:  # pragma: no cover
...
+        if self._span is None:  # pragma: no cover
```
```suggestion
        if self._span is None:
```

Please test this case by setting tags before `with span:`

AdolfoVillalobos:
@alexmojaki Thanks for the valuable feedback! My bad using "list" instead of "tuple".  

I also included a test case for setting tags when the `_span` attribute is `None`. 

I'm open to any other suggestions or improvements you might have. Thanks again!

logfire/_internal/main.py
```diff
@@ -1697,8 +1697,18 @@ def message_template(self) -> str | None:  # pragma: no cover
...
-    def tags(self) -> Sequence[str]:  # pragma: no cover
-        return self._get_attribute(ATTRIBUTES_TAGS_KEY, [])
+    def tags(self) -> tuple[str, ...]:
+        return self._get_attribute(ATTRIBUTES_TAGS_KEY, ())
+
+    @tags.setter
+    @handle_internal_errors()
+    def tags(self, new_tags: Sequence[str]) -> None:
+        """Set or add tags to the span."""
+        updated_tags: tuple[str, ...] = uniquify_sequence(tuple(new_tags))
```
```suggestion
        updated_tags: tuple[str, ...] = uniquify_sequence(new_tags)
```

My point about `list` before was that it shouldn't need preconverting at all.

logfire/_internal/main.py
```diff
@@ -1697,8 +1697,18 @@ def message_template(self) -> str | None:  # pragma: no cover
         return self._get_attribute(ATTRIBUTES_MESSAGE_TEMPLATE_KEY, None)
 
     @property
-    def tags(self) -> Sequence[str]:  # pragma: no cover
-        return self._get_attribute(ATTRIBUTES_TAGS_KEY, [])
+    def tags(self) -> tuple[str, ...]:
+        return self._get_attribute(ATTRIBUTES_TAGS_KEY, ())
```
alexmojaki, line 1730: Please test this default value

tests/test_logfire.py
```diff
@@ -2583,3 +2583,28 @@ def test_force_flush(exporter: TestExporter):
...
+        span.tags += ('d',)
+        assert span.tags == ('a', 'b', 'c', 'd')
+
+
+def test_tags_with_setter_and_non_initialized_span(exporter: TestExporter) -> None:
+    span = logfire.with_tags('tag1', 'tag2').span('test span')
+    assert span.tags == ('tag1', 'tag2')
+
+    span.tags = ('tag3', 'tag4')
+    assert span.tags == ('tag3', 'tag4')
```
alexmojaki: please test that these remain after opening the span

logfire/_internal/main.py
alexmojaki: but please test what happens when setting tags to a non-tuple value

alexmojaki: also we probably need to special-case setting to a string, e.g. `span.tags = "bar"` should behave like `span.tags = ["bar"]`

logfire/_internal/main.py
```diff
@@ -1697,8 +1697,18 @@ def message_template(self) -> str | None:  # pragma: no cover
...
+
+    @tags.setter
+    @handle_internal_errors()
+    def tags(self, new_tags: Sequence[str]) -> None:
+        """Set or add tags to the span."""
+        updated_tags: tuple[str, ...] = uniquify_sequence(tuple(new_tags))
+        if self._span is None:
+            self._otlp_attributes[ATTRIBUTES_TAGS_KEY] = updated_tags
+        else:
+            self._span.set_attribute(ATTRIBUTES_TAGS_KEY, updated_tags)
```
alexmojaki: please dedupe this logic with the message setter, probably by adding a `_set_attribute` method

alexmojaki (changes requested):

alexmojaki:
Also please update `NoopSpan`

alexmojaki:
and `logfire_api.LogfireSpan`

logfire/_internal/main.py
AdolfoVillalobos: Makes sense, my bad. 

If I understand correctly, we should allow setting tags using tuples, lists and str.  I added tests for these 3 cases, enforcing that the output is always a tuple.

In the case of lists, the setter will fail if the try to use a concatenation operation "+=". That kind of operations will only be available for tuples. 

My question is: Should we keep it like this, or should we also allow for concatenation operations in the case of lists? 

```python
# is valid
tags += ("a", "b")

# will fail
tags += ["a", "b"]
```

logfire/_internal/main.py
AdolfoVillalobos: Makes sense. I created the `_set_attribute` method.  I assume that we should not add its signature to the `logfire_api.LogfireSpan` declarations.  

Not adding it implies that pyright does not allow for "direct" testing (I think)

logfire/_internal/main.py
```diff
@@ -173,7 +173,8 @@ def _span(
 
             tags = (self._tags or ()) + tuple(_tags or ())
             if tags:
-                otlp_attributes[ATTRIBUTES_TAGS_KEY] = uniquify_sequence(tags)
+                updated_tags = uniquify_sequence(tags)
+                otlp_attributes[ATTRIBUTES_TAGS_KEY] = updated_tags
```
alexmojaki: please revert

logfire/_internal/main.py
```diff
@@ -1697,19 +1698,25 @@ def message_template(self) -> str | None:  # pragma: no cover
         return self._get_attribute(ATTRIBUTES_MESSAGE_TEMPLATE_KEY, None)
 
     @property
-    def tags(self) -> Sequence[str]:  # pragma: no cover
-        return self._get_attribute(ATTRIBUTES_TAGS_KEY, [])
+    def tags(self) -> tuple[str, ...]:
+        return tuple(self._get_attribute(ATTRIBUTES_TAGS_KEY, ()))
```
alexmojaki: rather store a tuple than convert on get

logfire/_internal/main.py
```diff
@@ -1793,6 +1800,13 @@ def _get_attribute(self, key: str, default: Any) -> Any:
         attributes = getattr(self._span, 'attributes', self._otlp_attributes)
         return attributes.get(key, default)
 
+    def _set_attribute(self, key: str, value: Any) -> None:
```
alexmojaki, line 1830:
```suggestion
    def _set_attribute(self, key: str, value: otel_types.AttributeValue) -> None:
```

logfire-api/logfire_api/_internal/main.pyi
```diff
@@ -932,7 +932,9 @@ class LogfireSpan(ReadableSpan):
     @property
     def message_template(self) -> str | None: ...
     @property
-    def tags(self) -> Sequence[str]: ...
+    def tags(self) -> tuple[str, ...]: ...
+    @tags.setter
+    def tags(self, new_tags: Sequence[str]) -> None: ...
```
alexmojaki, line 937: The actual class in `logfire_api/__init__.py` needs updating so that setting tags does nothing rather than raising an error, and this needs testing.

logfire/_internal/main.py
```diff
@@ -1828,8 +1842,8 @@ def message_template(self) -> str:  # pragma: no cover
         return ''
 
     @property
-    def tags(self) -> Sequence[str]:  # pragma: no cover
```
alexmojaki, line 1860: See the comment on `@message.setter`, tags have the same need, and testing in `test_internal_exception_span`

tests/test_logfire.py
```diff
@@ -1548,50 +1548,6 @@ def test_readable_span_signature():
     }
 
 
-def test_tags(exporter: TestExporter) -> None:
```
alexmojaki: Why was this test moved? If something changed, I can't see what.

tests/test_logfire.py
```diff
@@ -2583,3 +2539,109 @@ def test_force_flush(exporter: TestExporter):
...
+    # Call __enter__ to initialize _span, check that tags remain unchanged
+    with span:
+        assert span.tags == ('tag3', 'tag4')
+
+    # Check that tags remain unchanged
+    assert span.tags == ('tag3', 'tag4')
+
+
+def test_tags_setter_valid_type_extensions() -> None:
+    span = logfire.span('test span')
```
alexmojaki:
```suggestion
    span = logfire.span('test span')
    assert span.tags == ()
```

alexmojaki (changes requested):

tests/test_logfire.py
AdolfoVillalobos: I moved it so that all the tag tests are "near each other." If that's an issue, I can revert it.

tests/test_logfire.py
AdolfoVillalobos: I added tests checking default values in many cases.

logfire-api/logfire_api/_internal/main.pyi
AdolfoVillalobos: My bad! I've now updated the `__init__.py` and `__init__.pyi` files to account for the new setter and signature changes.

logfire/_internal/main.py
AdolfoVillalobos: I've added:

```python
    # This is required to make `span.tags = ` not raise an error.
    @tags.setter
    def tags(self, new_tags: Sequence[str]) -> None:
        pass
       
```
        
on the `NoopSpan` class so that setting `span.tags = "bar"` does not raise errors and does nothing. I hope I understood correctly.

tests/test_logfire.py
alexmojaki: let's keep diffs minimal and clean, both to make this PR more readable and for future `git blame`s. i don't think it's very important for the tests to be near each other, but you can also move the new tests here.

alexmojaki (approved):
Thank you!

PR #501, opened 2024-10-15
PYD-1326: basic live view sql explanation in docs

ChristopherGS:
https://linear.app/pydantic/issue/PYD-1326/live-view-temporary-bandage-part-1


ChristopherGS:
Yes this means that we have the schema in two places. No that's not perfect. Yes, it's way better than it was before.

docs/guides/web-ui/live.md
```diff
@@ -4,6 +4,57 @@ The live view is the main view of Logfire, where you can see traces in real-time
...
+    service_instance_id text,
+    process_pid integer
+)
+```
+
+So for example, to view your warnings and errors you could query:
+- `level > 'info'`
+
+To see just exceptions you could do:
+- `is_exception = true`
```
alexmojaki:
```suggestion
- `is_exception`
```

Issue #490, opened 2024-10-10
uv.lock has pinned pyscopg-binary==3.2.3, which can't be build on Mac M1


AdolfoVillalobos:
### Question

I'm using uv in my workflow to submit a PR, and am having problems with the `psycopg-binary` dependency.

## Error:

```bash
error: distribution psycopg-binary==3.2.3 @ registry+https://pypi.org/simple can't be installed because it doesn't have a source distribution or wheel for the current platform
make: *** [install] Error 2
```

## Root

According to the psycopg documentation at https://www.psycopg.org/psycopg3/docs/basic/install.html#binary-installation:

>... as of Psycopg 3.1.20, ARM64 macOS binary packages (for Apple M1 machines) are no longer available for macOS.

In the logfire project, the dev dependencies have the psycopg-binary version pinned through the uv.lock file, which is preventing me from pushing changes.

## Questions

Although this may not be a logfire issue, I have 2 questions:

1. Is it necessary to keep the uv.lock file in the repo? 
2. If yes, is it possible to make the `psycopg-binary` dependency more flexible? I installed version 3.1.18 of `psycopg-binary` after upgrading to macOS 15. However, pre-commit is still forcing me to use version 3.2.3, which I'm unable to do.

alexmojaki:
> Is it necessary to keep the uv.lock file in the repo?

It actually might not be. @Kludex is there a strong reason to keep it?

@AdolfoVillalobos see if you can get things working with `psycopg` without the binary?

AdolfoVillalobos:
> > Is it necessary to keep the uv.lock file in the repo?
> 
> It actually might not be. @Kludex is there a strong reason to keep it?
> 
> @AdolfoVillalobos see if you can get things working with `psycopg` without the binary?

@alexmojaki I removed `psycopg-binary` from the dev dependencies and replaced it with plain `psycopg`. Then, I was able to run the tests without issues.  

I could submit a PR implementing this change when we decide what to do with the `uv.lock` file.

Kludex:
> It actually might not be. @Kludex is there a strong reason to keep it?

Predictable development environment? Less annoying for contributors since the tests will always pass locally?

AdolfoVillalobos:
> > It actually might not be. @Kludex is there a strong reason to keep it?
> 
> Predictable development environment? Less annoying for contributors since the tests will always pass locally?

I agree that makes sense. Thanks! That's why I proposed flexibility on the dev dependencies instead, as well

[...]

PR #502, opened 2024-10-15
Pin `psycopg-binary` dependency for easier dev env setup

sydney-runkle:
Closes https://github.com/pydantic/logfire/issues/490

Also, update deps with `uv sync --upgrade`, including a `pyright` bump hence the changes re type checking comments.


sydney-runkle:
@Viicos, this might be a good one for you to take a look at - I got blocked with the `pyright` upgrade, and I know you're familiar with the recent changes...

tests/otel_integrations/test_psycopg.py
```diff
@@ -23,7 +24,7 @@ def test_check_version():
 
 
 def test_instrument_psycopg():
-    original_connect = psycopg.connect
+    original_connect = cast('Callable[..., psycopg.Connection[Tuple[Any, ...]]]', psycopg.connect)
```
sydney-runkle: This is probably wrong, I was just experimenting...

Viicos:
I remember [this thread](https://github.com/psycopg/psycopg/issues/708) from the `psycopg` library. Until the latest version gets used, I'd suggest adding a `# pyright: reportUnknownMemberType=false` at the top of file.

sydney-runkle:
> Until the latest version gets used

Ah ok, as in `psycopg[binary]` `latest`?

I think the pin might not actually be necessary here, waiting on a response to https://github.com/pydantic/logfire/issues/490#issuecomment-2414700487

Viicos:
And unfortunately `psycopg` has implemented the generic `Connection` class in a way that is not compatible with `pyright` (and I believe the typing spec as well). So we may encounter more issues in the future

sydney-runkle:
Ah ok, good to know. Thanks for the feedback @Viicos!

sydney-runkle:
No longer needed, original issue closed.

PR #503, opened 2024-10-15
Test on variety of `pydantic` versions 🩷 

sydney-runkle:
Add tests with a `pydantic` versions matrix

Not sure if we need to deal with coverage here, happy to remove that part if we want.


Kludex:
Should we test OTel packages min/max as well? 🤔
Also, should we test everything, or just the pydantic plugin?

This job needs to be added to the `check` job btw.

sydney-runkle:
Good question. I'll defer to you and @alexmojaki re Otel. Will add to check when I'm back at my computer.

I think just testing the plugin makes sense given that pydantic isn't a dependency.

alexmojaki:
> Not sure if we need to deal with coverage here, happy to remove that part if we want.

We have code like `if get_version(pydantic.__version__) < get_version('2.5.0'):  # pragma: no cover` (not the only such case) so:

1. Please keep the coverage stuff.
2. Those pragmas should be removed
3. Please test as far back as 2.4, and also one version of v1 (search the repo for `pydantic v1`)

> Should we test OTel packages min/max as well? 🤔

We definitely should, in another PR

> Also, should we test everything, or just the pydantic plugin?

Everything, just to be safe. These are fast. It's not just the plugin anyway, it's also about encoding pydantic objects as JSON.

Is there not a way to do this basically with just the GHA matrix, so we don't duplicate a bunch of stuff?

.github/workflows/main.yml
```diff
@@ -98,6 +98,49 @@ jobs:
...
           include-hidden-files: true
 
+  test-pydantic:
+    name: test with pydantic ${{ matrix.pydantic-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # keep up to date with 3-4 latest pydantic versions
+        pydantic-version: ["2.6.4", "2.7.4", "2.8.2", "2.9.2"]
```
alexmojaki:
```suggestion
        pydantic-version: ["1.10", 2.4, 2.6, 2.8, 2.9]
```

sydney-runkle:
> Please keep the coverage stuff.

Done

> Those pragmas should be removed

Done

> Please test as far back as 2.4, and also one version of v1 (search the repo for pydantic v1)

Done

> Everything, just to be safe. These are fast. It's not just the plugin anyway, it's also about encoding pydantic objects as JSON.

Sounds great, thanks for clarifying.

> Should we test OTel packages min/max as well? 🤔

I've opened https://github.com/pydantic/logfire/issues/507 to track this.

sydney-runkle:
Need to investigate failing coverage tomorrow. Not sure why that's the case now that we test on the variety of pydantic versions.

sydney-runkle:
Do we want to do the full cross product of python x pydantic versions here?

.github/workflows/main.yml
```diff
@@ -58,12 +58,28 @@ jobs:
...
+            - python-version: '3.12'
+              pydantic-version: '2.5.3'
+            - python-version: '3.12'
+              pydantic-version: '2.6.4'
+            - python-version: '3.12'
+              pydantic-version: '2.7.4'
+            - python-version: '3.12'
+              pydantic-version: '2.8.2'
+            - python-version: '3.12'
+              pydantic-version: '2.9.2'
```
alexmojaki:
```suggestion
              pydantic-version: '1.10'
            - python-version: '3.12'
              pydantic-version: '2.4'
            - python-version: '3.12'
              pydantic-version: '2.6'
            - python-version: '3.12'
              pydantic-version: '2.8'
            - python-version: '3.12'
              pydantic-version: '2.9'
```

If we want to cut down the number of pydantic versions, which we might, this is my take.

I definitely think that new patch releases shouldn't require us to update these numbers. I know there probably won't be a 2.8.3 but there might be a 2.9.3 or 1.10.19, and this is just simpler anyway.

.github/workflows/main.yml
```diff
@@ -58,12 +58,28 @@ jobs:
...
 
   test:
-    name: test on ${{ matrix.python-version }}
+    name: test on Python ${{ matrix.python-version }} and pydantic v${{ matrix.pydantic-version }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        pydantic-version: 'main'
```
alexmojaki:
```suggestion
        pydantic-version: ['main']
```

sydney-runkle: Thanks, was wondering why this was red.

.github/workflows/main.yml
sydney-runkle: Yeah but there are known bugs in the earliest versions of patches...

.github/workflows/main.yml
```diff
@@ -58,12 +58,28 @@ jobs:
           fi
 
   test:
-    name: test on ${{ matrix.python-version }}
+    name: test on Python ${{ matrix.python-version }} and pydantic v${{ matrix.pydantic-version }}
```
alexmojaki:
```suggestion
    name: test on Python ${{ matrix.python-version }} and pydantic ${{ matrix.pydantic-version }}
```

alexmojaki: just to avoid 'vmain'.

.github/workflows/main.yml
sydney-runkle: When you do `uv pip install pydantic==2.4`, it installs `2.4.0`, which seems odd, but was why I explicitly specified the versions here?

alexmojaki: Oh, I wrongly assumed how `pip install pydantic==2.9` would be interpreted. Basically I want it to install the latest patch version.

sydney-runkle: Np, experimenting with `~=` now

alexmojaki:
Looks like the coverage still isn't working. Once that's fixed, please search the repo for `pydantic v1` and remove associated pragmas.

.github/workflows/main.yml
sydney-runkle: We can do `uv pip install 'pydantic==2.4.*'

I think we should test all of the versions, there aren't that many, and they're all within a year...

alexmojaki: 👍

sydney-runkle:
It's not immediately evident to me why we're missing these lines on v1 testing.

tests/test_logfire_api.py
```diff
@@ -29,6 +32,7 @@ def import_logfire_api_with_logfire() -> ModuleType:
     return importlib.reload(logfire_api)
 
 
+@pytest.mark.skipif(get_version(pydantic_version) < get_version('2.5.0'), reason='Skipping on Pydantic versions < v2.5')
```
alexmojaki: Let's just skip the part that needs to be skipped, which i'm assuming is instrument_pydantic

alexmojaki:
Thanks, sorry it was so painful!

Issue #79, opened 2024-05-02
Links to GitHub code source


Kludex:
### Description

You'll be able to go to your GitHub repository directly from the **Logfire UI**, and see the code of a logfire call or exception.

Please let us know if you are interested in this, so we can prioritize it.

Kludex:
Documented here: https://logfire.pydantic.dev/docs/guides/advanced/link-to-code-source/

alexmojaki:
This still needs correcting: https://logfire.pydantic.dev/docs/reference/api/logfire/#logfire.CodeSource.root_path

Something like https://github.com/pydantic/logfire/pull/451#discussion_r1804453006

Kludex:
This is implemented, and @alexmojaki 's comment above was corrected.

PR #505, opened 2024-10-16
Document Code Source links

Kludex:
If it's easier, please apply the changes that you wish. I'm okay with anything. 🙏


sydney-runkle:
There's a related issue to this right? Can we link it as closing https://github.com/pydantic/logfire/issues/79?

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
+!!! warning "Experimental"
+    This feature is experimental and may change in the future.
+
+    Please give us feedback on [Slack][slack].
```
sydney-runkle: Does it make sense to have an `experimental` namespace like we do in `pydantic` for things like this that might explicitly change?

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+
+We support linking to the source code on GitHub, or any other VCS provider.
```
sydney-runkle:
```suggestion
We support linking to the source code on GitHub or any other VCS provider.
```

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+
+![Link to GitHub](../../images/guide/link-to-github.gif)
+
+## Usage
+
+To enable source code linking with Logfire, you need to configure your project. Here's how:
```
sydney-runkle:
```suggestion
To enable source code linking with Logfire, you need to configure your project's code source. Here's an example:
```

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+logfire.configure(
+    code_source=logfire.CodeSource(
+        repository='https://github.com/pydantic/logfire',
+        revision='main',
+        root_path='.',
+    )
+)
+```
+
+You can check more about it on the [`logfire.CodeSource`][logfire.CodeSource] on our API reference.
```
sydney-runkle:
```suggestion
You can learn more in our [`logfire.CodeSource`][logfire.CodeSource] API reference.
```

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+
+| Parameter | Description |
+| --- | --- |
+| `repository` | The URL of the repository e.g. `https://github.com/pydantic/logfire` |
+| `revision` | The specific branch, tag, or commit hash to link to e.g. `main` |
+| `root_path` | The path to the root of the repository. If your code is in a subdirectory, you can specify it here. |
+
+## Alternative Configuration
+
+You can configure these settings using the [`OTEL_RESOURCE_ATTRIBUTES`][otel-resource-attributes] environment variable:
```
sydney-runkle:
```suggestion
Alternatively, you can configure these settings using the [`OTEL_RESOURCE_ATTRIBUTES`][otel-resource-attributes] environment variable:
```

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+
+You can check more about it on the [`logfire.CodeSource`][logfire.CodeSource] on our API reference.
+
+| Parameter | Description |
+| --- | --- |
+| `repository` | The URL of the repository e.g. `https://github.com/pydantic/logfire` |
+| `revision` | The specific branch, tag, or commit hash to link to e.g. `main` |
+| `root_path` | The path to the root of the repository. If your code is in a subdirectory, you can specify it here. |
+
+## Alternative Configuration
```
sydney-runkle, line 32: Seems like this pattern comes up a lot (re our settings vs otel env vars). Maybe we should start using tab panels to explain how to set things (one tab for logfire, one tab for otel env var)?

sydney-runkle (commented):
Cool feature! Left some ideas for changes and questions.

docs/guides/advanced/link-to-code-source.md
Kludex: Answered in person.

docs/guides/advanced/link-to-code-source.md
Kludex: Yes, that would be great.

docs/guides/advanced/link-to-code-source.md
alexmojaki: What about directing people to the help page instead of just slack?

docs/guides/advanced/link-to-code-source.md
alexmojaki:
```suggestion
Here's an example:
```

This configuration is not at the project level anyway.

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+## Usage
+
+To enable source code linking with Logfire, you need to configure your project. Here's how:
+
+```python
+import logfire
+
+logfire.configure(
+    code_source=logfire.CodeSource(
+        repository='https://github.com/pydantic/logfire',
```
alexmojaki: This (combined with "We support linking to the source code on GitHub, or any other VCS provider") makes it look like it'll just work if they provide a non-github link, which it likely won't.

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+        revision='main',
```
alexmojaki: This is usually not good practice and will lead to stale links, we should at least point this out.

I think ideally there would be a way to say 'use the service_version as the revision' and then we'd also encourage them to use service versions.

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+        root_path='.',
```
alexmojaki: Should this just be the default?

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+    )
+)
+```
+
+You can check more about it on the [`logfire.CodeSource`][logfire.CodeSource] on our API reference.
+
+| Parameter | Description |
```
alexmojaki: This table feels redundant with the example. What about adding comments to the example? Maybe those clickable bubbles?

docs/guides/advanced/link-to-code-source.md
```diff
@@ -0,0 +1,45 @@
...
+| --- | --- |
+| `repository` | The URL of the repository e.g. `https://github.com/pydantic/logfire` |
+| `revision` | The specific branch, tag, or commit hash to link to e.g. `main` |
+| `root_path` | The path to the root of the repository. If your code is in a subdirectory, you can specify it here. |
+
+## Alternative Configuration
+
+Alternatively, you can configure these settings using the [`OTEL_RESOURCE_ATTRIBUTES`][otel-resource-attributes] environment variable:
```
alexmojaki:
```suggestion
For other OpenTelemetry SDKs, you can configure these settings using resource attributes, e.g. by setting the [`OTEL_RESOURCE_ATTRIBUTES`][otel-resource-attributes] environment variable:
```

alexmojaki (changes requested):

docs/guides/advanced/link-to-code-source.md
Kludex: yes, bubbles.

docs/guides/advanced/link-to-code-source.md
Kludex: I can change this in a future PR, but I guess it makes sense to be the default yes.

docs/guides/advanced/link-to-code-source.md
Kludex: What do you suggest?

docs/guides/advanced/link-to-code-source.md
Kludex: better now?

docs/guides/advanced/link-to-code-source.md
alexmojaki: It looks like our URL construction strategy works for gitlab (e.g. https://gitlab.com/rluna-gitlab/gitlab-ce/blob/v16.0.10/yarn.lock#L13) so i guess we support that too? maybe we just say we support github, gitlab, and any other VCS provider that uses the same URL format? we clearly don't support ANY provider.

docs/guides/advanced/link-to-code-source.md
alexmojaki: seems like it should be optional in the data class but the frontend needs to handle defaulting to it so that the env var isn't required

docs/guides/advanced/link-to-code-source.md
Kludex: Ah, ok. I see.

docs/guides/advanced/link-to-code-source.md
Kludex: The frontend already handles it, defaulting to it.

sydney-runkle (approved):
Nice, thanks!

docs/guides/advanced/link-to-code-source.md
Kludex: changed

PR #508, opened 2024-10-17
Add `local` parameter to `logfire.configure()`

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/195


logfire/_internal/config.py
```diff
@@ -889,7 +901,6 @@ def get_tracer_provider(self) -> ProxyTracerProvider:
         Returns:
             The tracer provider.
         """
-        self.warn_if_not_initialized('No logs or spans will be created')
```
alexmojaki, line 892: I moved these warnings to prevent `logfire.instrument_fastapi` emitting similar warnings 3 times.

logfire/_internal/integrations/aiohttp_client.py
```diff
@@ -2,10 +2,18 @@
...
+
+def instrument_aiohttp_client(logfire_instance: Logfire, **kwargs: Any):
     """Instrument the `aiohttp` module so that spans are automatically created for each client request.
 
     See the `Logfire.instrument_aiohttp_client` method for details.
     """
-    AioHttpClientInstrumentor().instrument(**kwargs)  # type: ignore[reportUnknownMemberType]
+    AioHttpClientInstrumentor().instrument(  # type: ignore[reportUnknownMemberType]
+        **{
+            'tracer_provider': logfire_instance.config.get_tracer_provider(),
```
alexmojaki, line 15: I'm torn on how to do this. On the one hand I don't want to completely prevent users from passing custom providers. We even do that in our own code (passing a NoopMeterProvider to disable fastapi metrics), and I've recommended another user to do the same. On the other hand I don't want to encourage or document it. So this makes it secretly possible to override the logfire providers with custom ones. Which is a mess for type checking.

logfire/_internal/main.py
```diff
@@ -1262,7 +1265,13 @@ def instrument_sqlalchemy(self, **kwargs: Unpack[SQLAlchemyInstrumentKwargs]) ->
         from .integrations.sqlalchemy import instrument_sqlalchemy
 
         self._warn_if_not_initialized_for_instrumentation()
-        return instrument_sqlalchemy(**kwargs)
+        return instrument_sqlalchemy(
+            **{  # type: ignore
```
alexmojaki, line 1269: This is the same logic as the other instrumentations, I just realized at this point that this approach was a bit easier and less tedious.

logfire/_internal/integrations/aiohttp_client.py
adriangb: I mean we're using kwargs already so this seems like a step better

adriangb (approved):
Thanks this looks great!

PR #509, opened 2024-10-17
Add fastapi arguments attributes directly on the root otel span

alexmojaki:
Having a separate "FastAPI arguments" span is a pain. It costs extra money and requires writing a join in SQL to combine info about the root span (particularly the duration) with the argument values. I want to get rid of it.

As a first step, this PR takes the same attributes that are set on the fastapi args span and sets them on the root span created by OTEL. This allows people to use the root span in queries and avoid joining in SQL, and should ease the transition to removing the fastapi args span.

In a followup change I will remove the fastapi args span. We might want to treat this as a breaking change and release version 2.0. We might also want to add a parameter (off by default) that keeps the span, although the only reasons I can think of to keep it are quite weak:

1. Backwards compatibility in dashboard/alert queries
2. Grouping together spans that come from resolving arguments (e.g. db queries)
3. Seeing how long it takes to resolve arguments


Kludex:
Nice!

logfire/_internal/integrations/fastapi.py
```diff
@@ -58,7 +58,6 @@ def instrument_fastapi(
         dict[str, Any] | None,
     ]
     | None = None,
-    use_opentelemetry_instrumentation: bool = True,
```
alexmojaki, line 61: This option was a bad idea on my part, clutters the docs and API, and doesn't really have a purpose any more now that OTEL kwargs are passed through.

alexmojaki:
After chatting with @dmontagu I think it's likely that we won't remove the fastapi args span soon, but this change should be useful anyway.

JacobHayes:
Ah, this might also fix https://github.com/pydantic/logfire/issues/533

alexmojaki:
> Separately, I do find it somewhat useful to track the db queries that come from parsing arguments and resolving `Depends`

But @JacobHayes if you've instrumented db queries then you'll still see spans for those queries even if we drop the fastapi args span, they just won't all be grouped together. Otherwise you'll see fastapi args taking a long time to resolve but you won't know why.

PR #513, opened 2024-10-17
Remove unnecessary line from changelog

sydney-runkle:


alexmojaki (approved):
Thanks!

sydney-runkle:
Test failures called by an `openai` update. Looking into this now.

PR #514, opened 2024-10-18
Fix openai tests due to `openai` upgrade -> v1.52.0

sydney-runkle:
Also, update `pyright` in the process and bump dev dependencies.


tests/otel_integrations/test_openai.py
```diff
@@ -277,24 +277,7 @@ def test_sync_chat_completions(instrumented_client: openai.Client, exporter: Tes
...
-                            'usage': {
-                                'completion_tokens': 1,
-                                'prompt_tokens': 2,
-                                'total_tokens': 3,
-                                'completion_tokens_details': None,
-                                'prompt_tokens_details': None,
-                            },
-                        }
-                    ),
+                    'response_data': '{"message":{"content":"Nine","refusal":null,"role":"assistant","audio":null,"function_call":null,"tool_calls":null},"usage":{"completion_tokens":1,"prompt_tokens":2,"total_tokens":3,"completion_tokens_details":null,"prompt_tokens_details":null}}',
```
Kludex: The `IsJson` was cool here... 👀

Kludex: Can you bring it back? I think only the `audio` field was added?

Kludex (changes requested):

tests/otel_integrations/test_openai.py
sydney-runkle: Yeah, will try

Kludex (approved):

PR #534, opened 2024-10-23
add logger name to `LogfireLoggingHandler` spans

samuelcolvin:
As requested by @OliverFarren today on our call earlier today.

@OliverFarren I checked and, and attributes sent through via the `extra` kwarg do get set as attributes.


alexmojaki (approved):
@samuelcolvin upon further consideration I think it's better to stick to this rather than touching scope.

PR #535, opened 2024-10-23
instrument docs searches

ChristopherGS:
Instruments searches in the docs (cookieless)


docs/overrides/main.html
```diff
@@ -3,5 +3,87 @@
...
+
+                // Don't track if term hasn't changed or is too short
+                if (searchTerm === lastSearchTerm || searchTerm.length < 3) {
+                    return;
+                }
+
+                searchTimeout = setTimeout(function() {
+                    const now = Date.now();
+                    // Only track if:
+                    // 1. Results exist
```
alexmojaki: If someone searches for something and it's not in our docs, that seems potentially valuable, although I agree it's likely to be because they searched for nonsense. What if we tracked, but included the number of search results in the captured event, and filtered on that number in posthog?

docs/overrides/main.html
```diff
@@ -3,5 +3,87 @@
...
+                        });
+                        lastSearchTerm = searchTerm;
+                        lastSearchTime = now;
+                    }
+                }, DEBOUNCE_MS);
+            });
+
+            // Track submitted searches
+            searchForm.addEventListener('submit', function(e) {
+                e.preventDefault();
```
alexmojaki: Why?

docs/overrides/main.html
```diff
@@ -3,5 +3,87 @@
...
+                clearTimeout(searchTimeout); // Cancel any pending typed search
+                const searchTerm = searchInput.value.trim();
+
+                // Only track if:
+                // 1. Term exists
+                // 2. Term is different from last search OR it's been a while
```
alexmojaki: IIUC, when I type something and press enter, it will either send a 'typed' event or a 'submitted' event for the final text, but never both, and which one it sends depends on how long I wait before pressing enter. that seems weird.

docs/overrides/main.html
```diff
@@ -3,5 +3,87 @@
...
+                }
+            });
+
+            // Track result clicks (unchanged)
+            searchResults.addEventListener('click', function(e) {
+                var clickedLink = e.target.closest('a');
+                if (clickedLink) {
+                    posthog.capture('search_result_click', {
+                        search_term: searchInput.value.trim(),
+                        clicked_url: clickedLink.href,
```
alexmojaki: If I search for `foo` then the links include `?h=foo` at the end for highlighting, probably want to trim that so that we can sensibly group by the search result URLs.

Also I suggest splitting out the URL fragment as another attribute.

docs/overrides/main.html
```diff
@@ -3,5 +3,87 @@
...
+                        clicked_text: clickedLink.innerText.trim()
```
alexmojaki: Sometimes a single `li.md-search-result__item` can contain multiple `a.md-search-result__link` elements, where the first contains an `h1` that would be great to capture here as a human readable title of the page they're going to, and subsequent links contain longer text (which will be hard to do analytics on) and a URL fragment pointing to a specific section of that page that matches the search term.

docs/overrides/main.html
```diff
@@ -3,5 +3,87 @@
...
+                // Don't track if term hasn't changed or is too short
+                if (searchTerm === lastSearchTerm || searchTerm.length < 3) {
+                    return;
+                }
+
+                searchTimeout = setTimeout(function() {
+                    const now = Date.now();
+                    // Only track if:
+                    // 1. Results exist
+                    // 2. Enough time has passed since last search
```
alexmojaki: Probably best to leave it, but the managing of timeouts should mean this is redundant, right? Might be worth a comment for clarity.

docs/overrides/main.html
ChristopherGS: Keeps the search experience in MkDocs' client-side system (doesn't try and submit the form to a server)

docs/overrides/main.html
ChristopherGS: I've updated so the behaviour is more logical:

- Quick typing + enter = submit event
- Slow typing = typed event
- Slow typing + enter = typed event followed by submit event

docs/overrides/main.html
ChristopherGS: Good idea, done

docs/overrides/main.html
```diff
@@ -3,5 +3,85 @@
...
+// Helper function to track search events
+function trackSearchEvent(searchTerm, searchType, resultCount) {
+    posthog.capture('documentation_search', {
+        search_term: searchTerm,
+        search_type: searchType,
+        result_count: resultCount,
+        has_results: resultCount > 0
+    });
+    lastSearchTerm = searchTerm;
+    lastSearchTime = Date.now();
```
alexmojaki: surely this sets global variables separate from those declared in the event listener?

docs/overrides/main.html
alexmojaki: please add a comment to the code explaining in more detail. i don't fully understand, but it does look like it improves something, because when i try submitting multiple searches in the prod docs they don't work after the first.

ChristopherGS: Added. Sadly this won't fix things in prod, it'll just make sure the JS doesn't interfere with existing mkdocs search functionality

docs/overrides/main.html
ChristopherGS: Done

docs/overrides/main.html
ChristopherGS: shall we see how this performs and then take a view?

docs/overrides/main.html
ChristopherGS: tidied

docs/overrides/main.html
ChristopherGS: tidied

docs/overrides/main.html
```diff
@@ -3,5 +3,90 @@
 {% block content %}
 {{ super() }}
 <script src="/flarelytics/client.js"></script>
+<script>
+    if (!window.__posthogInitialized) {
```
alexmojaki, line 9: it seems impossible for `window.__posthogInitialized` to be true here since this doesn't happen in a callback or anything

docs/overrides/main.html
```diff
@@ -3,5 +3,90 @@
...
+
+        try {
+            posthog.init('phc_WEVBsfcilPS9u6e0CY22d8DqtFt0zalAP9A6iyCBHNH', {
+                api_host: 'https://eu.i.posthog.com',
+                persistence: 'memory'
+            });
+            window.__posthogInitialized = true;
+        } catch (error) {
+            console.error('PostHog initialization failed:', error);
```
alexmojaki, line 21: we still do everything else if there's an error here. in the worst case, i worry that search might actually break

docs/overrides/main.html
```diff
@@ -3,5 +3,90 @@
...
+            posthog.capture('documentation_search', {
+                search_term: term,
+                search_type: type,
+                result_count: resultCount,
+                has_results: resultCount > 0,
+                timestamp: Date.now()
+            });
+
+            lastSearch = { term, time: Date.now() };
+        }, 100);
```
alexmojaki: please extract these numbers to named constants. in particular i can't tell if this 100 is meant to actually be 1000, and another reader might not notice the difference

docs/overrides/main.html
```diff
@@ -3,5 +3,90 @@
...
+    if (!elements.form || !elements.input || !elements.results) return;
+
+    let searchTimeout;
+    let lastSearch = { term: '', time: 0 };
+
+    // Track search with debouncing
+    function trackSearch(term, type) {
+        if (term.length < 3 || term === lastSearch.term || Date.now() - lastSearch.time < 1000) return;
+
+        setTimeout(() => {
```
alexmojaki, line 49: please add a comment explaining why setTimeout is used here. i'm guessing it's to give mkdocs time to render search results? is 100ms reliably enough?

docs/overrides/main.html
alexmojaki: it looks like you did exactly what i had in mind, and it looks like it works well

docs/overrides/main.html
```diff
@@ -3,5 +3,90 @@
...
+document.addEventListener('DOMContentLoaded', function() {
+
+    // Setup
+    const elements = {
+        form: document.querySelector('.md-search__form'),
+        input: document.querySelector('[data-md-component="search-query"]'),
+        results: document.querySelector('.md-search__output')
+    };
+
+    if (!elements.form || !elements.input || !elements.results) return;
```
alexmojaki:
```suggestion
    if (!elements.form || !elements.input || !elements.results || !window.__posthogInitialized) return;
```

docs/overrides/main.html
alexmojaki: added a suggestion separately that i think covers this. also i suppose that event listeners should execute in isolation so maybe there was no risk to begin with.

docs/overrides/main.html
ChristopherGS: moved into callback

docs/overrides/main.html
ChristopherGS: now I `return` here, so safer

docs/overrides/main.html
ChristopherGS: done with explanation

docs/overrides/main.html
ChristopherGS: Added, also bumped to 200ms which I reckon is enough

alexmojaki (approved):
Thanks!

PR #543, opened 2024-10-25
@instrument without needing source code

alexmojaki:
Changes the `@logfire.instrument` decorator implementation to a normal wrapper instead of AST modification, especially so that it can be used without access to the source code.

A side effect of the old implementation was that functions decorated with `@logfire.instrument` (or nested within such a function) were not auto-traced. This will no longer be the case, so more functions will be auto-traced. Use `@logfire.no_auto_trace` anywhere on functions you want to exclude, especially the instrumented function.

This change also means that decorated async generator functions won't support the `.asend` method properly - the generator will only receive `None`. But `instrument` shouldn't be used any generators anyway unless the generator is being used as a context manager, so new warnings and documentation about this have been added.


tests/test_auto_trace.py
```diff
@@ -73,34 +73,36 @@ def test_auto_trace_sample(exporter: TestExporter) -> None:
...
                 'parent': {'trace_id': 1, 'span_id': 1, 'is_remote': False},
                 'start_time': 2000000000,
                 'end_time': 3000000000,
                 'attributes': {
                     'code.filepath': 'foo.py',
                     'code.lineno': 123,
-                    'code.function': 'async_gen',
-                    'logfire.msg_template': 'Calling async_gen via @instrument',
+                    'code.function': 'async_gen.<locals>.inner',
+                    'logfire.msg_template': 'Calling tests.auto_trace_samples.foo.async_gen.<locals>.inner',
```
davidhewitt, line 102: Is there a reason the template isn't just something like `Calling {func}`?

logfire/_internal/instrument.py
```diff
@@ -20,118 +22,117 @@
...
+            def open_span(*_: P.args, **__: P.kwargs):
+                return logfire._fast_span(final_span_name, attributes)  # type: ignore
+
+        if inspect.isgeneratorfunction(func):
+            if not allow_generator:
+                warnings.warn(GENERATOR_WARNING_MESSAGE, stacklevel=2)
+
+            def wrapper(*func_args: P.args, **func_kwargs: P.kwargs):  # type: ignore
+                with open_span(*func_args, **func_kwargs):
+                    yield from func(*func_args, **func_kwargs)
```
davidhewitt, line 80: Is this more correct? Because it avoids keeping the span on the stack when the generator is suspended?

```suggestion
                while True:
                    with open_span(*func_args, **func_kwargs):
                        value = func(*func_args, **func_kwargs)
                    yield
```

... I think there are probably things to fix like sending values and `StopIteration`.

davidhewitt (commented):
Overall this seems reasonable to me, and fixes support for Rust functions 🤤 

> A side effect of the old implementation was that functions decorated with `@logfire.instrument` (or nested within such a function) were not auto-traced. This will no longer be the case, so more functions will be auto-traced. Use `@logfire.no_auto_trace` anywhere on functions you want to exclude, especially the instrumented function.

I think I might be missing a technical reason why this has to be the case. Can we just make the AST transformer recognise `@logfire.instrument` applications? (Why is `@logfire.no_auto_trace` any better, it's just a different name?)

alexmojaki:
> I think I might be missing a technical reason why this has to be the case. Can we just make the AST transformer recognise `@logfire.instrument` applications? (Why is `@logfire.no_auto_trace` any better, it's just a different name?)

See https://pydantic.slack.com/archives/C05AF4A4WRM/p1730115881076909

We can if we're willing to accept some inconsistency/imperfection. `no_auto_trace` is a sufficiently unique name that we can assume the meaning of `@no_auto_trace`. Do we detect `@instrument` without the `logfire.` prefix and also assume it comes from logfire (full on static analysis is probably not going to happen)? Or should `@logfire.instrument` and `@instrument` behave differently? Are we OK with decorating with `@` behaving differently from applying the function manually after the definition?

tests/test_auto_trace.py
alexmojaki: Auto-tracing has always included the fully qualified name. The name wasn't qualified in this test before because that function was instrumented with a hardcoded template.

If you mean 'why is the template the full message rather than something template-like' it's because auto-tracing is optimised to not go through message formatting, and these are still low cardinality enough for span names. Searching for `span_name = 'Calling {func}'` wouldn't be very useful. It's possible to filter on `otel_scope_name = 'logfire.auto_tracing'` instead.

logfire/_internal/instrument.py
alexmojaki: No, not just because it needs fixing for things, but because we generally actually want the span to be open while the generator is suspended.

davidhewitt: Ah, we need to avoid closing the span, of course. So the context manager might be incorrect, but maybe there's another way to achieve this (e.g. by manipulating the OTEL trace context)?

davidhewitt:
Understood. It feels like there should be a way to work around it at runtime somehow, e.g. by setting auto-traced `__code__` to have a special `__logfire_auto_traced__` attribute, which `instrument` could at least to try to detect.

(Or maybe `instrument` could set some some context manager which auto-traced functions could then inspect to decide if they need to avoid the auto-tracing. But this seems higher overhead so probably unacceptable.)

But I assume you have already thought about this problem much harder than I am, so I am 👍 to accept there is no way.

davidhewitt (approved):
It feels like the generator could be improved, but that could be a follow-up I think.

Issue #542, opened 2024-10-25
LLM tool call not shown for streamed response


jackmpcollins:
### Description

The Logfire UI nicely shows the tool call by an LLM for non-streamed responses

<img width="669" alt="image" src="https://github.com/user-attachments/assets/d363085f-ca73-4af9-a31f-9d74c0294836">

But for streamed responses the Assistant box is empty.

<img width="663" alt="image" src="https://github.com/user-attachments/assets/6550e674-5c68-4127-8dac-2cff1dd6d080">

Code to reproduce

```python
# Test logfire streamed responsea

from openai import Client

import logfire

logfire.configure()
logfire.instrument_openai()

client = Client()

response = client.chat.completions.create(
    model="gpt-4o",
    messages=[{"role": "user", "content": "Create a Superhero named Monkey Boy."}],
    stream=True,
    stream_options={"include_usage": True},
    tool_choice={"type": "function", "function": {"name": "return_superhero"}},
    tools=[
        {
            "type": "function",
            "function": {
                "name": "return_superhero",
                "parameters": {
                    "properties": {
                        "name": {"title": "Name", "type": "string"},
                        "age": {"title": "Age", "type": "integer"},
                        "power": {"title": "Power", "type": "string"},
                        "enemies": {
                            "items": {"type": "string"},
                            "title": "Enemies",
                            "type": "array",
                        },
                    },
                    "required": ["name", "age", "power", "enemies"],
                    "type": "object",
                },
            },
        },
    ],
)
for chunk in response:
    print(chunk)
```

Related (closed) issue: https://github.com/pydantic/logfire/issues/54

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="0.50.1"
platform="macOS-15.0.1-arm64-arm-64bit"
python="3.10.12 (main, Jul 15 2023, 09:54:16) [Clang 14.0.3 
(clang-1403.0.22.14.1)]"
[related_packages]
requests="2.32.3"
pydantic="2.8.2"
openai="1.52.0"
protobuf="4.25.3"
rich="13.7.1"
tomli="2.0.1"
executing="2.0.1"
opentelemetry-api="1.25.0"
opentelemetry-exporter-otlp-proto-common="1.25.0"
opentelemetry-exporter-otlp-proto-http="1.25.0"
opentelemetry-instrumentation="0.46b0"
opentelemetry-proto="1.25.0"
opentelemetry-sdk="1.25.0"
[...]

PR #545, opened 2024-10-27
Log LLM tool call for streamed response

jackmpcollins:
The current LLM streamed response parsing only collects the message content/text. This PR generalizes this to allow collecting any state needed to generate the response_data. Streamed OpenAI chat completions now include the tool calls in the Logfire UI. Behaviour for Anthropic and for OpenAI completions (non-chat) remains the same.

Fixes https://github.com/pydantic/logfire/issues/542

image


logfire/_internal/integrations/llm_providers/types.py
```diff
@@ -1,13 +1,28 @@
...
 from typing_extensions import LiteralString
 
 
+class StreamState(ABC):
+    """Keeps track of the state of a streamed response."""
+
+    @abstractmethod
+    def record_chunk(self, chunk: Any) -> None:
+        """Update the state based on a chunk from the streamed response."""
+        ...  # pragma: no cover
```
alexmojaki:
```suggestion
```

logfire/_internal/integrations/llm_providers/types.py
```diff
@@ -1,13 +1,28 @@
...
+
+    @abstractmethod
+    def get_response_data(self) -> Any:
+        """Returns the response data for including in the log."""
+        ...  # pragma: no cover
```
alexmojaki:
```suggestion
```

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -1,16 +1,17 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import openai
 from openai._legacy_response import LegacyAPIResponse
+from openai.lib.streaming.chat._completions import ChatCompletionStreamState
```
alexmojaki: I'm concerned about this private import. Do we have other options? Can you check if this works for older openai versions?

@Kludex shouldn't this have been flagged by pre-commit?

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -72,10 +70,42 @@ def content_from_completions(chunk: Completion | None) -> str | None:
...
+    def get_response_data(self) -> Any:
+        try:
+            final_completion = self._stream_state.current_completion_snapshot
+        except AssertionError:
+            # AssertionError is raised when there is no completion snapshot
+            # Return empty content to show an empty Assistant response in the UI
+            return {'combined_chunk_content': ''}
+        return {
+            'message': final_completion.choices[0].message if final_completion.choices else None,
+            'usage': final_completion.usage,
```
alexmojaki: I worry about this having a significantly different shape from the other response data dicts.

logfire/_internal/integrations/llm_providers/openai.py
Kludex: Which hook? This is already in `logfire._internal`.

alexmojaki: shouldn't pyright complain about private imports from third-party packages?

Kludex: https://github.com/microsoft/pyright/issues/9008

😅

logfire/_internal/integrations/llm_providers/openai.py
jackmpcollins: This is the same shape as the existing non-streamed chat completion, which is why it displays nicely in the UI.
https://github.com/pydantic/logfire/blob/af707e2e56fd026fde50fe3b6f93703129adbb2a/logfire/_internal/integrations/llm_providers/openai.py#L81-L91

I'll admit it's not exactly the same: the message object here is a `ParsedChatCompletionMessage`, a subclass of the non-streamed chat completion message `ChatCompletionMessage`. It has the "parsed" and tool call "parsed_arguments" fields added. These "ParsedX" classes are those returned by [the `client.beta.chat.completions.parse` method](https://platform.openai.com/docs/guides/structured-outputs/introduction).

Aside: it would be handy to have the response dicts that have special handling on the frontend be documented or codified (e.g. using TypedDicts), ideally with the order of precedence. For example I found that "combined_chunk_content" takes precedence over "message", which is why I excluded it.

logfire/_internal/integrations/llm_providers/openai.py
jackmpcollins: This was added in https://github.com/openai/openai-python/releases/tag/v1.40.0 released in August. So definitely makes sense to add handling for versions of the `openai` package that do not have it. What would be the best way to do that? / Is there an example of something similar to this in logfire already? My initial thought is to try/except the import and fall back to using the `OpenaiCompletionStreamState` if this is not available.

In terms of other options, there isn't a public function for this in the openai package. Similar code for recombining stream chunks into a message object could be added in logfire. It does seem like a broadly useful capability, so I will open an issue for it with openai.

jackmpcollins: openai issue: https://github.com/openai/openai-python/issues/1833

alexmojaki: > My initial thought is to try/except the import and fall back to using the `OpenaiCompletionStreamState` if this is not available.

That sounds right. And thanks for opening an issue.

logfire/_internal/integrations/llm_providers/openai.py
alexmojaki: ah, i didn't realise we were already using this shape. yes, creating some types sounds good. @dmontagu any thoughts? should we consider these shapes stable?

tests/otel_integrations/test_openai.py
```diff
@@ -63,15 +198,15 @@ def request_handler(request: httpx.Request) -> httpx.Response:
...
+                            cc_chunk.Choice(index=0, delta=cc_chunk.ChoiceDelta(content=' is secret', role='assistant'))
                         ],
                         created=1,
                         model='gpt-4',
                         object='chat.completion.chunk',
                     ),
                     cc_chunk.ChatCompletionChunk(
                         id='3',
-                        choices=[cc_chunk.Choice(index=2, delta=cc_chunk.ChoiceDelta(content=None, role='assistant'))],
+                        choices=[cc_chunk.Choice(index=0, delta=cc_chunk.ChoiceDelta(content=None, role='assistant'))],
```
alexmojaki: why did these change? why are both indices 0 now?

tests/otel_integrations/test_openai.py
```diff
@@ -513,14 +648,186 @@ def test_sync_chat_empty_response_choices(instrumented_client: openai.Client, ex
                     'logfire.span_type': 'log',
                     'logfire.tags': ('LLM',),
                     'duration': 1.0,
-                    'response_data': '{"combined_chunk_content":"","chunk_count":0}',
+                    'response_data': '{"message":null,"usage":null}',
```
alexmojaki, line 649: why the change?

tests/otel_integrations/test_openai.py
```diff
@@ -632,8 +939,8 @@ async def test_async_chat_completions_stream(
                     'logfire.span_type': 'log',
                     'logfire.tags': ('LLM',),
                     'duration': 1.0,
-                    'response_data': '{"combined_chunk_content":"The answer is secret","chunk_count":2}',
-                    'logfire.json_schema': '{"type":"object","properties":{"request_data":{"type":"object"},"async":{},"duration":{},"response_data":{"type":"object"}}}',
+                    'response_data': '{"message":{"content":"The answer is secret","refusal":null,"role":"assistantassistantassistant","audio":null,"function_call":null,"tool_calls":null,"parsed":null},"usage":null}',
```
alexmojaki: why is the role `assistantassistantassistant`?

tests/otel_integrations/test_openai.py
jackmpcollins: These chunks are all part of the first item in the "choices" list, so should all have index 0.

```python
from openai import Client

response = Client().chat.completions.create(
    model="gpt-4",
    messages=[{"role": "user", "content": "Just say: The answer is secret."}],
    stream=True,
    # stream_options={"include_usage": True},
    # n=2,
)
for chunk in response:
    print(chunk.model_dump_json(exclude_none=True))

{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":"","role":"assistant"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":"The"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":" answer"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":" is"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":" secret"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":"."},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{},"finish_reason":"stop","index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
```

tests/otel_integrations/test_openai.py
jackmpcollins: In the hardcoded test response the "role" field should be null except for in the first chunk. Fixed now in https://github.com/pydantic/logfire/pull/545/commits/7045b77b916731dde33dc9ed60658f8fbc27f65a

```python
from openai import Client

response = Client().chat.completions.create(
    model="gpt-4",
    messages=[{"role": "user", "content": "Just say: The answer is secret."}],
    stream=True,
    # stream_options={"include_usage": True},
    # n=2,
)
for chunk in response:
    print(chunk.model_dump_json(exclude_none=True))

{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":"","role":"assistant"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":"The"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":" answer"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":" is"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":" secret"},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{"content":"."},"index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASBkUgeHXBMarEwWnFLbT4vYMsldh","choices":[{"delta":{},"finish_reason":"stop","index":0}],"created":1731281710,"model":"gpt-4-0613","object":"chat.completion.chunk"}
```

tests/otel_integrations/test_openai.py
jackmpcollins: This is now the natural response_data for when the response has an empty "choices" list because of the switch to piecing together the message. In this case there are no pieces.

tests/otel_integrations/test_openai.py
alexmojaki: Still seems like the roles shouldn't be concatenated like that when they're present in the response data. Is that concatenation in our code, or openai's?

logfire/_internal/integrations/llm_providers/openai.py
alexmojaki: Will you add a try/except to the import?

tests/otel_integrations/test_openai.py
jackmpcollins: This behavior is from the openai code. They update the current snapshot of the ChatCompletion message by treating it as a dict and determining how to update each field based on its type. There is currently no special handling of the "role" field -- strings are concatenated.
https://github.com/openai/openai-python/blob/646a579cdb305a9d3fba6c5f9a96011c5e2c2882/src/openai/lib/streaming/_deltas.py#L27-L28

jackmpcollins: ollama "openai-compatible" API provides the role in every message so is not compatible with this. I'll open an issue with them.

```python
from openai import Client

client = Client(
    base_url="http://localhost:11434/v1",
    api_key="ollama",
)
response = client.chat.completions.create(
    model="llama3.1",
    messages=[{"role": "user", "content": "Just say: The answer is secret."}],
    stream=True,
)
for chunk in response:
    print(chunk.model_dump_json(exclude_none=True))
{"id":"chatcmpl-804","choices":[{"delta":{"content":"The","role":"assistant"},"index":0}],"created":1731398387,"model":"llama3.1","object":"chat.completion.chunk","system_fingerprint":"fp_ollama"}
{"id":"chatcmpl-804","choices":[{"delta":{"content":" answer","role":"assistant"},"index":0}],"created":1731398388,"model":"llama3.1","object":"chat.completion.chunk","system_fingerprint":"fp_ollama"}
{"id":"chatcmpl-804","choices":[{"delta":{"content":" is","role":"assistant"},"index":0}],"created":1731398388,"model":"llama3.1","object":"chat.completion.chunk","system_fingerprint":"fp_ollama"}
{"id":"chatcmpl-804","choices":[{"delta":{"content":" secret","role":"assistant"},"index":0}],"created":1731398388,"model":"llama3.1","object":"chat.completion.chunk","system_fingerprint":"fp_ollama"}
{"id":"chatcmpl-804","choices":[{"delta":{"content":".","role":"assistant"},"index":0}],"created":1731398388,"model":"llama3.1","object":"chat.completion.chunk","system_fingerprint":"fp_ollama"}
{"id":"chatcmpl-804","choices":[{"delta":{"content":"","role":"assistant"},"finish_reason":"stop","index":0}],"created":1731398388,"model":"llama3.1","object":"chat.completion.chunk","system_fingerprint":"fp_ollama"}
```

The message still appears in the logfire UI, but has the concatenated role.

image

logfire/_internal/integrations/llm_providers/openai.py
jackmpcollins: Added now in b5b26e8752514fab527b1487717c1d3d57661ad9. I just try-excepted the whole class definition, which is similar to how `_ProxyGauge` dependency on `Gauge` is done here https://github.com/jackmpcollins/logfire/blob/b5b26e8752514fab527b1487717c1d3d57661ad9/logfire/_internal/metrics.py#L309-L322  Please let me know if you have a preferred way to do this.

alexmojaki (approved):
Thanks! Are you happy for me to merge?

alexmojaki:
cc @willbakst in case you want to take a look

jackmpcollins:
> Thanks! Are you happy for me to merge?

Yes, happy to merge! Thanks

Issue #440, opened 2024-09-24
auto-tracing breaks numpy


samuelcolvin:
### Description

Running the following

```py
import logfire

logfire.configure()
logfire.install_auto_tracing(modules=['numpy'])

import numpy
```

Gives an error:

```
10:58:46.166 Calling numpy._globals._NoValueType.__new__
10:58:46.167 Calling numpy._utils.set_module
10:58:46.167 Calling numpy._utils.set_module.<locals>.decorator
10:58:46.181 Calling numpy._core._exceptions._display_as_base
10:58:46.181 Calling numpy._core._exceptions._display_as_base
10:58:46.182 Calling numpy._core._exceptions._display_as_base
10:58:46.182 Calling numpy._core._exceptions._display_as_base
10:58:46.182 Calling numpy._core._exceptions._display_as_base
10:58:46.182 Calling numpy._core._exceptions._display_as_base
10:58:46.183 Calling numpy.dtypes._add_dtype_helper
10:58:46.183 Calling numpy.dtypes._add_dtype_helper
10:58:46.183 Calling numpy.dtypes._add_dtype_helper
10:58:46.183 Calling numpy.dtypes._add_dtype_helper
10:58:46.184 Calling numpy.dtypes._add_dtype_helper
10:58:46.184 Calling numpy.dtypes._add_dtype_helper
10:58:46.184 Calling numpy.dtypes._add_dtype_helper
10:58:46.184 Calling numpy.dtypes._add_dtype_helper
10:58:46.184 Calling numpy.dtypes._add_dtype_helper
10:58:46.184 Calling numpy.dtypes._add_dtype_helper
10:58:46.184 Calling numpy.dtypes._add_dtype_helper
10:58:46.185 Calling numpy.dtypes._add_dtype_helper
10:58:46.185 Calling numpy.dtypes._add_dtype_helper
10:58:46.185 Calling numpy.dtypes._add_dtype_helper
10:58:46.185 Calling numpy.dtypes._add_dtype_helper
10:58:46.185 Calling numpy.dtypes._add_dtype_helper
10:58:46.185 Calling numpy.dtypes._add_dtype_helper
10:58:46.185 Calling numpy.dtypes._add_dtype_helper
10:58:46.186 Calling numpy.dtypes._add_dtype_helper
10:58:46.186 Calling numpy.dtypes._add_dtype_helper
10:58:46.186 Calling numpy.dtypes._add_dtype_helper
[...]

PR #550, opened 2024-10-28
Preserve docstrings in auto-tracing

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/440

### DIFF
--- a/logfire/_internal/ast_utils.py
+++ b/logfire/_internal/ast_utils.py
@@ -68,21 +68,36 @@ class BaseTransformer(ast.NodeTransformer):
         # Replace the body of the function with:
         #     with <logfire_method_call_node>:
         #         <original body>
+        body = node.body.copy()
+        new_body: list[ast.stmt] = []
+        if (
+            body
+            and isinstance(body[0], ast.Expr)
+            and isinstance(body[0].value, ast.Constant)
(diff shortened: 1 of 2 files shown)


logfire/_internal/ast_utils.py
```diff
@@ -68,21 +68,36 @@ def rewrite_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef, qualnam
         # Replace the body of the function with:
         #     with <logfire_method_call_node>:
         #         <original body>
+        body = node.body.copy()
+        new_body: list[ast.stmt] = []
+        if (
+            body
+            and isinstance(body[0], ast.Expr)
```
davidhewitt, line 75: I suppose it is guaranteed that `0` must exist (even if it is just `pass` or `...`)?

davidhewitt:
All looks good to me, sorry for the late review 👍

logfire/_internal/ast_utils.py
alexmojaki: I'm pretty sure it is, but I started the condition with `body and` just in case.

davidhewitt: Ah right, at this point Rust's lack of "empty collection is not truthy" has eroded this from my mind 😂

PR #556, opened 2024-10-30
fix docs redirects

samuelcolvin:
We need to be careful about adding redirects for any changes of URL.

image

### DIFF
--- a/logfire/integrations/logging.py
+++ b/logfire/integrations/logging.py
@@ -51,7 +51,7 @@ if TYPE_CHECKING:
 
 
 class LogfireLoggingHandler(LoggingHandler):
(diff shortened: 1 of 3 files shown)


Kludex:
Is there a way to enforce this?

Kludex (approved):

samuelcolvin:
> Is there a way to enforce this?

Not that I know of, but would be really good if we could.

Kludex:
> > Is there a way to enforce this?
> 
> Not that I know of, but would be really good if we could.

I think we can add a pre-commit hook if a `.md` file under `docs` changes only the location, and ask for a change on the `redirects`.

PR #557, opened 2024-10-30
Add `logfire.configure(environment=...)`

Kludex:


logfire/_internal/config.py
```diff
@@ -679,6 +692,8 @@ def _initialize(self) -> None:
                 )
             if self.service_version:
                 otel_resource_attributes[ResourceAttributes.SERVICE_VERSION] = self.service_version
+            if self.environment:
+                otel_resource_attributes[RESOURCE_ATTRIBUTES_DEPLOYMENT_ENVIRONMENT_NAME] = self.environment
```
Kludex, line 724: It's unfortunate that `ResourceAttributes.DEPLOYMENT_ENVIRONMENT` exists, but `ResourceAttributes.DEPLOYMENT_ENVIRONMENT_NAME` doesn't. And the first is deprecated...

https://opentelemetry.io/docs/specs/semconv/attributes-registry/deployment/

logfire/_internal/config.py
```diff
@@ -252,6 +254,7 @@ def configure(  # noqa: D417
         service_name: Name of this service. Defaults to the `LOGFIRE_SERVICE_NAME` environment variable.
         service_version: Version of this service. Defaults to the `LOGFIRE_SERVICE_VERSION` environment variable, or the
             current git commit hash if available.
+        environment: The environment this service is running in. Defaults to the `LOGFIRE_ENVIRONMENT` environment variable.
```
```suggestion
        environment: The environment this service is running in, e.g. `'staging'` or `'prod'`. Sets the `deployment.environment.name` resource attribute. Useful for filtering within projects in the Logfire UI. Defaults to the `LOGFIRE_ENVIRONMENT` environment variable.
```

but with newlines

logfire/_internal/config.py
alexmojaki: Maybe open an otel issue? And put this link in code.

logfire/_internal/config.py
alexmojaki: and link to otel docs

alexmojaki (approved):

logfire/_internal/config.py
```diff
@@ -252,13 +253,24 @@ def configure(  # noqa: D417
...
-        service_version: Version of this service. Defaults to the `LOGFIRE_SERVICE_VERSION` environment variable, or the
-            current git commit hash if available.
+        send_to_logfire: Whether to send logs to logfire.dev.
+
+            Defaults to the `LOGFIRE_SEND_TO_LOGFIRE` environment variable if set, otherwise defaults to `True`.
+            If `if-token-present` is provided, logs will only be sent if a token is present.
+        token: The project token.
+
+            Defaults to the `LOGFIRE_TOKEN` environment variable.
+        service_name: Name of this service.
```
```suggestion
            Defaults to the `LOGFIRE_TOKEN` environment variable.

        service_name: Name of this service.
```

can we do this to make it more readable? otherwise it's confusing which lines are grouped together.

alexmojaki (approved):

PR #561, opened 2024-10-30
Check `.logfire` for creds to respect `'if-token-present'` setting

sydney-runkle:
Fix https://github.com/pydantic/logfire/issues/527


tests/test_configure.py
```diff
@@ -1297,6 +1297,28 @@ def test_send_to_logfire_if_token_present_not_empty(capsys: pytest.CaptureFixtur
...
+            "logfire_api_url": "http://dash.localhost:8000/"
+        }
+        """
+    )
+    with requests_mock.Mocker() as request_mocker:
+        request_mocker.get(
+            'https://logfire-api.pydantic.dev/v1/info',
+            json={'project_name': 'myproject', 'project_url': 'http://dash.localhost:8000/'},
+        )
+        configure(send_to_logfire='if-token-present', data_dir=tmp_path)
```
alexmojaki, line 1341:
```suggestion
        configure(send_to_logfire='if-token-present', data_dir=tmp_path)
        wait_for_check_token_thread()
```

logfire/_internal/config.py
```diff
@@ -801,31 +801,35 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
             if isinstance(self.metrics, MetricsOptions):
                 metric_readers = list(self.metrics.additional_readers)
 
-            if (self.send_to_logfire == 'if-token-present' and self.token is not None) or self.send_to_logfire is True:
+            credentials = LogfireCredentials.load_creds_file(self.data_dir)  # pragma: no branch
```
```suggestion
            credentials = LogfireCredentials.load_creds_file(self.data_dir)
```

tests/test_configure.py
```diff
@@ -1297,6 +1297,29 @@ def test_send_to_logfire_if_token_present_not_empty(capsys: pytest.CaptureFixtur
...
+            "token": "foobar",
+            "project_name": "myproject",
+            "project_url": "http://dash.localhost:8000/",
+            "logfire_api_url": "http://dash.localhost:8000/"
+        }
+        """
+    )
+    with requests_mock.Mocker() as request_mocker:
+        request_mocker.get(
+            'https://logfire-api.pydantic.dev/v1/info',
```
alexmojaki, line 1338: /v1/info gets requested by `_initialize_credentials_from_token`, which isn't called when the token is retrieved from a file. maybe it should (e.g. to tell the user that their free usage has been exhausted, which is a new possibility), but it doesn't now and hasn't in the past. that's why this test isn't working.

sydney-runkle: Ohhh, thanks so much @alexmojaki, that makes a lot more sense. Is there another endpoint that I should test here that would make more sense for this context?

alexmojaki: i think the token should be checked even when it comes from a file

logfire/_internal/config.py
```diff
@@ -801,16 +801,20 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
             if isinstance(self.metrics, MetricsOptions):
                 metric_readers = list(self.metrics.additional_readers)
 
+            creds_from_file = LogfireCredentials.load_creds_file(self.data_dir)
+            if creds_from_file:
+                self.token = creds_from_file.token
```
alexmojaki: No, a directly configured token should take priority

alexmojaki: not just directly, a token configured in any way (env var, pyproject.toml) should take priority over the creds file

sydney-runkle: Makes sense, fixed.

logfire/_internal/config.py
```diff
@@ -801,16 +801,21 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
             if isinstance(self.metrics, MetricsOptions):
                 metric_readers = list(self.metrics.additional_readers)
 
+            creds_from_file = LogfireCredentials.load_creds_file(self.data_dir)
+            # token from creds file takes the lowest priority
+            if self.token is None and creds_from_file is not None:
```
```suggestion
            if not self.token and creds_from_file is not None:
```

logfire/_internal/config.py
```diff
@@ -801,16 +801,21 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
             if isinstance(self.metrics, MetricsOptions):
                 metric_readers = list(self.metrics.additional_readers)
 
+            creds_from_file = LogfireCredentials.load_creds_file(self.data_dir)
```
alexmojaki: This should probably be avoided if `send_to_logfire` is false

alexmojaki: or if `self.token` is set

samuelcolvin:
Please can we get this merged and released, I need it for what I'm working on 🙏.

sydney-runkle:
Sure thing, will finish up today!

sydney-runkle:
Still have a few follow up questions to resolve with @alexmojaki, then will push across the line.

logfire/_internal/config.py
```diff
@@ -801,65 +801,73 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
...
+                show_project_link: bool = self.console and self.console.show_project_link or False
 
+                # try loading credentials (and thus token) from file if a token is not already available
+                # this takes the lowest priority, behind the token passed to `configure` and the environment variable
                 if self.token is None:
-                    if (credentials := LogfireCredentials.load_creds_file(self.data_dir)) is None:  # pragma: no branch
+                    credentials = LogfireCredentials.load_creds_file(self.data_dir)
+
+                    # if we still don't have a token, try initializing a new project and writing a new creds file
+                    # note, we only do this if `send_to_logfire` is explicitly `True`
```
```suggestion
                    # note, we only do this if `send_to_logfire` is explicitly `True`, not 'if-token-present'
```

alexmojaki:
1 and 3 are OK, it's good to always check the token now because it may be rejected due to exceeded usage.

The test is failing because it previously only made requests using the user token `fake_user_token` stored in a toml file. Those requests created the project and a write token `fake_token`. It now checks `fake_token` and the tests don't expect that. You just need to tweak the assertion that's in a loop to be more specific.

sydney-runkle:
> The test is failing because it previously only made requests using the user token fake_user_token stored in a toml file. Those requests created the project and a write token fake_token. It now checks fake_token and the tests don't expect that. You just need to tweak the assertion that's in a loop to be more specific.

Great, thanks for the feedback! Done!

tests/test_configure.py
```diff
@@ -1122,9 +1138,12 @@ def test_initialize_project_create_project(tmp_dir_cwd: Path, tmp_path: Path, ca
 
         logfire.configure(send_to_logfire=True)
 
-        for request in request_mocker.request_history:
+        for request in request_mocker.request_history[:-1]:
             assert request.headers['Authorization'] == 'fake_user_token'
 
+        # we check that fake_token is valid now when we configure the project
+        assert request_mocker.request_history[-1].headers['Authorization'] == 'fake_token'
```
alexmojaki, line 1146:
```suggestion
        wait_for_check_token_thread()
        assert request_mocker.request_history[-1].headers['Authorization'] == 'fake_token'
```

sydney-runkle:
Thanks for the feedback :). I think I understand our auth patterns better after working in this space a bit more :)

PR #569, opened 2024-11-04
fix: declare variables which should not be managed by inline-snapshot

15r10nk:
things which can be changed with the **next** inline-snapshot version to make clear who is responsible for the source code (inline-snapshot or the developer).

### DIFF
--- a/tests/otel_integrations/test_redis.py
+++ b/tests/otel_integrations/test_redis.py
@@ -2,7 +2,7 @@ import sys
 from typing import Any, Iterator
 
 import pytest
-from inline_snapshot import snapshot
+from inline_snapshot import Is, snapshot
 from opentelemetry.instrumentation.redis import RedisInstrumentor
 from opentelemetry.trace import Span
 from redis import Connection, Redis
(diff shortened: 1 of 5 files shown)


alexmojaki:
I don't understand this. What's the difference between `Is(foo)` and `foo` when `foo` is a variable name?

15r10nk:
@alexmojaki, you are fast :smile:. 

I published this to show what I'm currently working on https://github.com/15r10nk/inline-snapshot/issues/123.

This makes only sense if you know that `Is(...)` will be a part of the next inline-snapshot version which allows the developer to take back control and to change code inside snapshots.

This makes it explicit who controls the source code and inline-snapshot will no longer try to change you code back with "update".

The value inside `Is(...)` can also change if the snapshot is evaluated multiple times and allow some constructs which are not possible today.

samuelcolvin:
thanks, personally I think this looks good, but @alexmojaki and @Kludex mostly maintain this repo, so it's up to them.

alexmojaki:
Closing because I think the underlying issue about long reports is resolved with the newest inline-snapshot and I don't think it's worth fixing merge conflicts here.

PR #570, opened 2024-11-04
Allow `extract_args` to be an iterable of argument names

alexmojaki:


logfire/_internal/instrument.py
```diff
@@ -112,6 +100,63 @@ def wrapper(*func_args: P.args, **func_kwargs: P.kwargs) -> R:
...
+    extract_args: bool | Iterable[str],
+    func: Callable[P, R],
+) -> Callable[P, ContextManager[Any]]:
+    final_span_name: str = span_name or attributes[ATTRIBUTES_MESSAGE_TEMPLATE_KEY]  # type: ignore
+
+    # This is the fast case for when there are no arguments to extract
+    def open_span(*_: P.args, **__: P.kwargs):  # type: ignore
+        return logfire._fast_span(  # type: ignore
+            final_span_name, attributes
+        )
```
Kludex: Really not important... Really.
```suggestion
        return logfire._fast_span(final_span_name, attributes)  # type: ignore
```

Kludex (approved):
Nice! 👍

Kludex:
@dmontagu we can use this on the scheduler/worker when this is released.

PR #572, opened 2024-11-05
Revert "Temporarily stop using pydantic main for testing (#538)"

sydney-runkle:
This reverts commit 6d7bb92fe83485250ff99a5c31a2e2455617779d.

### DIFF
--- a/logfire/_internal/json_schema.py
+++ b/logfire/_internal/json_schema.py
@@ -68,7 +68,7 @@ def type_to_schema() -> dict[type[Any], JsonDict | Callable[[Any, set[int]], Jso
                 pydantic.NameEmail: {'type': 'string', 'x-python-datatype': 'NameEmail'},
                 pydantic.SecretStr: {'type': 'string', 'x-python-datatype': 'SecretStr'},
                 pydantic.SecretBytes: {'type': 'string', 'x-python-datatype': 'SecretBytes'},
-                pydantic.AnyUrl: {'type': 'string', 'x-python-datatype': 'AnyUrl'},
+                pydantic.AnyUrl: {'type': 'string', 'x-python-datatype': 'Url'},
                 pydantic.BaseModel: _pydantic_model_schema,
             }
         )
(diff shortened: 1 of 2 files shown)


logfire/_internal/json_schema.py
```diff
@@ -68,15 +68,11 @@ def type_to_schema() -> dict[type[Any], JsonDict | Callable[[Any, set[int]], Jso
                 pydantic.NameEmail: {'type': 'string', 'x-python-datatype': 'NameEmail'},
                 pydantic.SecretStr: {'type': 'string', 'x-python-datatype': 'SecretStr'},
                 pydantic.SecretBytes: {'type': 'string', 'x-python-datatype': 'SecretBytes'},
+                pydantic.AnyUrl: {'type': 'string', 'x-python-datatype': 'AnyUrl'},
```
sydney-runkle: Can decide on how many of these you want to add 👍

alexmojaki:
```suggestion
                pydantic.AnyUrl: {'type': 'string', 'x-python-datatype': 'Url'},
```

hopefully that means the test changes can be reverted. the old tests make more sense because `repr(pydantic.AnyUrl('http://example.com'))` is `Url('http://example.com/')`.

sydney-runkle: @alexmojaki, with `pydantic` `main`, `repr(pydantic.AnyUrl('http://example.com'))` is `AnyUrl('http://example.com')`, which is more intuitive

alexmojaki: Let's go with `Url` for now so that tests pass and we can merge this and deal with the other new test failures from the otel release, we can revisit afterwards

sydney-runkle: Hmm, I still think things will fail here, as the `repr` is going to be different across different pydantic versions...

alexmojaki: i don't think the repr determines the output, it's just something we'd like to mimic

Issue #529, opened 2024-10-22
`logfire.instrument_httpx()` breaks OpenAI SDK


samuelcolvin:
### Description

Once I've run

```py
logfire.instrument_httpx()
```

if I then try to create an OpenAI client I get

```
  File "/Users/samuel/code/.../.venv/lib/python3.12/site-packages/openai/_client.py", line 337, in __init__
    super().__init__(
  File "/Users/samuel/code/.../.venv/lib/python3.12/site-packages/openai/_base_client.py", line 1420, in __init__
    raise TypeError(
TypeError: Invalid `http_client` argument; Expected an instance of `httpx.AsyncClient` but got <class 'httpx.AsyncClient'>
```

This is not our fault, but we could create a PR to OpenAI to try to fix it.

We should also allow

```
logfire.instrument_httpx(client)
```

So you can easily instrument one client but not all.

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="1.2.0"
platform="macOS-14.6.1-arm64-arm-64bit"
python="3.12.6 (main, Sep 20 2024, 17:42:55) [Clang 15.0.0 (clang-1500.0.40.1)]"
[related_packages]
requests="2.32.3"
pydantic="2.9.2"
openai="1.51.2"
protobuf="4.25.5"
rich="13.9.2"
executing="2.1.0"
opentelemetry-api="1.27.0"
opentelemetry-exporter-otlp-proto-common="1.27.0"
opentelemetry-exporter-otlp-proto-http="1.27.0"
opentelemetry-instrumentation="0.48b0"
opentelemetry-instrumentation-httpx="0.48b0"
opentelemetry-proto="1.27.0"
opentelemetry-sdk="1.27.0"
opentelemetry-semantic-conventions="0.48b0"
opentelemetry-util-http="0.48b0"
```

Kludex:
How our tests didn't get this? 🤔 

Can we use VCR instead of mocking things?

alexmojaki:
Duplicate of https://github.com/pydantic/logfire/issues/264 and https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2609

alexmojaki:
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2364#issuecomment-2429551345

alexmojaki:
This is supposedly fixed now in OTEL contrib, we need to check when they next release, but nothing needs to be done on our side.

PR #575, opened 2024-11-06
Allow instrumenting a single httpx client, fix tests for OTel 1.28

alexmojaki:


sydney-runkle:
Could this be related to https://github.com/pydantic/logfire/issues/529?

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "logfire"
 version = "2.1.2"
-description = "The best Python observability tool! 🪵🔥"
+description = "The best Python observability tool!"
```
alexmojaki, line 8: Pyright was failing to parse this file because of the emojis: https://github.com/microsoft/pyright/issues/9412

logfire/_internal/integrations/httpx.py
```diff
@@ -25,15 +25,26 @@ class HTTPXInstrumentKwargs(TypedDict, total=False):
         skip_dep_check: bool
 
 
-def instrument_httpx(logfire_instance: Logfire, **kwargs: Unpack[HTTPXInstrumentKwargs]) -> None:
+def instrument_httpx(
+    logfire_instance: Logfire, client: httpx.Client | None, **kwargs: Unpack[HTTPXInstrumentKwargs]
```
Kludex: I don't see the `httpx.AsyncClient` mentioned anywhere in the PR. What's the plan for it? Can't we just add `| httpx.AsyncClient`?

logfire/_internal/integrations/httpx.py
```diff
@@ -25,15 +25,26 @@ class HTTPXInstrumentKwargs(TypedDict, total=False):
...
-            'meter_provider': logfire_instance.config.get_meter_provider(),
-            **kwargs,
-        }
-    )
+    final_kwargs: dict[str, Any] = {
+        'tracer_provider': logfire_instance.config.get_tracer_provider(),
+        'meter_provider': logfire_instance.config.get_meter_provider(),
+        **kwargs,
+    }
+    del kwargs
```
Kludex: Why are we doing this?

pyproject.toml
Kludex: This is sad.

pyproject.toml
```diff
@@ -147,9 +147,9 @@ dev = [
     "numpy<1.24.4; python_version < '3.9'",
     "pytest-recording>=0.13.2",
     "uvicorn>=0.30.6",
-    # Logfire API
     "logfire-api",
     "requests",
+    "setuptools>=75.3.0",
```
Kludex, line 152: Why do we need to add setuptools?

alexmojaki: It's used in this code in `utils.py`:

```python
def get_version(version: str) -> Version:
    """Return a packaging.version.Version object from a version string.

    We check if `packaging` is available, falling back to `setuptools._vendor.packaging` if it's not.
    """
    try:
        from packaging.version import Version

    except ImportError:  # pragma: no cover
        # Trigger the sys.path change mentioned below, but discard this.
        from setuptools._vendor.packaging.version import Version
```

and it's no longer a dependency of OTEL.

We're essentially relying on either `packaging` or `setuptools` being installed. Currently we're safe because `packaging` is a dependency of `opentelemetry-instrumentation`, but there isn't actually much reason for `opentelemetry-instrumentation` to be a required dependency.

alexmojaki: Anyway, regardless of `packaging` being installed/required, we still need that logic because of the sys.path changes made by setuptools, so we need it as a dev dependency.

logfire/_internal/integrations/httpx.py
alexmojaki: Added a comment

Kludex (approved):
Maybe we want to test with `AsyncClient` as well?

But all good. Thanks.

Kludex:
I've created https://github.com/pydantic/logfire/issues/581 to document this. I can handle that. 👍

pyproject.toml
alexmojaki: The issue is closed, upgrading pyright should allow reverting this.

PR #576, opened 2024-11-07
pin deps in CI

samuelcolvin:
Also use the new [dev dependency groups in uv](https://docs.astral.sh/uv/concepts/dependencies/#development-dependencies).

@alexmojaki @Kludex I think you’re both off today, and I need to get CI passing, so I’m going to merge this to pin dependencies and thereby get tests passing.

As I said on slack, I think pinning deps in CI is the best route anyway, but happy to discuss other routes when your back, you can consider this a temporary fix until we discuss next week.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v3
         with:
-          version: "0.4.12"
+          version: "0.4.30"
           enable-cache: true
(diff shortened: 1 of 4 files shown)


Kludex:
Since we are doing this, it would be good to have dependabot as well.

Kludex:
The previous approach had the pro argument that we are faster to find out about incompatibilities caused by the optional dependencies. Which was the reason we made it that way.

PR #582, opened 2024-11-12
Fix pydantic tests again

alexmojaki:
I don't know why the last two commits in https://github.com/pydantic/logfire/commits/pydantic-main-again/ didn't make it into https://github.com/pydantic/logfire/pull/572


sydney-runkle:
thanks @alexmojaki for moving this across the line, have a good evening!

Issue #580, opened 2024-11-12
Daily CI with unpinned deps


alexmojaki:
Following up on https://github.com/pydantic/logfire/pull/576

At the bare minimum we need a job that runs daily with the latest versions of all dependencies so that we can detect breakages.

Previously regular CI ran with unpinned dependencies which forced us to fix the problem ASAP. That had downsides, but I want to keep the upsides. I want to have some kind of system so that when the job fails we treat it with similar urgency to before, so that new users aren't immediately faced with a critical error caused by an outstanding bug. Relevant developers should definitely get an email about the job failing, but that's not quite enough. I may miss the email in a sea of notifications, or I may think someone else will deal with it. For now let's just say that if any of us see that CI has failed we should open an urgent issue if there isn't one already, and check with the team what the plan is.

Kludex:
We can also add a weekly dependabot job?

Kludex:
Daily also works (the previous PR is closed).

alexmojaki:
That also works, but I also want daily CI for other things like testing a wider variety of versions as in https://github.com/pydantic/logfire/issues/507 without slowing down PRs.

Kludex:
I forgot that `uv.lock` doesn't seem supported by dependabot yet.

alexmojaki:
Reopening because based on https://github.com/orgs/community/discussions/25351 the current workflow will only send an email to a single person, which isn't loud enough.

alexmojaki:
(I know there's a TODO in that file in the PR which is probably because of this, I hadn't noticed until merging)

PR #583, opened 2024-11-12
Daily CI with unpinned deps

sydney-runkle:
FIxes https://github.com/pydantic/logfire/issues/580

### DIFF
--- a/.github/workflows/daily_deps_test.yml
+++ b/.github/workflows/daily_deps_test.yml
@@ -1,0 +1,62 @@
+name: Daily CI Job
+
+on:
+  schedule:
+    - cron: '0 12 * * *' # Daily at midnight UTC
+
+  # Can be triggered manually from the actions tab, if needed
+  workflow_dispatch:
+
+jobs:
+  test:
+    name: test on Python ${{ matrix.python-version }} and pydantic ${{ matrix.pydantic-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        pydantic-version: ['main']
+        include:
+          - python-version: '3.12'
+            pydantic-version: '2.4'
+          - python-version: '3.12'
+            pydantic-version: '2.5'
+          - python-version: '3.12'
+            pydantic-version: '2.6'
+          - python-version: '3.12'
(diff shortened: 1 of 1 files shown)


.github/workflows/daily_deps_test.yml
```diff
@@ -0,0 +1,63 @@
+name: Daily CI Job
+
+on:
+  schedule:
+    - cron: '0 12 * * *' # Daily at midnight UTC
+  pull_request: {} # temporary: run on PRs so we can test this workflow
```
sydney-runkle: Should be removed before we merge :)

.github/workflows/daily_deps_test.yml
```diff
@@ -0,0 +1,63 @@
...
+          - python-version: '3.12'
+            pydantic-version: '2.5'
+          - python-version: '3.12'
+            pydantic-version: '2.6'
+          - python-version: '3.12'
+            pydantic-version: '2.7'
+          - python-version: '3.12'
+            pydantic-version: '2.8'
+          - python-version: '3.12'
+            pydantic-version: '2.9'
```
Kludex, line 31: Why do we do this only with Pydantic?

alexmojaki:
Seems like it's working because tests are already failing due to another OTEL release. The assertion diffs look related to https://github.com/open-telemetry/opentelemetry-python/pull/4251

.github/workflows/daily_deps_test.yml
alexmojaki: I think we will want a matrix for other deps too but it doesn't need to be a first step. We're still using the latest version of everything.

I also think that having this full matrix here means that regular CI only needs 2.4 and the latest release (pinned) and anything else needed for coverage (eventually v1 when some form of that works) so that it can run quickly and cheaply. Not a blocker though.

sydney-runkle: @alexmojaki,

Wouldn't it make sense to just test `2.4` and `main` on the daily CI, and leave all of these for the PR tests?

alexmojaki: No, the daily CI is a good place to put things that take longer that we don't want to wait for to merge a PR.

.github/workflows/daily_deps_test.yml
alexmojaki:
```suggestion
```

PR #587, opened 2024-11-13
Add span links

Kludex:
I need to add tests, but I'll wait for @alexmojaki 's review.

I've implemented both `logfire.span(_links=...)` and `span.add_link`. I was planning to only implement `add_link` because it's easier to use, since you don't need to create a `Link`, but the `add_link` docstring from the OTel package actually recommends to use `_links`... So I've implemented both.

We can already see the links on Logfire itself:

Screenshot 2024-11-13 at 12 21 04

## Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/logfire/_internal/main.py
+++ b/logfire/_internal/main.py
@@ -154,6 +154,7 @@ class Logfire:
         _tags: Sequence[str] | None = None,
         _span_name: str | None = None,
         _level: LevelName | int | None = None,
+        _links: Sequence[trace_api.Link] = (),
     ) -> LogfireSpan:
         try:
             stack_info = get_user_stack_info()
@@ -200,6 +201,7 @@ class Logfire:
                 otlp_attributes,
                 self._spans_tracer,
                 json_schema_properties,
+                links=_links,
             )
         except Exception:
             log_internal_error()
@@ -492,6 +494,7 @@ class Logfire:
         _tags: Sequence[str] | None = None,
         _span_name: str | None = None,
         _level: LevelName | None = None,
+        _links: Sequence[trace_api.Link] = (),
         **attributes: Any,
     ) -> LogfireSpan:
         """Context manager for creating a span.
@@ -510,6 +513,7 @@ class Logfire:
             _span_name: The span name. If not provided, the `msg_template` will be used.
             _tags: An optional sequence of tags to include in the span.
             _level: An optional log level name.
+            _links: An optional sequence of links to include in the span.
             attributes: The arguments to include in the span and format the message template with.
                 Attributes starting with an underscore are not allowed.
         """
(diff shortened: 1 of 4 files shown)


logfire/_internal/main.py
```diff
@@ -1863,6 +1871,10 @@ def set_attributes(self, attributes: dict[str, Any]) -> None:
         for key, value in attributes.items():
             self.set_attribute(key, value)
 
+    def add_link(self, context: SpanContext, attributes: otel_types.Attributes = None) -> None:
+        if self._span is not None:  # pragma: no branch
```
alexmojaki: the other branch should be handled, i.e. the link should be stored somewhere in case the span is started later

logfire/_internal/main.py
```diff
@@ -154,6 +154,7 @@ def _span(
         _tags: Sequence[str] | None = None,
         _span_name: str | None = None,
         _level: LevelName | int | None = None,
+        _links: Sequence[trace_api.Link] = (),
```
```suggestion
        _links: Sequence[trace_api.Link | trace_api.SpanContext | trace_api.Span | ReadableSpan | LogfireSpan] = (),
```

Making users import and construct link objects feels clunky. They should easily be able to pass any of these things, from which we can get a span context and construct a link, assuming they don't need attributes. We could maybe even allow a typed dict or a 2-tuple. None of this is important now, and it might clutter things up, just a thought.

alexmojaki:
> but the `add_link` docstring from the OTel package actually recommends to use `_links`

Not sure how much benefit there is to custom OTEL samplers having access to links... but I guess setting a link at the beginning also means that you have the link on the pending span in case the span never closes.

It bothers me that `add_link` does extra validation and stuff inside OTEL that doesn't happen with `start_span(links=[Link(...)])`.

alexmojaki:
Anyway I think this is fine in general and you can go forward with writing tests.

logfire/_internal/main.py
Kludex: I like an analogous API to the `add_link`: `_links: Sequence[tuple[trace_api.SpanContext, otel_types.Attributes]]`

logfire/_internal/main.py
```diff
@@ -154,6 +154,7 @@ def _span(
         _tags: Sequence[str] | None = None,
         _span_name: str | None = None,
         _level: LevelName | int | None = None,
+        _links: Sequence[tuple[SpanContext, otel_types.Attributes]] = (),
```
Kludex, line 157: I've changed the API to match `add_link`. I think this makes more sense.

Kludex:
I don't know where to document this... But I'll document this feature on a future PR.

alexmojaki (approved):
Thanks. Maybe needs a test including attributes. Documentation can happen after dogfooding.

Issue #521, opened 2024-10-19
dataclass fields with `field(repr=False)` should not be serialized/shown


samuelcolvin:
### Description

so those fields are not displayed in attributes.

PR #592, opened 2024-11-13
Respect repr on fields when logging a dataclass

dmontagu:
Requested by @samuelcolvin.

I think it maybe makes sense to have a way for users to control this behavior more explicitly/independently of other use cases (e.g., via a magic `__pydantic_logfire_dump__` method?). But this seems reasonable to me for now. Not sure if we need to be concerned about it being a breaking change, or need to provide a way to opt into it or something.

- Closes https://github.com/pydantic/logfire/issues/521

### DIFF
--- a/tests/test_json_args.py
+++ b/tests/test_json_args.py
@@ -5,7 +5,7 @@ import re
 import sys
 from collections import deque
 from collections.abc import Sequence
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import date, datetime, time, timedelta
 from decimal import Decimal
 from enum import Enum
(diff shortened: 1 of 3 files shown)


Kludex (approved):

tests/test_json_args.py
```diff
@@ -71,6 +71,12 @@ class MyPydanticComplexDataclass:
     t: MyPydanticDataclass
 
 
+@dataclass
+class MyReprDataclass:
+    in_repr: int
+    not_in_repr: int = field(repr=False)
```
alexmojaki: this should be a complex object that would appear in the json schema.

then ensure it doesn't appear in the json schema.

dmontagu: good catch, it does misbehave :(

dmontagu: Fixed, I think. And I check there was nowhere else we use the dataclasses.fields function, so I think it should be good now.

Issue #528, opened 2024-10-22
If I'm missing a package, I should be told what to install


samuelcolvin:
### Description

if I run

```py
logfire.instrument_httpx()
```

Without the httpx otel package installed, I get

```
...
  File "/Users/samuel/code/.../.venv/lib/python3.12/site-packages/logfire/_internal/integrations/httpx.py", line 5, in <module>
    from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
ModuleNotFoundError: No module named 'opentelemetry.instrumentation.httpx'
```

I should get an exception that tells me what to install.

Something like

> install the httpx instrumentation package with `logfire[httpx]` or directly with `opentelemetry-instrumentation-httpx`

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="1.2.0"
platform="macOS-14.6.1-arm64-arm-64bit"
python="3.12.6 (main, Sep 20 2024, 17:42:55) [Clang 15.0.0 (clang-1500.0.40.1)]"
[related_packages]
requests="2.32.3"
pydantic="2.9.2"
openai="1.51.2"
protobuf="4.25.5"
rich="13.9.2"
executing="2.1.0"
opentelemetry-api="1.27.0"
opentelemetry-exporter-otlp-proto-common="1.27.0"
opentelemetry-exporter-otlp-proto-http="1.27.0"
opentelemetry-instrumentation="0.48b0"
opentelemetry-proto="1.27.0"
opentelemetry-sdk="1.27.0"
opentelemetry-semantic-conventions="0.48b0"
```

PR #593, opened 2024-11-13
Let user know what they should install

Kludex:
- Closes https://github.com/pydantic/logfire/issues/528


alexmojaki (approved):
I'm trusting you here lol

PR #600, opened 2024-11-14
Make metric instrument methods compatible with older OTel versions

alexmojaki:


Kludex (approved):
Testing different versions of OTel would have saved us on this one.

PR #606, opened 2024-11-15
Daily CI - notify about failure via slack notification

sydney-runkle:
Closes https://github.com/pydantic/logfire/issues/580

* Removes extra `pydantic` version testing from daily CI
* Adds integration for sending slack message if any tests fail


alexmojaki (approved):
Thanks!

PR #607, opened 2024-11-18
Support `logfire.instrument` without arguments

### LINKED ISSUE #537: `@logfire.instrument` should work in bear mode
### Description

e.g.

```py
@logfire.instrument
def foobar(a, b, c):
    ...
```

E.g. not require `@logfire.instrument()`.

Kludex:
- Closes #537

### DIFF
--- a/logfire/_internal/main.py
+++ b/logfire/_internal/main.py
@@ -8,7 +8,19 @@ import traceback
 import warnings
 from functools import cached_property
 from time import time
-from typing import TYPE_CHECKING, Any, Callable, ContextManager, Iterable, Literal, Sequence, TypeVar, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ContextManager,
+    Iterable,
+    Literal,
+    Sequence,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
 
 import opentelemetry.context as context_api
 import opentelemetry.trace as trace_api
@@ -528,6 +540,7 @@ class Logfire:
             _links=_links,
         )
 
+    @overload
     def instrument(
         self,
         msg_template: LiteralString | None = None,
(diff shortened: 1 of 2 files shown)


logfire/_internal/main.py
```diff
@@ -528,14 +540,27 @@ def span(
...
         msg_template: LiteralString | None = None,
         *,
         span_name: str | None = None,
         extract_args: bool | Iterable[str] = True,
         allow_generator: bool = False,
-    ) -> Callable[[Callable[P, R]], Callable[P, R]]:
+    ) -> Callable[[Callable[P, R]], Callable[P, R]]: ...
+
+    @overload
+    def instrument(self, msg_template: Callable[P, R]) -> Callable[P, R]: ...
```
```suggestion
    def instrument(self, func: Callable[P, R]) -> Callable[P, R]: ...
```

logfire/_internal/main.py
```diff
@@ -528,14 +540,27 @@ def span(
...
+
+    def instrument(
+        self,
+        msg_template: Callable[P, R] | LiteralString | None = None,
```
alexmojaki, line 558:
```suggestion
        msg_template_or_func: Callable[P, R] | LiteralString | None = None,
```

logfire/_internal/main.py
```diff
@@ -557,6 +582,8 @@ def my_function(a: int):
             allow_generator: Set to `True` to prevent a warning when instrumenting a generator function.
                 Read https://logfire.pydantic.dev/docs/guides/advanced/generators/#using-logfireinstrument first.
         """
+        if callable(msg_template):
+            return instrument(self, tuple(self._tags), None, span_name, extract_args, allow_generator)(msg_template)
```
```suggestion
        if callable(msg_template_or_func):
            return self.instrument()(msg_template_or_func)
```

Kludex:
I didn't do what the suggestions suggest because it's a breaking change... People could be doing `msg_template=...`.

logfire/_internal/main.py
Kludex: This is not possible. The parameter names must match.

logfire/_internal/main.py
Kludex: This I can do.
```suggestion
        if callable(msg_template):
            return self.instrument()(msg_template)
```

alexmojaki:
OK leave the main signature as is but use the name `func` in the overload and silence pyright about it.

Kludex:
> OK leave the main signature as is but use the name `func` in the overload and silence pyright about it.

The problem is the typing issue is not on the overload method, but on the main function itself...

Screenshot 2024-11-18 at 10 56 28

We may make typing mistakes in the `instrument` without noticing if we do this.

alexmojaki:
What if you just ignore that specific error code?

Kludex:
> What if you just ignore that specific error code?

I'm not sure if this is working as expected. But I can do it.

alexmojaki (approved):

PR #610, opened 2024-11-19
Handle errors in auto-tracing better

alexmojaki:
1. When a file has invalid syntax, skip auto-tracing so that it can be imported normally and produce a normal traceback.
2. Handle internal errors when modifying and compiling the AST so that the app still works and just logs an error about auto-tracing failing.

For both of these, compiling had to be moved from the import loader to the import finder.


davidhewitt (approved):
Looks like good improvements to me 👍

PR #615, opened 2024-11-20
Release v2.4.0

alexmojaki:
Bumping version to v2.4.0.

### DIFF
--- a/logfire-api/logfire_api/_internal/main.pyi
+++ b/logfire-api/logfire_api/_internal/main.pyi
@@ -39,7 +39,7 @@ from opentelemetry.util import types as otel_types
 from starlette.applications import Starlette
 from starlette.requests import Request as Request
 from starlette.websockets import WebSocket as WebSocket
-from typing import Any, Callable, ContextManager, Iterable, Literal, Sequence, TypeVar
+from typing import Any, Callable, ContextManager, Iterable, Literal, Sequence, TypeVar, overload
 from typing_extensions import LiteralString, ParamSpec, Unpack
 from wsgiref.types import WSGIApplication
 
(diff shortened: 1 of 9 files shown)


logfire-api/logfire_api/_internal/main.pyi
```diff
@@ -223,28 +223,10 @@ class Logfire:
...
-            span_name: The span name. If not provided, the `msg_template` will be used.
-            extract_args: By default, all function call arguments are logged as span attributes.
-                Set to `False` to disable this, or pass an iterable of argument names to include.
-            allow_generator: Set to `True` to prevent a warning when instrumenting a generator function.
-                Read https://logfire.pydantic.dev/docs/guides/advanced/generators/#using-logfireinstrument first.
-        """
+    @overload
+    def instrument(self, msg_template: LiteralString | None = None, *, span_name: str | None = None, extract_args: bool | Iterable[str] = True, allow_generator: bool = False) -> Callable[[Callable[P, R]], Callable[P, R]]: ...
+    @overload
+    def instrument(self, func: Callable[P, R]) -> Callable[P, R]: ...
```
alexmojaki: @Kludex apparently stubgen doesn't work with docstrings and overloads?

Kludex: We need to add the flag `--include-docstrings`.

Kludex: Ah... Sorry. Got it.

Kludex: Sad.

Kludex: Maybe because of the wrong `func` name?

Kludex: What if the `func` becomes `msg_template`, does it work?

alexmojaki: no that doesn't help.

```python
def overload(func):
    """Decorator for overloaded functions/methods.

    In a stub file, place two or more stub definitions for the same
    function in a row, each decorated with @overload.

    For example::

        @overload
        def utf8(value: None) -> None: ...
        @overload
        def utf8(value: bytes) -> bytes: ...
        @overload
        def utf8(value: str) -> bytes: ...

    In a non-stub file (i.e. a regular .py file), do the same but
    follow it with an implementation.  The implementation should *not*
    be decorated with @overload::

        @overload
        def utf8(value: None) -> None: ...
        @overload
        def utf8(value: bytes) -> bytes: ...
        @overload
        def utf8(value: str) -> bytes: ...
        def utf8(value):
            ...  # implementation goes here
```

so stubgen is doing what it's supposed to do.

but ruff doesn't want me to put docstrings on overloads: https://docs.astral.sh/ruff/rules/overload-with-docstring/

so i don't know what the expectation here is. i guess we leave it?

Kludex: drop that ruff rule

Kludex (approved):

Issue #484, opened 2024-10-09
Improve search for alternative backends/clients


alexmojaki:
Ideas based on https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1728466950149999:

1. Searching for "opentelemetry" should return https://logfire.pydantic.dev/docs/guides/advanced/alternative-clients/ and https://logfire.pydantic.dev/docs/guides/advanced/alternative-backends/ as the first results
2. https://logfire.pydantic.dev/docs/guides/advanced/alternative-backends/ should mention the word 'server' and any other keywords likely to be searched
3. https://logfire.pydantic.dev/docs/why-logfire/opentelemetry/ should link to both those pages.
4. https://logfire.pydantic.dev/docs/why-logfire/opentelemetry/ should make the alternative clients point clearer - I think that's what "supports cross-language data integration" is trying to say but I'm not sure.

dmontagu:
Closing — three PRs at this came and went, and docs have since moved to the unified-docs pipeline, so the specific pages and search behaviour described here no longer apply.

PR #618, opened 2024-11-21
Add how to guide section

ChristopherGS:
Inspired by: https://fastapi.tiangolo.com/how-to/general/

https://linear.app/pydantic/issue/PYD-1363/create-index-or-general-how-to-guide-page

### DIFF
--- a/docs/guides/how-to/general.md
+++ b/docs/guides/how-to/general.md
@@ -11,7 +11,7 @@ analyze and investigate your system. For details of the data model, see
 our [schema docs](../web-ui/explore.md#records-schema)
 
 ### Disable Logfire During Tests
-Logfire makes it very easy to test the emitted logs and spans. See the [testing with logfire](../../advanced/testing.md) docs.
+Logfire makes it very easy to test the emitted logs and spans. See the [testing with logfire](../advanced/testing.md) docs.
 If you encounter difficulties disabling logfire during testing (and are using pytest),
 ensure that in your conftest.py you set:
 
@@ -22,11 +22,11 @@ logfire.configure(send_to_logfire=False)
 ```
 
 ### Setting an Environment (e.g. dev, prod)
-With Logfire you can distinguish which environment you are sending data to. See the [environments](../../advanced/environments.md) page.
+With Logfire you can distinguish which environment you are sending data to. See the [environments](../advanced/environments.md) page.
 
 ### Querying Logs via API
 Logfire provides a web API for programmatically running arbitrary SQL queries against the data in your Logfire projects.
-See the [Query API docs](../../advanced/query-api.md)
+See the [Query API docs](../advanced/query-api.md)
 
 
 ### Use logfire Without Sending Data to a Server
@@ -40,9 +40,9 @@ logfire.configure(send_to_logfire=False)
 ### Use an Alternative Backend
 
 Logfire uses the OpenTelemetry standard. This means that you can configure the SDK to export to any
-backend that supports OpenTelemetry. See our [alternative backends docs](../../advanced/alternative-backends.md)
+backend that supports OpenTelemetry. See our [alternative backends docs](../advanced/alternative-backends.md)
 
 
 ### Combine Logs from Multiple Processes/Worker Machines Under One Span
 The standard way to do this is with something called “context propagation”, which
-we have a [docs page](../..reference/api/propagate.md) for.
+we have a [docs page](../../reference/api/propagate.md) for.
--- a/docs/guides/how-to/general.md
+++ b/docs/guides/how-to/general.md
@@ -11,7 +11,7 @@ analyze and investigate your system. For details of the data model, see
 our [schema docs](../web-ui/explore.md#records-schema)
 
 ### Disable Logfire During Tests
-Logfire makes it very easy to test the emitted logs and spans. See the [testing with logfire](../../advanced/testing) docs.
+Logfire makes it very easy to test the emitted logs and spans. See the [testing with logfire](../../advanced/testing.md) docs.
 If you encounter difficulties disabling logfire during testing (and are using pytest),
 ensure that in your conftest.py you set:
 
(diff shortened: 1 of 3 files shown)


Kludex:
That's a thing I don't like about our docs and FastAPI ones: pages describing what you have in other pages.

It adds a lot of burden, since we forget about those pages.

ChristopherGS:
> That's a thing I don't like about our docs and FastAPI ones: pages describing what you have in other pages.
> 
> It adds a lot of burden, since we forget about those pages.

As discussed, this is based on analysis of a lot of customer queries in the public slack, where people have struggled to find things in the docs. I agree that it does add some maintenance burden, but I think the trade-off in usability will be worth it.

docs/guides/how-to/general.md
```diff
@@ -0,0 +1,48 @@
+Here are several pointers to other places in the docs, for general or frequent questions.
+
+
+### Set a Service Name
+To see your service names in the logfire UI, you just need to set the `service_name` parameter in
+[`logfire.configure()`][logfire.configure(service_name)].
```
alexmojaki, line 6: This is easy to search for, it's one of the first parameters of `logfire.configure`, and the name is self-explanatory. Is this section helping? Did people ask about this? I think what's missing is some motivation (e.g. a screenshot showing the column of bubbles/tags for different services) and an explanation of when you'd use it (different kinds of processes).

docs/guides/how-to/general.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Understanding the SQL Schema
+With Logfire, you can use the UI to run arbitrary SQL queries against your trace and metric data to
+analyze and investigate your system. For details of the data model, see
+our [schema docs](../web-ui/explore.md#records-schema)
+
+### Disable Logfire During Tests
```
alexmojaki, line 13: The linked testing page starts by saying that `send_to_logfire=False` by default under pytest.

I know this was asked about in slack at some point, but this behaviour was implemented since then.

docs/guides/how-to/general.md
```diff
@@ -0,0 +1,48 @@
...
+
+logfire.configure(send_to_logfire=False)
+```
+
+### Setting an Environment (e.g. dev, prod)
+With Logfire you can distinguish which environment you are sending data to. See the [environments](../advanced/environments.md) page.
+
+### Querying Logs via API
+Logfire provides a web API for programmatically running arbitrary SQL queries against the data in your Logfire projects.
+See the [Query API docs](../advanced/query-api.md)
```
alexmojaki, line 29: These and "Use an Alternative Backend" don't feel like they're adding additional discoverability on top of search. All these sections, especially 'Environments', are very close to this page in the left navigation bar.

docs/guides/how-to/general.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Use logfire Without Sending Data to a Server
+To disable sending data to a server and just emit traces to stdout:
+```py title="main.py"
+import logfire
+
+logfire.configure(send_to_logfire=False)
+```
+
+### Use an Alternative Backend
```
alexmojaki, line 40: I think the points in https://github.com/pydantic/logfire/issues/484 would help more

docs/guides/how-to/general.md
```diff
@@ -0,0 +1,48 @@
...
+Logfire provides a web API for programmatically running arbitrary SQL queries against the data in your Logfire projects.
+See the [Query API docs](../advanced/query-api.md)
+
+
+### Use logfire Without Sending Data to a Server
+To disable sending data to a server and just emit traces to stdout:
+```py title="main.py"
+import logfire
+
+logfire.configure(send_to_logfire=False)
```
alexmojaki, line 37: Similar comments to what I said about `service_name`. This feels redundant.

I'm guessing this is inspired by https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1722341938689489

The user was asking about hosting their own backend, the end result was https://logfire.pydantic.dev/docs/guides/advanced/alternative-backends/ (both the docs and a change in behaviour) and those docs mention `send_to_logfire=False`.

docs/guides/how-to/general.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+### Use an Alternative Backend
+
+Logfire uses the OpenTelemetry standard. This means that you can configure the SDK to export to any
+backend that supports OpenTelemetry. See our [alternative backends docs](../advanced/alternative-backends.md)
+
+
+### Combine Logs from Multiple Processes/Worker Machines Under One Span
```
alexmojaki, line 46: This one is actually a bit helpful, but I should emphasise that significantly more is needed: https://linear.app/pydantic/issue/PYD-1364/how-to-guide-web-app-deployed-on-gcp#comment-c5caa957

Kludex:
> Like @Kludex I feel weird about this whole page. This feels randomly inserted in relation to everything else. The existence of [logfire.pydantic.dev/docs/guides](https://logfire.pydantic.dev/docs/guides/) doesn't help.

Same for https://logfire.pydantic.dev/docs/guides/advanced/, or https://logfire.pydantic.dev/docs/guides/web-ui/...

alexmojaki:
I actually find it weird that any of those section headers in the nav bar actually link to pages.

Kludex:
> I actually find it weird that any of those section headers in the nav bar actually link to pages.

I'm happy to remove all of them.

Kludex:
I don't think it makes sense to have a section that duplicates information across the documentation.

docs/guides/how-to/general.md
ChristopherGS: I think this is classic "we're too used to our own product" issues. When I first joined this wasn't obvious to me initially - none of our examples show it (which is what most users look at)

alexmojaki: When I say the name is self-explanatory, I mean that it's obvious that the `service_name` parameter sets the service name. Nothing in this section tells the user more than that, i.e. what the service name means and how to set it. If when you first joined you were wondering "how do I get those bubbles I've been seeing", this wouldn't tell you that. A new user might not even know those bubbles are possible. I'd like to solve that. Maybe adding more content here along those lines would help, maybe not.

PR #619, opened 2024-11-21
Test pydantic 2.10 regularly

alexmojaki:
And only test pydantic `main` in daily CI so that regular CI is stable.


sydney-runkle (approved):
Nice, thanks

PR #624, opened 2024-11-26
Add Airflow integration documentation

Kludex:


docs/integrations/airflow.md
```diff
@@ -0,0 +1,128 @@
+# Airflow
+
+Airflow has a native OpenTelemetry integration for [traces] and [metrics].
+
+Airflow creates an exporter internally, which sends data to the configured backend.
+
+Unfortunately, it's not possible to send data to Logfire directly, yet... But we are working on it. :wink:
+
+??? question "Why it's not possible to send data directly?"
```
alexmojaki:
```suggestion
??? question "Why isn't it possible to send data directly?"
```

docs/integrations/airflow.md
```diff
@@ -0,0 +1,128 @@
...
+
+    We need the ability to set the authorization header to send data to **Logfire**.
+    See our [Alternative clients section](../guides/advanced/alternative-clients.md).
+
+## Setup
+
+To enable the OpenTelemetry integration in Airflow, you'll need to have an [OpenTelemetry Collector] running, and then
+configure Airflow to send data to it.
+
+### OpenTelemetry Collector
```
alexmojaki, line 55: This feels like it should be on the alternative clients page or something.

docs/integrations/airflow.md
```diff
@@ -0,0 +1,128 @@
...
+      http:
+        endpoint: "0.0.0.0:4318"
+
+exporters:  # (2)!
+  debug:  # (3)!
+  otlphttp:
+    endpoint: https://logfire-api.pydantic.dev/
+    compression: gzip
+    tls:
+      insecure: true
```
alexmojaki: ?

docs/integrations/airflow.md
alexmojaki: Especially because there are other reasons to use a collector

docs/integrations/airflow.md
```diff
@@ -0,0 +1,128 @@
...
+
+To enable the OpenTelemetry integration in Airflow, you'll need to have an [OpenTelemetry Collector] running, and then
+configure Airflow to send data to it.
+
+### OpenTelemetry Collector
+
+The OpenTelemetry Collector is a vendor-agnostic agent that can collect traces and metrics from your applications and
+send them to various backends.
+
+In this case, we are interested in sending data to **Logfire**. :fire:
```
alexmojaki: This feels like too many words to say you should send data from airflow to a collector which in turn sends it to logfire. And it starts with "Unfortunately, it's not possible to send data to Logfire directly" and takes too long to make it clear that you can send it indirectly.

docs/integrations/airflow.md
Kludex: Hmmm... Alternative backends maybe?

Maybe just a `OpenTelemetry Collector` section? 🤔 And then we explain what you can do with the collector (what should I add? sampling?) and I can also explain it can be used to transition from the current backend to Logfire...

docs/integrations/airflow.md
```diff
@@ -0,0 +1,128 @@
...
+      http:
+        endpoint: "0.0.0.0:4318"
+
+exporters:  # (2)!
+  debug:  # (3)!
+  otlphttp:
+    endpoint: https://logfire-api.pydantic.dev/
+    compression: gzip
+    tls:
+      insecure: true
```
Kludex:
```suggestion
```

docs/integrations/airflow.md
Kludex: from the copy paste 😓

docs/integrations/airflow.md
alexmojaki: Yes actually alternative backends makes more sense. Or a new page.

@adriangb can you propose anything to say about why a collector is a good idea?

I don't think we necessarily have to sell it. But it'd be nice for readers to have some idea why the page/section exists (without having to come from the airflow docs) and to feel like it's something they might want to look into if they have time but that it's not needed.

adriangb: I do think having an extensive section on the otel collector makes sense.

There's a lot of reasons you might want it including:
- centralizing configuration management including redaction, auth tokens for logfire or other upstream services, sending telemetry to multiple places, sampling to reduce costs, etc.
- scraping metrics from prometheus enabled services
- enriching with metadata like k8s host information

And more I don't want to go into in this comment. There's _a lot_ of talks, blog posts, etc. on the otel collector.

I'm happy to contribute to writing this section.

alexmojaki: > I'm happy to contribute to writing this section.

Thanks, that'd be great.

@Kludex for now can you please just make a new page/section with the minimum amount of info needed so that we can merge this PR?

Kludex: I'd prefer to have this section as is right now, because at the time, it's needed.

My PR was merged on Airflow, so the collector will not be needed on the next airflow release.

But also... We can leave the collector setup that's written here till Adrian works on the new section.

alexmojaki: OK, up to you.

docs/integrations/airflow.md
Kludex: I'll leave as is, and when @adriangb works on the OTel Collector section, we can remove this whole thing.

PR #642, opened 2024-12-04
Add NodeJS example to Alternative Clients

alexmojaki:


docs/guides/advanced/alternative-clients.md
```diff
@@ -49,6 +49,48 @@ exporter = OTLPSpanExporter(
 )
 ```
 
+## Example with NodeJS
+
+Create these files:
```
petyosi: I would rather use the following shell command:
```
npm init es6 -y # creates package.json with type module
npm install @opentelemetry/sdk-node
```

alexmojaki: Thanks

docs/guides/advanced/alternative-clients.md
```diff
@@ -49,6 +49,40 @@ exporter = OTLPSpanExporter(
...
+sdk.shutdown().catch(console.error);
+```
+
+Then run these commands:
+
+```sh
+export OTEL_EXPORTER_OTLP_ENDPOINT=https://logfire-api.pydantic.dev
+export OTEL_EXPORTER_OTLP_HEADERS='Authorization=your-write-token'
+
+npm init es6 -y # creates package.json with type module
```
petyosi, line 82: I would rather do the npm init/npm install before we create the main.js, otherwise opening the file above with the editor is likely to complain about missing packages.

alexmojaki: Then let the editor complain. I don't want to increase the number of code blocks and steps to follow.

PR #643, opened 2024-12-04
Remove pytest pin and update deps

alexmojaki:
Since https://github.com/15r10nk/inline-snapshot/issues/139 is fixed


alexmojaki:
Oops, forgot this was auto-merging. Oh well.

Updating pytest somehow revealed a gap in coverage in the failure case. I know that the pytest update improved executing's abilities in the context of pytest, but I don't know exactly what happened here. In any case I realized that the code was still saying that executing didn't work inside functions decorated with `@logfire.instrument` which is no longer true since it no longer uses AST modification. So I updated that.

PR #644, opened 2024-12-05
Re-seed random ID generator after `os.fork()`

alexmojaki:
`os.fork()` copied the `SeededRandomIdGenerator` state so that the parent and child processes would generate the same pseudorandom trace/span IDs, which is very bad. Demo script that this PR fixes:

```python
import os
import logfire

logfire.configure()

logfire.info('log', pids=(os.fork(), os.getpid()))
```

### DIFF
--- a/logfire/_internal/utils.py
+++ b/logfire/_internal/utils.py
@@ -374,6 +374,8 @@ class SeededRandomIdGenerator(IdGenerator):
 
     def __post_init__(self) -> None:
         self.random = random.Random(self.seed)
+        if self.seed is None:
+            os.register_at_fork(after_in_child=self.random.seed)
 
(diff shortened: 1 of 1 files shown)


phil65:
Just fyi, this broke logfire for me (using windows):

  File "E:\llmling\.venv\Lib\site-packages\logfire\_internal\utils.py", line 378, in __post_init__
    os.register_at_fork(after_in_child=self.random.seed)
    ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'os' has no attribute 'register_at_fork'

alexmojaki:
Thanks @phil65 for reporting, released another fix: https://github.com/pydantic/logfire/releases/tag/v2.6.2

PR #657, opened 2024-12-10
Add `logfire.instrument_aws_lambda`

Kludex:
Suggestions on how to test this?

### DIFF
--- a/docs/integrations/aws-lambda.md
+++ b/docs/integrations/aws-lambda.md
@@ -23,12 +23,10 @@ logfire.configure()  # (1)!
 def handler(event, context):
     return 'Hello from Lambda'
 
-logfire.instrument_aws_lambda()  # (2)!
+logfire.instrument_aws_lambda(handler)
 ```
 
 1. Remember to set the `LOGFIRE_TOKEN` environment variable on your Lambda function configuration.
-2. The `logfire.instrument_aws_lambda` function must be called after defining the handler function.
-    Otherwise, the handler function will not be instrumented.
 
 [`logfire.instrument_aws_lambda`][logfire.Logfire.instrument_aws_lambda] uses the **OpenTelemetry AWS Lambda Instrumentation** package,
 which you can find more information about [here][opentelemetry-aws-lambda].
--- a/docs/integrations/aws-lambda.md
+++ b/docs/integrations/aws-lambda.md
@@ -1,0 +1,36 @@
+# AWS Lambda
+
+The [`logfire.instrument_aws_lambda`][logfire.Logfire.instrument_aws_lambda] function can be used to
+instrument AWS Lambda functions to automatically send traces to **Logfire**.
+
+## Installation
+
+Install `logfire` with the `aws-lambda` extra:
+
+{{ install_logfire(extras=['aws-lambda']) }}
+
+## Usage
+
+To instrument an AWS Lambda function, call the `logfire.instrument_aws_lambda` function after defining
+the handler function:
+
+```python
+import logfire
+
+logfire.configure()  # (1)!
+
+
+def handler(event, context):
+    return 'Hello from Lambda'
+
+logfire.instrument_aws_lambda()  # (2)!
+```
+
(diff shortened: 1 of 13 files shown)


docs/integrations/aws-lambda.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+def handler(event, context):
+    return 'Hello from Lambda'
+
+logfire.instrument_aws_lambda()  # (2)!
+```
+
+1. Remember to set the `LOGFIRE_TOKEN` environment variable on your Lambda function configuration.
+2. The `logfire.instrument_aws_lambda` function must be called after defining the handler function.
```
alexmojaki: I think this should be enforced with code, e.g. require calling `logfire.instrument_aws_lambda(handler)` or check `globals()['handler']`.

alexmojaki (commented):
What have you learned from manual testing? Didn't it not work in the python runtime?

I think basic automated testing like asserting that something got patched is good enough.

Kludex:
> What have you learned from manual testing? Didn't it not work in the python runtime?

I made it work, I had to put the instrument method after the lambda function.

alexmojaki:
And does it do something more useful than `logfire.instrument`?

Kludex:
> And does it do something more useful than `logfire.instrument`?

It does.

Screenshot 2024-12-10 at 13 49 18

But... I wouldn't be able to explain how useful this information is.

docs/integrations/aws-lambda.md
Kludex: I tried to do it with `globals()`, but the function was not there, I guess I needed to go up the stack and get the locals up? It was easier to add the `lambda_handler`.

logfire/_internal/main.py
```diff
@@ -1451,12 +1452,31 @@ def instrument_sqlite3(
             },
         )
 
+    def instrument_aws_lambda(self, lambda_function: Any, **kwargs: Unpack[AwsLambdaInstrumentKwargs]) -> None:
```
alexmojaki: Can this be type-hinted?

Kludex: `Callable[[Any, Any], Any]`, or maybe `Callable[[dict[str, Any], dict[str, Any]], Any]`...

Kludex: I'm not 100% sure if the one on the right covers all cases...

alexmojaki: I think the basic one on the left is fine. Also maybe a note in docs about this param.

Kludex: What do you want to mention?

alexmojaki: I guess just mention what they're supposed to pass in

Kludex: Please suggest, and apply whatever you'd like to see.

The example contains the function as first parameter...

logfire/_internal/main.py
```diff
@@ -1451,12 +1452,33 @@ def instrument_sqlite3(
             },
         )
 
+    def instrument_aws_lambda(
+        self, lambda_function: LambdaFunction, **kwargs: Unpack[AwsLambdaInstrumentKwargs]
```
```suggestion
        self, lambda_function_handler: LambdaFunctionHandler, **kwargs: Unpack[AwsLambdaInstrumentKwargs]
```

Kludex: The examples on AWS use the `lambda_function` name for the function, I prefer to use the same name.

alexmojaki: > A function handler can be any name; however, the default name in the Lambda console is lambda_function.lambda_handler. This function handler name reflects the function name (lambda_handler) and the file where the handler code is stored (lambda_function.py).

https://docs.aws.amazon.com/lambda/latest/dg/python-handler.html

Kludex: ah yeah, `lambda_handler`, not `lambda_function`. mb

Kludex: Anything else in this PR besides this? I'll merge after fixing this otherwise.

PR #659, opened 2024-12-10
Update deps, fix failure caused by inline-snapshot

alexmojaki:
Fix https://github.com/pydantic/logfire/actions/runs/12255780781/job/34189633072

### DIFF
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -6,7 +6,7 @@ from typing import Any, cast
 import pytest
 import requests
 from dirty_equals import IsInt
-from inline_snapshot import snapshot
+from inline_snapshot import Is, snapshot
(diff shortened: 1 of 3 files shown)


tests/test_metrics.py
```diff
@@ -134,7 +134,7 @@ def test_create_metric_histogram(metrics_reader: InMemoryMetricReader) -> None:
                             'sum': 4300,
                             'scale': scale,
                             'zero_count': 0,
-                            'positive': {'offset': offset, 'bucket_counts': bucket_counts},
+                            'positive': {'offset': offset, 'bucket_counts': Is(bucket_counts)},
```
alexmojaki, line 137: To deal with https://github.com/15r10nk/inline-snapshot/issues/148

PR #662, opened 2024-12-10
Remove docs workaround and upgrade `mkdocs-material`

sydney-runkle:
Thanks @Viicos for the inspo here.


Kludex:
You don't need to bump all the packages for this, fyi

Kludex (approved):

Issue #655, opened 2024-12-09
Allow HTTPX instrumentation to capture more request info, and improve API docs


samuelcolvin:
### Description

We should add the following kwargs to [`instrument_httpx`](https://logfire.pydantic.dev/docs/reference/api/logfire/#logfire.Logfire.instrument_httpx):
* `capture_request_headers: bool | None = None`
* `capture_request_body: bool | None = None`
* `capture_response_headers: bool | None = None`
* `capture_response_body: bool | None = None`
* `capture_all: bool = False` - just sets all the above to `True`

We can raise an error if you use these together with `async_request_hook` etc.

We can add a note saying that capturing the body in production might increase the amount of data collected significantly.

Also [`HTTPXInstrumentKwargs`](https://github.com/pydantic/logfire/blob/743fb5d720ffa8b6b99d3f53a56d54bf10c5d8d5/logfire/_internal/integrations/httpx.py#L27-L32) is not included in docs which makes it basically undocumented.

We should probably do the same for requests.

alexmojaki:
@Kludex is currently working on headers in https://github.com/pydantic/logfire/pull/671 which is close to finishing.

After that comes capturing JSON bodies which @Kludex will also work on, I've provided him with a PoC implementation in slack.

Then https://github.com/pydantic/logfire/pull/668 can go through.

We discussed eventually capturing other types of bodies:

1. Forms, which could be done by walking up the stack to retrieve the `data` from a frame instead of parsing bytes
2. Arbitrary non-streamed bodies
3. Arbitrary streamed bodies

I briefly toyed with the last case, leaving the code here in case we want it:

```python
from dataclasses import dataclass
from typing import Iterable
import httpx
from httpx._content import IteratorByteStream
import logfire
from logfire.propagate import attach_context, get_context, ContextCarrier
from opentelemetry.trace import get_current_span, Span


def content():
    yield b"Hello, "
    yield b"world!"


logfire.configure()


@dataclass
class LoggingStream:
    stream: Iterable[bytes]
    ctx: ContextCarrier
    current_span: Span

    def __iter__(self):
        result = []
        for chunk in self.stream:
            result.append(chunk)
            yield chunk
        body = b"".join(result)
        if self.current_span.is_recording():
            # TODO this requires the bytes to be decodable as text
            self.current_span.set_attribute("http.body", body)
        else:
            with attach_context(self.ctx):
                logfire.info("Streamed body", body=body)


def request_hook(span, info):
    record_body(info)


[...]

PR #668, opened 2024-12-13
Document HTTPX kwargs

Kludex:


logfire/_internal/integrations/httpx.py
```diff
@@ -14,15 +20,9 @@
...
-    from typing import Awaitable, Callable, TypedDict, Unpack
+    from typing import TypedDict, Unpack
 
     import httpx
-    from opentelemetry.trace import Span
-
-    RequestHook = Callable[[Span, httpx.Request], None]
-    ResponseHook = Callable[[Span, httpx.Request, httpx.Response], None]
-    AsyncRequestHook = Callable[[Span, httpx.Request], Awaitable[None]]
-    AsyncResponseHook = Callable[[Span, httpx.Request, httpx.Response], Awaitable[None]]
```
Kludex, line 25: It's not `httpx.Request`. It's a `RequestInfo` which is a NamedTuple (???)...

alexmojaki:
From https://github.com/pydantic/logfire/issues/655#issue-2727407545:

> Also [HTTPXInstrumentKwargs](https://github.com/pydantic/logfire/blob/743fb5d720ffa8b6b99d3f53a56d54bf10c5d8d5/logfire/_internal/integrations/httpx.py#L27-L32) is not included in docs which makes it basically undocumented.

I think the desire was for `HTTPXInstrumentKwargs` to actually link somewhere like other type hints. I know that this would be tricky, but as a compromise I think the docstring in the `Logfire` class needs more info, at least a link to the main integration page.

docs/integrations/httpx.md
```diff
@@ -63,6 +63,152 @@ Let's see a minimal example below. You can run it with `python main.py`:
     asyncio.run(main())
     ```
 
+## Hooks
+
+The `logfire.instrument_httpx()` also accepts the following hooks:
```
alexmojaki:
```suggestion
The `logfire.instrument_httpx()` method also accepts the following hooks:
```

docs/integrations/httpx.md
```diff
@@ -63,6 +63,152 @@ Let's see a minimal example below. You can run it with `python main.py`:
...
+
+import logfire
+
+
+def set_request_headers(span: Span, request: RequestInfo):
+    method, url, headers, stream, extensions = request
+    attributes: defaultdict[str, str] = defaultdict(list)
+    for key, value in headers.items():
+        key = key.lower()
+        attributes[f'http.request.header.{key}'].append(value)
```
alexmojaki, line 97: I started working on capturing headers. Currently that code is stashed. Once pushed this will become a bad example.

Kludex:
> From [#655 (comment)](https://github.com/pydantic/logfire/issues/655#issue-2727407545):
> 
> > Also [HTTPXInstrumentKwargs](https://github.com/pydantic/logfire/blob/743fb5d720ffa8b6b99d3f53a56d54bf10c5d8d5/logfire/_internal/integrations/httpx.py#L27-L32) is not included in docs which makes it basically undocumented.
> 
> I think the desire was for `HTTPXInstrumentKwargs` to actually link somewhere like other type hints. I know that this would be tricky, but as a compromise I think the docstring in the `Logfire` class needs more info, at least a link to the main integration page.

So... You mean to create a `logfire.integrations.httpx` module? Not internal? I'm not sure I understand... 🤔

docs/integrations/httpx.md
```diff
@@ -63,6 +63,152 @@ Let's see a minimal example below. You can run it with `python main.py`:
...
+    span.set_attributes(attributes)
+
+
+def request_hook(span: Span, request: RequestInfo):
+    """This is called when using a `httpx.Client`."""
+    set_request_headers(span, request)
+
+
+async def async_request_hook(span: Span, request: RequestInfo):
+    """This is called when using a `httpx.AsyncClient`."""
```
alexmojaki, line 107: Two additional notes:

1. `async_request_hook` has to be a coroutine function, i.e. defined with `async def`, even if it only does sync things.
2. When instrumenting a client, the parameter is always called `request_hook`. To make it an async hook, it has to be defined as async. The type of client isn't taken into account.

I think we can automate dealing with this in the SDK. Maybe it should be an OTEL issue.

alexmojaki:
I think the original request was that `HTTPXInstrumentKwargs` itself should

> > From [#655 (comment)](https://github.com/pydantic/logfire/issues/655#issue-2727407545):
> > > Also [HTTPXInstrumentKwargs](https://github.com/pydantic/logfire/blob/743fb5d720ffa8b6b99d3f53a56d54bf10c5d8d5/logfire/_internal/integrations/httpx.py#L27-L32) is not included in docs which makes it basically undocumented.
> > 
> > 
> > I think the desire was for `HTTPXInstrumentKwargs` to actually link somewhere like other type hints. I know that this would be tricky, but as a compromise I think the docstring in the `Logfire` class needs more info, at least a link to the main integration page.
> 
> So... Should we create the `logfire.integrations.httpx` module? Not internal... 🤔

That's one possibility. But I don't know what you'd do about types like `RequestHook`. We could type it ourselves instead of importing from OTEL. They have https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/httpx/httpx.html#opentelemetry.instrumentation.httpx.RequestInfo but not `RequestHook` as a type in docs. Also their `RequestInfo` is actually wrong, specifically `headers` is actually `httpx.Headers`.

But we don't *need* to have formal docs of all the types. I'm just saying that the docstring needs *something*. And no, it doesn't need to block this PR.

docs/integrations/httpx.md
Kludex: Can you suggest a useful example? I couldn't think of any...

docs/integrations/httpx.md
```diff
@@ -63,6 +63,152 @@ Let's see a minimal example below. You can run it with `python main.py`:
...
+- `async_request_hook`: called before sending an async request.
+- `async_response_hook`: called after receiving an async response.
+
+You can use these hooks to add custom attributes that are not automatically captured by **Logfire**.
+
+### (Async) Request Hook
+
+The `request_hook` and `async_request_hook` hooks are called before sending a request. They receive two arguments:
+
+```py title="main.py" hl_lines="24-27"
```
alexmojaki, line 81:
```suggestion
```py title="main.py"
```

docs/integrations/httpx.md
```diff
@@ -63,6 +63,152 @@ Let's see a minimal example below. You can run it with `python main.py`:
...
+
+    Once a generator is consumed, it cannot be consumed again. This means that if you consume the stream
+    in the `request_hook` or `async_request_hook`, the request will fail.
+
+
+### (Async) Response Hook
+
+The `response_hook` and `async_response_hook` hooks are called after receiving a response. They receive three arguments:
+
+```py title="main.py" hl_lines="29-32"
```
alexmojaki, line 141:
```suggestion
```py title="main.py"
```

docs/integrations/httpx.md
```diff
@@ -63,6 +63,152 @@ Let's see a minimal example below. You can run it with `python main.py`:
...
+asyncio.run(main())
+```
+
+`RequestInfo` is a [`NamedTuple`][typing.NamedTuple] with the following fields:
+
+- `method`: The HTTP method of the request.
+- `url`: The URL of the request.
+- `headers`: The headers of the request.
+- `stream`: The stream of the request.
+- `extensions`: The extensions of the request.
```
alexmojaki:
```suggestion
`RequestInfo` is a [`NamedTuple`][typing.NamedTuple] with the following attributes: `method`, `url`, `headers`, `stream`, and `extensions`.
```

Maybe I'm just fussy, but if we don't have more useful descriptions of these fields then I don't like the redundancy.

docs/integrations/httpx.md
Kludex: Why?

docs/integrations/httpx.md
```diff
@@ -63,6 +63,146 @@ Let's see a minimal example below. You can run it with `python main.py`:
...
+
+asyncio.run(main())
+```
+
+`ResponseInfo` is a [`NamedTuple`][typing.NamedTuple] with the following fields:
+
+- `status_code`: The status code of the response.
+- `headers`: The headers of the response.
+- `stream`: The stream of the response.
+- `extensions`: The extensions of the response.
```
Kludex:
```suggestion
`ResponseInfo` is a [`NamedTuple`][typing.NamedTuple] with the following attributes: `status_code`, `headers`, `stream`, and `extensions`.
```

docs/integrations/httpx.md
alexmojaki: Why are those lines highlighted? It looks weird. I thought it wasn't intentional.

docs/integrations/httpx.md
alexmojaki: Capturing URL query params as proper attributes?

Kludex:
This PR is outdated. Let me craft something better..

PR #671, opened 2024-12-13
Add `capture_(request|response)_headers` to `instrument_httpx`

Kludex:


logfire/_internal/integrations/httpx.py
```diff
@@ -44,14 +97,105 @@ def instrument_httpx(
...
+
+
+def capture_request_headers(span: Span, request: RequestInfo) -> None:
+    capture_headers(span, cast('httpx.Headers', request.headers), 'request')
+
+
+def capture_headers(span: Span, headers: httpx.Headers, request_or_response: Literal['request', 'response']) -> None:
+    span.set_attributes(
+        {
+            f'http.{request_or_response}.header.{header_name}': headers.get_list(header_name)
```
Kludex, line 205: @alexmojaki note that we comply with the OTel specification here, but the Python OTel itself doesn't comply, and they actually replace `-` by `_`.

logfire/_internal/integrations/httpx.py
```diff
@@ -44,14 +97,105 @@ def instrument_httpx(
         'meter_provider': logfire_instance.config.get_meter_provider(),
         **kwargs,
     }
-    del kwargs  # make sure only final_kwargs is used
```
alexmojaki, line 47: Looks like this should be restored

logfire/_internal/integrations/httpx.py
```diff
@@ -44,14 +97,105 @@ def instrument_httpx(
...
+        while inspect.isawaitable(result):
+            result = await result
+
+
+def run_hook(hook: Callable[P, Any] | None, *args: P.args, **kwargs: P.kwargs) -> None:
+    if hook:
+        hook(*args, **kwargs)
+
+
+def capture_response_headers(span: Span, request: RequestInfo, response: ResponseInfo) -> None:
```
alexmojaki:
```suggestion
def capture_response_headers(span: Span, response: ResponseInfo) -> None:
```

logfire/_internal/integrations/httpx.py
```diff
@@ -44,14 +97,105 @@ def instrument_httpx(
...
-        )
-    else:
+
+    if client is None:
+        request_hook = cast('RequestHook | None', final_kwargs.get('request_hook'))
+        response_hook = cast('ResponseHook | None', final_kwargs.get('response_hook'))
+        async_request_hook = cast('AsyncRequestHook | None', final_kwargs.get('async_request_hook'))
+        async_response_hook = cast('AsyncResponseHook | None', final_kwargs.get('async_response_hook'))
+
+        if capture_request_headers:  # pragma: no cover
```
alexmojaki, line 111: non-blocking, but things here are getting complicated enough that this is worth testing. it could either make a real request or a mini local server could be set up in another thread.

alexmojaki (commented):
Need to handle and test passing a sync hook for an async client when headers aren't being captured

logfire/_internal/integrations/httpx.py
Kludex: But the response hook receives a request parameter... 🤔

logfire/_internal/integrations/httpx.py
Kludex: I didn't remove it tho, I moved up 🤔

logfire/_internal/integrations/httpx.py
alexmojaki: yes, but that's for the user hook, it's not being used here

logfire/_internal/integrations/httpx.py
alexmojaki: you moved it down

logfire/_internal/integrations/httpx.py
Kludex: Oh, right. Good catch.

logfire/_internal/integrations/httpx.py
Kludex: Ah, I think your first commit moved it down, and then I thought I moved up.

I've readded the line where it was before.

logfire/_internal/main.py
```diff
@@ -1158,8 +1158,39 @@ def instrument_asyncpg(self, **kwargs: Unpack[AsyncPGInstrumentKwargs]) -> None:
...
+        capture_request_headers: bool = False,
+        capture_response_headers: bool = False,
+        **kwargs: Unpack[AsyncClientKwargs],
+    ) -> None: ...
+
+    @overload
     def instrument_httpx(
-        self, client: httpx.Client | httpx.AsyncClient | None = None, **kwargs: Unpack[HTTPXInstrumentKwargs]
+        self,
+        client: None,
```
alexmojaki: Does this have to be here? If so, can it default to None?

logfire/_internal/main.py
```diff
@@ -1158,8 +1158,39 @@ def instrument_asyncpg(self, **kwargs: Unpack[AsyncPGInstrumentKwargs]) -> None:
...
+        capture_request_headers: bool = False,
+        capture_response_headers: bool = False,
+        **kwargs: Unpack[HTTPXInstrumentKwargs],
```
alexmojaki, line 1185: Still want to get rid of these Unpacks and kwargs eventually.

alexmojaki (approved):

logfire/_internal/main.py
Kludex: Oh yeah, we don't need it. 👍

Kludex: We need it. I've added the default `None`.

PR #675, opened 2024-12-13
Add APIs for getting the current logfire span

dmontagu:
This is a proposed implementation for how we can make it possible to get access to the current `LogfireSpan`, for making use of the various methods/etc. on it that aren't on the result of `opentelemetry.trace.get_current_span()` when you don't have access to the value.

The motivating use case for this is having access to methods of `LogfireSpan` inside functions instrumented with `@logfire.instrument`. Right now, there is no way to get access to the `LogfireSpan`, as `opentelemetry.trace.get_current_span` doesn't return that. This PR makes it so that `logfire.current_logfire_span()` returns the `LogfireSpan` if the current otel span corresponds to a `LogfireSpan`, and something API-compatible if it does not. (I also added `logfire.current_span()` as a more easily discovered re-export of `opentelemetry.trace.get_current_span`, don't necessarily need to keep that.)

----

Note that we've discussed making it so that when you call `opentelemetry.trace.get_current_span()` you get a `LogfireSpan` back, but the naive approach to this would have the consequence that spans created by other instrumentations would also end up being `LogfireSpan`s, with the associated modified method behaviors/overhead/etc. This may be a good idea, but it's not immediately clear to me.

The good news is that the implementation in this PR would be backwards compatible if we were to make that change, and most of the logic brought in in this PR would just get removed, just the `current_span` and `current_logfire_span` functions would remain.

### DIFF
--- a/logfire/__init__.py
+++ b/logfire/__init__.py
@@ -21,6 +21,7 @@ from ._internal.exporters.file import load_file as load_spans_from_file
 from ._internal.main import Logfire, LogfireSpan
 from ._internal.scrubbing import ScrubbingOptions, ScrubMatch
 from ._internal.utils import suppress_instrumentation
+from .current_span import current_logfire_span, current_span
 from .integrations.logging import LogfireLoggingHandler
 from .integrations.structlog import LogfireProcessor as StructlogProcessor
 from .version import VERSION
@@ -101,6 +102,8 @@ __all__ = (
     'ConsoleOptions',
     'CodeSource',
     'PydanticPlugin',
+    'current_span',
+    'current_logfire_span',
     'configure',
     'span',
     'instrument',
--- a/logfire/current_span.py
+++ b/logfire/current_span.py
@@ -1,0 +1,55 @@
+from __future__ import annotations as _annotations
+
+import warnings
+from typing import TYPE_CHECKING
+
+from opentelemetry.trace import get_current_span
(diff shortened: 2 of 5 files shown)


dmontagu:
I haven't added any tests yet, mostly because I don't want to put in the effort if the approach is going to be rejected.

adriangb (changes requested):
I think we should use the existing OTEL APIs

logfire/current_span.py
samuelcolvin, line 1: are we sure this should be public.

logfire/__init__.py
```diff
@@ -21,6 +21,7 @@
 from ._internal.main import Logfire, LogfireSpan
 from ._internal.scrubbing import ScrubbingOptions, ScrubMatch
 from ._internal.utils import suppress_instrumentation
+from .current_span import current_logfire_span, current_span
```
samuelcolvin, line 24: I think it could be confusing or cause weird side effects that the imported `current_span` overrides the `current_span` module.

samuelcolvin (commented):
I would suggest we add just `current_logfire_span`, but perhaps for consistency it should be called `get_current_logfire_span`?

Is there a noticeable memory or CPU overhead? I assume not, but we should perhaps do some basic checks.

Let's see what @alexmojaki says, but I'm strongly in favour of moving quickly, even if it's with an imperfect solution.

PR #679, opened 2024-12-16
Document Requests kwargs

Kludex:
Should we actually remove the kwargs and use the parameters? It's just a bad experience in every sense to have kwargs.


logfire/__init__.py
```diff
@@ -150,9 +151,11 @@ def loguru_handler() -> Any:
...
     'suppress_instrumentation',
+    'SamplingOptions',
+    'MetricsOptions',
+    # Integrations
     'StructlogProcessor',
     'LogfireLoggingHandler',
     'loguru_handler',
-    'SamplingOptions',
-    'MetricsOptions',
+    'CeleryInstrumentKwargs',
```
alexmojaki: I think we should in general avoid adding things to `__init__` if they only serve as type hints.

alexmojaki (changes requested):
Normal arguments are definitely better. Each integration can also have `**kwargs: Any` which is documented as something like 'other arguments passed to OTEL, intended for forward compatibility'.

Kludex:
@alexmojaki please confirm if this is okay for you. If it's, I'll add the others.

logfire/_internal/main.py
```diff
@@ -1174,17 +1174,26 @@ def instrument_httpx(
         self._warn_if_not_initialized_for_instrumentation()
         return instrument_httpx(self, client, **kwargs)
 
-    def instrument_celery(self, **kwargs: Unpack[CeleryInstrumentKwargs]) -> None:
+    def instrument_celery(self, **kwargs: Any) -> None:
```
alexmojaki, line 1227: What about `skip_dep_check`?

logfire/_internal/main.py
```diff
@@ -1238,18 +1247,34 @@ def instrument_django(
             **kwargs,
         )
 
-    def instrument_requests(self, excluded_urls: str | None = None, **kwargs: Any) -> None:
+    def instrument_requests(
+        self,
+        excluded_urls: str | None = None,
+        request_hook: Callable[[Span, requests.PreparedRequest], None] | None = None,
```
alexmojaki, line 1303: Nice of otel to pass the actual request object this time lol

alexmojaki (commented):
Yes, this approach looks good

logfire/_internal/main.py
Kludex: I removed on purpose. Is it really important? People shouldn't use it anyway... Unless specific cases like `psycopg-binary`...

Kludex:
let me know if this is okay, and I'll apply to the others.

logfire/_internal/main.py
alexmojaki: No it's not that important but I wasn't sure if it was intentional. Let's leave it out.

PR #681, opened 2024-12-16
Document sample usage in docs for Opentelemetry AIOHTTP URL filter

rishabhc32:
Add documentation demonstrating how to use AIOHTTP OpenTelemetry request hooks with a sample.

### DIFF
--- a/docs/integrations/aiohttp.md
+++ b/docs/integrations/aiohttp.md
@@ -40,7 +40,7 @@ if __name__ == "__main__":
 The keyword arguments of `logfire.instrument_aiohttp_client()` are passed to the `AioHttpClientInstrumentor().instrument()` method of the OpenTelemetry aiohttp client Instrumentation package, read more about it [here][opentelemetry-aiohttp].
 
 ## Masking URL in AIOHTTP span
-To mask a URL for sensitive data you create add a AIOHTTP OpenTelemetry request hook.
+To mask a URL for sensitive data you add an AIOHTTP OpenTelemetry URL filter.
 
 ```python
 import aiohttp
@@ -51,17 +51,11 @@ from yarl import URL
 def mask_url(url: URL) -> str:
     ...
 
-
-def otel_request_hook(span: Span, params: aiohttp.TraceRequestStartParams):
-    if span and span.is_recording():
-        masked_url = mask_url(params.url)
-        span.set_attribute("http.url", masked_url)
-
 ```
 
-Now configure logfire to use this request hook
+Now configure logfire to use this URL filter
 ```python
-logfire.instrument_aiohttp_client(request_hook=otel_request_hook)
+logfire.instrument_aiohttp_client(url_filter=mask_url)
 ```
 
 [aiohttp]: https://docs.aiohttp.org/en/stable/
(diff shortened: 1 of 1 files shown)


alexmojaki:
Unfortunately this doesn't work if you really want your data to be safe, at least if you're sending to Logfire. When the span starts it immediately creates a pending span, before the request hook runs. This span contains the original URL. If you run `select kind, attributes from records_all` in the explore view, you'll see something like this:

Screenshot 2024-12-16 at 20 31 16

rishabhc32:
@alexmojaki What is the recommended method to scrub for sensitive data in URL? Because some external vendor only support authentication via query parameters.

rishabhc32:
@alexmojaki This should work?
```python
def mask_url(url: URL) -> str:
    ...


def otel_request_hook(span: Span, params: aiohttp.TraceRequestStartParams):
    if span and span.is_recording():
        masked_url = mask_url(params.url)
        span.set_attribute("http.url", masked_url)

```

logfire.instrument_aiohttp_client(request_hook=otel_request_hook, url_filter=mask_url)
```

**Comment by alexmojaki** [comment]
That looks promising, didn't know about that. Is the request hook still needed in that case?

**Comment by rishabhc32** [comment]
@alexmojaki Yeah the request hook can be removed in this.

Have update the PR as well. Seems to work fine.

File: docs/integrations/aiohttp.md [hunk 1]
```diff
@@ -39,6 +39,24 @@ if __name__ == "__main__":
...
+def mask_url(url: URL) -> str:
+    sensitive_keys = {"username", "password", "token", "api_key", "api_secret", "apikey"}
+    masked_query = {key: "*****" if key in sensitive_keys else value for key, value in url.query.items()}
+    return str(url.with_query(masked_query))
+```
+
+Now configure logfire to use this URL filter
+```python
+logfire.instrument_aiohttp_client(url_filter=mask_url)
+```
```
alexmojaki:
```suggestion
## Hiding sensitive URL parameters

The `url_filter` keyword argument can be used to modify the URL that's recorded in spans. Here's an example of how to use this to redact query parameters:

```python
from yarl import URL

def mask_url(url: URL) -> str:
    sensitive_keys = {"username", "password", "token", "api_key", "api_secret", "apikey"}
    masked_query = {key: "*****" if key in sensitive_keys else value for key, value in url.query.items()}
    return str(url.with_query(masked_query))

logfire.instrument_aiohttp_client(url_filter=mask_url)
```

docs/integrations/aiohttp.md
```diff
@@ -39,6 +39,20 @@ if __name__ == "__main__":
...
+```python
+from yarl import URL
+
+def mask_url(url: URL) -> str:
+    sensitive_keys = {"username", "password", "token", "api_key", "api_secret", "apikey"}
+    masked_query = {key: "*****" if key in sensitive_keys else value for key, value in url.query.items()}
+    return str(url.with_query(masked_query))
+
+logfire.instrument_aiohttp_client(url_filter=mask_url)
+
```
rishabhc32, line 55: Hi @alexmojaki The code block was not closed here. Now it is not properly formatted in the docs.

alexmojaki: Thanks, opened https://github.com/pydantic/logfire/pull/683

PR #682, opened 2024-12-17
httpx: support `capture_request_json_body`

Kludex:
Target is the headers branch.


logfire/_internal/integrations/httpx.py
```diff
@@ -206,3 +242,35 @@ def capture_headers(span: Span, headers: httpx.Headers, request_or_response: Lit
...
+            return body.decode('utf-8')
+    return body.decode(charset, errors='replace')
+
+
+def capture_request_body(span: Span, request: RequestInfo) -> None:
+    content_type = cast('httpx.Headers', request.headers).get('content-type', '').lower()
+    if not content_type.startswith('application/json'):
+        return
+    if not isinstance(request.stream, httpx.ByteStream):
+        return  # pragma: no cover
```
alexmojaki: To test:

```python
import httpx


def content():
    yield b"Hello, "
    yield b"world!"


r = httpx.post("https://httpbin.org/post", content=content())
print(r.text)
```

logfire/_internal/integrations/httpx.py
```diff
@@ -206,3 +242,35 @@ def capture_headers(span: Span, headers: httpx.Headers, request_or_response: Lit
...
+    with suppress(UnicodeDecodeError, LookupError):
+        return body.decode(charset)
+    if charset.lower() not in ('utf-8', 'utf8'):
+        with suppress(UnicodeDecodeError):
+            return body.decode('utf-8')
+    return body.decode(charset, errors='replace')
+
+
+def capture_request_body(span: Span, request: RequestInfo) -> None:
+    content_type = cast('httpx.Headers', request.headers).get('content-type', '').lower()
```
alexmojaki, line 253: Another sign that we should provide the correct types ourselves.

logfire/_internal/integrations/httpx.py
```diff
@@ -206,3 +242,35 @@ def capture_headers(span: Span, headers: httpx.Headers, request_or_response: Lit
...
+    if not content_type.startswith('application/json'):
+        return
+    if not isinstance(request.stream, httpx.ByteStream):
+        return  # pragma: no cover
+
+    body = decode_body(list(request.stream)[0], content_type)
+
+    # Alex, does this needs to be `http.request.json_body`? Otherwise I'd prefer to follow the analogous notation to
+    # the headers, e.g. `http.request.header.content-type` -> `http.request.body.json`
+    attr_name = 'http.request.body.json'
```
alexmojaki, line 261: I think this is good. `http.request.body.size` exists, this fits with that. In the future we could use `.form`, `.streamed`, `.bytes`, whatever.

tests/otel_integrations/test_httpx.py
```diff
@@ -247,6 +248,117 @@ async def test_async_httpx_client_instrumentation_with_capture_headers(
...
+        assert span.context
+        trace_id = span.context.trace_id
+        async with httpx.AsyncClient(transport=create_transport()) as client:
+            logfire.instrument_httpx(client, capture_request_headers=True, capture_request_json_body=True)
+            try:
+                response = await client.post(
+                    'https://example.org/', headers={'Content-Type': 'application/json'}, json={'hello': 'world'}
+                )
+            finally:
+                HTTPXClientInstrumentor().uninstrument()
```
alexmojaki: We shouldn't need these lines since we're instrumenting single clients.

tests/otel_integrations/test_httpx.py
```diff
@@ -247,6 +248,117 @@ async def test_async_httpx_client_instrumentation_with_capture_headers(
...
+
+async def test_async_httpx_client_capture_full_request(exporter: TestExporter):
+    with logfire.span('test span') as span:
+        assert span.context
+        trace_id = span.context.trace_id
+        async with httpx.AsyncClient(transport=create_transport()) as client:
+            logfire.instrument_httpx(client, capture_request_headers=True, capture_request_json_body=True)
+            try:
+                response = await client.post(
+                    'https://example.org/', headers={'Content-Type': 'application/json'}, json={'hello': 'world'}
```
alexmojaki:
```suggestion
                    'https://example.org/', json={'hello': 'world'}
```

tests/otel_integrations/test_httpx.py
```diff
@@ -247,6 +248,117 @@ async def test_async_httpx_client_instrumentation_with_capture_headers(
...
+                )
+            finally:
+                HTTPXClientInstrumentor().uninstrument()
+            traceparent_header = response.headers['traceparent']
+            assert f'{trace_id:032x}' == traceparent_header.split('-')[1]
```
alexmojaki: Can this along with these lines:

```python
    with logfire.span('test span') as span:
        assert span.context
        trace_id = span.context.trace_id
```

be extracted into a common context manager?

tests/otel_integrations/test_httpx.py
alexmojaki: That's just because the content-type should be implied by the json kwarg. But we also need to test the content type actually missing. Add `content=b'{"hello": "world"}'` to one of the older tests without setting a content type to ensure that it doesn't capture anything.

logfire/_internal/integrations/httpx.py
Kludex: My PR on OTel got 1 approval already. It will get merged soon. 🙏

tests/otel_integrations/test_httpx.py
Kludex: I wanted to test with other content types, as the parameterize shows.

alexmojaki: yes, but the content type isn't missing in the parametrized tests.

logfire/_internal/integrations/httpx.py
Kludex: This doesn't removes the pragma tho

alexmojaki: ah, because of the content type. maybe this check should come first anyway.

tests/otel_integrations/test_httpx.py
Kludex: I read this wrongly, sorry. Fixed.

logfire/_internal/integrations/httpx.py
Kludex: moved

tests/otel_integrations/test_httpx.py
Kludex: I don't know. Can it? 👀

alexmojaki (approved):

PR #684, opened 2024-12-17
Simplify conditions in constructing httpx hooks

alexmojaki:
Also handle internal errors, including those from the user's hooks.


Kludex (approved):
Wanna merge on my branch already?

PR #689, opened 2024-12-17
docs: update FastStream example a bit

Lancetnik:
I'm sorry to interrupt. Just wanted to make the example a little more consistent with our usage recommendations.

Thank you for adding such usage to the documentation!


Kludex (approved):
Sure! Thanks.

But I think I didn't come up with the example, I think I stole it from your documentation 🤔

PR #695, opened 2024-12-18
Add documentation about LlamaIndex

Kludex:


docs/integrations/llms/llama-index.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Create index from documents
+index = VectorStoreIndex.from_documents(documents)
+
+# Initialize the LLM
+query_engine = index.as_query_engine(llm=Anthropic())
+
+# Get response
+response = query_engine.query('How can I use Pydantic models?')
+print(str(response))
```
dmontagu, line 69: should it show something below this? 😄 Not sure what the response is

dmontagu (approved):
Generally this seems good, but should we be instrumenting the things LlamaIndex does that aren't just making calls to the LLM providers?

Like, should we be instrumenting/discussing how to instrument the web page reading to build the index? Is there any instrumentation for llama_index that goes beyond the instrumentation of the LLM calls?

It would be nice to explain/show what sorts of things will show up in logfire if you enable the LLM instrumentation and use llama_index, like, what do you see in logfire as a result of that `query_engine.query` call?

But I mean I think this is a reasonable starting point, I don't see a reason to delay merging this, even if we can add more.

docs/integrations/llms/llama-index.md
```diff
@@ -0,0 +1,70 @@
+---
+integration: logfire
+---
+
+**Logfire** supports instrumenting calls to different LLMs with one extra line of code.
+Since LlamaIndex supports multiple LLMs, you can use **Logfire** to instrument calls to those LLMs.
```
dmontagu, line 6:
```suggestion
The way we recommend instrumenting LlamaIndex is to instrument the underlying LLM, and to rely on that instrumentation to ensure that calls made by LlamaIndex end up in Logfire.

**Logfire** supports instrumenting calls to different LLMs with one extra line of code.
Since LlamaIndex supports multiple LLMs, you can use **Logfire** to instrument calls to those LLMs.
```

I think we can be a bit more explicit here that we aren't specifically instrumenting LlamaIndex, we are just relying on its usage of something that is instrumented. That said, if there is an actual LlamaIndex instrumentation maybe we should add that. Not sure if there is or if the only out-of-the-box thing available is the LLM instrumentation.

alexmojaki: https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-llamaindex

Kludex: I've added that one:

Screenshot 2024-12-23 at 11 42 04

Kludex:
I'm merging https://github.com/pydantic/logfire/pull/771 instead. FYI

PR #696, opened 2024-12-18
Use `OPEN_SPANS` and custom `record_exception` logic for all spans

dmontagu:
I'm working toward splitting `LogfireSpan` into a dedicated context manager, and a subclass of `ReadableSpan`.

This PR makes some of the changes we want to make along the way. In particular:

* I ensure _all_ spans created by the logfire `_ProxyTracer` are tracked in `OPEN_SPANS`, and therefore should get properly closed on exit (even if created in generators, etc.), not just LogfireSpans
* I ensure we use our custom logic for `record_exception` for all spans created by the logfire `_ProxyTracer`
	* To do this, I needed to move `_record_exception` and `_set_exception_status` to `tracer` to prevent circular imports. They are now used in multiple modules, so I removed the leading `_`, but they are only used in `_internal` modules so I think that's okay.
	* As an implementation detail, I modified `_MaybeDeterministicTimestampSpan` to do the `OPEN_SPANS` manipulation and override `record_exception`, since we were always returning spans wrapped with that class from `_ProxyTracer.start_span`. Given its expanded usage, I also renamed that class to `_LogfireWrappedSpan`. Given there are now behaviors we want to always override, I think we are paying less marginal overhead for setting the timestamp deterministically, so I think it's fine to rename the class and not even aspire to only use it when modifying the timestamp generation logic.
* I renamed `user_attributes` to `prepare_otlp_attributes` — this seems like a better name to me, and I found myself wanting to use `user_attributes` during my attempts to do more of the downstream refactors we've discussed, so I just figured it made sense to change that here.
* I made the signature of `LogfireSpan.end()` compatible with `trace_api.Span` (required no other changes), since I think we may eventually want to rework things so `LogfireSpan` is returned by `_ProxyTracer.start_span`. Even if we don't, I think it's a fine change.

I believe this PR has no user-facing API changes.


logfire-api/logfire_api/_internal/main.pyi
```diff
@@ -1034,7 +1034,7 @@ class LogfireSpan(ReadableSpan):
     def message(self) -> str: ...
     @message.setter
     def message(self, message: str): ...
-    def end(self) -> None:
+    def end(self, end_time: int | None = None) -> None:
```
adriangb, line 1037: I came to check that this is a backwards compatible change and it is 👍🏻

logfire-api/logfire_api/_internal/tracer.pyi
```diff
@@ -13,7 +14,9 @@ from opentelemetry.trace.status import Status, StatusCode
 from opentelemetry.util import types as otel_types
 from threading import Lock
 from typing import Any, Callable, Mapping, Sequence
-from weakref import WeakKeyDictionary
+from weakref import WeakKeyDictionary, WeakSet
+
+OPEN_SPANS: WeakSet[_LogfireWrappedSpan]
```
adriangb, line 19: No value?

```suggestion
OPEN_SPANS: WeakSet[_LogfireWrappedSpan] = WeakSet()
```

?

logfire-api/logfire_api/_internal/tracer.pyi
```diff
@@ -33,10 +36,11 @@ class ProxyTracerProvider(TracerProvider):
     def force_flush(self, timeout_millis: int = 30000) -> bool: ...
 
 @dataclass
-class _MaybeDeterministicTimestampSpan(trace_api.Span, ReadableSpan):
+class _LogfireWrappedSpan(trace_api.Span, ReadableSpan):
```
adriangb: Rename makes sense 👍🏻 

Should we make it a `__slots__` data class (we should do it for all of them), using dynamic kwargs for python < 3.10, etc.? Can we another PR

logfire-api/logfire_api/_internal/tracer.pyi
adriangb: Oh sorry realizing now this is a `.pyi` file...

logfire/_internal/tracer.py
```diff
@@ -101,13 +115,17 @@ def force_flush(self, timeout_millis: int = 30000) -> bool:
 
 
 @dataclass
```
adriangb: slots comment applies here

adriangb (approved):

logfire/_internal/tracer.py
```diff
@@ -151,7 +169,9 @@ def record_exception(
         escaped: bool = False,
     ) -> None:
         timestamp = timestamp or self.ns_timestamp_generator()
-        return self.span.record_exception(exception, attributes, timestamp, escaped)
+        record_exception(self.span, exception, attributes=attributes, timestamp=timestamp, escaped=escaped)
+
+    __hash__ = object.__hash__  # necessary to be able to add this to OPEN_SPANS
```
alexmojaki: i think rather use `@dataclass(eq=False)`.

logfire/_internal/tracer.py
```diff
@@ -101,13 +115,17 @@ def force_flush(self, timeout_millis: int = 30000) -> bool:
...
     """Span that overrides end() to use a timestamp generator if one was provided."""
 
     span: Span
     ns_timestamp_generator: Callable[[], int]
 
+    def __post_init__(self):
+        OPEN_SPANS.add(self)
+
     def end(self, end_time: int | None = None) -> None:
+        OPEN_SPANS.remove(self)
```
alexmojaki:
```suggestion
        OPEN_SPANS.discard(self)
```

logfire/_internal/main.py
```diff
@@ -2041,7 +2025,7 @@ def end(self) -> None:
                         ATTRIBUTES_JSON_SCHEMA_KEY, attributes_json_schema(self._json_schema_properties)
                     )
 
-                self._span.end()
+            self._span.end(end_time)
```
alexmojaki: why is this dedented?

logfire/_internal/config.py
alexmojaki, line 1: `span.__exit__(None, None, None)` in this file should probably be changed to `span.end()`

logfire-api/logfire_api/_internal/tracer.pyi
dmontagu: I'd rather wait until we finish the refactoring to make it a `__slots__` dataclass, just in case. But I think that makes sense once things are stable, maybe it would be fine to do it now I'd just rather not be thinking about it _yet_. Can change it now if you insist though.

logfire/_internal/main.py
dmontagu: not sure if it was intentional, if it was, I would guess it was because it's using the wrapped spans `.end()`, which isn't error-suppressed by otel and is for a span we didn't create, so this would preserve the error behavior. But anyway I can just re-indent it

logfire/_internal/config.py
dmontagu: kind of disturbing it didn't caught by type checking..

dmontagu: oh I see, the inherited `__exit__` just calls `end`

PR #700, opened 2024-12-19
Capture httpx response JSON bodies

alexmojaki:


tests/otel_integrations/test_httpx.py
```diff
@@ -348,3 +348,213 @@ def test_missing_opentelemetry_dependency() -> None:
...
+                    'http.flavor': '1.1',
+                    'network.protocol.version': '1.1',
+                    'http.response.header.host': ('example.org',),
+                    'http.response.header.accept': ('*/*',),
+                    'http.response.header.accept-encoding': ('gzip, deflate',),
+                    'http.response.header.connection': ('keep-alive',),
+                    'http.response.header.user-agent': (IsStr(),),
+                    'http.response.header.content-length': (IsStr(),),
+                    'http.response.header.content-type': ('application/json',),
+                    'http.response.header.traceparent': ('00-00000000000000000000000000000001-0000000000000003-01',),
```
Kludex, line 406: Isn't it a bit weird that you have both the information about the request headers, body and response headers in a single span, but the response body on another one? 🤔

alexmojaki: By the time the body is read, the main span is already closed, which is why a new span is needed.

The best we can do for now is to copy the attributes from the main span into the new span so that the new span has everything. But we can't put the body in the main span.

Kludex (approved):

logfire/_internal/integrations/httpx.py
```diff
@@ -176,34 +191,96 @@ async def new_hook(span: Span, request: RequestInfo) -> None:
...
+            if should_capture_json:
+                capture_response_json(logfire_instance, response, True)
             await run_async_hook(hook, span, request, response)
 
     return new_hook
 
 
+def capture_response_json(logfire_instance: Logfire, response_info: ResponseInfo, is_async: bool) -> None:
+    headers = cast('httpx.Headers', response_info.headers)
+    if not headers.get('content-type', '').lower().startswith('application/json'):
```
Kludex, line 233: I think the `lower` is redundant.

Kludex (approved):

PR #701, opened 2024-12-19
Support for AnthropicBedrock client

stephenhibbert:
Added support for passing AnthropicBedrock and AsyncAnthropicBedrock clients to `instrument_anthropic()`

### DIFF
--- a/docs/integrations/llms/anthropic.md
+++ b/docs/integrations/llms/anthropic.md
@@ -104,7 +104,8 @@ Shows up like this in Logfire:
   <figcaption>Anthropic streaming response</figcaption>
 </figure>
 
-# Amazon Bedrock
+## Amazon Bedrock
+
 You can also log Anthropic LLM calls to Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
 
 ```python
--- a/docs/integrations/llms/anthropic.md
+++ b/docs/integrations/llms/anthropic.md
@@ -103,3 +103,17 @@ Shows up like this in Logfire:
   ![Logfire Anthropic Streaming](../../images/logfire-screenshot-anthropic-stream.png){ width="500" }
   <figcaption>Anthropic streaming response</figcaption>
 </figure>
+
+# Amazon Bedrock
+You can also run inference on AWS with Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
+
+```
+import anthropic
+import logfire
+
+client = anthropic.AnthropicBedrock()
+
+logfire.configure()
+logfire.instrument_anthropic_bedrock(client)
+```
+Or if you don't have access to the client instance you can use `logfire.instrument_anthropic_bedrock()`) to instrument both the `anthropic.Anthropic` and `anthropic.AsyncAnthropic` classes.
--- a/docs/integrations/llms/anthropic.md
+++ b/docs/integrations/llms/anthropic.md
@@ -107,6 +107,13 @@ Shows up like this in Logfire:
 # Amazon Bedrock
 You can also log Anthropic LLM calls to Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
 
+To use this integration, you'll need the AWS SDK dependencies:
+- `boto3`
+- `botocore`
+
+You can install them along with other required packages using:
+```pip install -U logfire[bedrock]```
+
 ```
 import anthropic
 import logfire
--- a/docs/integrations/llms/anthropic.md
+++ b/docs/integrations/llms/anthropic.md
@@ -105,7 +105,7 @@ Shows up like this in Logfire:
 </figure>
 
 # Amazon Bedrock
-You can also run inference on AWS with Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
+You can also log Anthropic LLM calls to Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
 
 ```
 import anthropic
@@ -114,6 +114,5 @@ import logfire
 client = anthropic.AnthropicBedrock()
 
 logfire.configure()
-logfire.instrument_anthropic_bedrock(client)
+logfire.instrument_anthropic(client)
 ```
-Or if you don't have access to the client instance you can use `logfire.instrument_anthropic_bedrock()`) to instrument both the `anthropic.Anthropic` and `anthropic.AsyncAnthropic` classes.
--- a/docs/integrations/llms/anthropic.md
+++ b/docs/integrations/llms/anthropic.md
@@ -107,18 +107,15 @@ Shows up like this in Logfire:
 # Amazon Bedrock
 You can also log Anthropic LLM calls to Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
 
-To use this integration, you'll need the AWS SDK dependencies:
-- `boto3`
-- `botocore`
-
-You can install them along with other required packages using:
-```pip install -U logfire[bedrock]```
-
 ```
 import anthropic
 import logfire
 
-client = anthropic.AnthropicBedrock()
+client = anthropic.AnthropicBedrock(
+    aws_region='us-east-1',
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -53,6 +53,7 @@ dependencies = [
     "typing-extensions >= 4.1.0",
     "tomli >= 2.0.1; python_version < '3.11'",
     "executing >= 2.0.1",
+    "boto3>=1.35.84",
```
alexmojaki: This shouldn't be a required dependency.

stephenhibbert: Good point, I've now addressed this, and made it optional with the `bedrock` extra and included in the `dev` dependencies for tests

stephenhibbert:
image

@alexmojaki  - do you think we should add any other metadata about this LLM being on Bedrock to help users distinguish between this and the Anthropic API?

docs/integrations/llms/anthropic.md
```diff
@@ -103,3 +103,23 @@ Shows up like this in Logfire:
   ![Logfire Anthropic Streaming](../../images/logfire-screenshot-anthropic-stream.png){ width="500" }
   <figcaption>Anthropic streaming response</figcaption>
 </figure>
+
+# Amazon Bedrock
+You can also log Anthropic LLM calls to Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
+
+To use this integration, you'll need the AWS SDK dependencies:
+- `boto3`
+- `botocore`
```
alexmojaki: Let's stick to only recommending the `bedrock` extra so that we can add other dependencies to that if needed.

docs/integrations/llms/anthropic.md
```diff
@@ -103,3 +103,23 @@ Shows up like this in Logfire:
...
+
+You can install them along with other required packages using:
+```pip install -U logfire[bedrock]```
+
+```
```
alexmojaki:
```suggestion
```python
```

logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -83,9 +83,14 @@ def on_response(response: ResponseT, span: LogfireSpan) -> ResponseT:
...
 
-def is_async_client(client: type[anthropic.Anthropic] | type[anthropic.AsyncAnthropic]):
-    """Returns whether or not the `client` class is async."""
-    if issubclass(client, anthropic.Anthropic):
-        return False
-    assert issubclass(client, anthropic.AsyncAnthropic), f'Expected Anthropic or AsyncAnthropic type, got: {client}'
-    return True
+def is_async_client(client_type: Any) -> bool:
+    """Returns whether the `client` class is async."""
+    if isinstance(client_type, type):
```
alexmojaki: `is_async_client` is already only called with a class as the argument.

logfire/_internal/main.py
```diff
@@ -1094,7 +1100,13 @@ def instrument_anthropic(
         import logfire
         import anthropic
 
+        # For regular Anthropic client
         client = anthropic.Anthropic()
+        # Or for Bedrock client
+        client = anthropic.AnthropicBedrock(
+            aws_region='us-east-1', aws_access_key='access-key', aws_secret_key='secret-key'
```
```suggestion
            aws_region='us-east-1',
            aws_access_key='access-key',
            aws_secret_key='secret-key',
```

docs/integrations/llms/anthropic.md
```diff
@@ -103,3 +103,23 @@ Shows up like this in Logfire:
...
+
+# Amazon Bedrock
+You can also log Anthropic LLM calls to Amazon Bedrock using the `AmazonBedrock` and `AsyncAmazonBedrock` clients.
+
+To use this integration, you'll need the AWS SDK dependencies:
+- `boto3`
+- `botocore`
+
+You can install them along with other required packages using:
+```pip install -U logfire[bedrock]```
```
alexmojaki: this should probably be using the special extra installation block that other pages use.

docs/integrations/llms/anthropic.md
alexmojaki: actually, do we need an extra at all? boto3 and botocore are required to use anthropic bedrock at all, not just for the logfire part.

uv.lock
alexmojaki, line 1: It looks like you need to run `uv sync` again, currently `make install` installs the stubs.

logfire/_internal/main.py
alexmojaki: Actually I don't think we need the bedrock example at all.

alexmojaki (changes requested):

docs/integrations/llms/anthropic.md
stephenhibbert: I was taking inspiration from the `anthropic-sdk-python` docs. I agree I think we can remove the extra and reference the anthropic extra directly and then we're not tracking the versions twice for no reason and risking they get out of sync.

`pip install -U anthropic[bedrock]`

https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#aws-bedrock
https://github.com/anthropics/anthropic-sdk-python/blob/93cbbbde964e244f02bf1bd2b579c5fabce4e267/pyproject.toml#L39

alexmojaki: I don't think we even need to mention `anthropic[bedrock]`, it's not like we tell users to install `anthropic`. We assume that they already know how to use `AnthropicBedrock` and just need to instrument it.

docs/integrations/llms/anthropic.md
```diff
@@ -103,3 +103,20 @@ Shows up like this in Logfire:
   ![Logfire Anthropic Streaming](../../images/logfire-screenshot-anthropic-stream.png){ width="500" }
   <figcaption>Anthropic streaming response</figcaption>
 </figure>
+
+# Amazon Bedrock
```
alexmojaki:
```suggestion
## Amazon Bedrock

```

pyproject.toml
```diff
@@ -269,8 +272,8 @@ filterwarnings = [
     'ignore:(?s).*aiosqlite/core\.py.*get_loop\(future\).call_soon_threadsafe.*RuntimeError. Event loop is closed:pytest.PytestUnhandledThreadExceptionWarning',
     # This problem is fixed in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2013, but not the warning.
     "ignore:The '__version__' attribute is deprecated and will be removed in Flask 3.1:DeprecationWarning",
-    "ignore:The `capture_request_headers` parameter is deprecated. Use `capture_headers` instead.:DeprecationWarning",
-    "ignore:The `capture_response_headers` parameter is deprecated. Use `capture_headers` instead.:DeprecationWarning",
+    # ignore DeprecationWarning from httpx - Marcelo knows why this is happening
+    "ignore:The 'app' shortcut is now deprecated.*:DeprecationWarning:httpx*:",
```
alexmojaki: please revert

docs/integrations/llms/anthropic.md
stephenhibbert: Removed as discussed

logfire/_internal/integrations/llm_providers/anthropic.py
stephenhibbert: Removed in latest

pyproject.toml
```diff
@@ -269,8 +272,6 @@ filterwarnings = [
     'ignore:(?s).*aiosqlite/core\.py.*get_loop\(future\).call_soon_threadsafe.*RuntimeError. Event loop is closed:pytest.PytestUnhandledThreadExceptionWarning',
     # This problem is fixed in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2013, but not the warning.
     "ignore:The '__version__' attribute is deprecated and will be removed in Flask 3.1:DeprecationWarning",
-    "ignore:The `capture_request_headers` parameter is deprecated. Use `capture_headers` instead.:DeprecationWarning",
-    "ignore:The `capture_response_headers` parameter is deprecated. Use `capture_headers` instead.:DeprecationWarning",
```
alexmojaki: please keep

alexmojaki (approved):
Thanks!

PR #712, opened 2024-12-20
Remove double record exception

dmontagu:
We are currently double-recording exceptions in LogfireSpan's due to some convoluted logic that this PR simplifies


logfire/_internal/main.py
```diff
@@ -1948,13 +1948,13 @@ def __init__(self, span: trace_api.Span) -> None:
         self._token = context_api.attach(trace_api.set_span_in_context(self._span))
 
     def __enter__(self) -> FastLogfireSpan:
+        self._span.__enter__()
```
dmontagu: This is maybe unnecessary but seems reasonable to add, and I hope this class will be refactored/simplified soon anyway.

logfire/_internal/main.py
```diff
@@ -1990,6 +1990,7 @@ def __enter__(self) -> LogfireSpan:
                     attributes=self._otlp_attributes,
                     links=self._links,
                 )
+            self._span.__enter__()
```
dmontagu, line 2148: Same note as above, this might not be necessary but seems fine to add and arguably more correct(?)

logfire/_internal/main.py
```diff
@@ -1999,14 +2000,17 @@ def __enter__(self) -> LogfireSpan:
...
-        assert self._span is not None
-        _exit_span(self._span, exc_value)
-
-        self.end()
+        if self._span.is_recording():
+            with handle_internal_errors():
+                if self._added_attributes:
+                    self._span.set_attribute(
+                        ATTRIBUTES_JSON_SCHEMA_KEY, attributes_json_schema(self._json_schema_properties)
+                    )
```
dmontagu, line 2167: I just moved the logic from `end()` here, noting we can drop a check since we already know `self._span is not None`, and we don't currently rely on the `LogfireSpan` methods being called at the same time that the otel methods would be called.

logfire/_internal/tracer.py
```diff
@@ -171,6 +171,12 @@ def record_exception(
         timestamp = timestamp or self.ns_timestamp_generator()
         record_exception(self.span, exception, attributes=attributes, timestamp=timestamp, escaped=escaped)
 
+    def __exit__(self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: Any) -> None:
+        if self.is_recording() and isinstance(exc_value, BaseException):
+            self.record_exception(exc_value, escaped=True)
```
dmontagu: This is basically just the logic that used to be in `main._exit_span`. The parent class's version of this just calls `self.end()`, and I kept that below.

tests/test_logfire.py
```diff
@@ -3171,3 +3173,23 @@ def test_suppress_scopes(exporter: TestExporter, metrics_reader: InMemoryMetricR
...
+        return record_exception(*args, **kwargs)
+
+    with patch('logfire._internal.tracer.record_exception', patched_record_exception), patch(
+        'logfire._internal.main.record_exception', patched_record_exception
+    ):
+        with pytest.raises(RuntimeError):
+            with logfire.span('foo'):
+                raise RuntimeError('error')
+
+    assert n_calls_to_record_exception == 1
```
dmontagu: I confirmed this test fails on `main` and makes two calls instead of one.

logfire/_internal/main.py
```diff
@@ -2032,26 +2036,6 @@ def message(self) -> str:
...
-        if self._span is None:  # pragma: no cover
-            raise RuntimeError('Span has not been started')
-        if self._span.is_recording():
-            with handle_internal_errors():
-                if self._added_attributes:
-                    self._span.set_attribute(
-                        ATTRIBUTES_JSON_SCHEMA_KEY, attributes_json_schema(self._json_schema_properties)
-                    )
-
-                self._span.end(end_time)
```
dmontagu, line 2209: We used to only call `self._span.end` here if `self._span` was recording. Is that actually the right behavior though? The changes I made ensure that `end()` always gets called, even on a non-recording span, and that's why the timestamps changed below (because it generates another timestamp in the unconditional call to `end()`).

I can go back to only calling `span.end()` if the span is recording but I'm not sure if that's correct..

logfire/_internal/tracer.py
```diff
@@ -171,6 +171,12 @@ def record_exception(
...
+
+        self.end()
```
dmontagu:
```suggestion
        if self.is_recording():
            if isinstance(exc_value, BaseException):
                self.record_exception(exc_value, escaped=True)
            self.end()
```
I confirmed that this would get us back to the old timestamp behavior, again, not sure if this is better or worse than the version currently in this PR. But I think one of these two options should be okay

dmontagu: I decided to make this change because, 1, it preserves the old behavior, and 2, I think it marginally reduces overhead. We can always just dedent the call to `self.end()` if we think the method should be called even if the inner span is not a recording span.

logfire/_internal/main.py
alexmojaki: Given how optimised this class is, I'd rather not add it here, but maybe it can stay in LogfireSpan.

logfire/_internal/main.py
alexmojaki:
```python
import opentelemetry.trace

import logfire

logfire.configure()

tracer = opentelemetry.trace.get_tracer(__name__)
span = tracer.start_span('foo')
assert span.is_recording()
span.end()
assert not span.is_recording()
span.end()  # Logs a warning: Calling end() on an ended span.
```

So keeping the current behaviour seems best.

tests/test_logfire.py
alexmojaki: It looks like the doubling was introduced by https://github.com/pydantic/logfire/pull/696

alexmojaki (approved):

logfire/_internal/main.py
dmontagu: I'll note that the problem in your example isn't that end was called on a non-recording span, but that it was called on an ended span; I was more concerned with calling the `__exit__` for NonRecordingSpans than I was with double-calls to `end`.

But this isn't very important to me and I can just go back to the way it was. If it ever causes issues we can obviously change the behavior.

alexmojaki: Well `NonRecordingSpan.end` currently does nothing, and `Span` currently logs the warning on double-end if and only if `is_recording` returns False.

dmontagu: yeah my point was — if someone subclasses NonRecordingSpan and has that produced by the tracer that our proxy tracer wraps, and it implements `end`, then you could hit this. That's sort of what I mean by "correct" behavior — like respecting the APIs for the sake of sophisticated user/third-party otel code that logfire would layer on top of. Obviously that's a rare scenario though; I changed it back for now.

dmontagu:
@alexmojaki I see you approved this, but just to confirm, am I good to merge this now? I can make additional changes if there's anything left you prefer

logfire/_internal/tracer.py
```diff
@@ -171,6 +177,12 @@ def record_exception(
         timestamp = timestamp or self.ns_timestamp_generator()
         record_exception(self.span, exception, attributes=attributes, timestamp=timestamp, escaped=escaped)
 
+    def __exit__(self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: Any) -> None:
+        if self.is_recording():  # pragma: no branch
```
alexmojaki: Please test this case by putting `span.end()` before the end of a `with` block, and remove the pragma.

dmontagu: I can't find a good way to hit this because the only place `logfire._internal.tracer._LogfireWrappedSpan.__exit__` is called is inside `logfire._internal.main.LogfireSpan.__exit__`, and it only calls that if the wrapped span is not recording. Any suggestions of what to change or how to achieve this? (E.g., I could unconditionally call the span exit in the `LogfireSpan.__exit__` function, but not sure if that's preferable.)

The closest I've gotten is:
```python
def test_exit_ended_span():
    # Ensure that doing this does not emit a warning
    logfire.configure(send_to_logfire=False, console=False)

    span = logfire.span('foo').__enter__()
    with span._span:
        get_current_span().end()
```
But I question whether this is a test worth adding, given it requires use of the internal API. But I can add it if you prefer, or maybe you have some ideas of how to cover this without using internal APIs that I'm just failing to see.

alexmojaki:
```python
from opentelemetry import trace

import logfire

logfire.configure()
tracer = trace.get_tracer(__name__)

with tracer.start_span('test') as span:
    span.end()
```

dmontagu: 🙏

tests/test_logfire.py
```diff
@@ -3192,3 +3192,13 @@ def patched_record_exception(*args: Any, **kwargs: Any) -> Any:
...
+
+
+def test_exit_ended_span():
+    # Ensure that doing this does not emit a warning
+    logfire.configure(send_to_logfire=False, console=False)
+
+    tracer = get_tracer(__name__)
+
+    with tracer.start_span('test') as span:
+        span.end()
```
alexmojaki:
```suggestion
def test_exit_ended_span(exporter: TestExporter):
    tracer = get_tracer(__name__)

    with tracer.start_span('test') as span:
        # Ensure that doing this does not emit a warning about calling end() twice when the block exits.
        span.end()
    
    assert ...
```

PR #719, opened 2024-12-23
httpx: support `capture_headers`

Kludex:


logfire/_internal/integrations/httpx.py
```diff
@@ -1,25 +1,26 @@
...
+import warnings
 from contextlib import suppress
 from email.message import Message
-from typing import TYPE_CHECKING, Any, Callable, Literal, Mapping, cast, overload
+from typing import TYPE_CHECKING, Any, Callable, Literal, Mapping, cast
 
 import httpx
 import opentelemetry.sdk.trace
 
+from logfire.integrations.httpx import (
```
alexmojaki: Move this under TYPE_CHECKING so that the `except ImportError:` will still work if `logfire.integrations.httpx` starts importing from otel

logfire/_internal/integrations/httpx.py
```diff
@@ -111,6 +75,21 @@ def instrument_httpx(
 
     See the `Logfire.instrument_httpx` method for details.
     """
+    capture_request_headers = kwargs.get('capture_request_headers')
+    capture_response_headers = kwargs.get('capture_response_headers')
+
+    if capture_request_headers is not None:
+        warnings.warn(
```
alexmojaki:
```suggestion
        warn_at_user_stacklevel(
```

logfire/_internal/main.py
```diff
@@ -1212,8 +1208,7 @@ def instrument_httpx(
         Args:
             client: The `httpx.Client` or `httpx.AsyncClient` instance to instrument.
                 If `None`, the default, all clients will be instrumented.
-            capture_request_headers: Set to `True` to capture all request headers.
-            capture_response_headers: Set to `True` to capture all response headers.
+            capture_headers: Set to `True` to capture all HTTP headers.
```
alexmojaki, line 1211: Let's link to the integrations/guide page from the main body of this docstring so that (among other things) people know how to capture specific headers.

alexmojaki (approved):

logfire/_internal/integrations/httpx.py
Kludex: It still works like this, cause `httpx` is supposed to be installed here, and `opentelemetry.trace` as well.

alexmojaki: i'm referring to this:

```
# TODO(Marcelo): When https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3098/ gets merged,
# and the next version of `opentelemetry-instrumentation-httpx` is released, we can just do a reimport:
# from opentelemetry.instrumentation.httpx import RequestInfo as RequestInfo
# from opentelemetry.instrumentation.httpx import ResponseInfo as ResponseInfo
# from opentelemetry.instrumentation.httpx import RequestHook as RequestHook
# from opentelemetry.instrumentation.httpx import ResponseHook as ResponseHook
```

Kludex: got it

Kludex:
ah, you approved.

I've applied all comments

PR #720, opened 2024-12-23
Refactor httpx content type handling

alexmojaki:
Some initial work towards capturing text bodies. Currently nothing here should affect users (except that it can handle variations of `application/json` such as `application/ld+json` which is a good fix) but I see overlap with https://github.com/pydantic/logfire/pull/719 so I want to merge this.

### DIFF
--- a/logfire/_internal/integrations/httpx.py
+++ b/logfire/_internal/integrations/httpx.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import inspect
 from contextlib import suppress
-from email.message import Message
+from email.headerregistry import ContentTypeHeader
+from email.policy import EmailPolicy
+from functools import lru_cache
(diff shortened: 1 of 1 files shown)


logfire/_internal/integrations/httpx.py
```diff
@@ -390,3 +387,59 @@ def capture_request_form_data(span: Span, request: RequestInfo) -> None:
...
+
+@lru_cache
+def is_text_type(content_type: str) -> bool:
+    header = content_type_header_from_string(content_type)
+    if header.maintype == 'text':
+        return True
+    if header.maintype != 'application':
+        return False
+
+    return bool(content_type_subtypes(header.subtype) & TEXT_SUBTYPES)
```
alexmojaki, line 421: This isn't used yet.

PR #724, opened 2024-12-24
docs: add "Detect Service is Down" section

### LINKED ISSUE #677: Not available alert
### Description

Base on user of solution(or baser on each specific time), it's possible to detect when solution go down and us alert

Kludex:
@alexmojaki Feel free to modify the PR as you wish.

- Closes #677


docs/how-to-guides/detect-service-is-down.md
```diff
@@ -0,0 +1,58 @@
+For now, **Logfire** doesn't have a built-in way to detect if a service is down. In the sense that we don't ping
```
alexmojaki:
```suggestion
For now, **Logfire** doesn't have a built-in way to detect if a service is down, in the sense that we don't ping
```

alexmojaki (approved):
If we add many more guides then I think we'll have to move this somewhere else, maybe as a section in the alerts page. But for now I guess there's not too many guides.

PR #726, opened 2024-12-24
Improve Span Links API

### LINKED ISSUE #158: Support for Span link with good DX interface and demo presentation
### Description

I couldn't find anywhere in docs where it mentioned about span links. That might be something logfire is missing at the moment.
https://opentelemetry.io/docs/concepts/signals/traces/#span-links

It'd be nice to also have this on example, linking a span from frontend, to the backend. I'm interested how that would look in the dashboard as well.
Currently otel sdk interface for span link is some kind of horrible. but I'm questioning if that can be designed better or not.

Kludex:
Currently, it's not very ergonomic to create a span link.

Let's assume we have 2 services:

<details>
<summary>Main Service</summary>

```py
from typing import Any

from opentelemetry import trace
from opentelemetry.propagate import extract
from pydantic import TypeAdapter
from redis import Redis

import logfire
import logfire.propagate

logfire.configure(service_name='subscriber')
logfire.instrument_redis()

TA = TypeAdapter[Any](Any)


def main():
    client = Redis()
    logfire.info('Subscriber is running')

    def watch_key(key: str):
        pubsub = client.pubsub()
        pubsub.subscribe(key)

        for message in pubsub.listen():
            if message['type'] == 'message':
                with logfire.span('process_message'):
                    data = TA.validate_json(message['data'])
                    context = extract(data.pop('headers', {}))
                    span = next(iter(context.values()))
                    with logfire.span('process_data', _links=[(span.get_span_context(), None)]):
                        logfire.info(f'Received data: {data}')

    watch_key('key')


main()
```

</details>

And...

<details>
<summary>Service A</summary>

```py
from typing import Any

from pydantic import TypeAdapter
from redis import Redis

import logfire
import logfire.propagate

logfire.configure(service_name='service_a')
logfire.instrument_redis()

TA = TypeAdapter[Any](Any)

client = Redis()


@logfire.instrument()
def service_a():
    logfire.info('Service A is running')

    data: dict[str, Any] = {'data': 'Hello from Service A', 'headers': logfire.propagate.get_context()}
    client.publish('key', TA.dump_json(data))


service_a()
```

</details>

As you see, in the code, we set the headers with `logfire.propagate.get_context()`, and then on the subscriber, we need to run:

```py
context = extract(data.pop('headers', {}))
span = next(iter(context.values()))
with logfire.span('process_data', _links=[(span.get_span_context(), None)]):
    ...
```

We can for sure improve this. Either by passing `SpanContext` to `_links`, or a `Link`.

This PR proposes creating:
1. `logfire.propagate.build_span_link`: Creates a span link from a `ContextCarrier`, so it makes easier to use the propagate API we created. But... We can also build a `span_context`, instead of `Link` itself. It may be a bit more useful. I'm open to rework this.
2. Adds `Link` and `SpanContext` to the `_links` parameter.

---

I need to add tests here. I've first implemented this to bring back the discussion.

- Closes #158

### DIFF
--- a/logfire/_internal/main.py
+++ b/logfire/_internal/main.py
@@ -25,7 +25,7 @@ import opentelemetry.context as context_api
 import opentelemetry.trace as trace_api
 from opentelemetry.metrics import CallbackT, Counter, Histogram, UpDownCounter
 from opentelemetry.sdk.trace import ReadableSpan, Span
-from opentelemetry.trace import SpanContext, Tracer
+from opentelemetry.trace import Link, SpanContext, Tracer
 from opentelemetry.util import types as otel_types
 from typing_extensions import LiteralString, ParamSpec
 
@@ -165,7 +165,7 @@ class Logfire:
         _tags: Sequence[str] | None = None,
         _span_name: str | None = None,
         _level: LevelName | int | None = None,
-        _links: Sequence[tuple[SpanContext, otel_types.Attributes]] = (),
+        _links: Sequence[Link | SpanContext | tuple[SpanContext, otel_types.Attributes]] = (),
     ) -> LogfireSpan:
         try:
             stack_info = get_user_stack_info()
@@ -505,7 +505,7 @@ class Logfire:
         _tags: Sequence[str] | None = None,
         _span_name: str | None = None,
         _level: LevelName | None = None,
-        _links: Sequence[tuple[SpanContext, otel_types.Attributes]] = (),
+        _links: Sequence[Link | SpanContext | tuple[SpanContext, otel_types.Attributes]] = (),
         **attributes: Any,
     ) -> LogfireSpan:
         """Context manager for creating a span.
@@ -524,7 +524,7 @@ class Logfire:
             _span_name: The span name. If not provided, the `msg_template` will be used.
             _tags: An optional sequence of tags to include in the span.
             _level: An optional log level name.
-            _links: An optional sequence of links to other spans. Each link is a tuple of a span context and attributes.
+            _links: An optional sequence of links to other spans.
             attributes: The arguments to include in the span and format the message template with.
                 Attributes starting with an underscore are not allowed.
         """
@@ -1991,13 +1991,22 @@ class LogfireSpan(ReadableSpan):
         otlp_attributes: dict[str, otel_types.AttributeValue],
         tracer: Tracer,
         json_schema_properties: JsonSchemaProperties,
-        links: Sequence[tuple[SpanContext, otel_types.Attributes]],
+        links: Sequence[Link | SpanContext | tuple[SpanContext, otel_types.Attributes]],
     ) -> None:
         self._span_name = span_name
         self._otlp_attributes = otlp_attributes
         self._tracer = tracer
         self._json_schema_properties = json_schema_properties
-        self._links = list(trace_api.Link(context=context, attributes=attributes) for context, attributes in links)
+
+        self._links: list[Link] = []
+        for link in links:
+            if isinstance(link, Link):
+                self._links.append(link)
+            elif isinstance(link, SpanContext):
+                self._links.append(trace_api.Link(context=link))
+            else:
+                context, attributes = link
+                self._links.append(trace_api.Link(context=context, attributes=attributes))
 
         self._added_attributes = False
         self._token: None | object = None
@@ -2099,7 +2108,7 @@ class LogfireSpan(ReadableSpan):
 
     def add_link(self, context: SpanContext, attributes: otel_types.Attributes = None) -> None:
         if self._span is None:
-            self._links += [trace_api.Link(context=context, attributes=attributes)]
+            self._links += [trace_api.Link(context=context)]
         else:
             self._span.add_link(context, attributes)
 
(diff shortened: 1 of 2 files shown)


logfire/_internal/main.py
```diff
@@ -2099,7 +2108,7 @@ def set_attributes(self, attributes: dict[str, Any]) -> None:
 
     def add_link(self, context: SpanContext, attributes: otel_types.Attributes = None) -> None:
         if self._span is None:
-            self._links += [trace_api.Link(context=context, attributes=attributes)]
+            self._links += [trace_api.Link(context=context)]
```
alexmojaki: ?

logfire/propagate.py
```diff
@@ -70,3 +71,26 @@ def attach_context(carrier: ContextCarrier) -> Iterator[None]:
...
+
+    ```py
+    from logfire.propagate import build_span_link, get_context
+
+    logfire_context = get_context()
+
+    ...
+
+    # later on in another thread, process or service
+    link = build_span_link(logfire_context)
```
alexmojaki, line 89: Why wouldn't the user use `with attach_context` here?

logfire/_internal/main.py
Kludex: My bad.
```suggestion
            self._links += [trace_api.Link(context=context, attributes=attributes)]
```

logfire/propagate.py
Kludex: If it's attached, then it doesn't make sense to add a span link, does it?

alexmojaki: Right, I'm questioning adding (and recommending) `build_span_link`. In this situation it seems more sensible to use a parent-child relationship instead of a span link.

Kludex: I mean, if I change the comment to "later on a task queue", would that be better? Or what you suggest instead?

By the way, I'm not sure `build_span_link` is the best API for this, if we `build_span_context` also works. I don't care which if the `span(..., _links=[...])` accepts one of them.

alexmojaki: > I mean, if I change the comment to "later on a task queue", would that be better?

No because regular distributed tracing still seems better.

> Or what you suggest instead?

I honestly don't know when span links are useful in general. I suppose in your bigger example there's already a parent `process_message`. If we assume that for whatever reason we really want to keep that as the parent, then we have to use a link. But then the docs should clarify that so that users only use links when it makes sense.

> By the way, I'm not sure `build_span_link` is the best API for this, if we `build_span_context` also works. I don't care which if the `span(..., _links=[...])` accepts one of them.

I like that idea. Maybe name it `get_span_context`.

logfire/propagate.py
```diff
@@ -70,3 +71,26 @@ def attach_context(carrier: ContextCarrier) -> Iterator[None]:
...
+    with logfire.span('process_data', _links=[link]):
+        ...
+    ```
+    """
+    context = propagate.extract(carrier=carrier)
+    span = cast(Span, next(iter(context.values())))
```
alexmojaki, line 95: The context can contain multiple values, e.g. baggage.

logfire/propagate.py
Kludex: > No because regular distributed tracing still seems better.

You disagree with https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3002 ?

> I honestly don't know when span links are useful in general.

Besides Airflow, I'm not sure who uses them... But I feel they are **supposed** to be useful when you "trigger" something e.g. you have an endpoint, and you trigger a veryyyy long task somewhere else - you want to be able to close the span from the endpoint, and have a link to this very long task.

> I like that idea. Maybe name it get_span_context.

👍

alexmojaki: > > No because regular distributed tracing still seems better.
> 
> You disagree with [open-telemetry/opentelemetry-python-contrib#3002](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3002) ?

Overall, yes. What made you post that?

> > I honestly don't know when span links are useful in general.
> 
> Besides Airflow, I'm not sure who uses them... But I feel they are **supposed** to be useful when you "trigger" something e.g. you have an endpoint, and you trigger a veryyyy long task somewhere else - you want to be able to close the span from the endpoint,

The endpoint span will still be closed. It'll just have children that starts after it ends, which feels weird but should be harmless.

> and have a link to this very long task.

The endpoint span won't have a link to the task span, only the other way around. Someone looking at the endpoint span has no easy way to find the task span.

logfire/propagate.py
```diff
@@ -70,3 +71,26 @@ def attach_context(carrier: ContextCarrier) -> Iterator[None]:
...
+    ```py
+    from logfire.propagate import build_span_link, get_context
+
+    logfire_context = get_context()
+
+    ...
+
+    # later on in another thread, process or service
+    link = build_span_link(logfire_context)
+    with logfire.span('process_data', _links=[link]):
```
alexmojaki, line 90:
```suggestion
    with logfire.span('outer span'):
        # can't use attach_context because we want to keep outer span as the parent
        with logfire.span('process_data', _links=[link]):
```

Kludex:
We didn't achieve consensus on span links. There's too much subjectivity involved here.

One conclusion we got is that: we can have the `get_span_context` method.

Issue #728, opened 2024-12-27
Logfire causes sqlalchemy.exc.NoInspectionAvailable to raise when testing simple FastAPI+SQLModel app


igorantonow314:
### Summary

I copied* [this](https://fastapi.tiangolo.com/tutorial/sql-databases/#create-with-herocreate-and-return-a-heropublic) example from FastAPI's docs, added testing using `TestClient` and logfire with `logfire.instrument_pydantic()`, and encountered an error. The app without the line `logfire.instrument_pydantic()` works fine. 
(*I removed irrelevant routes).

### Details

Here is the `index.py` code:
```python
from typing import Annotated

import logfire
from fastapi import Depends, FastAPI, HTTPException, Query
from fastapi.testclient import TestClient
from sqlmodel import Field, Session, SQLModel, create_engine, select



logfire.configure()
logfire.instrument_pydantic()


class HeroBase(SQLModel):
    name: str = Field(index=True)
    age: int | None = Field(default=None, index=True)


class Hero(HeroBase, table=True):
    id: int | None = Field(default=None, primary_key=True)
    secret_name: str


class HeroPublic(HeroBase):
    id: int


class HeroCreate(HeroBase):
    secret_name: str


class HeroUpdate(HeroBase):
    name: str | None = None
    age: int | None = None
    secret_name: str | None = None


sqlite_file_name = "database.db"
sqlite_url = f"sqlite:///{sqlite_file_name}"

connect_args = {"check_same_thread": False}
engine = create_engine(sqlite_url, connect_args=connect_args)


def create_db_and_tables():
    SQLModel.metadata.create_all(engine)


def get_session():
    with Session(engine) as session:
        yield session


SessionDep = Annotated[Session, Depends(get_session)]
app = FastAPI()


@app.on_event("startup")
def on_startup():
    create_db_and_tables()


@app.post("/heroes/", response_model=HeroPublic)
def create_hero(hero: HeroCreate, session: SessionDep):
    db_hero = Hero.model_validate(hero)
    session.add(db_hero)
    session.commit()
    session.refresh(db_hero)
    return db_hero




def test_create():
    with TestClient(app) as client:
        response = client.post('/heroes', json={"name": "some name", "age":123, "secret_name":"younewerguess"})
        assert response.status_code == 200
```

I ran `pytest index.py`  and got the error:
```
sqlalchemy.exc.NoInspectionAvailable: No inspection system is available for object of type <class 'index.Hero'>
```

part of trace:
```
[...]

PR #733, opened 2024-12-27
Handle errors from `sqlalchemy.inspect`

alexmojaki:
Closes https://github.com/pydantic/logfire/issues/728

### DIFF
--- a/tests/test_pydantic_plugin.py
+++ b/tests/test_pydantic_plugin.py
@@ -7,6 +7,7 @@ from unittest.mock import patch
 
 import cloudpickle
 import pytest
+import sqlmodel
 from dirty_equals import IsInt
 from inline_snapshot import snapshot
 from opentelemetry.sdk.metrics.export import AggregationTemporality, InMemoryMetricReader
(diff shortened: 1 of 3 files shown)


tests/test_pydantic_plugin.py
```diff
@@ -1266,3 +1267,39 @@ class MyModel(BaseModel):
 
     m = MyModel(x=1)
     assert cloudpickle.loads(cloudpickle.dumps(m)).model_dump() == m.model_dump() == {'x': 1}  # type: ignore
+
+
+def test_sqlmodel_pydantic_plugin(exporter: TestExporter) -> None:
+    logfire.instrument_pydantic()
```
Kludex, line 1273: What in the `instrument_pydantic` causes this?

Kludex (approved):

tests/test_pydantic_plugin.py
alexmojaki: I don't know.

alexmojaki: Looked into it more. In the `sqlmodel_validate` function, it calls `cls.__pydantic_validator__.validate_python` which triggers the plugin. That produces an object that's temporarily missing the `_sa_instance_state` attribute which causes `sa_inspect` to fail while the plugin is trying to log the object. Then `sqlmodel_validate` restores that attribute. So nothing stays broken and we shouldn't need to worry. In `logfire.info('x', x=Hero.model_validate({}))`, the span from the plugin doesn't have the right schema, but the `x` log does.

PR #735, opened 2024-12-27
Cache `is_attrs` per class to avoid repeated failed imports

alexmojaki:
Running `uv pip uninstall attrs` and then this script:

```python
import time
import logfire
logfire.configure()

with logfire.span("test") as span:
    span.set_attribute("i", b'3')
    start = time.time()

    for i in range(10000):
        span.set_attribute("i", b'3')

    print(time.time() - start)
```

is >10x slower before this PR. This is because Python's import system is slow and so `import attrs` takes a while to fail each time.

Alternatively we could make it only try to import once overall, but I think this is good enough and will work better if somehow `attrs` becomes importable later.


Kludex (approved):
Is it only with `attrs`? Should we think about making this logic a bit more generic?

alexmojaki:
Yes, it seems to only be with attrs right now. `is_sqlalchemy` first checks `hasattr(obj, '__mapper__')` so doesn't need to try importing sqlalchemy most of the time.

PR #742, opened 2024-12-30
Remove `rich` from `logfire auth`

Kludex:


alexmojaki:
Not all terminals support ANSI escape codes and `rich` does various things to deal with this. I'm not sure if we should try dealing with it ourselves. But if we don't then we should use escape codes with caution. In general we should probably avoid it outside of console logging where color can be disabled. `logfire auth` is one of the first interactions for many users and it doesn't look good if there's some garbage characters in a message, especially attached to a file path. And making some text slightly thicker just isn't that valuable.

alexmojaki:
Yes, or we need a longer conversation about dealing with terminals that don't automatically support ANSI, particularly in Windows.

alexmojaki (approved):

PR #744, opened 2024-12-30
Don't show `urllib3` when `requests` is installed on `logfire inspect`

Kludex:
This is mainly a refactor on the test related to the `logfire inspect`, but I added the conditional to avoid showing the `opentelemetry-instrumentation-urllib3` when `requests` is installed.


Kludex:
I'm going to merge this since I'm trying to iterate on small fixes here.

I'm happy to comply with after merge comments if added @alexmojaki

PR #746, opened 2024-12-30
Drop `rich` from `logfire inspect`

Kludex:


logfire/_internal/cli.py
```diff
@@ -175,26 +170,30 @@ def parse_inspect(args: argparse.Namespace) -> None:
     if packages.get('urllib3') and packages.get('requests'):
         del packages['urllib3']
 
+    # fmt: off
+    sys.stderr.write('The following packages from your environment have an OpenTelemetry instrumentation that is not installed:\n')
+    sys.stderr.write('\n')
+    sys.stderr.write(f' {"Package":<18}| OpenTelemetry instrumentation package\n')
+    sys.stderr.write(f' {"-" * 18}|{"-" * 45}\n')
```
alexmojaki, line 177: please name this magic number, and preferably compute it in case we add some long package name in the future

PR #748, opened 2024-12-30
Add `--ignore` to `logfire inspect`

Kludex:


tests/test_cli.py
```diff
@@ -235,18 +237,25 @@ def packages_from_output(output: str) -> set[str]:
...
+            'inspect',
             ['urllib3', 'requests'],
             {
                 'opentelemetry-instrumentation-requests',
                 'opentelemetry-instrumentation-urllib',
                 'opentelemetry-instrumentation-sqlite3',
             },
         ),
+        (
+            'inspect --ignore urllib --ignore sqlite3',
```
alexmojaki, line 249: i think i'd much rather write `inspect --ignore urllib,sqlite3`

Kludex: Click accepts both for lists. I never use the comma, but I've created a custom action to do it.

PR #755, opened 2024-12-31
Document the HTTPX parameters

Kludex:


docs/integrations/http-clients/httpx.md
```diff
@@ -146,5 +146,106 @@ client.get('https://httpbin.org/get')
...
+
+logfire.configure()
+logfire.instrument_httpx(capture_request_form_data=True)
+
+client = httpx.Client()
+client.post("https://httpbin.org/post", data={"key": "value"})
+```
+
+To capture the request body when the `Content-Type` header is set to `text/*`, you can set
+the `capture_request_text_body` parameter to `True`.
```
Kludex, line 188: It seems this is not true, is everything that can be text... I guess?

PR #761, opened 2024-12-31
Access `model_fields` on the model class

Viicos:
In V3, the instance attribute access will not work anymore.


alexmojaki:
Thanks!

Issue #554, opened 2024-10-29
Issue to send log from Google Cloud Run


m-lce:
### Description

I instrument a FastAPI server running in Google Cloud Run.
Locally it works perferctly but with the same Docker image in Google Cloud Run some messages did not reach the Logfire backend and there is no errors in the logs.
Do you know what could be the issue

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="1.3.2"
platform="Linux-6.11.5-arch1-1-x86_64-with-glibc2.35"
python="3.12.7 (main, Oct 29 2024, 11:47:05) [GCC 11.4.0]"
[related_packages]
requests="2.32.3"
pydantic="2.9.2"
fastapi="0.115.4"
openai="1.52.2"
protobuf="4.25.5"
rich="13.9.3"
executing="2.1.0"
opentelemetry-api="1.27.0"
opentelemetry-exporter-otlp="1.27.0"
opentelemetry-exporter-otlp-proto-common="1.27.0"
opentelemetry-exporter-otlp-proto-grpc="1.27.0"
opentelemetry-exporter-otlp-proto-http="1.27.0"
opentelemetry-instrumentation="0.48b0"
opentelemetry-instrumentation-aiohttp-client="0.48b0"
opentelemetry-instrumentation-asgi="0.48b0"
opentelemetry-instrumentation-dbapi="0.48b0"
opentelemetry-instrumentation-fastapi="0.48b0"
opentelemetry-instrumentation-grpc="0.48b0"
opentelemetry-instrumentation-httpx="0.48b0"
opentelemetry-instrumentation-jinja2="0.48b0"
opentelemetry-instrumentation-pymongo="0.48b0"
opentelemetry-instrumentation-pymysql="0.48b0"
opentelemetry-instrumentation-requests="0.48b0"
opentelemetry-instrumentation-sqlite3="0.48b0"
opentelemetry-instrumentation-starlette="0.48b0"
opentelemetry-instrumentation-urllib="0.48b0"
opentelemetry-instrumentation-urllib3="0.48b0"
opentelemetry-proto="1.27.0"
opentelemetry-sdk="1.27.0"
opentelemetry-semantic-conventions="0.48b0"
opentelemetry-util-http="0.48b0"
```

alexmojaki:
Google Cloud Run is sending a `traceparent` header that OpenTelemetry picks up automatically, and it's also sampling traces. Here's some code to prevent that by effectively disabling distributed tracing:

```python
[...]

PR #773, opened 2025-01-03
Add `distributed_tracing` argument to `configure` and new docs about context propagation

alexmojaki:
Fixes https://github.com/pydantic/logfire/issues/554


mkdocs.yml
```diff
@@ -88,6 +88,7 @@ nav:
       - Implement Sampling Strategies: how-to-guides/sampling.md
       - Export your Logfire Data: how-to-guides/query-api.md
       - Scrub Sensitive Data: how-to-guides/scrubbing.md
+      - Trace across Processes: how-to-guides/distributed-tracing.md
```
dmontagu:
```suggestion
      - Distributed Tracing: how-to-guides/distributed-tracing.md
```
Not sure where the "Trace across Processes" name came from, I understand if "Distributed Tracing" seems too jargony, but reading "across Processes" makes me think it's more of a process-pool thing in a single python program, than an across-different-hardware-instances sort of thing.

If you don't like "Distributed Tracing", I would suggest:
```suggestion
      - Trace across Multiple Services: how-to-guides/distributed-tracing.md
```

dmontagu (approved):
Seems very good, one small note above

mkdocs.yml
alexmojaki: all the headings here complete the sentence 'How to'. Not sure I like that but I was going along with it.

PR #776, opened 2025-01-06
Downgrade pinned mkdocs-material

Kludex:


Kludex:
I don't understand why I need to do this to make it install locally, considering that the cloudflare worker is using our insiders as well.

alexmojaki (approved):

PR #781, opened 2025-01-07
Document HTTPX parameters

Kludex:


alexmojaki:
The page feels very verbose and repetitive. Is it because you want each section to work on its own?

Capturing bodies needs to mention something about bytes vs text.

Also I wouldn't call `capture_all` (or even the other params) 'fine tuning'.

alexmojaki (approved):
Thanks.

I think "capture everything" should be at the top of configuration.

PR #783, opened 2025-01-08
Generate trace IDs as ULIDs by default

adriangb:
The idea is to use a ULID to generate trace IDs.
This has the huge benefit of making trace IDs sortable and correlated with time.
For our backend this means much faster queries and much less data to index them.

This should not cause any issues with other backends, sampling or anything else.

This is [already the default for logs](https://github.com/open-telemetry/opentelemetry-specification/pull/3047) and [has been proposed to become the default for traces](https://github.com/open-telemetry/opentelemetry-specification/issues/1947).

**Why not use UUID v7 or set the bits for UUID v4 compatibility?**

Mainly for safety of current implementations.
Existing implementations (including the OTEL Python SDK) should expect no more than 64 bits or randomness (see [here](https://github.com/open-telemetry/opentelemetry-python/blob/29aad2efa30a35757ea78d90e4d3fd517bc00468/opentelemetry-sdk/src/opentelemetry/sdk/trace/sampling.py#L268-L270)).
Using a pure ULID (80 bits of randomness) ensures compatibility, using a UUID would leave us with only 7 (contiguous lower) bytes of randomness at most.
In the future if OTEL adopts https://www.w3.org/TR/trace-context-2/#random-trace-id-flag we can revisit.


tests/test_sampling.py
```diff
@@ -70,7 +70,7 @@ def test_sample_rate_config(exporter: TestExporter, config_kwargs: dict[str, Any
 
     # 1000 iterations of 2 spans -> 2000 spans
     # 30% sampling -> 600 spans (approximately)
-    assert len(exporter.exported_spans_as_dict()) == 634
+    assert len(exporter.exported_spans_as_dict()) == 588, len(exporter.exported_spans_as_dict())
```
alexmojaki, line 73:
```suggestion
    assert len(exporter.exported_spans_as_dict()) == snapshot(588)
```

tests/test_tail_sampling.py
```diff
@@ -345,7 +345,7 @@ def test_background_rate(config_kwargs: dict[str, Any], exporter: TestExporter):
     # None of them meet the tail sampling criteria.
     for _ in range(1000):
         logfire.info('info')
-    assert len(exporter.exported_spans) == 100 + 321
+    assert len(exporter.exported_spans) == 100 + 299, len(exporter.exported_spans)
```
alexmojaki:
```suggestion
    assert len(exporter.exported_spans) - 100 == snapshot(299)
```

alexmojaki: same below

logfire/_internal/ulid.py
```diff
@@ -0,0 +1,42 @@
...
+    - Our SDK / Python SDK's in general.
+    - The OTEL collector.
+
+    And both behave properly with 8 bytes of randomness because trace IDs were originally 64 bits
+    so to be compatible with old trace IDs nothing in OTEL can assume >8 bytes of randomness in trace IDs
+    unless they generated the trace ID themselves (e.g. the Go SDK _does_ expect >8 bytes of randomness internally).
+    """
+    # Timestamp: first 6 bytes of the ULID (48 bits)
+    # Note that it's not important that this timestamp is super precise or unique.
+    # It just needs to be monotonically increasing so that the ULID is sortable, at least for our purposes.
```
alexmojaki:
```suggestion
    # It just needs to be roughly monotonically increasing so that the ULID is sortable, at least for our purposes.
```

since this doesn't respect https://github.com/ulid/spec?tab=readme-ov-file#monotonicity

logfire/_internal/utils.py
```diff
@@ -384,7 +385,7 @@ def generate_span_id(self) -> int:
         return span_id
 
     def generate_trace_id(self) -> int:
-        trace_id = self.random.getrandbits(128)
+        trace_id = ulid(self.random)
         while trace_id == trace_api.INVALID_TRACE_ID:  # pragma: no cover
-            trace_id = self.random.getrandbits(128)
+            trace_id = ulid(self.random)
         return trace_id
```
alexmojaki: just `return ulid(self.random)`, no need to check for this case any more

tests/test_logfire.py
```diff
@@ -3221,15 +3222,28 @@ def test_exit_ended_span(exporter: TestExporter):
...
     """Test that SeededRandomIdGenerator generates trace and span ids without errors."""
-    with ExitStack() as stack:
-        for i in range(1024):
+    for i in range(32):
+        with ExitStack() as stack:
             for j in range(8):
                 stack.enter_context(logfire.span(f'span {i}:{j}'))
+                # make sure we get a mix of traces spread across ms
+                # test should take < 32 * 8 ms (0.25s) to run
+                sleep(1 / 1000 / 5)
```
alexmojaki: Rather inject ns_timestamp_generator into the ID generator, then the exact IDs will be deterministic in tests.

alexmojaki: and if you really want to generate lots of spans, this is significantly faster:

```python
    for i in range(1024 * 8):
        logfire.info('test', i=i)
    sorted_by_trace_id = [
        export['attributes']['i'] for export in sorted(exported, key=lambda span: span['context']['trace_id'])
    ]
```

alexmojaki:
The test was broken, it would always sort by the random part of the trace ID and never disambiguate by start time. Only 3.8 picked this up because f-string magic is disabled there. Otherwise the span name was always the f-string template.

adriangb:
> The test was broken, it would always sort by the random part of the trace ID and never disambiguate by start time. Only 3.8 picked this up because f-string magic is disabled there. Otherwise the span name was always the f-string template.

Ah good call. Should have used message!

adriangb:
Thanks for your help Alex!

PR #800, opened 2025-01-15
Add logfire.warning to mirror logging.warning

### LINKED ISSUE #793: Add `logfire.warning` to mirror stdlib `logging.warning`
### Description

The standard library's `logging` module [uses `warning` and deprecated `warn`](https://docs.python.org/3/library/logging.html#logging.Logger.warning). It'd be helpful to at least make `warning` available as `logfire.warning` and `logfire.Logfire.warning`.

This would make it a bit easier to write code that can accept either a `logging.Logger` or a `logfire.Logfire` instance (eg: library code or easier test injection).

I'm happy to submit a PR for this if you all are open to this. I see a [prior commit](https://github.com/pydantic/logfire/commit/5f88ab7ce7c6c281d9e60bf172747f0699501b32) replacing the _usage_ of a `warning` method with `warn`, but I think the referenced PR was from a prior repo / before it was public, so I'm not sure if `warning` was removed intentional before.

JacobHayes:
I didn't see any obvious docs to update - any I should take a look at?

Closes #793

### DIFF
--- a/logfire-api/logfire_api/_internal/main.pyi
+++ b/logfire-api/logfire_api/_internal/main.pyi
@@ -136,7 +136,7 @@ class Logfire:
 
                 Set to `True` to use the currently handled exception.
         """
-    def warn(self, msg_template: str, /, *, _tags: Sequence[str] | None = None, _exc_info: ExcInfo = False, **attributes: Any) -> None:
+    def warning(self, msg_template: str, /, *, _tags: Sequence[str] | None = None, _exc_info: ExcInfo = False, **attributes: Any) -> None:
         """Log a warning message.
 
         ```py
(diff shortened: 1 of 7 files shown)


logfire-api/logfire_api/_internal/main.pyi
alexmojaki, line 1: i think i'd prefer

```python
def warning(...):
    ...

warn = warning
```

JacobHayes: Works for me - are you ok without the note that "warn is an alias of warning"?

alexmojaki: I think you can still have a note in the `warning` docstring

PR #802, opened 2025-01-17
Fix conflict with ddtrace futures patching by renaming fn parameter

alexmojaki:
Fixes problem reported in https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1737061346996829

Before this PR, this:

```python
from concurrent.futures import ThreadPoolExecutor

from ddtrace.contrib.internal.futures.patch import patch

import logfire

# Removing one of these or changing the order will cause the error to disappear.
patch()
logfire.configure()

ThreadPoolExecutor().submit(lambda: None).result()
```

raises:

```pytb
  File "/Users/alex/Library/Application Support/JetBrains/PyCharm2024.3/scratches/scratch_990.py", line 11, in <module>
    ThreadPoolExecutor().submit(lambda: None).result()
  File "/Users/alex/.pyenv/versions/3.12.6/lib/python3.12/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/.pyenv/versions/3.12.6/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/Users/alex/.pyenv/versions/3.12.6/lib/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/ddtrace/contrib/internal/futures/threading.py", line 36, in _wrap_execution
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
TypeError: <lambda>() got an unexpected keyword argument 'carrier'
```

`Executor.submit` has the signature `(self, fn, /, *args, **kwargs)`. Although `fn` is now positional-only, ddtrace pays special attention to that name in https://github.com/DataDog/dd-trace-py/blob/26dc258020126a7527f973c26494f0dd1016dead/ddtrace/contrib/internal/futures/threading.py#L7-L23

So avoiding that name does the trick.

### DIFF
--- a/logfire/_internal/integrations/executors.py
+++ b/logfire/_internal/integrations/executors.py
@@ -28,15 +28,17 @@ try:
         """A wrapper around ThreadPoolExecutor.submit() that carries over OTEL context across threads."""
         fn = partial(fn, *args, **kwargs)
         carrier = get_context()
-        return submit_t_orig(s, _run_with_context, carrier=carrier, fn=fn, parent_config=None)
+        return submit_t_orig(s, _run_with_context, carrier=carrier, func=fn, parent_config=None)
 
     def submit_p(s: ProcessPoolExecutor, fn: Callable[..., Any], /, *args: Any, **kwargs: Any):
         """A wrapper around ProcessPoolExecutor.submit() that carries over OTEL context across processes."""
         fn = partial(fn, *args, **kwargs)
         carrier = get_context()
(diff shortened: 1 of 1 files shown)


nairb774:
Thanks for quickly fixing this - really appreciate it!

PR #804, opened 2025-01-17
Fix compatibility with Pydantic dropping 3.8

Viicos:


alexmojaki:
Are we dropping 3.8 or not? This seems indecisive.

alexmojaki:
I spoke to @samuelcolvin about it and he seemed happy to do so.

PR #811, opened 2025-01-20
Add process runtime information

Kludex:
https://opentelemetry.io/docs/specs/semconv/resource/process/#python-runtimes

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -712,6 +712,9 @@ class LogfireConfig(_LogfireConfigData):
             otel_resource_attributes: dict[str, Any] = {
                 ResourceAttributes.SERVICE_NAME: self.service_name,
                 ResourceAttributes.PROCESS_PID: os.getpid(),
+                ResourceAttributes.PROCESS_RUNTIME_NAME: sys.implementation.name,
+                ResourceAttributes.PROCESS_RUNTIME_VERSION: get_runtime_version(),
+                ResourceAttributes.PROCESS_RUNTIME_DESCRIPTION: sys.version,
                 # Having this giant blob of data associated with every span/metric causes various problems so it's
                 # disabled for now, but we may want to re-enable something like it in the future
                 # RESOURCE_ATTRIBUTES_PACKAGE_VERSIONS: json.dumps(collect_package_info(), separators=(',', ':')),
@@ -1558,6 +1561,13 @@ def sanitize_project_name(name: str) -> str:
 
 def default_project_name():
     return sanitize_project_name(os.path.basename(os.getcwd()))
+
+
+def get_runtime_version() -> str:
+    version_info = sys.implementation.version
+    if version_info.releaselevel == 'final' and not version_info.serial:
+        return '.'.join(map(str, version_info[:3]))
+    return '.'.join(map(str, version_info))
 
 
 class LogfireNotConfiguredWarning(UserWarning):
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -1567,7 +1567,7 @@ def get_runtime_version() -> str:
     version_info = sys.implementation.version
     if version_info.releaselevel == 'final' and not version_info.serial:
         return '.'.join(map(str, version_info[:3]))
-    return '.'.join(map(str, version_info))
+    return '.'.join(map(str, version_info))  # pragma: no cover
 
 
 class LogfireNotConfiguredWarning(UserWarning):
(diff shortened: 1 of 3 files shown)


logfire/_internal/config.py
```diff
@@ -1560,5 +1563,12 @@ def default_project_name():
     return sanitize_project_name(os.path.basename(os.getcwd()))
 
 
+def get_runtime_version() -> str:
```
adriangb, line 1567: A doc string and some tests would be nice. Probably split it into a getter and a parser function so it can be tested.

Kludex: tests are already there

adriangb: with all possible inputs of `sys.implementation.version` e.g. pre-release versions? It'd be nice to see _explicitly_ what the expected inputs are.

Viicos: The issue is you can't instantiate the `sys.version_info` class to define custom values. So we could define a mock class acting like it but I don't know if it is worth it?

adriangb: I just want:

```python
def get_runtime_version(version_info: ?) -> str:
    if version_info.releaselevel == 'final' and not version_info.serial:
        return '.'.join(map(str, version_info[:3]))
    return '.'.join(map(str, version_info))  # pragma: no cover
```

And then tests along the lines of:

```python
def test_get_runtime_version_pre_release() -> None:
    version_info = sys.VersionInfo(???)
    version = get_runtime_version(version_info)
    assert version == ??
```

Maybe with parametrize

Viicos: Yes but you can't instantiate version infos, meaning we'd have to do something like:

```python
class MockVersionInfo(tuple[int, int, int, str, int]):
    @property
    def releaselevel(self) -> str: ...
    ...

get_runtime_version(MockVersionInfo(...))
```

Viicos:
```python
sys.version_info(3, 9, 19, 'final', 0)
#> TypeError: 'sys.version_info' object is not callable
```

adriangb: Is it an actual type?

```python
VersionInfo = tuple[int, int, int, str, int]

def get_runtime_version(version_info: VersionInfo) -> None:
    ...

Or worst case:

class SysVersionInfoProvider(Protocol):
    def __call__(self) -> VersionInfo: ...

def get_runtime_version(version_info: SysVersionInfoProvider) -> None:
    ...
```

Viicos: `sys.version_info` is an instance of the `version_info` class (confusingly of the same name, meaning my previous example doesn't make sense), but it can't be instantiated:

```python
type(sys.version_info)
#> <class 'sys.version_info'>
type(sys.version_info)(3, 9, 19, 'final', 0)
#> TypeError: cannot create 'sys.version_info' instances
```

So we could write a parametrized test, but it requires a mock class:

```python
class MockVersionInfo(tuple[int, int, int, str, int]):
    @property
    def releaselevel(self) -> str: ...
    ...

@pytest.mark.parametrize(
    ['version_info', 'expected'],
    [
        (MockVersionInfo(3, 14, 0, 'alpha', 2), '3.14.0.alpha.2'),
        ...
    ]
)
def test_get_runtime_version(version_info, expected):
    assert get_runtime_version(version_info) == expected
```

logfire/_internal/config.py
```diff
@@ -712,6 +712,9 @@ def _initialize(self) -> None:
             otel_resource_attributes: dict[str, Any] = {
                 ResourceAttributes.SERVICE_NAME: self.service_name,
                 ResourceAttributes.PROCESS_PID: os.getpid(),
+                ResourceAttributes.PROCESS_RUNTIME_NAME: sys.implementation.name,
+                ResourceAttributes.PROCESS_RUNTIME_VERSION: get_runtime_version(),
+                ResourceAttributes.PROCESS_RUNTIME_DESCRIPTION: sys.version,
```
Kludex, line 718:
```suggestion
                # https://opentelemetry.io/docs/specs/semconv/resource/process/#python-runtimes
                ResourceAttributes.PROCESS_RUNTIME_NAME: sys.implementation.name,
                ResourceAttributes.PROCESS_RUNTIME_VERSION: get_runtime_version(),
                ResourceAttributes.PROCESS_RUNTIME_DESCRIPTION: sys.version,
```

Kludex:
Can we merge this @adriangb ?

adriangb:
Sure go for it

PR #818, opened 2025-01-24
Support Pyodide

samuelcolvin:


pyproject.toml
```diff
@@ -248,17 +248,7 @@ quote-style = "single"
 typeCheckingMode = "strict"
 reportUnnecessaryTypeIgnoreComment = true
 reportMissingTypeStubs = false
-exclude = [
```
samuelcolvin, line 251: @Kludex I can revert this if you want, but we're excluding many more things than we're including, so this seemed more sensible.

Kludex: It's okay.

Kludex (approved):

PR #820, opened 2025-01-26
Parallelize tests using pytest-xdist

sorasful:
Hi there! 

I've noticed that the tests took a little while to run. So I've decided to see if we could parallelize them, I used `pytest-xdist`. 

For very few tests (3), concurrency was a problem so I've decided to run them sequentially. I followed this issue to implement this : 

https://github.com/pytest-dev/pytest-xdist/issues/385 

Now everything runs smoothly and quickly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.env
+.idea/
 .DS_Store
 .logfire
 .python-version
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 *.env
-.idea/
 .DS_Store
 .logfire
 .python-version
--- a/tests/otel_integrations/test_celery.py
+++ b/tests/otel_integrations/test_celery.py
@@ -61,6 +61,7 @@ def celery_worker(celery_app: Celery) -> Iterator[WorkController]:
         yield worker
 
 
+@pytest.mark.xdist_group(name='sequential')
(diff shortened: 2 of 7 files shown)


tests/otel_integrations/test_celery.py
```diff
@@ -61,6 +61,7 @@ def celery_worker(celery_app: Celery) -> Iterator[WorkController]:
         yield worker
 
 
+@pytest.mark.xdist_group(name='sequential')
```
alexmojaki: I'm removing this one because I think this test is just flaky, I can't see evidence that xdist makes this fail more often.

.gitignore
```diff
@@ -1,4 +1,5 @@
 *.env
+.idea/
```
alexmojaki: @Kludex I'm guessing you don't want this?

alexmojaki:
@sorasful awesome, thanks for the speedup! Is there a way to make it use `xdist` by default, e.g. when I run `pytest` in the terminal, or run pytest through my IDE? Maybe a setting in pyproject.toml?

.gitignore
Kludex: Yep, I don't agree with adding environment specific to the `.gitignore`. Users can and should add `.idea` to the global `.gitignore`.

Kludex (approved):

.gitignore
alexmojaki:
```suggestion
```

Issue #825, opened 2025-01-29
Instrumenting HTTPX somehow breaks CORS in pyodide


samuelcolvin:
### Description

See https://pydantic.run/store/2394806888132a13

Code is 

```py
# /// script
# dependencies = ['logfire[httpx]', 'httpx']
# ///
import logfire
import httpx

logfire.configure(send_to_logfire=False)
logfire.instrument_httpx(capture_all=True)

r = httpx.get('https://global-power-plants.datasettes.com/global-power-plants.db')
r.raise_for_status()
print('datasize:', len(r.content))
```

If I comment out `logfire.instrument_httpx(capture_all=True)` it works fine.

### Python, Logfire & OS Versions, related packages (not required)

```TOML

```

alexmojaki:
> Access to XMLHttpRequest at 'https://global-power-plants.datasettes.com/global-power-plants.db' from origin 'https://pydantic.run' has been blocked by CORS policy: Request header field traceparent is not allowed by Access-Control-Allow-Headers in preflight response.

So we need to prevent context propagation in HTTP clients under pyodide.

alexmojaki:
Closing because I'm guessing we don't actually need to fix this in all cases, it's enough to be able to make requests to specific domains for demonstrations.

PR #826, opened 2025-01-29
add `logfire.logfire_info()`

samuelcolvin:
While creating #825 I realised there's no nice way to get the `logfire info` output if you don't have a terminal.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -20,10 +20,16 @@ body:
 
         **Don't worry if you can't run this command or don't have this information, we'll help you if we can.**
 
-        Please run the following command and copy the output below:
+        Please run the following command in your terminal:
 
         ```bash
         logfire info
+        ```
+
+        Or in python run:
+
+        ```python
+        import logfire; print(logfire.logfire_info())
         ```
 
       render: TOML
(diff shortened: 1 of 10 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -20,10 +20,16 @@ body:
...
         **Don't worry if you can't run this command or don't have this information, we'll help you if we can.**
 
-        Please run the following command and copy the output below:
+        Please run the following command in your terminal:
 
         ```bash
         logfire info
         ```
 
+        Or in python run:
```
samuelcolvin, line 29: this will be wrong until we make a release.

logfire-api/logfire_api/__init__.py
```diff
@@ -229,3 +229,7 @@ def __init__(self, *args, **kwargs) -> None: ...
 
         class LogfireLoggingHandler:
             def __init__(self, *args, **kwargs) -> None: ...
+
+        def logfire_info() -> str:
+            """Show versions of logfire, OS and related packages."""
+            raise NotImplementedError('this method is not implement in logfire-api')
```
Kludex: Logfire API is meant to be a no-op. Why would it not be possible to call it with logfire-api?

samuelcolvin: okay, I'll make it return an empty string.

Kludex (approved):

PR #827, opened 2025-01-29
Remove None from text that should be printed in console

hramezani:
Fixes https://github.com/pydantic/logfire/pull/827
https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1738124276068249

### DIFF
--- a/logfire/_internal/exporters/console.py
+++ b/logfire/_internal/exporters/console.py
@@ -118,9 +118,9 @@ class SimpleConsoleSpanExporter(SpanExporter):
             parts += [('\n', '')] + details_parts
 
         if self._console:
-            self._console.print(Text.assemble(*parts))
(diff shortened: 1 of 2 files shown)


alexmojaki:
Needs a test, and should check `isinstance(text, str)` instead of just `text is not None`.

hramezani:
> Needs a test, and should check `isinstance(text, str)` instead of just `text is not None`.

Replaced with `isinstance(text, str)`. I didn't add a test because I couldn't reproduce the problem. do you know how can I create such a span?

alexmojaki:
Not sure what exactly happened in the user's case, but this produces the error:

```python
import logfire

logfire.info('hi', **{'code.filepath': 3, 'code.lineno': None})
```

hramezani:
Replaced by https://github.com/pydantic/logfire/pull/831

PR #829, opened 2025-01-29
Add `logfire.add_non_user_code_prefix` function for library developers

### LINKED ISSUE #745: add `def non_user_code_prefix`
### Description

Add

```py
def non_user_code_prefix(path: str | Path) -> None:
    if isinstance(path, Path):
        path = str(path)
    logfire._internal.stack_info.NON_USER_CODE_PREFIXES += (path,)
```

Equivilant to

```py
logfire._internal.stack_info.NON_USER_CODE_PREFIXES += (str(Path(__file__).parent.absolute()),)
```

dmontagu:
Adds a public API for adding items to the tuple of non-user-code-prefixes, so that the frames are excluded in logfire spans. This will be used in `pydantic_ai`; see occurrences of the following in that codebase:
```python
# while waiting for https://github.com/pydantic/logfire/issues/745
try:
    import logfire._internal.stack_info
except ImportError:
    pass
else:
    from pathlib import Path

    logfire._internal.stack_info.NON_USER_CODE_PREFIXES += (str(Path(__file__).parent.absolute()),)
```

Closes #745

### DIFF
--- a/logfire-api/logfire_api/__init__.py
+++ b/logfire-api/logfire_api/__init__.py
@@ -193,6 +193,8 @@ except ImportError:
         def no_auto_trace(x):
             return x
 
+        def add_non_user_code_prefix(*args, **kwargs) -> None: ...
+
         @contextmanager
         def suppress_instrumentation():
             yield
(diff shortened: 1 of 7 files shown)


alexmojaki:
I've decided against the `logfire.testing` idea because that module requires pytest. I think the module thing would be good but I guess it's not essential and I'd like to merge this.

PR #845, opened 2025-02-05
Handle null arguments using JSON schema

alexmojaki:
Just a little simplification of our data model and logic that I've wanted.


adriangb (approved):
Looks good to me!

PR #847, opened 2025-02-07
Use Algolia UI widgets for docs search

petyosi:
This follows the logic of https://github.com/pydantic/pydantic-ai/pull/845. 

image


petyosi:
@Kludex clicking on the search results does not take you to the page because of the URL mismatch between the branch deployment and prod. So you will have to take a leap of faith here.

Kludex (approved):
faith 🙏

PR #854, opened 2025-02-11
Catch more errors when checking for sqlalchemy objects

alexmojaki:
https://pydanticlogfire.slack.com/archives/C06EDRBSAH3/p1739205757483649

In general I think we need to start being more aggressive with catching exceptions here, we're working with arbitrary objects that may have all sorts of weirdness, and logging tracebacks about internal errors doesn't look good.


Kludex (approved):
Hard to be an observability package.

PR #858, opened 2025-02-12
Set log level to warning instead of error for 4xx HTTPExceptions from FastAPI/Starlette

alexmojaki:


Kludex:
Why is it even warning and not info?

alexmojaki:
I think this is more intuitive/expected but we can reduce it further to info if we want.

PR #860, opened 2025-02-13
Add option to disable printing tags to console

dmontagu:
This adds an option to `ConsoleOptions` to just disable the printing of tags to the console entirely.

This was in response to a user-reported complaint about long tags polluting their console logs. I'm not totally sure if this is an adequate solution to their issue, and maybe we shouldn't merge this if it isn't. But it was fairly fast to implement so I figured I'd just open it.


alexmojaki:
Thanks!

PR #863, opened 2025-02-13
Initial OTEL logs infra

alexmojaki:
Some of the boilerplate we need to configure OTEL logs/events in logfire, similar to what we have for tracing and metrics. Does nothing by default and the configuration is tucked away in advanced options. Not comprehensive, e.g. doesn't have scrubbing.

### DIFF
--- a/tests/test_otel_logs.py
+++ b/tests/test_otel_logs.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 from opentelemetry._logs import LogRecord, SeverityNumber, get_logger
 from opentelemetry.sdk._logs.export import InMemoryLogExporter
 
+import logfire
+
 
 def test_otel_logs(logs_exporter: InMemoryLogExporter) -> None:
     record = LogRecord(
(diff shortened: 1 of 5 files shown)


tests/test_otel_logs.py
```diff
@@ -0,0 +1,107 @@
...
+        observed_timestamp=2,
+        trace_id=3,
+        span_id=4,
+        severity_text='INFO',
+        severity_number=SeverityNumber.INFO,
+        body='body',
+        attributes={'key': 'value'},
+    )
+
+    logfire.suppress_scopes('scope1')
```
adriangb, line 28: I guess this method already exists, so not for this PR, but may I suggest we somehow allow restricting scopes by level? E.g. "only allow error or above for scope1, allow info or above for scope2"

adriangb (commented):
Looks good to me.

I am curious to see how you handle tail sampling and such...

tests/test_otel_logs.py
alexmojaki: It's been requested and considered, just not prioritized.

Issue #867, opened 2025-02-17
Calling `logfire.warn` with a Django object raises exception


joshbrooks:
### Description

I'm calling `logfire.warn` on a Django "User" object with `logfire.warn(request.user)`. What I'd expect is that I'd get the equivalent of calling `str(request.user)` in the log

```
In [2]: User.objects.first()
Out[2]: <User: eskola>

In [3]: str(User.objects.first())
Out[3]: 'eskola'
```

Instead, I get an exception as the serialization of the Django object is not supported

```
Feb 17 11:01:20 Caught an internal error in Logfire. Your code should still be running fine, just with less telemetry. This is just logging the internal error.
Feb 17 11:01:20 Traceback (most recent call last):
Feb 17 11:01:20   File "/var/www/education/env/lib/python3.12/site-packages/django/views/generic/base.py", line 143, in dispatch
Feb 17 11:01:20     return handler(request, *args, **kwargs)
Feb 17 11:01:20            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Feb 17 11:01:20   File "/var/www/education/source/gather/gather/surveydata/views.py", line 603, in put
Feb 17 11:01:20     couchdoc = self.create_couchdoc(request, rid=rid, uid=uid)
Feb 17 11:01:20                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Feb 17 11:01:20   File "/var/www/education/source/gather/gather/surveydata/views.py", line 564, in create_couchdoc
Feb 17 11:01:20     logfire.warn(request.user)
Feb 17 11:01:20   File "/var/www/education/env/lib/python3.12/site-packages/logfire/_internal/main.py", line 423, in warning
Feb 17 11:01:20     self.log('warn', msg_template, attributes, tags=_tags, exc_info=_exc_info)
Feb 17 11:01:20   File "/var/www/education/env/lib/python3.12/site-packages/logfire/_internal/main.py", line 681, in log
Feb 17 11:01:20     msg, extra_attrs, msg_template = logfire_format_with_magic(
Feb 17 11:01:20                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
Feb 17 11:01:20   File "/var/www/education/env/lib/python3.12/site-packages/logfire/_internal/formatter.py", line 330, in logfire_format_with_magic
[...]

PR #868, opened 2025-02-17
Fix PyCharm type hinting

alexmojaki:
Some stupid bug means that this is required for PyCharm to give warnings that would prevent mistakes like https://github.com/pydantic/logfire/issues/867


alexmojaki:
It seems to know all the types, it just stops warning about them.

PR #877, opened 2025-02-20
PYD-1550: add section on OTEL Collector

adriangb:


mkdocs.yml
```diff
@@ -95,6 +95,7 @@ nav:
       - Trace across Multiple Services: how-to-guides/distributed-tracing.md
       - Detect Service is Down: how-to-guides/detect-service-is-down.md
       - Suppress Spans and Metrics: how-to-guides/suppress.md
+      - OpenTelemetry Collector: how-to-guides/otel-collector.md
```
Kludex, line 98: I think the title should be changed to something more "How To Guides". Like, the idea of those are to be problem oriented, and the title of this one looks more just "normal documentation".

Kludex: Maybe "Set up a OpenTelemetry Collector"

adriangb: I guess the question is do we want:
1. A docs section about the Collector listing several use cases.
2. A docs section for each use case as a how to guide.

I'm kind of going for (1) here, happy to put it somewhere else if you can suggest a good spot.

Kludex: Advanced? it's a bit hidden tho

Kludex: @ChristopherGS

ChristopherGS: Wherever gets it shipped fastest :)

Advanced is probably a good enough start. We have the Algolia search now so people will be able to find.

I'm also fine to have a catch-all "How to use with an OTel Collector" / "How to migrate via an OTel Collector" guide which later we split out into more focused how-tos.

adriangb: @Kludex let's just leave it here then if we're going for fastest shipping to avoid bikeshed

docs/how-to-guides/otel-collector.md
```diff
@@ -0,0 +1,283 @@
...
+subjects:
+- kind: ServiceAccount
+  name: otel-collector
+  namespace: default
+```
+
+Apply this configuration via `kubectl apply -f rbac.yaml`.
+
+Now we can create the deployment for the collector itself.
+There are several options for deploying the OTel collector, including:
```
MRoci: perhaps worth mentioning [otel doc](https://opentelemetry.io/docs/platforms/kubernetes/collector/components)  as a reference to the deployments options

MRoci (approved):
LGTM with 1 nitpick

adriangb:
Addressed.

@Kludex wdyt?

Kludex:
Go ahead.

PR #881, opened 2025-02-21
create compliance page

ChristopherGS:
Mention our SOC2 Type I certification, which we were awarded formally yesterday


docs/compliance.md
Kludex, line 1: Can we have it more visible? First page maybe?

Kludex: Also, why are we not putting this on soocial media????

ChristopherGS: Yes, we'll do socials

ChristopherGS: I think the current location is good - people aren't searching for compliance as often as integrations / how-tos. And as the guy who is always looking for this kind of thing, I know it's usually down at the bottom

ChristopherGS: Added to the first page also

PR #899, opened 2025-02-25
internal: add a marker for the first span event for each exception

adriangb:
This will allow us to differentiate in the backend if this is the span an exception originated inside of vs. subsequent spans it tears through as it goes up the stack

### DIFF
--- a/logfire/_internal/tracer.py
+++ b/logfire/_internal/tracer.py
@@ -384,6 +384,18 @@ def record_exception(
         stacktrace = ''.join(traceback.format_exception(type(exception), exception, exception.__traceback__))
         attributes[SpanAttributes.EXCEPTION_STACKTRACE] = stacktrace
 
+    if not getattr(exception, '_logfire_recorded', False):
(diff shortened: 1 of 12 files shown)


logfire/_internal/tracer.py
```diff
@@ -384,6 +384,18 @@ def record_exception(
         stacktrace = ''.join(traceback.format_exception(type(exception), exception, exception.__traceback__))
         attributes[SpanAttributes.EXCEPTION_STACKTRACE] = stacktrace
 
+    if not getattr(exception, '_logfire_recorded', False):
+        # This exception has already been recorded, mark it as such in attributes so that
+        # our backend can replicate the behavior of exceptions as OTEL logs, see
+        # https://github.com/open-telemetry/opentelemetry-specification/pull/4430#issue-2876146448
+        # and linked issues.
+        attributes['logfire.exception_first_recorded'] = True
```
adriangb, line 392: @alexmojaki the idea is that we only explode exception span events if this attribute is set

PR #901, opened 2025-02-26
Account for new EU region

adriangb:
This goes hand in hand with a backend PR (private) that adds a region prefix to tokens.

### DIFF
--- a/logfire/_internal/cli.py
+++ b/logfire/_internal/cli.py
@@ -24,7 +24,7 @@ from logfire.propagate import ContextCarrier, get_context
 
 from ..version import VERSION
 from .auth import DEFAULT_FILE, HOME_LOGFIRE, DefaultFile, is_logged_in, poll_for_token, request_device_code
-from .config import REGIONS, LogfireCredentials
+from .config import REGIONS, LogfireCredentials, get_base_url_from_token
 from .config_params import ParamManager
 from .tracer import SDKTracerProvider
 from .utils import read_toml_file
@@ -50,8 +50,9 @@ def parse_whoami(args: argparse.Namespace) -> None:
     """Show user authenticated username and the URL to your Logfire project."""
     data_dir = Path(args.data_dir)
     param_manager = ParamManager.create(data_dir)
-    base_url = param_manager.load_param('base_url', args.logfire_url)
+    base_url = param_manager.load_param('base_url', _get_logfire_url(args.logfire_url, args.region))
     token = param_manager.load_param('token')
+    base_url = base_url or get_base_url_from_token(token)
 
     if token:
         credentials = LogfireCredentials.from_token(token, args._session, base_url)
@@ -198,17 +199,24 @@ def parse_auth(args: argparse.Namespace) -> None:
     """
     logged_in = False
 
+    logfire_url = _get_logfire_url(args.logfire_url, args.region)
     if DEFAULT_FILE.is_file():
         data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
-        if args.logfire_url and is_logged_in(data, args.logfire_url):  # pragma: no branch
+        if logfire_url and is_logged_in(data, logfire_url):  # pragma: no branch
             logged_in = True
-        elif not args.logfire_url:
+        elif not logfire_url:
             logged_in = any(is_logged_in(data, url) for url in data['tokens'])
     else:
         data: DefaultFile = {'tokens': {}}
 
     if logged_in:
-        sys.stderr.write(f'You are already logged in. (Your credentials are stored in {DEFAULT_FILE})\n')
+        sys.stderr.writelines(
+            (
+                f'You are already logged in. (Your credentials are stored in {DEFAULT_FILE})\n',
+                'If you would like to log in using a different account, use the --region argument:\n',
+                'logfire --region <region> auth\n',
+            )
+        )
         return
 
     sys.stderr.writelines(
@@ -219,15 +227,20 @@ def parse_auth(args: argparse.Namespace) -> None:
             '\n',
         )
     )
-    if not args.logfire_url:
-        # TODO input validation
-        sys.stderr.write('Logfire is available in multiple data regions. Please select one:\n')
-        for i, (region_id, region_data) in enumerate(REGIONS.items(), start=1):
-            sys.stderr.write(f'{i}. {region_id.upper()} (GCP region: {region_data["gcp_region"]})\n')
-        selected_region = int(input(f'Selected region [{"/".join(str(i) for i in range(1, len(REGIONS) + 1))}]: '))
+    if not logfire_url:
+        selected_region = -1
+        while not (1 <= selected_region <= len(REGIONS)):
+            sys.stderr.write('Logfire is available in multiple data regions. Please select one:\n')
+            for i, (region_id, region_data) in enumerate(REGIONS.items(), start=1):
+                sys.stderr.write(f'{i}. {region_id.upper()} (GCP region: {region_data["gcp_region"]})\n')
+
+            try:
+                selected_region = int(
+                    input(f'Selected region [{"/".join(str(i) for i in range(1, len(REGIONS) + 1))}]: ')
+                )
+            except ValueError:
+                selected_region = -1
         logfire_url = list(REGIONS.values())[selected_region - 1]['base_url']
-    else:
-        logfire_url = cast(str, args.logfire_url)
 
     device_code, frontend_auth_url = request_device_code(args._session, logfire_url)
     frontend_host = urlparse(frontend_auth_url).netloc
@@ -258,7 +271,7 @@ def parse_auth(args: argparse.Namespace) -> None:
 
 def parse_list_projects(args: argparse.Namespace) -> None:
     """List user projects."""
-    logfire_url = args.logfire_url
+    logfire_url = _get_logfire_url(args.logfire_url, args.region)
     projects = LogfireCredentials.get_user_projects(session=args._session, logfire_api_url=logfire_url)
     if projects:
         sys.stderr.write(
@@ -288,7 +301,9 @@ def _write_credentials(project_info: dict[str, Any], data_dir: Path, logfire_api
 def parse_create_new_project(args: argparse.Namespace) -> None:
     """Create a new project."""
     data_dir = Path(args.data_dir)
-    logfire_url = args.logfire_url
+    logfire_url = _get_logfire_url(args.logfire_url, args.region)
+    if logfire_url is None:
+        _, logfire_url = LogfireCredentials._get_token_data()  # type: ignore
     project_name = args.project_name
     organization = args.org
     default_organization = args.default_org
@@ -306,7 +321,9 @@ def parse_create_new_project(args: argparse.Namespace) -> None:
 def parse_use_project(args: argparse.Namespace) -> None:
     """Use an existing project."""
     data_dir = Path(args.data_dir)
-    logfire_url = args.logfire_url
+    logfire_url = _get_logfire_url(args.logfire_url, args.region)
+    if logfire_url is None:
+        _, logfire_url = LogfireCredentials._get_token_data()  # type: ignore
     project_name = args.project_name
     organization = args.org
 
@@ -387,6 +404,13 @@ def _pretty_table(header: list[str], rows: list[list[str]]):
     return '\n'.join(lines) + '\n'
 
 
+def _get_logfire_url(logfire_url: str | None, region: str | None) -> str | None:
+    if logfire_url is not None:
+        return logfire_url
+    if region is not None:
+        return REGIONS[region]['base_url']
+
+
 class SplitArgs(argparse.Action):
     def __call__(
         self,
@@ -410,7 +434,9 @@ def _main(args: list[str] | None = None) -> None:
 
     parser.add_argument('--version', action='store_true', help='show the version and exit')
     global_opts = parser.add_argument_group(title='global options')
-    global_opts.add_argument('--logfire-url', default='', help=argparse.SUPPRESS)
+    url_or_region_grp = global_opts.add_mutually_exclusive_group()
+    url_or_region_grp.add_argument('--logfire-url', help=argparse.SUPPRESS)
+    url_or_region_grp.add_argument('--region', choices=REGIONS, help='the region to use')
     parser.set_defaults(func=lambda _: parser.print_help())  # type: ignore
     subparsers = parser.add_subparsers(title='commands', metavar='')
 
--- a/logfire/_internal/cli.py
+++ b/logfire/_internal/cli.py
@@ -24,7 +24,7 @@ from logfire.propagate import ContextCarrier, get_context
 
 from ..version import VERSION
 from .auth import DEFAULT_FILE, HOME_LOGFIRE, DefaultFile, is_logged_in, poll_for_token, request_device_code
-from .config import LogfireCredentials
+from .config import REGIONS, LogfireCredentials
 from .config_params import ParamManager
 from .tracer import SDKTracerProvider
 from .utils import read_toml_file
@@ -196,15 +196,20 @@ def parse_auth(args: argparse.Namespace) -> None:
 
     This will authenticate your machine with Logfire and store the credentials.
     """
-    logfire_url = cast(str, args.logfire_url)
+    logged_in = False
 
     if DEFAULT_FILE.is_file():
         data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
-        if is_logged_in(data, logfire_url):  # pragma: no branch
-            sys.stderr.write(f'You are already logged in. (Your credentials are stored in {DEFAULT_FILE})\n')
-            return
+        if args.logfire_url and is_logged_in(data, args.logfire_url):  # pragma: no branch
+            logged_in = True
+        elif not args.logfire_url:
+            logged_in = any(is_logged_in(data, url) for url in data['tokens'])
     else:
         data: DefaultFile = {'tokens': {}}
+
+    if logged_in:
+        sys.stderr.write(f'You are already logged in. (Your credentials are stored in {DEFAULT_FILE})\n')
+        return
 
     sys.stderr.writelines(
         (
@@ -214,6 +219,15 @@ def parse_auth(args: argparse.Namespace) -> None:
             '\n',
         )
     )
+    if not args.logfire_url:
+        # TODO input validation
+        sys.stderr.write('Logfire is available in multiple data regions. Please select one:\n')
+        for i, (region_id, region_data) in enumerate(REGIONS.items(), start=1):
+            sys.stderr.write(f'{i}. {region_id.upper()} (GCP region: {region_data["gcp_region"]})\n')
+        selected_region = int(input(f'Selected region [{"/".join(str(i) for i in range(1, len(REGIONS) + 1))}]: '))
+        logfire_url = list(REGIONS.values())[selected_region - 1]['base_url']
+    else:
+        logfire_url = cast(str, args.logfire_url)
 
     device_code, frontend_auth_url = request_device_code(args._session, logfire_url)
     frontend_host = urlparse(frontend_auth_url).netloc
--- a/logfire/_internal/cli.py
+++ b/logfire/_internal/cli.py
@@ -50,11 +50,11 @@ def parse_whoami(args: argparse.Namespace) -> None:
     """Show user authenticated username and the URL to your Logfire project."""
     data_dir = Path(args.data_dir)
     param_manager = ParamManager.create(data_dir)
-    base_url = param_manager.load_param('base_url', _get_logfire_url(args.logfire_url, args.region))
+    base_url = param_manager.load_param('base_url', args.logfire_url)
     token = param_manager.load_param('token')
-    base_url = base_url or get_base_url_from_token(token)
 
     if token:
+        base_url = base_url or get_base_url_from_token(token)
         credentials = LogfireCredentials.from_token(token, args._session, base_url)
         if credentials:
             credentials.print_token_summary()
@@ -199,15 +199,16 @@ def parse_auth(args: argparse.Namespace) -> None:
     """
     logged_in = False
 
-    logfire_url = _get_logfire_url(args.logfire_url, args.region)
+    logfire_url = args.logfire_url
     if DEFAULT_FILE.is_file():
         data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
-        if logfire_url and is_logged_in(data, logfire_url):  # pragma: no branch
-            logged_in = True
-        elif not logfire_url:
-            logged_in = any(is_logged_in(data, url) for url in data['tokens'])
     else:
         data: DefaultFile = {'tokens': {}}
+
+    if logfire_url:
+        logged_in = is_logged_in(data, logfire_url)
+    else:
+        logged_in = any(is_logged_in(data, url) for url in data['tokens'])
 
     if logged_in:
         sys.stderr.writelines(
@@ -271,8 +272,7 @@ def parse_auth(args: argparse.Namespace) -> None:
 
 def parse_list_projects(args: argparse.Namespace) -> None:
     """List user projects."""
-    logfire_url = _get_logfire_url(args.logfire_url, args.region)
-    projects = LogfireCredentials.get_user_projects(session=args._session, logfire_api_url=logfire_url)
+    projects = LogfireCredentials.get_user_projects(session=args._session, logfire_api_url=args.logfire_url)
     if projects:
         sys.stderr.write(
             _pretty_table(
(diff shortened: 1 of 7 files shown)


logfire/_internal/config.py
```diff
@@ -865,7 +878,9 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
                     # note, we only do this if `send_to_logfire` is explicitly `True`, not 'if-token-present'
                     if self.send_to_logfire is True and credentials is None:
                         credentials = LogfireCredentials.initialize_project(
-                            logfire_api_url=self.advanced.base_url,
+                            # TODO: what happens here? Seems like we need to assume a region to create the new project?
+                            # Should the SDK have the region from when the user logged in to the SDK?
+                            logfire_api_url=self.advanced.generate_base_url(None),
```
adriangb: @alexmojaki I'm not sure what the flow is here... creating a new project?

logfire/_internal/config_params.py
```diff
@@ -98,7 +98,9 @@ class _DefaultCallback:
 """Whether to enable the f-string magic feature. On by default for Python 3.11 and above."""
 IGNORE_NO_CONFIG = ConfigParam(env_vars=['LOGFIRE_IGNORE_NO_CONFIG'], allow_file_config=True, default=False, tp=bool)
 """Whether to show a warning message if logfire if used without calling logfire.configure()"""
-BASE_URL = ConfigParam(env_vars=['LOGFIRE_BASE_URL'], allow_file_config=True, default=LOGFIRE_BASE_URL)
+BASE_URL = ConfigParam(env_vars=['LOGFIRE_BASE_URL'], allow_file_config=True, default=None, tp=Union[str, None])
+REGION = ConfigParam(env_vars=['LOGFIRE_REGION'], allow_file_config=True, default='us')
```
adriangb: Maybe we also want the default to be `None` here?

I guess when a user logs in with the CLI or creates a new account prompted by the CLI they need to tell the CLI which region they want to log into, then the CLI opens the appropriate site.

logfire/_internal/config.py
Viicos: I'm having trouble understanding the use case for this code path:

https://github.com/pydantic/logfire/blob/d5cedcc125993c7818cb91fd4d38e4390ab84476/logfire/_internal/config.py#L874-L885

It runs if:
- No token was explicitly provided with `logfire.configure()`
- No credentials are available in the data directory (either an explicit one provided with `logfire.configure()`, or the default `~/.logfire/`).

Calling `initialize_project()` (indirectly) calls `_get_user_token()`:

https://github.com/pydantic/logfire/blob/d5cedcc125993c7818cb91fd4d38e4390ab84476/logfire/_internal/config.py#L1321-L1326

which looks for credentials in the default `~/.logfire/` data directory. This is guaranteed to raise here if no custom data directory was specified with `configure()`, as per the second condition above. If a custom data directory was specified, it feels weird to still look for creds in the default data dir?

It feels to me that we should instead raise an exception here instead of trying to initialize a project, as we don't have any credentials to use?

alexmojaki: @Viicos user tokens are higher level than write tokens. User tokens are created in `~/.logfire/` and are not associated with any project. They are used to create write tokens, which are required to write to a specific project. Write tokens are stored in a local `.logfire` directory, separate from the one in `~`. Yes this is confusing. But you can see it in action if you use logfire in a fresh directory.

`_get_user_token` will need to be changed to not require specifying `logfire_api_url` when there's only one URL in the file, which is usually the case. But that doesn't necessarily need to be in this PR.

logfire/_internal/config_params.py
alexmojaki: Yes I think we will need something like that. I'm not sure what the solution is, but if users have already signed up for an account and then get sent to a site where they don't have an account, that's a bad UX.

But I think we should be able to merge something that doesn't need this `REGION` param and purely helps with the case where a token is specified.

adriangb: If the token is specified / we have their credentials we should be able to derive the region from that (platform PR we worked on together).

alexmojaki: Yes but this line is about the opposite of deriving the region from the token so I think it can be removed from here.

For user credentials we can store multiple credentials in a single file for different regions, so without additional info we still won't know which region to pick.

Viicos:
I changed the logic related to regions and tokens in the last commit. The implementation in `LogfireCredentials` was already slightly convoluted so I've tried making as little changes as possible:

## `logfire.configure()`

- if no token is provided and no credentials file is present:
    - try to create a new project. To do so, we need a user token (present in `~/.logfire/default.toml`) [^1]:
      - if an _explicit_ base URL was provided in `logfire.configure()`, try to find a token for this base URL. If none, prompt the user to `logfire auth`.
      - if no _explicit_ base URL was provided, iterate over the available user tokens:
        - if only one exists, use it (this will cover most use cases, as users are expected to have only one user token).
        - if multiple tokens exists, prompt the user into selecting one.


## CLI

Up until now, the `--logfire-url` (undocumented) option defaulted to `logfire.pydantic.dev`. In this PR, it defaults to an empty string so we need to update the logic to select a correct base URL.

For most CLI commands (e.g. `logfire list`), the logic is delegated to `LogfireCredentials._get_token_data()` and so the same behavior described below applies.

For `logfire auth`, we look into the user tokens (present in `~/.logfire/default.toml`):
- if an _explicit_ `--logfire-url` was provided, try to find a token for this base URL. If it exists, exit saying the user is already logged in.
- if no _explicit_ `--logfire-url` was provided, assume the user is logged in if any of the available user tokens is still valid (and then also exits).
- else, the user is not logged in. We prompt the user into selecting a region between eu/us, and determine the appropriate base URL to use to do the device code logic.

[^1]: The following logic describes what's in `LogfireCredentials._get_token_data()`.

logfire/_internal/config.py
```diff
@@ -158,8 +177,12 @@ class ConsoleOptions:
 class AdvancedOptions:
     """Options primarily used for testing by Logfire developers."""
 
-    base_url: str = 'https://logfire-api.pydantic.dev'
-    """Root URL for the Logfire API."""
+    base_url: str | None = None
+    """Base URL for the Logfire API.
+
+    This is the explicitly configured base URL. If not set, Logfire will infer
```
```suggestion
    If not set, Logfire will infer
```

logfire/_internal/config.py
```diff
@@ -173,6 +196,17 @@ class AdvancedOptions:
...
+    def generate_base_url(self, token: str) -> str:
+        if self.base_url is not None:
+            return self.base_url
+
+        match = PYDANTIC_LOGFIRE_TOKEN_PATTERN.match(token)
+        if not match:
+            # default to US region for tokens that were created before regions were added:
+            return 'https://logfire-api.pydantic.dev'
+
+        return f'https://api-{match.group("region")}.pydantic.dev'
```
```suggestion
            # default to US region for tokens that were created before regions were added:
            region = 'us'
            if match:
                region = match.group("region")
                if region not in REGIONS:
                    region = 'us'
        
        return REGIONS[region]['base_url']
```

logfire/_internal/config.py
```diff
@@ -864,8 +898,8 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
                     # note, we only do this if `send_to_logfire` is explicitly `True`, not 'if-token-present'
                     if self.send_to_logfire is True and credentials is None:
                         credentials = LogfireCredentials.initialize_project(
-                            logfire_api_url=self.advanced.base_url,
                             session=requests.Session(),
+                            logfire_api_url=self.advanced.base_url,
```
alexmojaki: please move back to keep the diff clean

logfire/_internal/config.py
```diff
@@ -1304,11 +1339,37 @@ def from_token(cls, token: str, session: requests.Session, base_url: str) -> Sel
...
+                ones are available, or use the only one available otherwise.
+
+        Returns:
+            A two-tuple, the first element being the token and the second element being the base API URL.
+        """
         if DEFAULT_FILE.is_file():  # pragma: no branch
             data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
-            if is_logged_in(data, logfire_api_url):  # pragma: no branch
-                return data['tokens'][logfire_api_url]['token']
+            if logfire_api_url is not None and is_logged_in(data, logfire_api_url):  # pragma: no branch
```
```suggestion
            if logfire_api_url is not None and is_logged_in(data, logfire_api_url):
```

logfire/_internal/config.py
```diff
@@ -1304,11 +1339,37 @@ def from_token(cls, token: str, session: requests.Session, base_url: str) -> Sel
...
+                    choices_str = '\n'.join(
+                        f'{i}. {d["token"]} ({url})' for i, (url, d) in enumerate(tokens_list, start=1)
+                    )
+                    int_choice = IntPrompt.ask(
+                        f'Multiple user tokens found. Please select one:\n{choices_str}\n',
+                        choices=[str(i) for i in range(1, len(data['tokens']) + 1)],
+                    )
+                    choice = tokens_list[int_choice - 1]
+                    if is_logged_in(data, choice[0]):
+                        return choice[1]['token'], choice[0]
```
```suggestion
                    url, token_data = tokens_list[int_choice - 1]
                    if is_logged_in(data, url):
                        return token_data['token'], url
```

logfire/_internal/config.py
```diff
@@ -1304,11 +1339,37 @@ def from_token(cls, token: str, session: requests.Session, base_url: str) -> Sel
...
         if DEFAULT_FILE.is_file():  # pragma: no branch
             data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
-            if is_logged_in(data, logfire_api_url):  # pragma: no branch
-                return data['tokens'][logfire_api_url]['token']
+            if logfire_api_url is not None and is_logged_in(data, logfire_api_url):  # pragma: no branch
+                return data['tokens'][logfire_api_url]['token'], logfire_api_url
+            elif logfire_api_url is None:
+                tokens_list = list(data['tokens'].items())
+                if len(tokens_list) == 1 and is_logged_in(data, tokens_list[0][0]):
+                    return tokens_list[0][1]['token'], tokens_list[0][0]
```
```suggestion
                if len(tokens_list) == 1:
                    return cls._get_token_data(tokens_list[0])
```

alexmojaki (changes requested):
I think that deriving the base URL from the write token is still entirely separate from dealing with user tokens, in which case this should be split into two PRs

logfire/_internal/cli.py
```diff
@@ -276,7 +302,9 @@ def _write_credentials(project_info: dict[str, Any], data_dir: Path, logfire_api
 def parse_create_new_project(args: argparse.Namespace) -> None:
     """Create a new project."""
     data_dir = Path(args.data_dir)
-    logfire_url = args.logfire_url
+    logfire_url = _get_logfire_url(args.logfire_url, args.region)
+    if logfire_url is None:
+        _, logfire_url = LogfireCredentials._get_token_data()  # type: ignore
```
Viicos: Perhaps `_get_token_data()` should be made "public".

Viicos:
> I think that deriving the base URL from the write token is still entirely separate from dealing with user tokens, in which case this should be split into two PRs

I'm not sure this is possible as changes seems to be tied up and inter-dependent.

I'll note that the logic is starting to be quite confusing, lots of indirection is happening especially on logic related to base URL creation (for instance in some CLI commands such as `projects list`, we use either `--logfire-url`, URL from `--region`, or `LogfireCredentials._get_token_data()` which looks into `~/.logfire` tokens — this logic is necessary but hard to follow).

logfire/_internal/cli.py
```diff
@@ -198,16 +198,28 @@ def parse_auth(args: argparse.Namespace) -> None:
...
         data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
-        if is_logged_in(data, logfire_url):  # pragma: no branch
-            sys.stderr.write(f'You are already logged in. (Your credentials are stored in {DEFAULT_FILE})\n')
-            return
+        if logfire_url and is_logged_in(data, logfire_url):  # pragma: no branch
+            logged_in = True
+        elif not logfire_url:
+            logged_in = any(is_logged_in(data, url) for url in data['tokens'])
     else:
         data: DefaultFile = {'tokens': {}}
```
alexmojaki: Suggestion:

```python
logfire_url = _get_logfire_url(args.logfire_url, args.region)
if DEFAULT_FILE.is_file():
    data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
else:
    data: DefaultFile = {'tokens': {}}

if logfire_url:
    logged_in = is_logged_in(data, logfire_url)
else:
    logged_in = any(is_logged_in(data, url) for url in data['tokens'])
```

logfire/_internal/cli.py
```diff
@@ -52,8 +51,9 @@ def parse_whoami(args: argparse.Namespace) -> None:
     """Show user authenticated username and the URL to your Logfire project."""
     data_dir = Path(args.data_dir)
     param_manager = ParamManager.create(data_dir)
-    base_url = param_manager.load_param('base_url', args.logfire_url)
+    base_url = param_manager.load_param('base_url', _get_logfire_url(args.logfire_url, args.region))
     token = param_manager.load_param('token')
+    base_url = base_url or get_base_url_from_token(token)
```
alexmojaki: I'm getting an error here when I have no url or token:

```pytb
  File "/Users/alex/.pyenv/versions/3.12.6/bin/logfire", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/alex/work/logfire/logfire/_internal/cli.py", line 518, in main
    _main(args)
  File "/Users/alex/work/logfire/logfire/_internal/cli.py", line 506, in _main
    namespace.func(namespace)
  File "/Users/alex/work/logfire/logfire/_internal/cli.py", line 57, in parse_whoami
    base_url = base_url or get_base_url_from_token(token)
  File "/Users/alex/work/logfire/logfire/_internal/config.py", line 1747, in get_base_url_from_token
    if match := PYDANTIC_LOGFIRE_TOKEN_PATTERN.match(token):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'NoneType'
```

logfire/_internal/cli.py
```diff
@@ -246,7 +272,7 @@ def parse_auth(args: argparse.Namespace) -> None:
 
 def parse_list_projects(args: argparse.Namespace) -> None:
     """List user projects."""
-    logfire_url = args.logfire_url
+    logfire_url = _get_logfire_url(args.logfire_url, args.region)
```
alexmojaki: It'd be nice to not repeat this for every command.

logfire/_internal/cli.py
```diff
@@ -52,8 +51,9 @@ def parse_whoami(args: argparse.Namespace) -> None:
     """Show user authenticated username and the URL to your Logfire project."""
     data_dir = Path(args.data_dir)
     param_manager = ParamManager.create(data_dir)
-    base_url = param_manager.load_param('base_url', args.logfire_url)
+    base_url = param_manager.load_param('base_url', _get_logfire_url(args.logfire_url, args.region))
```
alexmojaki: Not for this PR, but load_param should probably be used for all commands, so that the env var `LOGFIRE_BASE_URL` always works.

logfire/_internal/cli.py
alexmojaki: `LogfireCredentials.create_new_project` already calls `_get_token_data`, which makes sense, so you shouldn't need this.

logfire/_internal/cli.py
```diff
@@ -294,7 +322,9 @@ def parse_create_new_project(args: argparse.Namespace) -> None:
 def parse_use_project(args: argparse.Namespace) -> None:
     """Use an existing project."""
     data_dir = Path(args.data_dir)
-    logfire_url = args.logfire_url
+    logfire_url = _get_logfire_url(args.logfire_url, args.region)
+    if logfire_url is None:
+        _, logfire_url = LogfireCredentials._get_token_data()  # type: ignore
```
alexmojaki: Again, shouldn't be needed.

logfire/_internal/cli.py
alexmojaki: Same for region.

logfire/_internal/cli.py
Viicos: I initially wanted to set `namespace.logfire_url = _get_logfire_url(args.logfire_url, args.region)` once parsed (and before dispatching to the subcommand callback) but thought it would be confusing as this "overrides" the explicitly provided `logfire_url`. Anyway, I went with the change.

logfire/_internal/cli.py
Viicos: The issue is I need the URL to then call `_write_credentials()` :confused: :

https://github.com/pydantic/logfire/blob/4a93a7674813c977a30709a3ab44cfea05ab5ad8/logfire/_internal/cli.py#L305-L318

logfire/_internal/config.py
```diff
@@ -1661,6 +1719,39 @@ def _get_creds_file(creds_dir: Path) -> Path:
...
+            region = 'us'
+
+    token_repr = f'{region.upper()} ({url}) - '
+    if match:
+        # new_token, include prefix and 5 chars
+        token_repr += PYDANTIC_LOGFIRE_TOKEN_TEMPLATE.format(
+            version=match.group('version'),
+            region=match.group('region'),
+            token=match.group('token')[:5],
+        )
```
```suggestion
    token_repr = f'{region.upper()} ({url}) - '
    if match := PYDANTIC_LOGFIRE_TOKEN_PATTERN.match(token):
        # new_token, include prefix and 5 chars
        token_repr += match.group('safe_part') + match.group('token')[:5]
```

with

```python
PYDANTIC_LOGFIRE_TOKEN_PATTERN = re.compile(
    r'^(?P<safe_part>pylf_v(?P<version>[0-9]+)_(?P<region>[a-z]+)_)(?P<token>[a-zA-Z0-9]+)$'
)
```

logfire/_internal/cli.py
alexmojaki: ok that doesn't need to be fixed here but this is a sign of messy logic that doesn't need to be this way, the other place where `create_new_project` gets called ends up following a similar path including `write_creds_file`

logfire/_internal/config.py
```diff
@@ -1304,11 +1334,37 @@ def from_token(cls, token: str, session: requests.Session, base_url: str) -> Sel
         )
 
     @classmethod
-    def _get_user_token(cls, logfire_api_url: str) -> str:
-        if DEFAULT_FILE.is_file():  # pragma: no branch
+    def _get_token_data(cls, logfire_api_url: str | None = None) -> tuple[str, str]:
```
```suggestion
    def _get_user_token_data(cls, logfire_api_url: str | None = None) -> tuple[str, str]:
```

alexmojaki: this class is mostly about write tokens, so i'm not even sure this method should be here, but the name and the docs really need to emphasise the difference

logfire/_internal/config.py
```diff
@@ -1304,11 +1334,37 @@ def from_token(cls, token: str, session: requests.Session, base_url: str) -> Sel
...
+                    choices_str = '\n'.join(
+                        f'{i}. {_get_token_repr(url, d["token"])}' for i, (url, d) in enumerate(tokens_list, start=1)
+                    )
+                    int_choice = IntPrompt.ask(
+                        f'Multiple user tokens found. Please select one:\n{choices_str}\n',
+                        choices=[str(i) for i in range(1, len(data['tokens']) + 1)],
+                    )
+                    url, token_data = tokens_list[int_choice - 1]
+                    if is_logged_in(data, url):
+                        return token_data['token'], url
```
```suggestion
            if logfire_api_url is None:
                tokens_list = list(data['tokens'].items())
                if len(tokens_list) == 1:
                    return cls._get_token_data(tokens_list[0][0])
                elif len(tokens_list) >= 2:
                    choices_str = '\n'.join(
                        f'{i}. {_get_token_repr(url, d["token"])}' for i, (url, d) in enumerate(tokens_list, start=1)
                    )
                    int_choice = IntPrompt.ask(
                        f'Multiple user tokens found. Please select one:\n{choices_str}\n',
                        choices=[str(i) for i in range(1, len(data['tokens']) + 1)],
                    )
                    url, token_data = tokens_list[int_choice - 1]
                    if is_logged_in(data, url):
                        return token_data['token'], url
            elif is_logged_in(data, logfire_api_url):
                return data['tokens'][logfire_api_url]['token'], logfire_api_url
```

just switching the branches

logfire/_internal/cli.py
```diff
@@ -198,16 +198,29 @@ def parse_auth(args: argparse.Namespace) -> None:
 
     This will authenticate your machine with Logfire and store the credentials.
     """
-    logfire_url = cast(str, args.logfire_url)
+    logged_in = False
```
alexmojaki:
```suggestion
```

logfire/_internal/config.py
```diff
@@ -1661,6 +1720,35 @@ def _get_creds_file(creds_dir: Path) -> Path:
     return creds_dir / CREDENTIALS_FILENAME
 
 
+def _get_token_repr(url: str, token: str) -> str:
+    region = 'us'
```
alexmojaki, line 1724: you're not using this region

logfire/_internal/cli.py
```diff
@@ -398,7 +435,9 @@ def _main(args: list[str] | None = None) -> None:
 
     parser.add_argument('--version', action='store_true', help='show the version and exit')
     global_opts = parser.add_argument_group(title='global options')
-    global_opts.add_argument('--logfire-url', default=LOGFIRE_BASE_URL, help=argparse.SUPPRESS)
+    url_or_region_grp = global_opts.add_mutually_exclusive_group()
+    url_or_region_grp.add_argument('--logfire-url', help=argparse.SUPPRESS)
+    url_or_region_grp.add_argument('--region', choices=REGIONS, help='the region to use')
```
alexmojaki, line 437: `logfire auth -h` doesn't show `--region`

logfire/_internal/cli.py
```diff
@@ -445,6 +484,7 @@ def _main(args: list[str] | None = None) -> None:
     cmd_info.set_defaults(func=parse_info)
 
     namespace = parser.parse_args(args)
+    namespace.logfire_url = _get_logfire_url(namespace.logfire_url, namespace.region)
```
alexmojaki, line 484: ok this might have been a mistake, e.g. `logfire whoami` now asks me which region i want, when i want it to tell me based on the write token

tests/conftest.py
```diff
@@ -136,8 +136,37 @@ def default_credentials(tmp_path: Path) -> Path:
...
+        [tokens."https://logfire-us.pydantic.dev"]
+        token = "pylf_v1_us_0kYhc414Ys2FNDRdt5vFB05xFx5NjVcbcBMy4Kp6PH0W"
+        expiration = "1970-01-01T00:00:00"
+        """
+    )
+    return auth_file
+
+
+@pytest.fixture
+def multiple_credentials(tmp_path: Path) -> Path:
```
alexmojaki, line 161: These two new fixtures should be used in both test_cli and test_configure

logfire/_internal/cli.py
Viicos: Yes I mentioned this in https://github.com/pydantic/logfire/pull/901#issuecomment-2749324773, after the region launch we can revisit

alexmojaki: yes but what i'm saying is that unlike what you said there this logic is *not* necessary

logfire/_internal/config.py
Viicos: I'm not sure I understand: this one is used for old tokens which doesn't contain region information

alexmojaki: my mistake.

logfire/_internal/cli.py
Viicos: `--region` is attached to the global parser, it will show up when doing `logfire --help` (and should be specified as `logfire --region eu/us auth`).

alexmojaki: Can we not improve this?

alexmojaki: anyway i guess we leave it for another PR

logfire/_internal/cli.py
Viicos: It seems to me that something is inconsistent here: the `ParamManager` instance created in the `whoami` command (that then `load_param` the token) is reading from a `pyproject.toml` file. Is it expected for a token to be present in pyproject.toml files? Should we then also call `LogfireCredentials.load_creds_file()`?

tests/conftest.py
Viicos: `test_cli` monkeypatches `_get_user_token_data()` everywhere as `_get_user_token_data()` is unit tested in `test_configure` (and these fixtures are relevant to test it). Not ideal, but this would have introduced more changes to this PR.

logfire/_internal/cli.py
Viicos: > Can we not improve this?

I can relatively easily move the two mutually exclusive `--logfire-url` and `--region` options for each command.

logfire/_internal/cli.py
alexmojaki: the token can't be in pyproject, but everything else can, and the token can be in env vars. and yes we also want load_creds_file.

tests/conftest.py
alexmojaki: ok but at least we should test that `logfire whoami` doesn't ask the user for a region when there are multiple credentials.

logfire/_internal/cli.py
Viicos: Hum so as it stands whoami _will_ try to read `token` from the pyproject. I looked and it is the only place where `param_manager.load_param('token')` is used, so I can easily replace the logic with `load_creds_file()`.

PR #904, opened 2025-03-04
Forward 'get_code' attribute in LogfireLoader

Bobronium:
Reproduction:

```shell
mkdir -p package/subpackage

touch package/subpackage/__init__.py

cat >  package/__init__.py << 'EOL'
import logfire

logfire.configure()
logfire.install_auto_tracing(modules=["package.subpackage"], min_duration=0.01)
EOL

cat >  package/subpackage/__main__.py << 'EOL'
print("Hello world")
EOL

python -m package.subpackage
```

```py
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 148, in _get_module_details
  File "<frozen runpy>", line 159, in _get_module_details
  File ".venv/lib/python3.12/site-packages/logfire/_internal/auto_trace/import_hook.py", line 131, in __getattr__
    raise AttributeError(item)
AttributeError: get_code
```

Adding `get_code` resolves the issue. However, maybe a better fix would be to forward any attribute and let it the original instance raise the `AttributeError` in case it doesn't have required attribute.

### DIFF
--- a/logfire/_internal/auto_trace/import_hook.py
+++ b/logfire/_internal/auto_trace/import_hook.py
@@ -126,6 +126,6 @@ class LogfireLoader(Loader):
 
     def __getattr__(self, item: str):
         """Forward some methods to the plain spec's loader (likely a `SourceFileLoader`) if they exist."""
-        if item in {'get_filename', 'is_package'}:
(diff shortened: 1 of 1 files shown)


alexmojaki:
Thanks for the bug report! Will release soon.

PR #905, opened 2025-03-04
Fix auto-tracing with `python -m`

alexmojaki:
Closes https://github.com/pydantic/logfire/pull/904. Simply forwarding `get_code` wouldn't auto-trace because it would use the wrong code object.


Bobronium:
Cool stuff, thank you!

PR #911, opened 2025-03-05
Release v3.7.1

alexmojaki:
Bumping version to v3.7.1.


alexmojaki:
@bllchmbrs this link check will always fail on these release PRs because we create a link in the changelog that doesn't exist until after this merges and the release is completed. What can we do?

PR #926, opened 2025-03-13
Add `logfire.instrument_pydantic_ai()`

alexmojaki:


logfire/_internal/integrations/pydantic_ai.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic_ai import Agent
+from pydantic_ai.agent import InstrumentationSettings
+from pydantic_ai.models import Model
+from pydantic_ai.models.instrumented import InstrumentedModel
```
Kludex, line 8: Don't we check if it's imported and raise error to install?

alexmojaki: We do that for `opentelemetry.instrumentation.fastapi`, not `fastapi` itself etc. There's no instrumentation package here or extra here.

Kludex (approved):

PR #932, opened 2025-03-14
Stop implicitly emitting deprecated process runtime metrics

alexmojaki:


alexmojaki:
I'll resolve merge conflicts after the base gets merged, or maybe not at all until we want to make a major release.

PR #944, opened 2025-03-21
feat(formatter): check if expression contains await

### LINKED ISSUE #942: SyntaxError when using 'await' expressions inside f-strings
### Description

When using await expressions inside f-strings passed to logfire logging methods, a SyntaxError: 'await' outside function exception is raised, causing the application to crash. This occurs despite the code being executed within an async function context.

Here's an example : 

```
import asyncio
import logfire

async def get_content():
    await asyncio.sleep(0.1)
    return "response content"

async def main():
    logfire.configure()
    # This line triggers the error
    logfire.error(f"res error: {await get_content()}")

if __name__ == "__main__":
    asyncio.run(main())
```

It produces : 
```
  ...
  File ".../formatter.py", line 387, in compile_formatted_value
    value_code = compile(source, '<fvalue1>', 'eval')
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<fvalue1>", line 1
SyntaxError: 'await' outside function
```
The issue appears to be in logfire's handling of f-strings. It attempts to extract and re-evaluate the expressions in f-strings, but when it encounters an await expression, it attempts to evaluate it outside of an async context, leading to the SyntaxError.

```TOML
logfire="3.9.0"
platform="macOS-14.3-arm64-arm-64bit-Mach-O"
python="3.13.2 (main, Feb  4 2025, 14:51:09) [Clang 16.0.0 (clang-1600.0.26.6)]"
[related_packages]
requests="2.32.3"
protobuf="5.29.4"
rich="13.9.4"
executing="2.2.0"
opentelemetry-api="1.31.1"
opentelemetry-exporter-otlp-proto-common="1.31.1"
opentelemetry-exporter-otlp-proto-http="1.31.1"
opentelemetry-instrumentation="0.52b1"
opentelemetry-proto="1.31.1"
opentelemetry-sdk="1.31.1"
opentelemetry-semantic-conventions="0.52b1"
```

Sbargaoui:
Closes #942


tests/test_formatter.py
```diff
@@ -178,3 +178,36 @@ def test_internal_exception_formatting(caplog: pytest.LogCaptureFixture):
     assert len(caplog.records) == 1
     assert caplog.records[0].message.startswith('Caught an internal error in Logfire.')
     assert str(caplog.records[0].exc_info[1]) == 'bad scrubber'  # type: ignore
+
+
+def test_await_in_fstring():
```
alexmojaki: Rather follow the approach of tests in `test_logfire.py` that check for `FormattingFailedWarning`.

logfire/_internal/formatter.py
```diff
@@ -386,6 +386,12 @@ def get_node_source_text(node: ast.AST, ex_source: executing.Source):
     This happens sometimes due to Python bugs (especially for older Python versions)
     in the source positions of AST nodes inside f-strings.
     """
+    # Check for await expressions first, before attempting to unparse
+    if isinstance(node, ast.Await):
```
alexmojaki: There could also be an await deeper within the node

Sbargaoui: Agree, this is aligned with my initial idea of checking through the depth in `compile_formatted_value` : 
```
@lru_cache
def compile_formatted_value(node: ast.FormattedValue, ex_source: executing.Source) -> tuple[str, CodeType, CodeType]:
    """Returns three things that can be expensive to compute.

    1. Source code corresponding to the node value (excluding the format spec).
    2. A compiled code object which can be evaluated to calculate the value.
    3. Another code object which formats the value.
    """
    source = get_node_source_text(node.value, ex_source)

    # Check if the expression contains await before attempting to compile
    try:
        parsed = ast.parse(source, mode='eval')
        for sub_node in ast.walk(parsed):
            if isinstance(sub_node, ast.Await):
                # Signal that we can't compile this
                raise KnownFormattingError(
                    f'Cannot evaluate await expression in f-string: {source}. Pre-evaluate the expression before logging.'
                )
    except SyntaxError:
        # For other syntax errors during parsing, let the compile step handle it
        # as it might provide better error messages
        pass
        
   ...rest of the code
```

but the python 3.8 test failed because of the unavailability of ast.unparse : https://github.com/pydantic/logfire/actions/runs/13996222305/job/39191957330?pr=944

alexmojaki: these tests should just be skipped in 3.8. but also don't parse the source, you already have an ast node.

tests/test_formatter.py
Sbargaoui: Should I test the warning instead of the exception ? What exact warning message should be expected ? Should I keep the AST validation part of the test or is testing the warning behavior sufficient ?

alexmojaki: test that `logfire.info(f'{foo(await bar)}')` still logs `f'{foo(await bar)}'` while emitting a warning

tests/test_formatter.py
```diff
@@ -178,3 +182,41 @@ def test_internal_exception_formatting(caplog: pytest.LogCaptureFixture):
...
+    try:
+        loop.run_until_complete(test_log())
+    except SyntaxError:
+        pass
+    finally:
+        loop.close()
+
+    assert len(exporter.exported_spans) == 1
+    span = exporter.exported_spans[0]
+    assert span.attributes[ATTRIBUTES_MESSAGE_KEY] == snapshot('foo(await bar())')
```
alexmojaki: It should log `content data`.

Sbargaoui: didn't you mention _test that logfire.info(f'{foo(await bar)}') still logs f'{foo(await bar)}' while emitting a warning_ ?
just to get it right, you want the evaluated expression to be logged and not the source string ?

alexmojaki: yes i should have clarified. it should give up on the magic and give the actual information it has.

logfire/_internal/formatter.py
```diff
@@ -354,6 +354,19 @@ def compile_formatted_value(node: ast.FormattedValue, ex_source: executing.Sourc
...
     source = get_node_source_text(node.value, ex_source)
+
+    # Check if the expression contains await before attempting to compile
+    try:
+        for sub_node in ast.walk(node.value):
+            if isinstance(sub_node, ast.Await):
+                raise KnownFormattingError(
+                    f'Cannot evaluate await expression in f-string: {source}. Pre-evaluate the expression before logging.'
+                )
+    except SyntaxError:
```
alexmojaki: You're not parsing so there's no reason to expect this error

logfire/_internal/formatter.py
```diff
@@ -354,6 +354,19 @@ def compile_formatted_value(node: ast.FormattedValue, ex_source: executing.Sourc
     3. Another code object which formats the value.
     """
     source = get_node_source_text(node.value, ex_source)
+
+    # Check if the expression contains await before attempting to compile
+    try:
+        for sub_node in ast.walk(node.value):
```
alexmojaki: Since you're not using `source`, move this above it.

tests/test_formatter.py
```diff
@@ -178,3 +183,62 @@ def test_internal_exception_formatting(caplog: pytest.LogCaptureFixture):
...
+    from logfire._internal.formatter import compile_formatted_value
+
+    _, value_code, formatted_code = compile_formatted_value(node, MockSource())
+
+    assert isinstance(value_code, CodeType)
+    assert isinstance(formatted_code, CodeType)
+
+
+@pytest.mark.skipif(sys.version_info[:2] == (3, 8), reason='fstring magic is only for 3.9+')
+def test_await_in_fstring(exporter: TestExporter):
```
alexmojaki:
```suggestion
@pytest.mark.anyio
async def test_await_in_fstring(exporter: TestExporter):
```

Then you can just use await and not start an event loop. See the tests in test_logfire for example.

tests/test_formatter.py
```diff
@@ -178,3 +183,62 @@ def test_internal_exception_formatting(caplog: pytest.LogCaptureFixture):
...
+            )
+
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    try:
+        loop.run_until_complete(test_log())
+        assert len(exporter.exported_spans) == 1
+        span = exporter.exported_spans[0]
+        assert span.attributes is not None
+        assert span.attributes[ATTRIBUTES_MESSAGE_KEY] == snapshot('content data')
```
alexmojaki:
```suggestion
        assert exporter.exported_spans_as_dict() == snapshot(...)
```

logfire/_internal/formatter.py
Sbargaoui: Should the try-catch behavior be kept, or should we get rid of it?

alexmojaki: get rid of it, and the test

logfire/_internal/formatter.py
Sbargaoui: the raised KnownFormattingError uses source in the error msg

alexmojaki: sorry, i missed that

tests/test_formatter.py
```diff
@@ -178,3 +181,48 @@ def test_internal_exception_formatting(caplog: pytest.LogCaptureFixture):
...
+        return x
+
+    with pytest.warns(FormattingFailedWarning) as warnings:
+        logfire.info(f'{foo(await bar())}')
+        [warning] = warnings
+        assert str(warning.message) == snapshot(
+            '\n'
+            '    Ensure you are either:\n'
+            '      (1) passing an f-string directly, with inspect_arguments enabled and working, or\n'
+            '      (2) passing a literal `str.format`-style template, not a preformatted string.\n'
```
alexmojaki: Following this advice won't help the user. KnownFormattingError says "In other words this should correspond to errors that would be raised when using `str.format`," which doesn't apply here. The general logic should be similar to KnownFormattingError but with a different exception class and warning message.

tests/test_formatter.py
```diff
@@ -178,3 +181,50 @@ def test_internal_exception_formatting(caplog: pytest.LogCaptureFixture):
...
+        [warning] = warnings
+        assert str(warning.message) == snapshot(
+            '\n'
+            '    Cannot evaluate await expression in f-string. Pre-evaluate the expression before logging.\n'
+            '    For example, change:\n'
+            '      logfire.info(f"{await get_value()}")\n'
+            '    To:\n'
+            '      value = await get_value()\n'
+            '      logfire.info(f"{value}")\n'
+            '    The problem was: Cannot evaluate await expression in f-string: foo(await bar()). Pre-evaluate the expression before logging.'
```
alexmojaki:
```suggestion
            '    The problematic f-string value was: foo(await bar())'
```

logfire/_internal/formatter.py
```diff
@@ -336,6 +336,8 @@ def logfire_format_with_magic(
         return ''.join(chunk['v'] for chunk in chunks), extra_attrs, new_template
     except KnownFormattingError as e:
         warn_formatting(str(e) or str(e.__cause__))
+    except FStringAwaitError as e:
+        warn_fstring_await(str(e) or str(e.__cause__))
```
alexmojaki:
```suggestion
        warn_fstring_await(str(e))
```

alexmojaki (approved):
Thank you!

PR #959, opened 2025-03-26
Fix tests for system package

medaminezghal:
When you install psycopg2 from system packages or pip it will give you the same package as psycopg2-binary installed from pip. This will fix the test that use system site packages.


alexmojaki:
Run `make install` to install pre-commit so that future commits are checked and fixed for formatting before pushing.

In this case, the failed lint job says how the code should look.

alexmojaki:
You can also try `make format`

tests/otel_integrations/test_httpx.py
```diff
@@ -409,7 +409,9 @@ def test_httpx_client_capture_full(exporter: TestExporter):
                     'logfire.msg': 'POST example.org/foo',
                     'http.request.header.host': ('example.org:8080',),
                     'http.request.header.accept': ('*/*',),
-                    'http.request.header.accept-encoding': ('gzip, deflate',),
+                    'http.request.header.accept-encoding': (
+                        IsAnyStr(regex="^'gzip, deflate(?:, br|, zstd|, br, zstd)?'$"),
```
alexmojaki: why?

medaminezghal: @alexmojaki if you install httpx with brotli or zstandard the tests will fails

alexmojaki: are you maintaining some packaging of logfire? do you need to run tests in some environment where the dependencies may be different from ours? because we need to be able to add new tests without having to worry about breaking your jobs and remembering to tweak snapshots like this.

medaminezghal: @alexmojaki Yes I'm maintaining logfire, pydantic-ai in Arch Linux and some tests need modification in the source in order to pass

alexmojaki:
OK I couldn't really see a reason not to merge this but I don't know how sustainable this is, it will be up to you to fix other tests in the future that fail in your environment.

I'd also like to understand the use case for a specialised package over pip. Do all the dependencies of logfire and pydantic-ai have Arch Linux packages?

medaminezghal:
@alexmojaki Every dependencies of pydantic-ai and logfire exist in Arch and other distros. The main reason for me to use the packages provided by systems, it’s easier to update in all projects, it doesn’t need to install it in every project and create their virtual environment.

Issue #961, opened 2025-03-27
Improvements to tokens/credentials management and CLI after regions changes


Viicos:
The changes in https://github.com/pydantic/logfire/commit/2073a0c60dd6145c87c1c25d41a79bc7765bdc64 introduced new logic to infer regions (and indirectly the base API URL) from the tokens.

There is now some messy and repeated logic with some levels of indirection.

## The `LogfireCredentials` class

... has a bunch of class methods that repeat logic to get the user token, set the headers, etc.

One option I think would be worth exploring is to define a proper Logfire API client, that you would be instantiated with a chosen user token (this way calling these methods don't prompt for user input when multiple tokens are available in the `default.toml` file). The class methods are then made instance methods, and you don't need to pass the `logfire_url` everytime.

## The CLI

- I think we first need to decide if we keep `rich`. Imo I think it is worth the extra dependency: colors are nice and hard to manage manually, [prompts](https://rich.readthedocs.io/en/stable/prompt.html) are also painful to implement manually. The current tables we have could be better rendered:

![Image](https://github.com/user-attachments/assets/65540b85-a70d-4fd3-a678-ee55965679e0)

 - We currently write to `stderr`, and unless there's something I'm not aware of we should use `stdout` (or `print()` directly). Because some paths of the CLI make use of rich prompts, we also currently have a mix of `stdin` (rich) and `stderr` (us).

- The `--region`/`--logfire-url` options are set on the global parser, meaning users need to use `logfire --region us auth`. I think this is fine, but doing `logfire auth --help` doesn't show the global options' help, which can be confusing (for reference, `uv`/`ruff`'s CLIs show them on subcommand help, which I think is better UX — I'm not sure which rust crate is used to do so).

- It was discussed that the `whoami` command should show more info. Probably we can:
  - ignore the `--logfire-url`/`--region` top level options, and always display all the available user tokens in `default.toml` (and optionally if they are expired).
  - Show information about the source of the read/write token (env var, `logfire_credentials.json` file)

alexmojaki:
Out of those 4 bullet points, the last 2 are more important.

PR #981, opened 2025-04-03
Refactor user tokens, introduce Logfire client

Viicos:
This is a refactor in preparation of https://github.com/pydantic/logfire/issues/961.

I'm introducing a Logfire client, so that we can use it in the various `LogfireCredentials` methods without passing the base URL and session around. A proper user token structure is also introduced (which will also be useful for the whoami command improvements).

Early PR before adapting test to get some early feeback.


logfire/_internal/auth.py
```diff
@@ -26,12 +37,133 @@ class UserTokenData(TypedDict):
...
+    def add_token(self, base_url: str, token: UserTokenData) -> UserToken:
+        existing_token = next((t for t in self.user_tokens if t.base_url == base_url), None)
+        if existing_token:
+            token_index = self.user_tokens.index(existing_token)
+            self.user_tokens.remove(existing_token)
+        else:
+            token_index = len(self.user_tokens)
+
+        user_token = UserToken.from_user_token_data(base_url, token)
+        self.user_tokens.insert(token_index, user_token)
```
Viicos: This isn't really ideal logic, perhaps I should have the internal `UserTokenCollection` structure matching the toml file (i.e. mapping of url->token)?

logfire/_internal/cli.py
```diff
@@ -299,42 +302,41 @@ def _write_credentials(project_info: dict[str, Any], data_dir: Path, logfire_api
 def parse_create_new_project(args: argparse.Namespace) -> None:
     """Create a new project."""
     data_dir = Path(args.data_dir)
-    logfire_url = args.logfire_url
-    if logfire_url is None:  # pragma: no cover
-        _, logfire_url = LogfireCredentials._get_user_token_data()  # type: ignore
+    logfire_url: str | None = args.logfire_url
+    user_token = default_token_collection().get_token(logfire_url)
+    client = LogfireClient(user_token)
```
Viicos: Can also be added on the namespace to avoid repetition:

```python
namespace.get_client = lambda: LogfireClient(default_token_collection().get_token(namespace.logfire_url))
```

logfire/_internal/auth.py
```diff
@@ -26,12 +37,133 @@ class UserTokenData(TypedDict):
     expiration: str
 
 
-class DefaultFile(TypedDict):
-    """Content of the default.toml file."""
+class TokensFileData(TypedDict):
```
alexmojaki:
```suggestion
class UserTokensFileData(TypedDict):
```

logfire/_internal/auth.py
```diff
@@ -26,12 +37,133 @@ class UserTokenData(TypedDict):
...
+    token: str
+    base_url: str
+    expiration: str
+
+    @classmethod
+    def from_user_token_data(cls, base_url: str, token: UserTokenData) -> Self:
+        return cls(
+            token=token['token'],
+            base_url=base_url,
+            expiration=token['expiration'],
```
alexmojaki, line 75:
```suggestion
            base_url=base_url, **token
```

Not sure if the extra class method is helpful.

logfire/_internal/auth.py
```diff
@@ -26,12 +37,133 @@ class UserTokenData(TypedDict):
...
+
+    def __str__(self) -> str:
+        # TODO define in this module?
+        from .config import PYDANTIC_LOGFIRE_TOKEN_PATTERN, REGIONS
+
+        region = 'us'
+        if match := PYDANTIC_LOGFIRE_TOKEN_PATTERN.match(self.token):
+            region = match.group('region')
+            if region not in REGIONS:
+                region = 'us'
```
alexmojaki: i think this should be a separate method/function

logfire/_internal/auth.py
```diff
@@ -26,12 +37,133 @@ class UserTokenData(TypedDict):
...
+        return token_repr
+
+
+@dataclass
+class UserTokenCollection:
+    """A collection of user tokens."""
+    user_tokens: list[UserToken]
+
+    @classmethod
+    def from_tokens(cls, tokens: TokensFileData) -> Self:
```
alexmojaki:
```suggestion
    def from_file_data(cls, file_data: TokensFileData) -> Self:
```

logfire/_internal/auth.py
alexmojaki: Yes that sounds better

logfire/_internal/cli.py
```diff
@@ -197,16 +204,14 @@ def parse_auth(args: argparse.Namespace) -> None:
...
     """
-    logfire_url = args.logfire_url
+    logfire_url: str | None = args.logfire_url
+
     if DEFAULT_FILE.is_file():
-        data = cast(DefaultFile, read_toml_file(DEFAULT_FILE))
+        tokens_collection = default_token_collection()
     else:
-        data: DefaultFile = {'tokens': {}}
+        tokens_collection = UserTokenCollection.from_tokens({'tokens': {}})
```
alexmojaki: Should be a method on UserTokenCollection

logfire/_internal/auth.py
```diff
@@ -26,12 +37,133 @@ class UserTokenData(TypedDict):
...
+                token = self.user_tokens[int_choice - 1]
+            else:  # self.user_tokens == []
+                raise LogfireConfigError('No user tokens are available. Please run `logfire auth` to authenticate.')
+
+        if token.is_expired:
+            raise LogfireConfigError(f'User token {token} is expired. Pleas run `logfire auth` to authenticate.')
+        return token
+
+    def is_logged_in(self, base_url: str | None = None) -> bool:
+        if base_url is not None:
```
alexmojaki, line 174: The `base_url is None` case seems weird to me

logfire/_internal/cli.py
alexmojaki: or something like `LogfireClient.from_url(args.logfire_url)`

logfire/_internal/client.py
```diff
@@ -0,0 +1,102 @@
...
+        response = self._session.post(urljoin(self.base_url, endpoint), json=body)
+        UnexpectedResponse.raise_for_status(response)
+        return response
+
+    def get_user_organizations(self) -> list[dict[str, Any]]:
+        """Get the organizations of the logged-in user."""
+        try:
+            response = self._get('/v1/organizations')
+        except UnexpectedResponse as e:
+            raise LogfireConfigError('Error retrieving list of organizations') from e
```
alexmojaki: Looks like it should be something like this:

```suggestion
        response = self._get('/v1/organizations')
        LogfireConfigError.raise_for_status(response, 'Error retrieving list of organizations')
```

logfire/_internal/auth.py
```diff
@@ -26,12 +37,133 @@ class UserTokenData(TypedDict):
...
+        )
+
+    @property
+    def is_expired(self) -> bool:
+        return datetime.now(tz=timezone.utc) >= datetime.fromisoformat(self.expiration.rstrip('Z')).replace(
+            tzinfo=timezone.utc
+        )
+
+    def __str__(self) -> str:
+        # TODO define in this module?
```
alexmojaki: I think so, yes

alexmojaki (commented):
General idea looks good

logfire/_internal/cli.py
Viicos: You mean something like `UserTokenCollection.empty()`?

alexmojaki: yes

logfire/_internal/auth.py
Viicos: I agree it can feel weird, but the intent is to filter which tokens to consider from the collection. This is used in the CLI `auth` command, where the base URL is optional.

logfire/_internal/client.py
Viicos: Actually I think the client should raise a proper `ClientError` exception or similar, which should then be reraised as a `LogfireConfigError`. Because the client is only used during configuration, I left it as is for simplicity (this can be changed if the client gets used outside configuration logic at some point). However, I don't think the `LogfireConfigError` should be aware of HTTP errors as this will make it harder to change in the future if — as I mentioned — we want a proper client exception.

tests/test_configure.py
```diff
@@ -1445,92 +1442,6 @@ def test_load_creds_file_invalid_key(tmp_path: Path):
         LogfireCredentials.load_creds_file(creds_dir=tmp_path)
 
 
-def test_get_user_token_data_explicit_url(default_credentials: Path):
```
Viicos, line 1452: Moved to `test_auth.py`, and thanks to the refactor they do not require `mock.patch`ing anymore.

logfire/_internal/auth.py
```diff
@@ -1,37 +1,203 @@
...
+        tokens_list = list(self.user_tokens.values())
+        if base_url is not None:
+            token = next((t for t in tokens_list if t.base_url == base_url), None)
+            if token is None:
+                raise LogfireConfigError(
+                    f'No user token was found matching the {base_url} Logfire URL. '
+                    'Please run `logfire auth` to authenticate.'
+                )
+        else:
+            if len(tokens_list) == 1:
```
alexmojaki:
```suggestion
        elif len(tokens_list) == 1:
```

logfire/_internal/auth.py
```diff
@@ -1,37 +1,203 @@
...
+                int_choice = IntPrompt.ask(
+                    f'Multiple user tokens found. Please select one:\n{choices_str}\n',
+                    choices=[str(i) for i in range(1, len(tokens_list) + 1)],
+                )
+                token = tokens_list[int_choice - 1]
+            else:  # tokens_list == []
+                raise LogfireConfigError('No user tokens are available. Please run `logfire auth` to authenticate.')
+
+        if token.is_expired:
+            raise LogfireConfigError(f'User token {token} is expired. Pleas run `logfire auth` to authenticate.')
```
alexmojaki:
```suggestion
            raise LogfireConfigError(f'User token {token} is expired. Please run `logfire auth` to authenticate.')
```

logfire/_internal/auth.py
```diff
@@ -1,37 +1,203 @@
...
+                available, it will be used, otherwise the user will be prompted to choose
+                a token.
+
+        Raises:
+            LogfireConfigError: If no user token is found (no token matched the base URL,
+                the collection is empty, or the selected token is expired).
+        """
+        tokens_list = list(self.user_tokens.values())
+        if base_url is not None:
+            token = next((t for t in tokens_list if t.base_url == base_url), None)
```
alexmojaki:
```suggestion
            token = self.user_tokens.get(base_url)
```

logfire/_internal/auth.py
```diff
@@ -1,37 +1,203 @@
...
+        if base_url is not None:
+            tokens = (t for t in self.user_tokens.values() if t.base_url == base_url)
+        else:
+            tokens = self.user_tokens.values()
+        return any(not t.is_expired for t in tokens)
+
+    def add_token(self, base_url: str, token: UserTokenData) -> UserToken:
+        """Add a user token to the collection."""
+        user_token = UserToken.from_user_token_data(base_url, token)
+        self.user_tokens[base_url] = UserToken.from_user_token_data(base_url, token)
```
alexmojaki:
```suggestion
        self.user_tokens[base_url] = user_token = UserToken.from_user_token_data(base_url, token)
```

logfire/_internal/auth.py
```diff
@@ -1,37 +1,203 @@
...
+        # There's no standard library package to write TOML files, so we'll write it manually.
+        with path.open('w') as f:
+            for base_url, user_token in self.user_tokens.items():
+                f.write(f'[tokens."{base_url}"]\n')
+                f.write(f'token = "{user_token.token}"\n')
+                f.write(f'expiration = "{user_token.expiration}"\n')
+
+
+@cache
+def default_token_collection() -> UserTokenCollection | None:
```
alexmojaki: Is this worth caching? When is it accessed multiple times? Isn't there a risk of missing updates?

Why doesn't it return the empty collection by default?

logfire/_internal/cli.py
```diff
@@ -256,22 +262,19 @@ def parse_auth(args: argparse.Namespace) -> None:
...
     sys.stderr.write('Successfully authenticated!\n')
 
-    # There's no standard library package to write TOML files, so we'll write it manually.
-    with DEFAULT_FILE.open('w') as f:
-        for url, info in data['tokens'].items():
-            f.write(f'[tokens."{url}"]\n')
-            f.write(f'token = "{info["token"]}"\n')
-            f.write(f'expiration = "{info["expiration"]}"\n')
-
+    tokens_collection.dump(DEFAULT_FILE)
```
alexmojaki: IMO add_token itself should call `dump`

logfire/_internal/cli.py
```diff
@@ -61,12 +68,17 @@ def parse_whoami(args: argparse.Namespace) -> None:
             credentials.print_token_summary()
             return
 
-    current_user = LogfireCredentials.get_current_user(session=args._session, logfire_api_url=base_url)
-    if current_user is None:
+    token_collection = default_token_collection() or UserTokenCollection.empty()
+
+    try:
+        client = LogfireClient.from_url(base_url, token_collection=token_collection)
```
alexmojaki:
```suggestion
    try:
        client = LogfireClient.from_url(base_url)
```

seems that this is implied

logfire/_internal/cli.py
```diff
@@ -300,42 +303,39 @@ def _write_credentials(project_info: dict[str, Any], data_dir: Path, logfire_api
 def parse_create_new_project(args: argparse.Namespace) -> None:
     """Create a new project."""
     data_dir = Path(args.data_dir)
-    logfire_url = args.logfire_url
-    if logfire_url is None:  # pragma: no cover
-        _, logfire_url = LogfireCredentials._get_user_token_data()  # type: ignore
+    logfire_url: str | None = args.logfire_url
+    client = LogfireClient.from_url(logfire_url)
```
alexmojaki:
```suggestion
    client = LogfireClient.from_url(args.logfire_url)
```

prevent accidentally using `logfire_url`

logfire/_internal/auth.py
Viicos: > Why doesn't it return the empty collection by default?

No real reason, it is just a matter of being more explicit at the caller site (`default_token_collection() or UserTokenCollection.empty()`) vs convenience here. Because the default token collection to use throughout the application needs to be empty anyway when there's no file, I think I can return an empty collection directly.

Regarding caching, performance-wise it doesn't really matter but I think it could be surprising to have two different calls returning a different collection, even when the `DEFAULT_FILE` didn't change. There's a risk of missing updates, but at the same time I don't know if having it updating at runtime is desirable?

alexmojaki: i'm more worried about the cache being stale somehow

logfire/_internal/cli.py
Viicos: What if the token collection isn't created from a file? Do we always dump to `DEFAULT_FILE`? (I know right now no such thing is happening and we always use `~/.logfire/default.toml` as a reference, but this might change at some point?).

logfire/_internal/cli.py
Viicos: With the change in `default_token_collection()` now yes indeed.

logfire/_internal/client.py
```diff
@@ -0,0 +1,120 @@
...
+        UnexpectedResponse.raise_for_status(response)
+        return response
+
+    def get_user_organizations(self) -> list[dict[str, Any]]:
+        """Get the organizations of the logged-in user."""
+        try:
+            response = self._get('/v1/organizations/')
+        except UnexpectedResponse as e:
+            raise LogfireConfigError('Error retrieving list of organizations') from e
+        return response.json()
```
alexmojaki:
```suggestion
        return self._get('/v1/organizations/', description='retrieving list of organizations')
```

_get and _post can be refactored to allow writing this in most cases

logfire/_internal/client.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self, user_token: UserToken) -> None:
+        if user_token.is_expired:
+            raise RuntimeError('The provided user token is expired')
+        self.base_url = user_token.base_url
+        self._token = user_token.token
+        self._session = Session()
+        self._session.headers.update({'Authorization': self._token, 'User-Agent': UA_HEADER})
+
+    @classmethod
+    def from_url(cls, base_url: str | None, token_collection: UserTokenCollection | None = None) -> Self:
```
alexmojaki: Is a non-default token collection ever used?

tests/test_auth.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_get_user_token_explicit_url(default_credentials: Path) -> None:
+    token_collection = UserTokenCollection.from_tokens_file(default_credentials)
+
+    # https://logfire-us.pydantic.dev is the URL present in the default credentials fixture:
+    token = token_collection.get_token(base_url='https://logfire-us.pydantic.dev')
+    assert token.base_url == 'https://logfire-us.pydantic.dev'
+
+    with pytest.raises(LogfireConfigError):
```
alexmojaki: These are a good place for https://15r10nk.github.io/inline-snapshot/latest/extra/#inline_snapshot.extra.raises

logfire/_internal/cli.py
alexmojaki: Token collections can keep track of the associated file path if you want, and just always set it to DEFAULT_FILE. They should always be created from a file.

logfire/_internal/client.py
Viicos: Unfortunately `create_new_project()` needs access to the `UnexpectedResponse` exception before raising `LogfireConfigError`. I can fetch it from the `__cause__`, wdyt?

tests/test_cli.py
```diff
@@ -307,6 +308,8 @@ def new_find_spec(name: str) -> ModuleSpec | None:
 def test_auth(tmp_path: Path, webbrowser_error: bool, capsys: pytest.CaptureFixture[str]) -> None:
     auth_file = tmp_path / 'default.toml'
     with ExitStack() as stack:
+        stack.enter_context(patch('logfire._internal.auth.DEFAULT_FILE', auth_file))
+        # Necessary to assert that credentials where written to the `auth_file` (which happens from the `cli` module)
```
alexmojaki:
```suggestion
        # Necessary to assert that credentials are written to the `auth_file` (which happens from the `cli` module)
```

tests/test_cli.py
```diff
@@ -411,6 +414,8 @@ def test_auth_on_authenticated_user(default_credentials: Path, capsys: pytest.Ca
 def test_auth_no_region_specified(tmp_path: Path) -> None:
     auth_file = tmp_path / 'default.toml'
     with ExitStack() as stack:
+        stack.enter_context(patch('logfire._internal.auth.DEFAULT_FILE', auth_file))
+        # Necessary to assert that credentials where written to the `auth_file` (which happens from the `cli` module)
```
alexmojaki:
```suggestion
        # Necessary to assert that credentials are written to the `auth_file` (which happens from the `cli` module)
```

logfire/_internal/client.py
Viicos: No, I will remove it (although this can be added back later whenever we want to allow using custom token files).

tests/test_cli.py
```diff
@@ -452,8 +457,10 @@ def test_projects_list(default_credentials: Path, capsys: pytest.CaptureFixture[
     with ExitStack() as stack:
         stack.enter_context(
             patch(
-                'logfire._internal.config.LogfireCredentials._get_user_token_data',
-                return_value=('', 'https://logfire-us.pydantic.dev'),
+                'logfire._internal.auth.UserTokenCollection.get_token',
```
alexmojaki, line 460: it's not new but i don't love the patching being at this level

alexmojaki: also this really needs to be refactored, making the same change like this so many times is a red flag

logfire/_internal/client.py
alexmojaki: there's many ways to do it, e.g. make a lower level `_post_raw` method

tests/test_cli.py
Viicos: Agree, this refactor eliminated some use of patching, but not all. Related to the above, this can probably be solved by allowing a custom `UserTokenCollection` (e.g. by providing a different path than `DEFAULT_FILE`) to be provided.

logfire/_internal/client.py
Viicos: Imo this adds more indirection but I won't fight it

logfire/_internal/cli.py
Viicos: Yes I think we can add this file-backed constraint, and just use a tmpfile in tests if we need to instantiate a collection.

Viicos: Proposed a change in the last commit

logfire/_internal/auth.py
```diff
@@ -100,25 +100,30 @@ def __str__(self) -> str:
...
+    def empty(cls, path: Path) -> Self:
+        """Create an empty user token collection.
+
+        Args:
+            path: The path where the user tokens will be stored.
+        """
+        inst = cls.__new__(cls)
+        inst.user_tokens = {}
+        inst.path = path
+        return inst
```
Viicos: Alternatively, keep the dataclass-generated `__init__()`, and re-add the classmethod constructors (but then we have a public `__init__()` that shouldn't be used because `path` and `user_tokens` shouldn't be provided separately).

logfire/_internal/auth.py
```diff
@@ -1,37 +1,211 @@
...
+
+    user_tokens: dict[str, UserToken]
+    """A mapping between base URLs and user tokens."""
+
+    path: Path
+    """The path where the user tokens are stored."""
+
+    def __init__(self, path: Path) -> None:
+        self.path = path
+        data = cast(UserTokensFileData, read_toml_file(path))
```
alexmojaki:
```suggestion
    def __init__(self, path: Path = DEFAULT_FILE) -> None:
        self.path = path
        try:  # prefer try/except over checking if file exists - EAFP
            data = cast(UserTokensFileData, read_toml_file(path))
        except ...:
            data = {}
```

Then `default_token_collection()` can be replaced by `UserTokenCollection()`, and `empty` can be removed.

logfire/_internal/auth.py
```diff
@@ -1,37 +1,198 @@
...
+
+    def __init__(self, path: Path | None = None) -> None:
+        # FIXME: we can't set the default value of `path` to `DEFAULT_FILE`, otherwise
+        # `mock.patch()` doesn't work:
+        self.path = path if path is not None else DEFAULT_FILE
+        try:
+            data = cast(UserTokensFileData, read_toml_file(self.path))
+        except FileNotFoundError:
+            data: UserTokensFileData = {}
+        self.user_tokens = {url: UserToken(base_url=url, **data) for url, data in data.get('tokens', {}).items()}
```
alexmojaki: don't reuse the name `data`

alexmojaki (approved):
Thanks!

logfire/_internal/auth.py
Viicos:
```suggestion
        self.user_tokens = {url: UserToken(base_url=url, **token) for url, token in data.get('tokens', {}).items()}
```

logfire/_internal/utils.py
```diff
@@ -167,6 +171,8 @@ def span_to_dict(span: ReadableSpan) -> ReadableSpanDict:
 class UnexpectedResponse(RequestException):
     """An unexpected response was received from the server."""
 
+    response: Response  # type: ignore (guaranteed to exist)
```
Teo-coder296, line 174: It is guaranteed to be set

PR #983, opened 2025-04-07
Return None from LogfireSpan.context when logfire could not be imported

DouweM:
Fixes https://github.com/pydantic/pydantic-ai/issues/1399


alexmojaki:
Thanks, merging since this looks somewhat urgent, but opened https://github.com/pydantic/logfire/issues/989 as a followup to improve

PR #1001, opened 2025-04-12
Add codespell to pre-commit configuration

karta9821:
Integrate codespell to prevent future typos.


Kludex:
Thanks for the PR @karta9821 , but I don't think this is necessary in this project.

As @alexmojaki said, we have hyperlint already, but more than that, I prefer to keep the minimal set of tools in the repository.

PR #1027, opened 2025-04-21
Fix exporting of very large spans and payloads

alexmojaki:
Replaces https://github.com/pydantic/logfire/pull/1026

- Measures the OTLP trace export body size before compression. This should prevent an edge case where the SDK could send a highly compressed payload of many spans that the backend would reject for being too big. Instead it will split the payload in two as usual.
- When a single span exceeds the SDK's 5MB limit, just send it.
  - The backend has since increased its limit, so we shouldn't have been dropping it. This way the limits don't have to stay in sync.
  - The SDK was supposed to be logging something in this case about the dropped span. But it was doing so with logfire, inside a context where instrumentation is suppressed. So the data was being dropped silently. Now, when the backend rejects a large payload that was otherwise not even being sent, OTel logs the following using stdlib logging: `Failed to export batch code: 413, reason: Failed to buffer the request body: length limit exceeded`. We should try to make this message clearer some other time.

### DIFF
--- a/logfire/_internal/exporters/otlp.py
+++ b/logfire/_internal/exporters/otlp.py
@@ -197,6 +197,8 @@ class RetryFewerSpansSpanExporter(WrapperSpanExporter):
             return super().export(spans)
         except BodyTooLargeError:
             half = len(spans) // 2
+            # BodySizeCheckingOTLPSpanExporter should only raise BodyTooLargeError for >1 span,
+            # otherwise it should just try exporting it.
             assert half > 0
             res1 = self.export(spans[:half])
             res2 = self.export(spans[half:])
--- a/logfire/_internal/exporters/otlp.py
+++ b/logfire/_internal/exporters/otlp.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import contextlib
 import random
 import time
 import uuid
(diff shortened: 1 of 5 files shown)


logfire/_internal/exporters/otlp.py
```diff
@@ -20,23 +20,30 @@
...
+    max_body_size = 5 * 1024 * 1024
 
-    Also defers failed requests to a DiskRetryer.
-    """
+    def _serialize_spans(self, spans: Sequence[ReadableSpan]) -> bytes:
+        result = super()._serialize_spans(spans)  # type: ignore
+        if len(spans) > 1 and len(result) > self.max_body_size:
+            # Tell outer RetryFewerSpansSpanExporter to split in half
+            raise BodyTooLargeError(len(result), self.max_body_size)
+        return result
```
adriangb, line 40: My only piece of feedback is that this could be expensive and that instead of re-serializing we could transform to protobuf and then manipulate the in-memory representation of the protobuf since that's cheap to manipulate and cheap to calculate the size of objects and their children.

alexmojaki: Ideally yes, but not really worth it now.

adriangb: ok

PR #1047, opened 2025-04-30
forward preferred settings in `WrappingMetricExporter`

davidhewitt:
This seemed like a reasonable cleanup, reading `_preferred_temporality` in this fashion is exactly what `PeriodicExportingMetricReader` does.

### DIFF
--- a/logfire/_internal/exporters/wrapper.py
+++ b/logfire/_internal/exporters/wrapper.py
@@ -6,8 +6,7 @@ from typing import Any, Sequence, cast
 from opentelemetry import context
 from opentelemetry.sdk._logs import LogData, LogRecordProcessor
 from opentelemetry.sdk._logs.export import LogExporter, LogExportResult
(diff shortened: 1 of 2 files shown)


logfire/_internal/exporters/wrapper.py
```diff
@@ -36,10 +35,11 @@ class WrapperMetricExporter(MetricExporter):
     def __init__(
         self,
         exporter: MetricExporter,
-        preferred_temporality: dict[type, AggregationTemporality] | None = None,
-        preferred_aggregation: dict[type, Aggregation] | None = None,
     ) -> None:
-        super().__init__(preferred_temporality=preferred_temporality, preferred_aggregation=preferred_aggregation)  # type: ignore
+        super().__init__(  # type: ignore
+            preferred_temporality=exporter._preferred_temporality,
```
alexmojaki, line 40: Reading protected attributes doesn't really feel better

davidhewitt: Fair enough, I'll just update comments as part of #1048

Issue #1042, opened 2025-04-29
Exporting to logfire via an otel collector


davidhewitt:
### Description

We have a feature request (for the Rust SDK) to export to the logfire UI via an otel collector as a proxy.

The otel collector in this case is used to strip certain metadata for client applications before sending them to logfire, however there are any number of reasonable use-cases for a collector (e.g. double-writing to multiple projects, etc)

As of https://github.com/pydantic/logfire/issues/298#issuecomment-2271415494 it's possible to turn off the logfire exporter and send to an arbitrary OTLP (http) endpoint instead. This can be used to send data to a collector, however the data doesn't include logfire-specific metadata so is strictly worse if using the collector to then forward to logfire.

I think we could probably add an additional environment variable to make this configuration possible. I can think of at least two (similar) possible forms (either name can be bikeshed, whatever is preferred):
- `LOGFIRE_OTEL_ENDPOINT=<some endpoint>` - if set, then we would change the main logfire exporter to send to this endpoint instead. This is different from `LOGFIRE_BASE_URL` because we wouldn't be able to do token health checking on this endpoint. (We may not even have a valid token as only the collector needs it.)
- `LOGFIRE_EXPORT_METADATA_TO_ALL_PROCESSORS=1` - if set, we would send the full logfire metadata to all span processors (so with this plus `LOGFIRE_SEND_TO_LOGFIRE=0` and `OTEL_EXPORTER_OTLP_ENDPOINT=<some endpoint>` we send all data over otel)

PR #1048, opened 2025-04-30
allow `LOGFIRE_BASE_URL` to accept `grpc` endpoints

davidhewitt:
Related to #1042 

The idea (from offline discussion) is that we can set `LOGFIRE_BASE_URL` to a grpc endpoint and everything can "just work", with the caveat that we wouldn't be able to do any token checking.

This is sufficient for a local test script to work for me:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "logfire[grpc]",
# ]
# ///

import os

import logfire

os.environ['LOGFIRE_BASE_URL'] = 'grpc://localhost:4317'

logfire.configure(
    # Setting a service name is good practice in general, but especially
    # important for Jaeger, otherwise spans will be labeled as 'unknown_service'
    service_name='my_logfire_service',
)

with logfire.span('This is a span'):
    logfire.info('Logfire logs are also actually just spans!')
```

----

Steps to complete before merge:
 - [ ] Update documentation for OTEL collectors
 - [ ] Testing - not sure how much we want this covered? It seems painful to start up a full GRPC endpoint in testing, maybe there's a mock solution?

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -972,9 +972,6 @@ class LogfireConfig(_LogfireConfigData):
                             endpoint=base_url,
                             headers=headers,
                             compression=GrpcCompression.Gzip,
-                            # I'm pretty sure that this line here is redundant,
-                            # and that passing it to the QuietMetricExporter is what matters
-                            # because the PeriodicExportingMetricReader will read it from there.
                             preferred_temporality=METRICS_PREFERRED_TEMPORALITY,
                         )
                         log_exporter = GrpcOTLPLogExporter(
@@ -995,9 +992,6 @@ class LogfireConfig(_LogfireConfigData):
                             headers=headers,
                             session=session,
                             compression=Compression.Gzip,
-                            # I'm pretty sure that this line here is redundant,
-                            # and that passing it to the QuietMetricExporter is what matters
-                            # because the PeriodicExportingMetricReader will read it from there.
                             preferred_temporality=METRICS_PREFERRED_TEMPORALITY,
                         )
                         log_exporter = OTLPLogExporter(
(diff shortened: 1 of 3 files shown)


logfire/_internal/config.py
```diff
@@ -923,14 +935,70 @@ def check_token():
...
-                    session.headers.update(headers)
-                    span_exporter = BodySizeCheckingOTLPSpanExporter(
-                        endpoint=urljoin(base_url, '/v1/traces'),
-                        session=session,
-                        compression=Compression.Gzip,
-                    )
+                    headers['authorization'] = self.token
+                elif self.send_to_logfire is True and self.advanced.base_url is not None:
+                    # We may not need a token if we are sending to a custom
+                    # base URL
```
alexmojaki: We usually do though, let's only make this change for GRPC for now.

logfire/_internal/config.py
```diff
@@ -893,21 +893,33 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
...
+                                logfire_api_url=self.advanced.base_url,
+                                session=requests.Session(),
+                            )
+                            credentials.write_creds_file(self.data_dir)
+
+                        if credentials is not None:
+                            self.token = credentials.token
+                            self.advanced.base_url = self.advanced.base_url or credentials.logfire_api_url
+                    except LogfireConfigError:
+                        if self.advanced.base_url is not None:
```
Viicos: I think we should only allow this if it is a grpc url?

Alternatively, perhaps we could split the logic, and have different code paths depending on whether the endpoint is HTTP/gRPC

dmontagu:
Closing this draft as stale (conflicts + failing CI, untouched for a year). #1042 stays open to track collector/OTLP export.

PR #1066, opened 2025-05-09
Export first batch of spans more quickly

alexmojaki:
For the first 10 spans, use a delay of 100ms, then increase it to 500ms after. This makes the initial experience of the SDK more responsive.


logfire/_internal/exporters/dynamic_batch.py
```diff
@@ -0,0 +1,32 @@
...
+        self.final_delay = float(os.environ.get(OTEL_BSP_SCHEDULE_DELAY) or 500)
+        # Start with the configured value immediately if it's less than 100ms.
+        initial_delay = min(self.final_delay, 100)
+        super().__init__(BatchSpanProcessor(exporter, schedule_delay_millis=initial_delay))
+        self.num_processed = 0
+
+    def on_end(self, span: ReadableSpan) -> None:
+        self.num_processed += 1
+        if self.num_processed == 10:
+            assert isinstance(self.processor, BatchSpanProcessor)
```
adriangb: What happens if it's not? We just fail... Might as well force it to be a `BatchSpanProcessor` in `__init__`?

logfire/_internal/exporters/dynamic_batch.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    def __init__(self, exporter: SpanExporter) -> None:
+        self.final_delay = float(os.environ.get(OTEL_BSP_SCHEDULE_DELAY) or 500)
+        # Start with the configured value immediately if it's less than 100ms.
+        initial_delay = min(self.final_delay, 100)
+        super().__init__(BatchSpanProcessor(exporter, schedule_delay_millis=initial_delay))
+        self.num_processed = 0
+
+    def on_end(self, span: ReadableSpan) -> None:
```
adriangb, line 28: This is a hot path. Can we rig it so when it hits `self.num_processed == 10` it does `self.on_end = BatchSpanProcessor.on_end`?

adriangb (approved):

logfire/_internal/exporters/dynamic_batch.py
alexmojaki: we do, with `super().__init__(BatchSpanProcessor(exporter, schedule_delay_millis=initial_delay))`

this is for the type checker

adriangb: hmm okay. can you leave a comment saying as much?

logfire/_internal/exporters/dynamic_batch.py
alexmojaki: yes but i worry the benefit isn't really noticeable and it could make things harder to understand and debug

PR #1096, opened 2025-05-29
🐛 Fix verbose console formatting for enum types (#1073)

### LINKED ISSUE #1073: Fix verbose console formatting for some types
```python
import datetime
import decimal
from enum import Enum, IntEnum, StrEnum

import logfire


class MyEnum(StrEnum):
    foo = 'foo'


class MyEnum2(IntEnum):
    foo = 3


class MyEnum3(Enum):
    foo = 3


logfire.info(
    'hi',
    d=datetime.datetime.now(),
    v=decimal.Decimal('1.0'),
)
logfire.info(
    'hi',
    e1=MyEnum.foo,
    e2=MyEnum2.foo,
    e3=MyEnum3.foo,
)
```

Output:

```
11:07:08.675 hi
             │ scratch_1499.py:23 info
             │ d=2025-05-14T13:07:08.652694
             │ v=1.0
11:07:08.675 hi
             │ scratch_1499.py:24 info
             │ e1=<MyEnum.foo: 'foo'>
             │ e2=MyEnum2(<MyEnum2.foo: 3>)
             │ e3=MyEnum3(3)
```

All the values should look like reprs. The enums are an inconsistent mess.

Following up on https://github.com/pydantic/logfire/issues/1063

Previously these raised errors or did other nonsense, that was fixed in https://github.com/pydantic/logfire/pull/1071 and https://github.com/pydantic/logfire/pull/1072

sbhrwlr:
Fixes #1073 
**Change:**
Explicitly handled Enums by returning logfire_json_dumps(value), which ensures consistent output across all Enum types.

**How I Tested:**
Manually tested with all Enum types (`Enum`, `IntEnum`, `StrEnum`, `Flag`, `IntFlag`). Please validate the output below for each Enum variant.
Screenshot 2025-05-29 at 12 08 59 PM


alexmojaki:
Thanks! Please extend `test_other_json_schema_types` to cover enums.

Also note that this PR doesn't close the issue as decimals and datetimes are still incorrect.

sbhrwlr:
Screenshot 2025-05-29 at 10 57 19 PM
@alexmojaki Changes include a fix, format updates in tests, and new test cases for different enum types. Let me know if anything needs attention.

tests/test_json_args.py
```diff
@@ -978,8 +978,8 @@ class MyPydanticDataclass:
             'code.lineno': 123,
             'code.function': 'test_log_non_scalar_complex_args',
             'a': 1,
-            'complex_list': '["a",1,{"x":"x","y":"2023-01-01T00:00:00"},{"t":10},{"p":20}]',
-            'complex_dict': '{"k1":"v1","model":{"x":"x","y":"2023-01-01T00:00:00"},"dataclass":{"t":10},"pydantic_dataclass":{"p":20}}',
+            'complex_list': '["a",1,{"x":"x","y":"datetime.datetime(2023, 1, 1, 0, 0)"},{"t":10},{"p":20}]',
```
alexmojaki: This part can't change. The actual attributes should contain the kind of data you'd expect to find in JSON. Python reprs should only be in the console.

sbhrwlr: You're right, this behavior stems from Pydantic's data serialization approach. Refer `json_encoder.py` method `_pydantic_model_encoder`

### The Problem

When using the default `model_dump()`, Pydantic preserves Python objects in their native form:

```python
from pydantic import BaseModel
from fractions import Fraction
from datetime import datetime

class MyModel(BaseModel):
   x: Fraction
   y: datetime

model = MyModel(x=Fraction(1, 3), y=datetime.now())
```

Serialization behaviour comparison

```python
# Default mode - preserves Python objects
model.model_dump()
# Output: {'x': '1/3', 'y': datetime.datetime(2025, 5, 30, 0, 29, 36, 610411)}

# JSON mode - converts to JSON-serializable types
model.model_dump(mode="json")
# Output: {'x': '1/3', 'y': '2025-05-30T00:29:36.610411'}
```

### The Challenge

Using `model_dump(mode="json")` would fix serialization consistency but breaks the `test_recursive_objects` test. 
so should we use `model_dump(mode="json")` and change `test_recursive_objects`  accordingly?

alexmojaki: No, it's not about pydantic models. We always want datetimes to be represented as ISO strings in the attributes, even if they're at the top level.

sbhrwlr: @alexmojaki This behavior should be consistent across similar cases. For instance, I tested using `Fraction(3, 4)` at the root level. Instead of getting `'3/4'` as the output, the value returned was `Fraction(3, 4)` — which is its `repr()`.

alexmojaki: It just so happens that Pydantic dumps fractions to strings (which TBH I find odd in Python mode) and that logfire doesn't have any handling for fractions. This isn't intentional. It would make sense for logfire to always produce `3/4` too.

sbhrwlr:
@alexmojaki Tried these changes:
- **Introduced dedicated methods**: `_format_date`, `_format_datetime`, `_format_time`, selected via `x-python-datatype` in the schema.
- **Updated formatter** to parse string values using ISO format methods or use the object directly if already the correct type.
- **Updated tests** to ensure all edge cases and input types are handled.

Screenshot 2025-05-30 at 8 10 22 PM

alexmojaki:
Please check that it still looks right in the logfire UI, the x-python-datatype might change things.

sbhrwlr:
# Logfire UI vs Console Output Comparison

## Test Code
```python
logfire.info(
    'sample test new with datetime, date, time, and decimal',
    dt=datetime(2025, 5, 30, 12, 34, 56, 789000),
    da=date(2025, 5, 30),
    ti=time(12, 34, 56, 789000),
    dec=Decimal(1.0)
)
```

## Output Comparison

### Local Console Output
Local Console Screenshot

### Logfire UI Output  
Logfire UI Screenshot

## Key Observations
The screenshots demonstrate how the same logfire entry with datetime, date, time, and decimal values appears differently between the local console output and the Logfire web UI interface.

alexmojaki:
Thanks!

PR #1098, opened 2025-05-30
Update docs about query API and alerts

alexmojaki:


docs/how-to-guides/query-api.md
```diff
@@ -144,7 +144,7 @@ client library, such as `requests` in Python. Below are the general steps and an
 
 ### General Steps to Make a Direct HTTP Request
 
-1. **Set the Endpoint URL**: The base URL for the Logfire API is `https://logfire-api.pydantic.dev`.
+1. **Set the Endpoint URL**: The base URL for the Logfire API is `https://logfire-us.pydantic.dev` for accounts in the US region, and `https://logfire-eu.pydantic.dev` for accounts in the EU region.
```
Kludex, line 147: Can't we infer this for the user somehow?

alexmojaki: This code isn't using the SDK. We could maybe do something in the backend but we chose not to do that for writing data.

Kludex (approved):

Issue #1107, opened 2025-06-02
Pydantic RootModel not rendered correctly


alexmojaki:
```python
from pydantic import BaseModel, RootModel

import logfire

logfire.configure()


class InnerModel(BaseModel):
    name: str


class MyModel(BaseModel):
    child: InnerModel


MyRootModel = RootModel[MyModel]

r = MyRootModel(child=InnerModel(name='test'))
print(repr(r))  # RootModel[MyModel](root=MyModel(child=InnerModel(name='test')))
logfire.info('MyRootModel', r1=r, r2=r.root)
```

<img width="249" alt="Image" src="https://github.com/user-attachments/assets/d416a1b3-de93-4781-a308-c80238f96b21" />

<img width="240" alt="Image" src="https://github.com/user-attachments/assets/e0918999-dc5e-4a9a-b907-ad6b6bd9bd93" />

The data is the same and we shouldn't change that, but the JSON schemas differ which affects their appearance:

```json
{
  "type": "object",
  "properties": {
    "r": {
      "type": "object",
      "title": "RootModel[MyModel]",
      "x-python-datatype": "PydanticModel",
      "properties": {
        "root": {
          "type": "object",
          "title": "MyModel",
          "x-python-datatype": "PydanticModel",
          "properties": {
            "child": {
              "type": "object",
              "title": "InnerModel",
              "x-python-datatype": "PydanticModel"
            }
          }
        }
      }
    },
    "root": {
      "type": "object",
      "title": "MyModel",
      "x-python-datatype": "PydanticModel",
      "properties": {
        "child": {
          "type": "object",
          "title": "InnerModel",
          "x-python-datatype": "PydanticModel"
        }
      }
    }
  }
}
```

In particular `r1.child` should show as an `InnerModel`.

nagarajpoojari404:
May be we can remove the `root` field from `RootModel` type.
my solution seems bit hacky.

I am new to this repo, I appreciate any help

dhruv-ahuja:
@alexmojaki can I take this up if its not being actively worked on?

alexmojaki:
Yes, https://github.com/pydantic/logfire/pull/1108 looks abandoned, but it's worth taking a look at the discussion there.

PR #1108, opened 2025-06-02
fix: unwrap root field from RootModel schema to render correctly

nagarajRPoojari:
Fixes: #1107 

Issue: `_pydantic_model_schema` keeps `root` field of `RootModel` type resulting in incorrect rendering
Sol: recursively unwrap root field if object if `RootModel` type.


alexmojaki:
What does it look like when you run the script in the issue?

nagarajRPoojari:
Screenshot 2025-06-03 at 6 15 27 AM
Screenshot 2025-06-03 at 6 15 57 AM


I assume, this is what we are expecting.

nagarajpoojari404:
Screenshot 2025-06-03 at 6 56 52 AM
Screenshot 2025-06-03 at 6 57 03 AM


I assume, this is what we were expecting

logfire/_internal/json_schema.py
```diff
@@ -285,14 +285,21 @@ def _pydantic_model_schema(obj: Any, seen: set[int]) -> JsonDict:
     import pydantic
 
     assert isinstance(obj, pydantic.BaseModel)
+
+    class_name = obj.__class__.__name__
+
+    while isinstance(obj, pydantic.RootModel):
+        obj = obj.root  # type: ignore
```
alexmojaki:
```suggestion
    if isinstance(obj, pydantic.RootModel):
        return create_json_schema(obj.root, seen)
```

Also please add a test using a RootModel where the root is not a pydantic model (e.g. a list) which should fail without this change.

alexmojaki: ah, i see you were trying to pass the original class name. i guess just give up on that, it doesn't really matter to not have the `RootModel[]` since we don't have the `root=` anyway.

tests/test_json_args.py
```diff
@@ -1238,6 +1238,41 @@ class Model:
...
+                'attributes': {
+                    'logfire.span_type': 'log',
+                    'logfire.level_num': 9,
+                    'logfire.msg_template': 'hi',
+                    'logfire.msg': 'hi',
+                    'code.filepath': 'test_json_args.py',
+                    'code.function': 'test_pydantic_root_model_objects',
+                    'code.lineno': 123,
+                    'r': '{"child":"hi"}',
+                    'r2': '"test"',
```
alexmojaki, line 1264: Ah, we have a problem here. this should just be `'test'`. This PR isn't changing that, but `"r2":{}` in the schema means that the backend won't parse it as JSON and it will keep the extra quotes. Can you also change the JSON encoding code to match?

nagarajpoojari404: it occurs when we `json.dump` primitive types (e.g `RootModel[str]` ). 
We can't handle it while json encoding,  as we need to know its schema ( whether it is `{}` or not ).

may be we can try to convert `"' test '" ` -> `" test "` if `schema = {}` 
```python

safe_strip_json_quotes = lambda s: json.loads(s) if isinstance(s, str) and s.startswith('"') and s.endswith('"') else s
for key, value in json.loads(otlp_attributes[ATTRIBUTES_JSON_SCHEMA_KEY])['properties'].items():
       if not value:
       otlp_attributes[key] = safe_strip_json_quotes(otlp_attributes[key])         
```                    
    
 this sol kind a hacky, 
 need some input

alexmojaki: Understood. Then we need to ensure that the JSON schema is nontrivial.

alexmojaki (changes requested):
JSON schema mustn't have `"r2":{}`

alexmojaki:
https://github.com/pydantic/logfire/issues/1107 is done now

Issue #1102, opened 2025-05-31
Fixed a TODO in the httpx integration, should I create a PR ?


peter-avg:
### Question

Hello! 

I was working on #1059 when I came across a TODO in the httpx integration (logfire/integrations/httpx.py) made by @Kludex that was waiting on this [merge](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3098/). I believe I have fixed it, should I create a PR ? 

Here is what was requested as a TODO:
<img width="1055" alt="Image" src="https://github.com/user-attachments/assets/40da69ed-c8a4-41a2-a7ad-eea048072ad0" />

Issue #1059, opened 2025-05-07
Instrumentation with urllib3


steinwaywhw:
### Description

Would love to see integrations with urllib3. 

This is used in at least one OpenAPI auto-generated sync client. The specific one I'm using is a client for OpenFGA. For context, their auto-generated asynchronous client is based on aiohttp. 

For now I'm directly using `opentelemetry.instrumentation.urllib3`. But having it directly from LogFire would be even better.

peter-avg:
Would it be okay if I got involved with this ?

Br1an67:
Hi @peter-avg — are you still working on this? I noticed it's been a while since the last update. If not, I'd be happy to pick it up.

Added `logfire.instrument_urllib3()` following the existing `instrument_*` pattern, wrapping the OpenTelemetry urllib3 instrumentation.

PR: #1744

peter-avg:
Go right ahead! @Br1an67

G-ChandraSekhar:
Hi! I'd like to pick this up.

I noticed the two earlier attempts (#1744 and #2037) were both closed by their authors without
merging, so the SDK side is still missing. My plan is to add `logfire.instrument_urllib3()`
modeled directly on `instrument_requests()` — a thin wrapper around OTel's
`URLLib3Instrumentor` with `excluded_urls` / `request_hook` / `response_hook` / `url_filter`,
plus the `urllib3` extra, the `INSTRUMENTATION_TO_EXTRA` entry so `logfire run` suggests
`logfire[urllib3]`, docs, and VCR-based tests.

I'll open a PR shortly. Let me know if you'd like anything done differently.

Update: PR is up — #2449. Happy to address any feedback

PR #1109, opened 2025-06-02
updated httpx integration, patches #1102

peter-avg:
References #1102. I deleted the branch by mistake and so I am reopening the pull request.

## Context
I was working on https://github.com/pydantic/logfire/issues/1059 when I came across a TODO in the httpx integration (logfire/integrations/httpx.py) made by @Kludex that was waiting on this [merge](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3098/). The merged happened so I went ahead and made the changes.

## How I tested it
I ran `uv run pytest -s tests/otel-integrations/test_httpx.py` and all tests passed.

## Types of Changes
- [x] Code Improvement
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation

### DIFF
--- a/logfire/_internal/integrations/httpx.py
+++ b/logfire/_internal/integrations/httpx.py
@@ -148,11 +148,11 @@ def instrument_httpx(
             )
 
         tracer_provider = final_kwargs['tracer_provider']
-        instrumentor.instrument_client(client, tracer_provider, request_hook, response_hook)  # type: ignore[reportArgumentType]
+        instrumentor.instrument_client(client, tracer_provider, request_hook, response_hook)
 
 
 class LogfireHttpxInfoMixin:
-    headers: httpx.Headers
+    headers: httpx.Headers | None
 
     @property
     def content_type_header_object(self) -> ContentTypeHeader:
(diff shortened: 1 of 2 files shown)


logfire/integrations/httpx.py
```diff
@@ -1,44 +1,10 @@
...
-    headers: httpx.Headers
-    stream: httpx.SyncByteStream | httpx.AsyncByteStream | None
-    extensions: dict[str, Any] | None
-
-
-RequestHook = Callable[[Span, RequestInfo], None]
-ResponseHook = Callable[[Span, RequestInfo, ResponseInfo], None]
-AsyncRequestHook = Callable[[Span, RequestInfo], Awaitable[None]]
-AsyncResponseHook = Callable[[Span, RequestInfo, ResponseInfo], Awaitable[None]]
+from opentelemetry.instrumentation.httpx import (
```
alexmojaki, line 3: Need to be careful that these only get imported when type checking to not break for older versions of the instrumentation library.

logfire/_internal/integrations/httpx.py
```diff
@@ -148,26 +148,30 @@ def instrument_httpx(
             )
 
         tracer_provider = final_kwargs['tracer_provider']
-        instrumentor.instrument_client(client, tracer_provider, request_hook, response_hook)  # type: ignore[reportArgumentType]
+        instrumentor.instrument_client(client, tracer_provider, request_hook, response_hook)
 
 
 class LogfireHttpxInfoMixin:
-    headers: httpx.Headers
+    headers: httpx.Headers | None
```
alexmojaki, line 155: Rather make this an empty Headers object than let it be None. Definitely don't raise errors.

logfire/integrations/httpx.py
peter-avg: tests break with `if TYPE_CHECKING:`.

alexmojaki: if you just put `if TYPE_CHECKING:` here then I expect breakage because another file imports from here. But what if you do the same for that file? Are these types used at runtime?

peter-avg: I believe so. From what I understand they are used at the httpx integration at runtime and they are also used when type checking in main.

alexmojaki: ok, i see now. then I think this isn't actually worth it unless/until the httpx integration doesn't work with older versions of the otel library for other reasons.

peter-avg: Understood!

alexmojaki:
Closing since I think we can't actually make this change for a while based on discussion

Issue #608, opened 2024-11-19
Minor bug: `logfire.configure(token="")` actually tries to use an empty token


alexmojaki:
Probably need to do something a bit smarter, e.g. fallback to other config methods, or warn about the empty token without actually checking it against `/v1/info`.

PR #1110, opened 2025-06-03
Check empty tokens before making connection attempt

nagarajRPoojari:
fixes: #608 

sol: 
simple condition to check valid tokens before making any api call for connection.
if condition fails fallback to other method to load creds

P.S:  we can handle basic set of validations for a token ( apart from empty, None)

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -910,7 +910,6 @@ class LogfireConfig(_LogfireConfigData):
                         thread = Thread(target=check_token, name='check_logfire_token')
                         thread.start()
 
-                    assert self.token is not None
                     base_url = self.advanced.generate_base_url(self.token)
                     headers = {'User-Agent': f'logfire/{VERSION}', 'Authorization': self.token}
                     session = OTLPExporterHttpSession()


alexmojaki:
I think `logfire.configure(token=None)` and `logfire.configure(token='')` should probably be equivalent, i.e. both fallback to the environment variable as the next option, and I don't think this does that.

> P.S: we can handle basic set of validations for a token ( apart from empty, None)

No I think this is fine, which also means you can just use `if token` and drop the assert.

nagarajpoojari404:
> I think `logfire.configure(token=None)` and `logfire.configure(token='')` should probably be equivalent, i.e. both fallback to the environment variable as the next option, and I don't think this does that.
> 
> > P.S: we can handle basic set of validations for a token ( apart from empty, None)
> 
> No I think this is fine, which also means you can just use `if token` and drop the assert.

It works fine for both `None` & `''`.
And reason for using `is_valid_token` instead of just `if token` is to handle `''` from env as well

alexmojaki:
Please add a test with `configure(token=''` combined with a non-empty env-var LOGFIRE_TOKEN and make sure it uses the env var.

logfire/_internal/config.py
```diff
@@ -904,7 +907,7 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
                         self.token = credentials.token
                         self.advanced.base_url = self.advanced.base_url or credentials.logfire_api_url
 
-                if self.token is not None:
+                if is_valid_token(self.token):
```
```suggestion
                if self.token:
```

then the assert shouldn't be needed

nagarajpoojari404: yea,  earlier I thought abt handling `''` token from cred file also, it doesn't make sense.
committing

logfire/_internal/config.py
```diff
@@ -878,9 +878,12 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
                 credentials: LogfireCredentials | None = None
                 show_project_link: bool = self.console and self.console.show_project_link or False
 
+                def is_valid_token(token: str | None) -> bool:
+                    return token is not None and token != ''
+
```
```suggestion
```

logfire/_internal/config.py
```diff
@@ -878,9 +878,12 @@ def add_span_processor(span_processor: SpanProcessor) -> None:
...
                 # try loading credentials (and thus token) from file if a token is not already available
                 # this takes the lowest priority, behind the token passed to `configure` and the environment variable
-                if self.token is None:
+                if not is_valid_token(token=self.token):
```
```suggestion
                if not self.token:
```

logfire/_internal/config.py
```diff
@@ -910,6 +913,7 @@ def check_token():
                         thread = Thread(target=check_token, name='check_logfire_token')
                         thread.start()
 
+                    assert self.token is not None
```
```suggestion
```

alexmojaki:
Thanks!

Issue #492, opened 2024-10-10
Propagate tags to child spans


LumaC0:
### Description

A `logfire.span` argument that allows child spans and records to inherit tags. All child spans (a relationship that looks to be defined by index in `OPEN_SPANS`)  would inherit tags from surrounding spans that have `propagate_tags=True`. Spans that set the argument to False (the default) do not pass tags to child spans.

**Proposal**

```python

logfire.config()

with logfire.span("first span", _tags=["one"], _propagate_tags=True): # has tag ["one"]
    logfire.info("first") # has tag ["one"]

    with logfire.span("second span", _tags=["two"]): # has tags ["one", "two"]
        logfire.info("second", _tags=["t1"]) # has tag ["one", "t1"]

        with logfire.span("third span", _tags=["three"], _propagate_tags=True) # has tags ["one", "three"]
            logfire.info("third") # has tags ["one", "three"]

**Comment by alexmojaki** [comment]
Try `logfire.with_tags`

**Comment by alexmojaki** [comment]
Can't you define the instance in another file that both file1 and file2 import?

**Comment by LumaC0** [comment]
I could hack this together in each of my projects.

(I would instantiate a Logfire instance (not using with_tags) in file3 and just keep updating its internal _tags attribute) 

this is fine but it's not really a long term solution.

**Comment by alexmojaki** [comment]
OK, I guess you're using tags in a way that's more dynamic than what `with_tags` was intended for.

I would recommend creating a context variable and passing the value to `_tags` rather than modifying a global shared logfire instance. But I understand that this is still inconvenient. We should create an API that automates this, without requiring opening a span, and that also propagates things like attributes and levels.

**Comment by parashardhapola** [comment]
I usually define a convenience function in a logger config file and then use it as needed across the code base:

```python
def get_tagged_logfire(tags: list[str] | None = None) -> logfire.Logfire :
    if tags is None:
        tags = []
    if CONFIG.LOGFIRE_API_KEY:
        logfire.configure(
            token=CONFIG.LOGFIRE_API_KEY,
        )
    logger.configure(handlers=[logfire.loguru_handler()])
    
    tlg = logfire.with_tags(*tags)
    tlg.instrument_pydantic_ai()
    return tlg
```

However, instruments don't consume these tags; only direct logfire statements will inherit these.

alexmojaki:
[Baggage](https://logfire.pydantic.dev/docs/reference/advanced/baggage/) allows propagating attributes like this, but not tags yet.

[...]

PR #1128, opened 2025-06-06
Add baggage utilities

dmontagu:
Adds:
* A convenience argument (`add_baggage_to_attributes: bool = True`) to `logfire.configure` for adding a BaggageSpanProcessor which will convert baggage into attributes.
* Utility functions for setting and retrieving baggage (which affect propagated otel baggage even if `add_baggage_to_attributes` is `False`)

Shout out to https://github.com/mpfaffenberger/code_puppy, which I used for initial implementation.

### DIFF
--- a/logfire-api/logfire_api/_internal/baggage.pyi
+++ b/logfire-api/logfire_api/_internal/baggage.pyi
@@ -1,29 +1,25 @@
 from _typeshed import Incomplete
 from collections.abc import Iterator
 from contextlib import contextmanager
+from opentelemetry import context
+from opentelemetry.sdk.trace import Span, SpanProcessor
 
 __all__ = ['get_baggage', 'set_baggage']
 
 get_baggage: Incomplete
 
 @contextmanager
-def set_baggage(**bag: str) -> Iterator[None]:
+def set_baggage(**values: str) -> Iterator[None]:
     '''Context manager that attaches key/value pairs as OpenTelemetry baggage to the current context.
 
-    All values in `bag` must be strings, as OpenTelemetry baggage only supports string values.
+    See the [Baggage documentation](https://logfire.pydantic.dev/docs/reference/advanced/baggage/) for more details.
 
-    OpenTelemetry baggage is a way to propagate metadata across service boundaries in a distributed system, and is
-    included in headers of outgoing requests for which context propagation is configured.
-    This is intended to be used to propagate arbitrary context (like user ids, task ids, etc.) down to all nested spans.
-
-    Baggage is not _automatically_ converted into attributes on descendant spans, but this is a common usage pattern.
-    If you want baggage to be converted into attributes, use `logfire.configure(add_baggage_to_attributes=True)`.
-
-    Note: this function should always be used as a context manager; if you try to open and close it manually you may
+    Note: this function should always be used in a `with` statement; if you try to open and close it manually you may
     run into surprises because OpenTelemetry Baggage is stored in the same contextvar as the current span.
 
     Args:
-        bag: The key/value pairs to attach to baggage.
+        values: The key/value pairs to attach to baggage. These should not be large or sensitive.
+            Strings longer than 1000 characters will be truncated with a warning.
 
     Example usage:
 
@@ -37,3 +33,9 @@ def set_baggage(**bag: str) -> Iterator[None]:
             ...
     ```
     '''
+
+class NoForceFlushSpanProcessor(SpanProcessor):
+    def force_flush(self, timeout_millis: int = 30000) -> bool: ...
+
+class DirectBaggageAttributesSpanProcessor(NoForceFlushSpanProcessor):
+    def on_start(self, span: Span, parent_context: context.Context | None = None) -> None: ...
--- a/logfire-api/logfire_api/_internal/config.pyi
+++ b/logfire-api/logfire_api/_internal/config.pyi
@@ -23,6 +23,7 @@ from .utils import SeededRandomIdGenerator as SeededRandomIdGenerator, Unexpecte
 from _typeshed import Incomplete
 from collections.abc import Sequence
 from dataclasses import dataclass, field
+from logfire._internal.baggage import DirectBaggageAttributesSpanProcessor as DirectBaggageAttributesSpanProcessor
 from logfire.exceptions import LogfireConfigError as LogfireConfigError
 from logfire.sampling import SamplingOptions as SamplingOptions
 from logfire.sampling._tail_sampling import TailSamplingProcessor as TailSamplingProcessor
@@ -95,7 +96,7 @@ class CodeSource:
 
 class DeprecatedKwargs(TypedDict): ...
 
-def configure(*, local: bool = False, send_to_logfire: bool | Literal['if-token-present'] | None = None, token: str | None = None, service_name: str | None = None, service_version: str | None = None, environment: str | None = None, console: ConsoleOptions | Literal[False] | None = None, config_dir: Path | str | None = None, data_dir: Path | str | None = None, additional_span_processors: Sequence[SpanProcessor] | None = None, metrics: MetricsOptions | Literal[False] | None = None, scrubbing: ScrubbingOptions | Literal[False] | None = None, inspect_arguments: bool | None = None, sampling: SamplingOptions | None = None, add_baggage_to_attributes: bool = False, code_source: CodeSource | None = None, distributed_tracing: bool | None = None, advanced: AdvancedOptions | None = None, **deprecated_kwargs: Unpack[DeprecatedKwargs]) -> Logfire:
+def configure(*, local: bool = False, send_to_logfire: bool | Literal['if-token-present'] | None = None, token: str | None = None, service_name: str | None = None, service_version: str | None = None, environment: str | None = None, console: ConsoleOptions | Literal[False] | None = None, config_dir: Path | str | None = None, data_dir: Path | str | None = None, additional_span_processors: Sequence[SpanProcessor] | None = None, metrics: MetricsOptions | Literal[False] | None = None, scrubbing: ScrubbingOptions | Literal[False] | None = None, inspect_arguments: bool | None = None, sampling: SamplingOptions | None = None, add_baggage_to_attributes: bool = True, code_source: CodeSource | None = None, distributed_tracing: bool | None = None, advanced: AdvancedOptions | None = None, **deprecated_kwargs: Unpack[DeprecatedKwargs]) -> Logfire:
     """Configure the logfire SDK.
 
     Args:
@@ -143,7 +144,8 @@ def configure(*, local: bool = False, send_to_logfire: bool | Literal['if-token-
             Defaults to `True` if and only if the Python version is at least 3.11.
 
         sampling: Sampling options. See the [sampling guide](https://logfire.pydantic.dev/docs/guides/advanced/sampling/).
-        add_baggage_to_attributes: if True, any OpenTelemetry Baggage will be added to spans as attributes
+        add_baggage_to_attributes: Set to `False` to prevent OpenTelemetry Baggage from being added to spans as attributes.
+            See the [Baggage documentation](https://logfire.pydantic.dev/docs/reference/advanced/baggage/) for more details.
         code_source: Settings for the source code of the project.
         distributed_tracing: By default, incoming trace context is extracted, but generates a warning.
             Set to `True` to disable the warning.
@@ -176,12 +178,13 @@ class _LogfireConfigData:
     scrubbing: ScrubbingOptions | Literal[False]
     inspect_arguments: bool
     sampling: SamplingOptions
+    add_baggage_to_attributes: bool
     code_source: CodeSource | None
     distributed_tracing: bool | None
     advanced: AdvancedOptions
 
 class LogfireConfig(_LogfireConfigData):
-    def __init__(self, send_to_logfire: bool | Literal['if-token-present'] | None = None, token: str | None = None, service_name: str | None = None, service_version: str | None = None, environment: str | None = None, console: ConsoleOptions | Literal[False] | None = None, config_dir: Path | None = None, data_dir: Path | None = None, additional_span_processors: Sequence[SpanProcessor] | None = None, metrics: MetricsOptions | Literal[False] | None = None, scrubbing: ScrubbingOptions | Literal[False] | None = None, inspect_arguments: bool | None = None, sampling: SamplingOptions | None = None, add_baggage_to_attributes: bool = False, code_source: CodeSource | None = None, distributed_tracing: bool | None = None, advanced: AdvancedOptions | None = None) -> None:
+    def __init__(self, send_to_logfire: bool | Literal['if-token-present'] | None = None, token: str | None = None, service_name: str | None = None, service_version: str | None = None, environment: str | None = None, console: ConsoleOptions | Literal[False] | None = None, config_dir: Path | None = None, data_dir: Path | None = None, additional_span_processors: Sequence[SpanProcessor] | None = None, metrics: MetricsOptions | Literal[False] | None = None, scrubbing: ScrubbingOptions | Literal[False] | None = None, inspect_arguments: bool | None = None, sampling: SamplingOptions | None = None, add_baggage_to_attributes: bool = True, code_source: CodeSource | None = None, distributed_tracing: bool | None = None, advanced: AdvancedOptions | None = None) -> None:
         """Create a new LogfireConfig.
 
         Users should never need to call this directly, instead use `logfire.configure`.


docs/reference/advanced/baggage.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Why use Baggage to set attributes?
+
+Though tracing makes it a lot easier to understand and visualize the relationship between parent and child spans,
+it's still usually easiest to express filtering criteria on an _individual_ span when doing custom searches,
+aggregations, etc.
+
+As a concrete example, you might have received an issue report from a user with user_id `scolvin`, and want to
+search for all spans corresponding to slow database queries that happened in requests made by that user. You might find
+all slow database requests with the query `duration > '1 second' AND attributes ? 'db.statement'`, but trying to filter
```
alexmojaki:
```suggestion
all slow database requests with the query `duration > 1 AND attributes ? 'db.statement'`, but trying to filter
```

docs/reference/advanced/baggage.md
```diff
@@ -0,0 +1,66 @@
...
+Note that the `update_baggage` contextmanager will always update the OpenTelemetry Baggage used during propagation,
+regardless of whether you configure logfire to add the Baggage to span attributes.
+
+!!! info "Baggage and distributed tracing"
+    Baggage is propagated automatically during distributed tracing, but in order for the propagated baggage to be
+    added as _attributes_ in other services, you need to ensure a `BaggageSpanProcessor` is configured in those services.
+
+    If using the `logfire` library from Python, that is as simple as calling `logfire.configure(add_baggage_to_attributes=True)`,
+    but you can accomplish this with the opentelemetry SDK in many other languages as well. You can find more information
+    in the official [OpenTelemetry Baggage documentation]https://opentelemetry.io/docs/concepts/signals/baggage/).
```
alexmojaki:
```suggestion
    in the official [OpenTelemetry Baggage documentation](https://opentelemetry.io/docs/concepts/signals/baggage/).
```

docs/reference/advanced/baggage.md
```diff
@@ -0,0 +1,66 @@
...
+
+```python
+import logfire
+logfire.configure(add_baggage_to_attributes=True)
+```
+
+This will ensure that any OpenTelemetry Baggage is added as attributes to the spans handled by logfire.
+(Under the hood, this is just adding a [`BaggageSpanProcessor`](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/processor/opentelemetry-processor-baggage) to the list of additional span processors.)
+
+Once you've done this, you can use the [`update_baggage`][logfire.update_baggage] contextmanager to contextually update the Baggage (which will
```
alexmojaki: Might as well also mention get_baggage

logfire-api/logfire_api/_internal/baggage.pyi
alexmojaki, line 1: FYI you don't need to update .pyi files in PRs, it happens automatically during release.

tests/test_baggage.py
```diff
@@ -0,0 +1,119 @@
...
+    ],
+)
+def test_baggage_goes_to_span_attributes(
+    config_kwargs: dict[str, Any],
+    exporter: TestExporter,
+    add_baggage_to_attributes: bool,
+    expected: list[dict[str, Any]],
+):
+    # Use the default TailSamplingOptions.level of 'notice'.
+    # Set duration to None to not include spans with a long duration.
```
alexmojaki:
```suggestion
```

tests/test_baggage.py
```diff
@@ -0,0 +1,119 @@
...
+    logfire.configure(**config_kwargs, add_baggage_to_attributes=add_baggage_to_attributes)
+    with logfire.update_baggage({'a': 1}):
+        with logfire.span('outer'):
+            with logfire.update_baggage({'b': 2}):
+                with logfire.span('outer-middle'):
+                    with logfire.update_baggage({'a': 3}):
+                        with logfire.span('inner-middle'):
+                            # confirm the behavior of attributes that conflict with baggage:
```
alexmojaki: We might actually want to put all baggage under a special attribute or prefix so that conflicts aren't a problem. Even if we don't, attributes set directly on the span should definitely take priority.

docs/reference/advanced/baggage.md
alexmojaki, line 1: The docs need to mention distributed tracing, in particular the fact that all instrumented HTTP requests will have baggage attached as headers.

tests/test_baggage.py
alexmojaki, line 1: Need to test what happens in case of something sensitive, e.g. a baggage value named 'secret'. Not sure what the behaviour should be, but either way needs testing.

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     "opentelemetry-sdk >= 1.21.0, < 1.35.0",
     "opentelemetry-exporter-otlp-proto-http >= 1.21.0, < 1.35.0",
     "opentelemetry-instrumentation >= 0.41b0",
+    "opentelemetry-processor-baggage >= 0.41b0",
```
alexmojaki: We should just reimplement this, it's trivial: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/processor/opentelemetry-processor-baggage/src/opentelemetry/processor/baggage/processor.py

logfire/_internal/config.py
```diff
@@ -628,6 +633,12 @@ def _load_configuration(
             advanced = AdvancedOptions(base_url=param_manager.load_param('base_url'))
         self.advanced = advanced
 
+        if add_baggage_to_attributes:
+            additional_span_processors = [
```
alexmojaki: I think `test_config_serializable` should be extended to double check that this works as expected.

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,50 @@
...
+
+__all__ = (
+    'get_baggage',
+    'update_baggage',
+)
+
+
+def get_baggage() -> Mapping[str, object] | None:
+    """Get all OpenTelemetry baggage for the current context as a mapping of key/value pairs."""
+    return baggage.get_all()
```
alexmojaki:
```suggestion
get_baggage = baggage.get_all
```

In particular `baggage.get_all` never returns None.

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@contextmanager
+def update_baggage(bag: dict[str, object]) -> Iterator[None]:
```
alexmojaki:
```suggestion
def set_baggage(**values: object) -> Iterator[None]:
```

I definitely think keyword arguments are nicer here. I'm not sure about `set_baggage`, I see how `update` makes it clearer that it merges with existing baggage, but I think kwargs also help suggest this, and we can just document this. The symmetry of get/set is clean and easy to remember.

alexmojaki: We need to talk about what happens when you set a weird type. Right now it will happily store it in baggage and allow you to retrieve it in the same process, but it'll log a warning when trying to set the span attribute and then discard it, and it won't survive distributed tracing.

alexmojaki: Need to document that this has to actually be used as a context manager, if you try to manually open and close it you will likely run into weirdness because baggage is stored in the same contextvar as the current span and `suppress_instrumentation` meaning that it gets cleared when those context managers exit:

```python
import logfire

logfire.configure()

with logfire.span('span'):
    baggage_context_manager = logfire.update_baggage({'a': 1})
    baggage_context_manager.__enter__()
    print(logfire.get_baggage())  # {'a': 1} as expected

print(logfire.get_baggage())  # User probably expects to see {'a': 1} here, but it is empty
baggage_context_manager.__exit__(None, None, None)
```

alexmojaki:
Something related to keep in mind: https://github.com/pydantic/logfire/issues/492

logfire/_internal/baggage.py
dmontagu: When you say "a weird type" do you mean types that are handled properly if you call `logfire.info`, but not otherwise (e.g. BaseModel instances)? If so, yeah that's unfortunate. Probably good to fix, I'd be okay just erroring eagerly in that case for now if it helps us release faster and we think we can eventually fix it, but I'm also okay trying to get it working now. Not sure how involved it will be...

dmontagu: Agreed **kwargs is better.

I didn't like that with `set_baggage` (which I did consider) it's not clear that it is merging in instead of updating, but also, for both of them, I don't like that it doesn't indicate that it's a contextmanager. Not sure much can be done about that, I guess I'm okay with `set_baggage`. I liked just `with baggage(...):` but I guess that doesn't play nice with `get_baggage`.

tests/test_baggage.py
dmontagu: I agree with that, but I'm not quite sure how we'd make it work. I don't think baggage attributes should get a prefix from a user's perspective during query time, that's pretty ugly, but I see the potential for conflicts. I think in practice conflicts will be rare/non-existent enough to not be worth worrying about though.

logfire-api/logfire_api/_internal/baggage.pyi
dmontagu: I just ran the make command while trying to fix some issues with the `logfire_api` tests

docs/reference/advanced/baggage.md
dmontagu: 🤦

docs/reference/advanced/baggage.md
dmontagu: oh yeah it should have been `INTERVAL '1 second'` for my code but anyway I see that it needs to be `> 1` due to being a number and not a time interval anyway.

tests/test_baggage.py
alexmojaki: If all baggage was put in a single JSON attribute `logfire.baggage` and that was a popped attribute (maybe even into a user friendly column like `baggage`) it could significantly help with query performance.

logfire/_internal/baggage.py
dmontagu: I don't mind documenting it needs to be used as a contextmanager more clearly, but I think anyone using `__enter__` and `__exit__` manually should recognize that the behavior could be surprising, I don't think this is likely to affect anyone besides people that know (or should know) they are likely getting themselves into trouble. But either way I can make it clear in the docstring that it must be used as a contextmanager.

docs/reference/advanced/baggage.md
dmontagu: I've now made use of it here, but not sure what else to say (or if it's necessary to say anything). Feel free to tweak.

docs/reference/advanced/baggage.md
dmontagu: I have mentioned distributed tracing in the docs now, but not about instrumented HTTP requests in the docs, though I think I did mention it in the `set_baggage` docstring. (Given there are other ways of propagating I'm not sure how important it is to explicitly describe one specific way.)

logfire/_internal/baggage.py
dmontagu: Updated the set_baggage docstring to mention this

logfire-api/logfire_api/_internal/baggage.pyi
```diff
@@ -0,0 +1,39 @@
...
+
+__all__ = ['get_baggage', 'set_baggage']
+
+get_baggage: Incomplete
+
+@contextmanager
+def set_baggage(**bag: str) -> Iterator[None]:
+    '''Context manager that attaches key/value pairs as OpenTelemetry baggage to the current context.
+
+    All values in `bag` must be strings, as OpenTelemetry baggage only supports string values.
```
adriangb: Should we make any attempt to serialize it for users? E.g. if pydantic_core is installed dump non-string values to json?

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,94 @@
...
+    finally:
+        context.detach(token)
+
+
+class NoForceFlushSpanProcessor(SpanProcessor):
+    # The default SpanProcessor.force_flush returns None,
+    # which gets interpreted as False by the OTel SDK, meaning that the spans did not export successfully.
+    # Then SynchronousMultiSpanProcessor stops looping through processors and doesn't force flush the next one.
+    # OTel is dumb.
+    # This base class just means there's nothing to flush.
```
adriangb: Hmm this does sem unreasonable if true! Have we opened an upstream issue? I also don't think we should be including `OTel is dumb` in a docstring.

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,94 @@
...
+    def force_flush(self, timeout_millis: int = 30000) -> bool:
+        return True
+
+
+class DirectBaggageAttributesSpanProcessor(NoForceFlushSpanProcessor):
+    def on_start(self, span: Span, parent_context: context.Context | None = None) -> None:
+        existing_attrs = span.attributes or {}
+        span.set_attributes({k: v for k, v in _get_baggage_attrs(parent_context).items() if k not in existing_attrs})
```
adriangb: Good to see that we aren't overriding attributes with baggage.

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,94 @@
...
+        if attrs := _get_baggage_attrs(parent_context):
+            span.set_attribute('logfire.baggage', json.dumps(attrs))
+
+
+def _get_baggage_attrs(parent_context: context.Context | None = None) -> dict[str, str]:
+    """Get baggage as a dict of strings to set as attributes on a span.
+
+    The values are converted to strings because that's what happens when baggage is propagated
+    to different services, e.g. through HTTP headers.
+    This way the value will be consistent between services.
```
adriangb: Sad but okay

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def _safe_str(obj: Any) -> str:
+    try:
+        return str(obj)
+    except Exception:
+        try:
+            return f'<{type(obj).__name__} object>'
+        except Exception:  # pragma: no cover
+            return '<unknown (repr failed)>'
```
adriangb: Does this also get called during propagation? Otherwise `This way the value will be consistent between services` won't be true

logfire/_internal/config.py
```diff
@@ -341,6 +346,11 @@ def configure(  # noqa: D417
             Defaults to `True` if and only if the Python version is at least 3.11.
 
         sampling: Sampling options. See the [sampling guide](https://logfire.pydantic.dev/docs/guides/advanced/sampling/).
+        add_baggage_to_attributes: Set to `False` to prevent OpenTelemetry Baggage from being added to spans as attributes.
+            Set to `'direct'` to directly add each Baggage value as a separate span attribute with the same name.
+            Set to `'json'` to add Baggage as a single span attribute named `logfire.baggage` with a JSON string value
+            (this is the default).
```
adriangb: I don't love the idea of this `json` mode and certainly don't think it should be the default. What was the use case / user request that prompted this?

adriangb (approved):
I generally like the idea. I'm 👎🏻 on the `json` mode however. Left some comments.

logfire-api/logfire_api/_internal/baggage.pyi
dmontagu: My feeling is it’s an unfortunate loss of information to stringify things, and we can add support for non-string inputs to this function later if anyone expresses interest. So I’m happy to start with it like this. I guess we could try to do the json schema nonsense again for this data but 🤷‍♂️ strings seem enough to be very useful in the short term

alexmojaki: Note that these .pyi files are out of date, the real docstrings don't say this.

I think we should probably have a conversation about strings vs JSON before merging.

logfire/_internal/baggage.py
alexmojaki: No, I'm just trying to avoid exceptions in exotic cases. Do you think omitting the value would be better?

logfire/_internal/config.py
alexmojaki: Pros:

1. `logfire.baggage` can be a popped attribute, making queries on it more efficient, especially when baggage is small but lives on spans with other very large attributes.
2. When an existing span attribute and a baggage key have the same name, both values are recorded separately, no worries about conflicts.
3. Namespace separates things clearly. When you query `attributes->>'foo'` you know you will always be getting a direct span attribute and don't have to think about whether it might actually be baggage. If we turned on `direct` mode by default it could actually lead to changes in alert results, though it's unlikely.
4. Similarly it's easier to whitelist attributes specifically in baggage from scrubbing.

Cons:

1. We probably want to make a new db column `baggage` which is more friendly than `attributes->'logfire.baggage'`. This means more backend and frontend complexity (particularly considering things like autocomplete and column choosing) and more for users to learn.
2. Different from other SDKs.

logfire/_internal/baggage.py
alexmojaki: Opened https://github.com/open-telemetry/opentelemetry-python/issues/4631, updated comment.

logfire/_internal/config.py
adriangb: > logfire.baggage can be a popped attribute, making queries on it more efficient, especially when baggage is small but lives on spans with other very large attributes.

It'll actually worse long term because popping will be replaced with data shredding and having a single attribute will defeat data shredding making queries slower and storage more expensive.

> When an existing span attribute and a baggage key have the same name, both values are recorded separately, no worries about conflicts.

I don't think that's a real concern

alexmojaki: Interesting, I didn't know about the shredding thing. So do you think having a single `logfire.metrics` attribute is also bad?

What do you think the default should be? `direct` or `False`?

alexmojaki:
If we keep `json` mode, should I rename `direct` to `simple`?

logfire/_internal/config.py
adriangb: Yes it will be better to flatten as much as possible. Maybe even flattening nested objects into flat keys if they don't contain arrays.

logfire/_internal/baggage.py
adriangb: Well what would happen during propagation?

logfire/_internal/config.py
adriangb: I think the default should be direct

alexmojaki: Do you think it's fine if when the user upgrades their SDK their spans start getting extra attributes which (1) add bloat to spans that we can't optimize yet, and (2) query results may change?

I get that this may be a non-issue in practice, just checking.

adriangb: Yes I think it's okay. I think (2) is relatively unlikely and (1) the same impact can happen by users adding new attributes, etc. In general very few attributes are currently optimized 😢, that's what shredding is supposed to fix (it will automatically optimize up to something like 128 attributes for each file as long as they share a common type across the whole file)

adriangb:
> If we keep `json` mode, should I rename `direct` to `simple`?

How about `inline` or `flat`?

logfire/_internal/config.py
alexmojaki: OK. Do you think we should just remove `json` mode and convert this to a boolean?

logfire/_internal/baggage.py
alexmojaki: It raises the error:

```pytb
  File "/Users/alex/Library/Application Support/JetBrains/PyCharm2025.1/scratches/scratch_1650.py", line 14, in <module>
    httpx.get('https://example.com')
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/httpx/_api.py", line 195, in get
    return request(
           ^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/httpx/_api.py", line 109, in request
    return client.request(
           ^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/httpx/_client.py", line 825, in request
    return self.send(request, auth=auth, follow_redirects=follow_redirects)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/httpx/_client.py", line 914, in send
    response = self._send_handling_auth(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/httpx/_client.py", line 942, in _send_handling_auth
    response = self._send_handling_redirects(
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/httpx/_client.py", line 979, in _send_handling_redirects
    response = self._send_single_request(request)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/httpx/_client.py", line 1014, in _send_single_request
    response = transport.handle_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/instrumentation/httpx/__init__.py", line 987, in _handle_request_wrapper
    _inject_propagation_headers(headers, args, kwargs)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/instrumentation/httpx/__init__.py", line 335, in _inject_propagation_headers
    inject(_headers)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/propagate/__init__.py", line 121, in inject
    get_global_textmap().inject(carrier, context=context, setter=setter)
  File "/Users/alex/work/logfire/logfire/propagate.py", line 109, in inject
    return self.wrapped.inject(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/propagators/composite.py", line 69, in inject
    propagator.inject(carrier, context, setter=setter)
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/baggage/propagation/__init__.py", line 125, in inject
    baggage_string = _format_baggage(baggage_entries)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/baggage/propagation/__init__.py", line 135, in _format_baggage
    return ",".join(
           ^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/baggage/propagation/__init__.py", line 136, in <genexpr>
    quote_plus(str(key)) + "=" + quote_plus(str(value))
                                            ^^^^^^^^^^
  File "/Users/alex/Library/Application Support/JetBrains/PyCharm2025.1/scratches/scratch_1650.py", line 8, in __repr__
    raise ValueError('bad repr')
ValueError: bad repr
```

logfire/_internal/config.py
adriangb: I think having a literal with 2 choices isn't terrible. IMO we should remove JSON mode, we can always add it as another feature later, and leave a literal with `off` or `inline`, something like that.

adriangb:
Should we have a way to set an attribute for all spans within the current service without propagating?

I think one of the common requests is setting e.g. `user_id` for all spans in a request, but you probably _actively_ do _not_ want to send that to any downstream services!

logfire-api/logfire_api/__init__.pyi
```diff
@@ -14,7 +14,7 @@ from .version import VERSION as VERSION
...
...vancedOptions', 'ConsoleOptions', 'CodeSource', 'PydanticPlugin', 'configure', 'span', 'instrument', 'log', 'trace', 'debug', 'notice', 'info', 'warn', 'warning', 'error', 'exception', 'fatal', 'force_flush', 'log_slow_async_callbacks', 'install_auto_tracing', 'instrument_asgi', 'instrument_wsgi', 'instrument_pydantic', 'instrument_pydantic_ai', 'instrument_fastapi', 'instrument_openai', 'instrument_openai_agents', 'instrument_anthropic', 'instrument_asyncpg', 'instrument_httpx', 'instrument_celery', 'instrument_requests', 'instrument_psycopg', 'instrument_django', 'instrument_flask', 'instrument_starlette', 'instrument_aiohttp_client', 'instrument_aiohttp_server', 'instrument_sqlalchemy', 'instrument_sqlite3', 'instrument_aws_lambda', 'instrument_redis', 'instrument_pymongo', 'instrument_mysql', 'instrument_system_metrics', 'instrument_mcp', 'AutoTraceModule', 'with_tags', 'with_settings', 'suppress_scopes', 'shutdown', 'no_auto_trace', 'ScrubMatch', 'ScrubbingOptions', 'VERSION', 'add_non_user_code_prefix', 'suppress_instrumentation', 'StructlogProcessor', 'LogfireLoggingHandler', 'loguru_handler', 'SamplingOptions', 'MetricsOptions', 'logfire_info', 'get_baggage', 'set_baggage']
```
adriangb: I feel like this should be multi-line. I can't tell from the diff what changed. Not related to this PR.

logfire/_internal/baggage.py
```diff
@@ -1,22 +1,26 @@
...
     'set_baggage',
 )
 
+from .utils import truncate_string
 
 get_baggage = baggage.get_all
 """Get all OpenTelemetry baggage for the current context as a mapping of key/value pairs."""
 
+MAX_BAGGAGE_VALUE_LENGTH = 1000
+
```
adriangb, line 31: Comment on where this constant is coming from

logfire/_internal/baggage.py
```diff
@@ -44,6 +48,17 @@ def set_baggage(**values: str) -> Iterator[None]:
     """
     current_context = context.get_current()
     for key, value in values.items():
+        if not isinstance(value, str):  # type: ignore
+            warnings.warn(
+                f'Baggage value for key "{key}" is a {type(value).__name__}. Converting to string.', stacklevel=3
+            )
+            value = logfire_json_dumps(value)
```
adriangb: We said in the call we are _not_ going to support non-strings for now

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,98 @@
...
+        attrs: dict[str, str] = {}
+        for k, v in baggage.get_all(parent_context).items():
+            if k in existing_attrs:
+                continue  # TODO
+            if not isinstance(v, str):
+                # Since this happens for every span, don't try converting to string, which could be expensive.
+                warnings.warn(
+                    f'Baggage value for key "{k}" is of type "{type(v).__name__}", skipping setting as attribute.'
+                )
+                continue
```
adriangb, line 97: Why do we need this here? I don't think it matters at this point what the types are, attributes support "all" types

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,98 @@
...
+        return True
+
+
+class DirectBaggageAttributesSpanProcessor(NoForceFlushSpanProcessor):
+    def on_start(self, span: Span, parent_context: context.Context | None = None) -> None:
+        existing_attrs = span.attributes or {}
+        attrs: dict[str, str] = {}
+        for k, v in baggage.get_all(parent_context).items():
+            if k in existing_attrs:
+                continue  # TODO
```
adriangb: Maybe a warn?

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,98 @@
...
+                warnings.warn(
+                    f'Baggage value for key "{k}" is of type "{type(v).__name__}", skipping setting as attribute.'
+                )
+                continue
+            if len(v) > MAX_BAGGAGE_VALUE_LENGTH:
+                # Truncating is cheap though.
+                warnings.warn(
+                    f'Baggage value for key "{k}" is too long. Truncating to {MAX_BAGGAGE_VALUE_LENGTH} characters before setting as attribute.',
+                )
+                v = truncate_string(v, max_length=MAX_BAGGAGE_VALUE_LENGTH)
```
adriangb, line 103: Also don't think we need to truncate here, just let attribute truncation do it's thing later

adriangb (approved):
Left a couple nits about when to convert to strings (never, we don't allow anything but strings on the first iteration) and when to truncate (only when calling `set_baggage`.

I think all of this highlights the need for a `logfire.with_attributes(...)` because:
1. It won't be subject to the limitations of baggage (types, length, etc.) because it doesn't get propagated.
2. Many use cases want attributes in all spans in a trace, but don't necessarily (or even actively do not) want that to be propagated to other services.

logfire-api/logfire_api/__init__.pyi
alexmojaki: Ignore pyi files

adriangb: ok

logfire/_internal/baggage.py
alexmojaki: These are raw otel spans, they don't support all types. OTel will log a warning every time:

```python
from opentelemetry import baggage as otel_baggage, context

import logfire

logfire.configure()
context.attach(otel_baggage.set_baggage('a', {'b': 'c'}))

for _ in range(3):
    logfire.info('hi')
```

```
Invalid type dict for attribute 'a' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
Invalid type dict for attribute 'a' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
Invalid type dict for attribute 'a' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

By contrast, `warnings` only warns once.

Using logfire's JSON logic here is too expensive.

logfire/_internal/baggage.py
alexmojaki: Let's rather consider relaxing this later. Attribute truncation is much more generous and we don't want too much attached to every span.

alexmojaki:
Docs updates still coming, no need to review yet.

alexmojaki:
Now it's ready.

docs/reference/advanced/baggage.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Disabling
+
+If you don't want to add Baggage to span attributes, pass `add_baggage_to_attributes=False` to the
+[`logfire.configure()`][logfire.configure] function. This may slightly improve performance.
+The [`set_baggage`][logfire.set_baggage] contextmanager will still update the OpenTelemetry Baggage and propagate it to other services.
+
+## Baggage values requirements
+
+While baggage can technically contain any data type, only strings will be set as attributes on spans.
```
adriangb, line 57: I don't feel like this is the right direction to give users. I think we should be explicit that baggage can only contain strings because it is serialized across the wire and we should enforce that in type hints and runtime assertions / errors / warnings (I'm okay with doing best effort to dump non-strings to strings but I think that should basically be undocumented and emit a warning).

docs/reference/advanced/baggage.md
```diff
@@ -0,0 +1,90 @@
...
+
+then the values set in [`logfire.set_baggage`][logfire.set_baggage] in step 2 will be available as Baggage in the `fastapi` service,
+and by default those values will be added as attributes to the spans created in that service.
+
+This works by including the values in the `Baggage` HTTP header in the request made by `httpx`.
+This happens regardless of whether the request is received by a server that can read the Baggage or not,
+or whether either service sets Baggage as span attributes. So it's important to be careful about what you put in Baggage:
+
+- **Don't put sensitive information in Baggage**, as it may be sent to third party services. While span attributes from baggage are still [scrubbed](../../how-to-guides/scrubbing.md) by default, the Baggage header itself is not scrubbed.
+- **Don't put large values in Baggage**, as this may add bloat to HTTP headers. Large values may also be dropped by servers instead of being propagated. This is one reason why Logfire truncates Baggage values longer than 1000 characters.
```
adriangb, line 82: 👍🏻

logfire-api/logfire_api/_internal/baggage.pyi
```diff
@@ -0,0 +1,39 @@
...
+
+    OpenTelemetry baggage is a way to propagate metadata across service boundaries in a distributed system, and is
+    included in headers of outgoing requests for which context propagation is configured.
+    This is intended to be used to propagate arbitrary context (like user ids, task ids, etc.) down to all nested spans.
+
+    Baggage is not _automatically_ converted into attributes on descendant spans, but this is a common usage pattern.
+    If you want baggage to be converted into attributes, use `logfire.configure(add_baggage_to_attributes=True)`.
```
adriangb:
```suggestion
    Baggage is not _automatically_ converted into attributes on descendant spans, but this is a common usage pattern.
    If you want baggage to be converted into attributes, use `logfire.configure(add_baggage_to_attributes=True)`.
    This is the default.
    To disable this set `logfire.configure(add_baggage_to_attributes=False)`
```

logfire/_internal/baggage.py
```diff
@@ -0,0 +1,101 @@
...
+    with set_baggage(my_id='123'):
+        # All spans opened inside this block will have baggage '{"my_id": "123"}'
+        with set_baggage(my_session='abc'):
+            # All spans opened inside this block will have baggage '{"my_id": "123", "my_session": "abc"}'
+            ...
+    ```
+    """
+    current_context = context.get_current()
+    for key, value in values.items():
+        if isinstance(value, str):  # type: ignore  # prevent error in line below if user ignores type hints
```
adriangb: What if it's not a string? I think we should emit a warning and do a best-effort to make it a string

adriangb (approved):
generally looks good!

adriangb:
> I think all of this highlights the need for a `logfire.with_attributes(...)` because:
> 
> 1. It won't be subject to the limitations of baggage (types, length, etc.) because it doesn't get propagated.
> 2. Many use cases want attributes in all spans in a trace, but don't necessarily (or even actively do not) want that to be propagated to other services.

Any thoughts on this?

docs/reference/advanced/baggage.md
alexmojaki: > I'm okay with doing best effort to dump non-strings to strings but I think that should basically be undocumented and emit a warning

That's exactly what I had before.

Regardless of what `logfire.set_baggage` does, it will always be possible to use the OTel APIs so set other types. I'm essentially documenting what happens in that case.

logfire-api/logfire_api/_internal/baggage.pyi
alexmojaki: ignore pyi files

https://github.com/pydantic/logfire/pull/1128#discussion_r2135493755

adriangb: okay can you apply the edit to the right file please?

alexmojaki:
> > I think all of this highlights the need for a `logfire.with_attributes(...)` because:
> > 
> > 1. It won't be subject to the limitations of baggage (types, length, etc.) because it doesn't get propagated.
> > 2. Many use cases want attributes in all spans in a trace, but don't necessarily (or even actively do not) want that to be propagated to other services.
> 
> Any thoughts on this?

I want to wait for real use cases and user feedback.

logfire/_internal/baggage.py
alexmojaki: I removed this, but now it sounds like you want it back?

logfire-api/logfire_api/_internal/baggage.pyi
alexmojaki: This wording doesn't exist on the real function. These files weren't supposed to be generated and I'm not updating them each time.

alexmojaki:
Will merge on Monday and release right after

PR #1139, opened 2025-06-12
Add `logfire run` command

Kludex:
I need to document and ~test~ (it's tested).

To run it: `logfire run -m clai` or `logfire run main.py`.


logfire/_internal/cli/__init__.py
```diff
@@ -417,8 +346,8 @@ def __call__(
         namespace: argparse.Namespace,
         values: str | Sequence[Any] | None,
         option_string: str | None = None,
-    ):
-        if isinstance(values, str):  # pragma: no branch
+    ):  # pragma: no cover
```
alexmojaki: are you going to cover this?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,323 @@
...
+from rich.text import Text
+
+import logfire
+
+STANDARD_LIBRARY_PACKAGES = {'urllib', 'sqlite3'}
+
+# Map of instrumentation packages to the packages they instrument
+OTEL_INSTRUMENTATION_MAP = {
+    'opentelemetry-instrumentation-aio_pika': 'aio_pika',
+    'opentelemetry-instrumentation-aiohttp-client': 'aiohttp',
```
alexmojaki:
```suggestion
    'opentelemetry-instrumentation-aiohttp-client': 'aiohttp',
    'opentelemetry-instrumentation-aiohttp-server': 'aiohttp',
```

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,323 @@
...
+    """
+    instrumented: list[str] = []
+
+    # Process all installed OpenTelemetry packages
+    for pkg_name in installed_otel_packages:
+        if pkg_name in instrument_pkg_map.keys():  # pragma: no branch
+            base_pkg = pkg_name.replace('opentelemetry-instrumentation-', '')
+            # Handle special cases
+            if base_pkg == 'aiohttp-client':
+                base_pkg = 'aiohttp'  # pragma: no cover
```
alexmojaki: This is worth covering. The methods are `logfire.instrument_aiohttp_client()` and `logfire.instrument_aiohttp_server()`, this doesn't look right.

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,323 @@
...
+    'opentelemetry-instrumentation-remoulade': 'remoulade',
+    'opentelemetry-instrumentation-requests': 'requests',
+    'opentelemetry-instrumentation-sqlalchemy': 'sqlalchemy',
+    'opentelemetry-instrumentation-sqlite3': 'sqlite3',
+    'opentelemetry-instrumentation-starlette': 'starlette',
+    'opentelemetry-instrumentation-tornado': 'tornado',
+    'opentelemetry-instrumentation-tortoiseorm': 'tortoise_orm',
+    'opentelemetry-instrumentation-urllib': 'urllib',
+    'opentelemetry-instrumentation-urllib3': 'urllib3',
+    'pydantic-ai-slim': 'pydantic_ai',
```
alexmojaki, line 66: IIUC this means that if pydantic-ai-slim is installed, it will call `logfire.instrument_pydantic_ai()`, which is good. It's a bit hard to understand from looking at this, since this is quite a different case from the rest of the dict.

Other similar cases where we can instrument without an external package: openai, agents, (instrument_openai_agents), anthropic.

Also just set these to instrument langchain (before importing langchain, so either don't check if anything is installed, or only check if it's importable without importing):

```
os.environ['LANGSMITH_OTEL_ENABLED'] = 'true'
os.environ['LANGSMITH_TRACING'] = 'true'
```

alexmojaki: there's also logfire.instrument_pydantic(), but we should probably leave that.

Kludex: Just to be clear... The things I need to do:

1. Call `instrument_openai_agents`, `instrument_anthropic` and `instrument_openai`.
2. Set the langsmith enviornment variables.

logfire/_internal/cli/__init__.py
Kludex: I was not planning to.

logfire/_internal/cli/run.py
Kludex: Can I set the environment variables without checking anything then?

logfire/_internal/cli/__init__.py
alexmojaki: please do, especially since it was apparently covered before?

logfire/_internal/cli/run.py
alexmojaki: yes i don't think you need to check anything

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+    'openai-agents': 'openai_agents',
+}
+
+
+def parse_run(args: argparse.Namespace) -> None:  # pragma: no cover
+    # Initialize Logfire
+    logfire.configure()
+
+    summary = cast(bool, args.summary)
+    exclude = cast('set[str]', args.exclude)
```
alexmojaki:
```suggestion
    exclude = cast(set[str], args.exclude)
```

no more 3.8

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+
+    instrument_pkg_map = {otel_pkg: pkg for otel_pkg, pkg in OTEL_INSTRUMENTATION_MAP.items() if pkg not in exclude}
+
+    installed_pkgs = installed_packages()
+    installed_otel_pkgs = {pkg for pkg in instrument_pkg_map.keys() if pkg in installed_pkgs}
+
+    recommendations = recommended_instrumentation(instrument_pkg_map, installed_otel_pkgs, installed_pkgs)
```
alexmojaki: please dedupe these lines against the inspect code

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+
+            # Run the module
+            runpy.run_module(module_name, run_name='__main__', alter_sys=True)
+        finally:
+            # Restore original arguments
+            sys.argv = orig_argv
+    elif hasattr(args, 'script') and args.script:
+        # Script mode
+        script_path = args.script
+        script_args = getattr(args, 'args', [])
```
alexmojaki: it's weird that this is the same as module_args. Also does `args.args` not work in all cases?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+        cmd_str = f'python -m {module_name} {" ".join(module_args)}'
+
+        # Save original arguments
+        orig_argv = sys.argv.copy()
+
+        try:
+            # Set up args for the module
+            sys.argv = [f'-m {module_name}'] + module_args
+
+            logfire.info(f'Running command: {cmd_str}')
```
alexmojaki: avoid f-string magic since we don't know the user's deployment situation

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+
+            # Run the module
+            runpy.run_module(module_name, run_name='__main__', alter_sys=True)
```
alexmojaki, line 115: both runpy calls modify sys.argv. does this fit nicely with your own modifications to sys.argv? i.e. they wouldn't work correctly without your changes, but also don't make your changes redundant?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+            # Restore original arguments
+            sys.argv = orig_argv
+    else:
+        print('Usage: logfire run [-m MODULE] [args...] OR logfire run SCRIPT [args...]')
+        sys.exit(1)
+
+
+def is_uv_installed() -> bool:
+    """Check if uv package manager is installed and available in the PATH."""
+    return shutil.which('uv') is not None  # pragma: no cover
```
alexmojaki: surely this is covered?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+    for pkg_name in installed_otel_packages:
+        if pkg_name in instrument_pkg_map.keys():  # pragma: no branch
+            base_pkg = pkg_name.replace('opentelemetry-instrumentation-', '')
+
+            import_name = instrument_pkg_map[pkg_name]
+            instrument_attr = f'instrument_{import_name}'
+
+            try:
+                # Try to access the instrumentation function as an attribute of logfire
+                if hasattr(logfire, instrument_attr):
```
alexmojaki: Not really worth checking this if we catch all exceptions

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+
+def instrument_packages(installed_otel_packages: set[str], instrument_pkg_map: dict[str, str]) -> list[str]:
+    """Automatically instrument the installed OpenTelemetry packages.
+
+    Returns a list of packages that were successfully instrumented.
+    """
+    instrumented: list[str] = []
+
+    # Process all installed OpenTelemetry packages
+    for pkg_name in installed_otel_packages:
```
alexmojaki:
```suggestion
    for otel_pkg_name in installed_otel_packages:
```

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+        if pkg_name in instrument_pkg_map.keys():  # pragma: no branch
```
alexmojaki:
```suggestion
        if pkg_name in instrument_pkg_map:  # pragma: no branch
```

but do we need this check? is there a way for it to be false?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+            instrument_attr = f'instrument_{import_name}'
+
+            try:
+                # Try to access the instrumentation function as an attribute of logfire
+                if hasattr(logfire, instrument_attr):
+                    # The function exists, call it to instrument the package
+                    getattr(logfire, instrument_attr)()
+                    instrumented.append(base_pkg)
+            except Exception:  # pragma: no cover
+                continue
```
alexmojaki:
```suggestion
            except Exception:  # pragma: no cover
                continue
            instrumented.append(base_pkg)
```

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+    return instrumented
+
+
+def recommended_instrumentation(
```
alexmojaki: name and docstring need to make it clearer that this is about what's recommended *to install*

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+def instrumented_packages_text(
+    installed_otel_pkgs: set[str], instrumented_packages: list[str], installed_pkgs: set[str]
+) -> Text:
+    # Filter out special cases for display
+    if 'fastapi' in installed_pkgs:
+        installed_otel_pkgs.discard('opentelemetry-instrumentation-starlette')  # pragma: no cover
+    if 'requests' in installed_pkgs:
+        installed_otel_pkgs.discard('opentelemetry-instrumentation-urllib3')  # pragma: no cover
```
alexmojaki: why is this duplicated in the function above, and why is it not covered?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,325 @@
...
+
+    text = Text('Your instrumentation checklist:\n\n')
+    for pkg_name in sorted(installed_otel_pkgs):
+        base_pkg = _base_pkg_name(pkg_name)
+        if base_pkg in instrumented_packages:
+            text.append(f'✓ {base_pkg} (installed and instrumented)\n', style='green')
+        else:
+            text.append(f'⚠️ {base_pkg} (installed but not automatically instrumented)\n', style='yellow')
```
alexmojaki, line 231: When does this happen?

Kludex: I added this for FastAPI, Starlette and Flask that need to pass the app.

alexmojaki: i guess there's also otel packages that don't have a logfire method?

tests/test_cli.py
```diff
@@ -231,77 +237,89 @@ def test_clean_default_dir_is_not_a_directory(
...
+                'opentelemetry-instrumentation-fastapi': 'fastapi',
+                'opentelemetry-instrumentation-urllib': 'urllib',
+                'opentelemetry-instrumentation-sqlite3': 'sqlite3',
             },
-        ),
-        (
-            'inspect',
-            ['fastapi', 'starlette'],
+            {'fastapi'},
             {
```
alexmojaki:
```suggestion
            snapshot({
```

logfire/_internal/cli/run.py
Kludex: This is not correct, it should be inside.

logfire/_internal/cli/run.py
Kludex: If a new OpenTelemetry package is added, and we don't map it yet?

alexmojaki: but installed_otel_packages comes from instrument_pkg_map

logfire/_internal/cli/run.py
alexmojaki: that's because you had the `if hasattr`

tests/test_cli.py
Kludex: I didn't get this. Where should I add the snapshot?

alexmojaki: weird.

wrap expected results in snapshots. they work inside pytest params.

Kludex: what... I didn't know this.

logfire/_internal/cli/run.py
Kludex: I've removed the pragmas.

logfire/_internal/cli/run.py
Kludex: I made some changes around here, now it works properly.

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,352 @@
...
+@dataclass
+class InstrumentationContext:
+    instrument_pkg_map: dict[str, str]
+    installed_pkgs: set[str]
+    installed_otel_pkgs: set[str]
+    recommendations: set[tuple[str, str]]
+
+
+def parse_run(args: argparse.Namespace) -> None:  # pragma: no cover
+    # Initialize Logfire
```
alexmojaki:
```suggestion
def parse_run(args: argparse.Namespace) -> None:
```

logfire/_internal/cli/run.py
alexmojaki: same questions apply. please add comments explaining why sys.argv is changed, point out what runpy does, and the interaction.

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,352 @@
...
+        module_args = script_and_args
+
+        cmd_str = f'python -m {module_name} {" ".join(module_args)}'
+
+        # Save original arguments
+        orig_argv = sys.argv.copy()
+
+        try:
+            # Set up args for the module
+            sys.argv = [f'-m {module_name}'] + module_args
```
alexmojaki: why is `f'-m {module_name}'` a single arg?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,352 @@
...
+
+            logfire.info('Running command: {cmd_str}', cmd_str=cmd_str)
+
+            # Run the module
+            runpy.run_module(module_name, run_name='__main__', alter_sys=True)
+        finally:
+            # Restore original arguments
+            sys.argv = orig_argv
```
alexmojaki:
```suggestion
        with running_common_need_better_name(new_argv=[f'-m {module_name}'] + module_args, cmd_str=cmd_str):
            runpy.run_module(module_name, run_name='__main__', alter_sys=True)
```

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,352 @@
...
+            sys.path.insert(0, script_dir)
+
+        # Save original arguments
+        orig_argv = sys.argv.copy()
+
+        try:
+            # Set up args for the script
+            sys.argv = [script_path] + script_args
+
+            logfire.info(f'Running command: {cmd_str}')
```
alexmojaki: f-string

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,352 @@
...
+            ctx.installed_otel_pkgs, instrumented_packages, ctx.installed_pkgs
+        )
+        print_otel_summary(
+            console=console, instrumented_packages_text=instrumentation_text, recommendations=ctx.recommendations
+        )
+
+    # Get arguments from the script_and_args parameter
+    script_and_args = args.script_and_args
+
+    if hasattr(args, 'module') and (module_name := args.module):
```
alexmojaki:
```suggestion
    if module_name := args.module:
```

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,352 @@
...
+        sys.exit(1)
+
+
+def is_uv_installed() -> bool:
+    """Check if uv package manager is installed and available in the PATH."""
+    return shutil.which('uv') is not None
+
+
+def instrument_packages(installed_otel_packages: set[str], instrument_pkg_map: dict[str, str]) -> list[str]:
+    """Automatically instrument the installed OpenTelemetry packages.
```
alexmojaki:
```suggestion
    """Call every `logfire.instrument_x()` we can based on what's installed.
```

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,352 @@
...
+    for otel_pkg_name in installed_otel_packages:
+        base_pkg = otel_pkg_name.replace('opentelemetry-instrumentation-', '')
+
+        import_name = instrument_pkg_map[otel_pkg_name]
+        instrument_attr = f'instrument_{import_name}'
+
+        try:
+            # If the function exists, call it to instrument the package
+            getattr(logfire, instrument_attr)()
+        except Exception:  # pragma: no cover
```
alexmojaki: isn't this covered?

logfire/_internal/cli/run.py
Kludex: Because otherwise the `module_name` would be passed to the application.

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,336 @@
...
+        with alter_sys_argv(sys.argv, f'python {script_path} {" ".join(script_args)}'):
+            sys.argv = [script_path] + script_args
+            runpy.run_path(script_path, run_name='__main__')
+    else:
+        print('Usage: logfire run [-m MODULE] [args...] OR logfire run SCRIPT [args...]')
+        sys.exit(1)
+
+
+@contextmanager
+def alter_sys_argv(argv: list[str], cmd: str) -> Generator[None, None, None]:  # pragma: no cover
```
alexmojaki:
```suggestion
def alter_sys_argv(argv: list[str], cmd: str) -> Generator[None, None, None]:
```

why always the pragmas?

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,336 @@
...
+    orig_argv = sys.argv.copy()
+    sys.argv = argv
```
alexmojaki, line 135: might as well remove this argument

logfire/_internal/cli/run.py
```diff
@@ -0,0 +1,336 @@
...
+        script_path = script_and_args[0]
+        script_args = script_and_args[1:]
+
+        # Make sure the script directory is in sys.path
+        script_dir = os.path.dirname(os.path.abspath(script_path))
+        if script_dir not in sys.path:
+            sys.path.insert(0, script_dir)
+
+        with alter_sys_argv(sys.argv, f'python {script_path} {" ".join(script_args)}'):
+            sys.argv = [script_path] + script_args
```
alexmojaki:
```suggestion
        # Make sure the script directory is in sys.path
        script_dir = os.path.dirname(os.path.abspath(script_path))
        if script_dir not in sys.path:
            sys.path.insert(0, script_dir)

        with alter_sys_argv(sys.argv, f'python {" ".join(script_and_args)}'):
            sys.argv = script_and_args
```

alexmojaki:
`logfire run -m` needs to add to sys.path:

Screenshot 2025-07-10 at 11 37 46 (1)

and it needs to passthrough arguments after the module:
 
Screenshot 2025-07-14 at 17 09 46

alexmojaki:
@Kludex maybe we try merging this without `-m` first?

alexmojaki:
Note that `-m` still isn't quite perfect, but we may have to settle a bit.

Screenshot 2025-07-15 at 22 03 17

Kludex:
Hmmmm... Suggestions on how to solve that one? Maybe manually parsing instead of relying on `argparse`?

@alexmojaki Are you good with this PR?

PR #1151, opened 2025-06-18
Chapter on JS

petyosi:


docs/integrations/javascript/browser.md
```diff
@@ -0,0 +1,40 @@
+---
+integration: logfire
+---
+
+# Browser
```
GSemikozov, line 5: not sure, but might be helpful to mention if bundler configuration is needed to support ESM

docs/integrations/javascript/cloudflare.md
```diff
@@ -0,0 +1,77 @@
+---
+integration: logfire
+---
+
+# Cloudflare
+
+To instrument your Cloudflare Workers and send spans to Logfire, install the `@pydantic/logfire-cf-workers` and `@pydantic/logfire-api` NPM packages:
```
GSemikozov, line 7: do we need to specify which versions of Cloudflare Workers are supported?

docs/integrations/javascript/express.md
```diff
@@ -0,0 +1,58 @@
+---
+integration: logfire
+---
+
+# Express
+
+Instrumenting an Express application with Logfire is straightforward. You can use the `logfire` package to set up logging and monitoring for your Express routes.
+
+```ts title="app.ts"
+import express, type { Express } from 'express';
```
GSemikozov, line 14: as for compatibility - it works let's say with both v4 and v5?

docs/integrations/javascript/express.md
```diff
@@ -0,0 +1,58 @@
...
+
+logfire.configure();
+```
+
+The `logfire.configure` call should happen before importing the actual Express module, so your NPM start script should look like this in `package.json`. Note that we use `npx ts-node` to run the TypeScript code directly:
+
+```json title="package.json"
+"scripts": {
+  "start": "npx ts-node --require ./instrumentation.ts app.ts"
+A complete working example can be found in [examples/express](https://github.com/pydantic/logfire-js/tree/main/examples/express).
```
GSemikozov: is this red string is redundant diff?

docs/integrations/javascript/index.md
```diff
@@ -0,0 +1,36 @@
+# JavaScript
```
GSemikozov, line 1: in general: 1-line desc under each heading might be helpful, for example 'for client-side logging' or 'for edge worker environments' etc. WDYT?

docs/integrations/javascript/nextjs.md
```diff
@@ -0,0 +1,56 @@
+---
+integration: logfire
+---
+
+# Next.js
```
GSemikozov, line 5: again, maybe we have to mention compatibility (like Next.js 13+ with App Router). I remember had some pain with different versions of Next.js

docs/integrations/javascript/node.md
```diff
@@ -0,0 +1,39 @@
...
+});
+
+logfire.info("Hello from Node.js", {
+  "attribute-key": "attribute-value",
+}, {
+  tags: ["example", "example2"],
+});
+```
+
+Run the script with `node hello.js`, and you should see the log entry appear in
```
GSemikozov: how to persist logs? or it's not a correct question here?)

docs/integrations/javascript/browser.md
GSemikozov: And also maybe we can add minimum version requirements of Node, Deno, etc., if applicable.

GSemikozov (commented):
LGTM, just a few small questions / suggestions.

docs/integrations/javascript/cloudflare.md
petyosi: That's complicated to figure out, we're using a fork of the `otel-cf-workers` package, and I am not entirely sure if it has some minimum version requirements.

docs/integrations/javascript/index.md
petyosi: I'm not sure I understand what you mean - perhaps we can do an additional minor PR to add those.

docs/integrations/javascript/nextjs.md
petyosi: This is (mostly) outside of our control, and likely to become outdated, since the heavy lifting is done by the `@vercel/otel` package.

docs/integrations/javascript/node.md
petyosi: That's a nice catch, we don't want to talk about logs here (we're talking only about spans/traces). Fixed wording.

Issue #1025, opened 2025-04-18
Integration tests of Redis


medaminezghal:
### Question

Is it possible to change to integration tests of Redis from Testcontainers to FakeRedis?

Kludex:
@medaminezghal Why? I prefer to use the real package instead of a mock.

medaminezghal:
@Kludex I 've tested to mock with fakeredis (python package) and it works and I'm testing now mysqld-integration-test  (python package).

I think it make the tests faster and use less resources.

alexmojaki:
@medaminezghal thanks for trying but I think we should just keep things as is.

PR #1158, opened 2025-06-19
Remove testcontainers dependency from tests

medaminezghal:
This an example of using `fakeredis` instead of `testcontainer`.

If you like the idea I will do the same for `mysql` tests.

If it is necessary to use [real packages](https://github.com/pydantic/logfire/issues/1025) for tests just close this `PR`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -154,6 +154,7 @@ dev = [
     "mypy>=1.10.0",
     "celery>=5.4.0",
     "testcontainers>3.7.1",
+    "fakeredis>=2.30.0",
     "mysql-connector-python~=8.0",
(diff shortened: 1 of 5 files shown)


medaminezghal:
I think that test of `mysql-integration` doesn't need real `mysql` instance that's why I use mock.

alexmojaki:
Sorry, I was interested at first, but I'm not feeling it.

1. You're ignoring warnings and type errors
2. The celery tests are still slow.
3. The faking looks quite heavy. Even after creating a fake redis object and passing it to celery, you still also have to use mock.patch.
4. For mysql, everything is a magic mock, there's no sign of anything real left over.

PR #1159, opened 2025-06-19
Add `send_to_logfire_min_log_level` to `logfire.configure`

alexmojaki:
Example:

```python
import logfire

logfire.configure(send_to_logfire_min_log_level='info')

# spans aren't removed, even if they are debug spans
with logfire.span('span1', _level='debug'):
    logfire.info('info1')
    logfire.debug('debug2')
```

Screenshot 2025-06-19 at 18 16 19

Probably need a better name.


alexmojaki:
The current docs (https://logfire.pydantic.dev/docs/reference/api/logfire/#logfire.configure(advanced)) already have a problem of the `configure` args table having horizontal scroll:

Screenshot 2025-06-19 at 18 18 46

But this PR makes it worse (https://alex-send-to-logfire-min-log.logfire-docs.pages.dev/reference/api/logfire/#logfire.configure(advanced)):

Screenshot 2025-06-19 at 18 18 51

adriangb:
How about just `log_level`?

alexmojaki:
I think that sounds like it applies to all exporters, not just sending to logfire.

logfire/_internal/config.py
```diff
@@ -536,6 +547,9 @@ class _LogfireConfigData:
     sampling: SamplingOptions
     """Sampling options."""
 
+    send_to_logfire_min_log_level: int | LevelName | None
```
dmontagu, line 550: I think this should just be `min_log_level`; it feels perfectly reasonable to me for the "send_to_logfire_" to be implicit. I mean, is there a reasonable alternative interpretation of `min_log_level`?

If we must keep the `send`, then I'd at least suggest dropping the `to_logfire_`, I mean I'm assuming that this isn't specific to logfire, in that if you configure the SDK to send to a different (non-logfire) service's otel endpoint, it will still drop records below the specified level. But I think `min_log_level` is enough for the name.

The main thing I'll note is that I think there should be a way to contextually override the min log level, so e.g. you could add a middleware that looks for a specific header and if present sets a contextvar which is used to override the value of the min_log_level used by the `MinLogLevelFilterSpanExporter` etc. (As a way to get debug-level spans on a dynamic basis.)

But I'm okay adding that later, and I think even if we call this argument `min_log_level`, I don't foresee us regretting the choice or wanting to make a breaking API change.

I am open to being convinced otherwise on all of the above though.

adriangb: I think that's the same point I made here: https://github.com/pydantic/logfire/pull/1159#issuecomment-2993686401

Kludex: (I would prefer `log_level` instead of `min_log_level`, given that it's implicit that the `log_level` is the minimum.)

dmontagu: I personally feel `min_log_level` is a bit more clear, otherwise I start thinking it's something about the log level that it will record and/or emit by default

alexmojaki: > I mean I'm assuming that this isn't specific to logfire, in that if you configure the SDK to send to a different (non-logfire) service's otel endpoint, it will still drop records below the specified level

https://github.com/pydantic/logfire/pull/1159#issuecomment-2995630510

https://github.com/pydantic/logfire/pull/1159/files#diff-bd226b8c255428fa3ea7e81d5416cdf49c796f9e4d301ba16d3e9efe30e09ba7R353

https://pydantic.slack.com/archives/C0813FX362E/p1750085812697729?thread_ts=1749641756.391929&cid=C0813FX362E

PR #1179, opened 2025-06-24
Add Instrumenting a Flask application with Gunicorn doc

hramezani:


docs/integrations/web-frameworks/gunicorn.md
```diff
@@ -24,3 +24,39 @@ gunicorn myapp:app --config gunicorn_config.py
...
 
 Where `myapp:app` is your WSGI application and `gunicorn_config.py` is the configuration file where you defined the `post_fork` function.
+
+## Instrumenting a Flask application
+
+This section shows how to instrument a Flask application running under Gunicorn with Logfire.
+
+Here is the `Flask` application code (`myapp.py`):
+
+```py
```
alexmojaki:
```suggestion
Here is the `Flask` application code:

```py title="myapp.py"
```

same for below

alexmojaki:
@hramezani can you confirm that things work with this post_fork hook, and don't work without it?

Also see https://github.com/open-telemetry/opentelemetry-python/issues/3307#issuecomment-2410994381, maybe just mention that `--preload` is a problem.

hramezani:
The small flask example that I added works with `post_fork` in Gunicorn config. 
Also, I removed the `post_fork` function and moved the logfire configuration and instrumentation into flask application code `myapp.py` and it works there as well.

It works with `--preload` as well. 

What should we do here? Removing the `post_fork` suggestion?

cc @Kludex

alexmojaki:
No, I get the impression that it's still best to recommend configuring after forking, even if it's hard to reproduce the problem otherwise.

hramezani:
> No, I get the impression that it's still best to recommend configuring after forking, even if it's hard to reproduce the problem otherwise.

Thanks for the clarification. I pushed a commit and addressed your comments

alexmojaki:
@hramezani please can you link to this page from the flask and django docs?

hramezani:
> @hramezani please can you link to this page from the flask and django docs?

@alexmojaki Here is the [PR](https://github.com/pydantic/logfire/pull/1186)

PR #1186, opened 2025-06-24
Link to Gunicorn doc page from Django and Flask doc

hramezani:
Addresses https://github.com/pydantic/logfire/pull/1179#issuecomment-3000810063


alexmojaki:
thanks

PR #1189, opened 2025-06-25
add metrics to concepts

ChristopherGS:


docs/concepts.md
```diff
@@ -66,3 +66,27 @@ with logfire.span('Asking the user for their {question}', question='birthday'):
...
+import logfire
+
+logfire.configure()
+
+start = time.perf_counter()
+# … handle request …
+latency = time.perf_counter() - start
+
+# Emit a histogram metric with useful tags for slicing and alerting
+logfire.metric('http.request.duration', latency, unit='seconds', tags={'route': '/api/items', 'method': 'GET'})
```
alexmojaki: 👀

docs/concepts.md
```diff
@@ -1,9 +1,20 @@
+| Concept   | Description  |
+|-----------|--------------|
+| Span | Atomic unit of telemetry data |
+| Trace | Contains spans, tree structure shows parent/child relationships |
+| Metric | Values calculated using telemetry data |
+| Log | No duration, timestamped, emitted by services/loggers |
```
Kludex, line 14: I find it a bit weird to start the page with a table, and no link to the relevant sections. Maybe we can either add a text or link to the sections below?

docs/concepts.md
```diff
@@ -1,9 +1,20 @@
...
+
+## What is a Span? 
+
 A **span** is the building block of a trace. You might also think of spans as logs with extra functionality — a single row in our live view.
 
 !!! info
     Spans let you **add context** to your logs and **measure code execution time**. Multiple spans combine to form a trace, providing a complete picture of an operation's journey through your system.
 
-## Example #1
+### Example #1
```
Kludex: Can we actually change this name for what the example is about...?

GSemikozov (approved):
LGTM

docs/concepts.md
```diff
@@ -66,3 +77,56 @@ with logfire.span('Asking the user for their {question}', question='birthday'):
 By instrumenting your code with traces and spans, you can see how long operations take, identify bottlenecks,
 and get a high-level view of request flows in your system — all invaluable for maintaining the performance and
 reliability of your applications.
+
+## What is a Trace?
```
Viicos: Maybe a personal preference, but I find it clearer when the concept of traces is explained right after spans (that is _not_ after the examples), especially as they are  tightly coupled together.

It might also be worth explaining that the root span implicitly creates a trace?

Also wondering if we should emphasize more on the fact that traces are not limited to a single service — they can be propagated to completely different and isolated services (maybe with an example: Python API and SQL database).

docs/concepts.md
```diff
@@ -66,3 +77,56 @@ with logfire.span('Asking the user for their {question}', question='birthday'):
...
+
+def handle_request():
+    start = time.perf_counter()
+    # … handle the request …
+    duration_ms = (time.perf_counter() - start) * 1000
+    # Record the observed latency
+    request_duration.record(duration_ms)
+```
+
+Each call to request_duration.record() adds a sample to the histogram. On the backend you can visualise
```
Viicos:
```suggestion
Each call to `request_duration.record()` adds a sample to the histogram. On the backend you can visualise
```

docs/concepts.md
```diff
@@ -66,3 +77,56 @@ with logfire.span('Asking the user for their {question}', question='birthday'):
...
+p50/p95 latency, set SLOs, and trigger alerts whenever performance degrades. For more metrics examples
+see the [adding metrics guide](guides/onboarding-checklist/add-metrics.md).
+
+## What is a Log?
+
+Logs record something which happened in your application. Importantly, they do not have a duration, compared to spans and traces. 
+
+A log is a timestamped text record, either structured (recommended) or unstructured, with optional metadata. Of all telemetry signals, logs are the best known and have the largest footprint on our collective understanding. Most programming languages have built-in logging capabilities or well-known, widely used logging libraries. Most [Python users](https://docs.python.org/3/library/sys.html#sys.stdin) are familiar with `sys.stdin`, `sys.stdout` and `sys.stderr` for example.
```
Viicos: Not sure why `sys.stdin/out.err` was mentioned here, as they are just IO objects. Maybe better as:

```suggestion
A log is a timestamped text record, either structured (recommended) or unstructured, with optional metadata. Of all telemetry signals, logs are the best known and have the largest footprint on our collective understanding. Most programming languages have built-in logging capabilities or well-known, widely used logging libraries. Most Python users are familiar with the [`logging`][] module for example.
```

docs/concepts.md
summerscope: Added an overview section.

docs/concepts.md
summerscope: I've added an admonition for this @Viicos - let me know if you like that or you think we should also add a code example?

summerscope: (I also reordered as you suggested, which I thought was nice)

PR #1192, opened 2025-06-25
logfire export script

samuelcolvin:
Not adding docs to this until we agree it's a good idea.


dmontagu:
Closing this draft — it's been open well over a year without a decision on whether we want it. Easy to resurrect from the branch if we do.

Issue #1193, opened 2025-06-25
`AsyncLogfireQueryClient` should have an `info()` method


samuelcolvin:
that calls `/read-token-info`.

hramezani:
Here is the PR https://github.com/pydantic/logfire/pull/1204

PR #1204, opened 2025-06-27
Add info method to logfire query clients

hramezani:
Fixes https://github.com/pydantic/logfire/issues/1193

### DIFF
--- a/docs/how-to-guides/query-api.md
+++ b/docs/how-to-guides/query-api.md
@@ -68,10 +68,6 @@ Here's an example of how to use these clients:
         """
 
         async with AsyncLogfireQueryClient(read_token='<your_read_token>') as client:
-            # Get read token info
-            read_token_info = await client.info()
-            print(read_token_info)
-
             # Load data as JSON, in column-oriented format
             json_cols = await client.query_json(sql=query)
             print(json_cols)
@@ -91,6 +87,10 @@ Here's an example of how to use these clients:
             # JSON-serialized strings
             df_from_csv = pl.read_csv(StringIO(await client.query_csv(sql=query)))
             print(df_from_csv)
+
+            # Get read token info
+            read_token_info = await client.info()
+            print(read_token_info)
 
 
     if __name__ == '__main__':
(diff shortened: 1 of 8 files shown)


docs/how-to-guides/query-api.md
```diff
@@ -68,6 +68,10 @@ Here's an example of how to use these clients:
         """
 
         async with AsyncLogfireQueryClient(read_token='<your_read_token>') as client:
+            # Get read token info
```
alexmojaki: I'm not sure we need to document this at all, but if we do I think it should be less prominent, not in this usage example. This is not something I expect to be used much, I don't even know why @samuelcolvin wants it.

samuelcolvin: Because it's useful to go from a read token to a project and org name.

I think it should go in this example, but last.

logfire-api/logfire_api/experimental/query_client.pyi
```diff
@@ -15,6 +16,14 @@ class QueryExecutionError(RuntimeError):
 class QueryRequestError(RuntimeError):
     """Raised when the query request is invalid."""
 
+class ReadTokenInfo(TypedDict):
+    """Information about the read token."""
+    token_id: UUID
+    organization_id: UUID
```
samuelcolvin: I don't think anyone cares about token_id, org_id or project_id, and we might want to remove them.

So I think we should use a Pydantic model here with just the two name fields.

samuelcolvin: Also Pydantic model makes sure we get an error at runtime if the Schema changes.

samuelcolvin (commented):
Otherwise lgtm.

docs/how-to-guides/query-api.md
alexmojaki: I imagine it might be useful for generic things like your export script. But I expect users to know where their read tokens come from, so I can't imagine what users would do with this.

logfire-api/logfire_api/experimental/query_client.pyi
alexmojaki: logfire doesn't require pydantic.

the query client has additional 'dependencies': httpx and optionally pyarrow for some methods. but still not pydantic.

hramezani:
I addressed the comments.

alexmojaki (approved):

PR #1251, opened 2025-07-16
revert admonitions in .md files

GSemikozov:
In a previous PR when editing md files, the indents were incorrectly formatted in some places, which is why the description "fell out" of admonitions, so here I revert back correct formatting.

For instance, now:
Screenshot 2025-07-16 at 15 10 34

How it should be:
Screenshot 2025-07-16 at 15 10 44


alexmojaki:
???

GSemikozov:
> ???

added description.

Issue #710, opened 2024-12-20
Logfire displays time on the terminal that does not match the system time.


dlivxpr:
### Description

As the title suggests, I am using logfire while using pydantic-ai. However, when executing the code, the time displayed by logfire in the terminal is 8 hours later than the system time. I suspect it is a time zone issue, but I have checked the logfire documentation and found no time zone related settings. Is there a way to make logfire display log time in the local terminal as the system time?

### Python, Logfire & OS Versions, related packages (not required)

```TOML
logfire="2.8.0"
platform="Windows-11-10.0.26100-SP0"
python="3.13.1 (main, Dec  6 2024, 18:07:41) [MSC v.1942 64 bit (AMD64)]"
[related_packages]
requests="2.32.3"
pydantic="2.10.3"
fastapi="0.115.6"
openai="1.58.1"
protobuf="5.29.1"
rich="13.9.4"
executing="2.1.0"
opentelemetry-api="1.29.0"
opentelemetry-exporter-otlp-proto-common="1.29.0"
opentelemetry-exporter-otlp-proto-http="1.29.0"
opentelemetry-instrumentation="0.50b0"
opentelemetry-proto="1.29.0"
opentelemetry-sdk="1.29.0"
opentelemetry-semantic-conventions="0.50b0"
```

alexmojaki:
It intentionally uses the UTC timezone. Maybe we should make that configurable.

Kludex:
> Maybe we should make that configurable.

A `timezone` parameter to the `SimpleConsoleSpanExporter`?

---

I don't think this should be a UTC tho, the deployment environments are usually configured to UTC. I think it should just be local time.

PR #1255, opened 2025-07-17
Remove UTC timezone defaults for console logs

fswair:
Current logfire console using utc time for logs, as in **/logfire/_internal/exporters/console.py:198**
So, I have removed the code part `tz=timezone.utc` and now tz set as datetime defaults (local time).

Fixes: https://github.com/pydantic/logfire/issues/710


fswair:
See: https://github.com/pydantic/logfire/issues/710#issuecomment-2560812838

alexmojaki:
Thanks!

Issue #1254, opened 2025-07-16
JsonArgsValueFormatter._format_items does not handle None


safetythird:
### Description

When using Logfire to auto-instrument a PydanticAI Agent with verbose output on, I encountered the following error:

```
Exception while exporting Span.
Traceback (most recent call last):
  File "/Users/myusername/github/myapp/.venv/lib/python3.13/site-packages/opentelemetry/sdk/trace/export/__init__.py", line 109, in on_end
    self.span_exporter.export((span,))
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/myusername/github/myapp/.venv/lib/python3.13/site-packages/logfire/_internal/exporters/console.py", line 148, in export
    self.export_record(Record.from_span(span))
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/myusername/github/myapp/.venv/lib/python3.13/site-packages/logfire/_internal/exporters/console.py", line 417, in export_record
    self._print_span(span)
    ~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/myusername/github/myapp/.venv/lib/python3.13/site-packages/logfire/_internal/exporters/console.py", line 181, in _print_span
    self._print_arguments(span, indent_str)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/Users/myusername/github/myapp/.venv/lib/python3.13/site-packages/logfire/_internal/exporters/console.py", line 268, in _print_arguments
    value = json_args_value_formatter(value, schema=schema)
  File "/Users/myusername/github/myapp/.venv/lib/python3.13/site-packages/logfire/_internal/json_formatter.py", line 62, in __call__
    self._format(indent_current, True, value, schema)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/myusername/github/myapp/.venv/lib/python3.13/site-packages/logfire/_internal/json_formatter.py", line 72, in _format
    self._format_items('{', ': ', '}', True, indent_current, value, None)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1256, opened 2025-07-18
🐛 Fix str with safe json loads for exceptional values

sbhrwlr:
Validation Handling in JSON Field Parsing
- Added logic to parse string values using json.loads.
- If the parsed value is a primitive (int, float, bool, or None), skip schema processing.
- Ensures quoted primitives like "123" or "null" are preserved as strings.
- Prevents schema application on non-JSON or raw primitives.


alexmojaki:
also please check/fix the case where the value is a list but the schema says object, or the schema says array but the value is a dict.

sbhrwlr:
@alexmojaki As you suggested to check other variations like list/dict and corresponding to object/array types, I tried to use the piece of code you provided in #1254 foo is not accepting dict with object type as well, can you guide me with this or if this is expected for attributes types.

```python
import json

import logfire

tracer = logfire.DEFAULT_LOGFIRE_INSTANCE.config.get_tracer_provider().get_tracer('tracer')
schema = {'type': 'object', 'properties': {'foo': {'type': 'object'}}}
data = {"id": 1, "name": "Alice", "active": True}
span = tracer.start_span(
    'foo',
    attributes={'foo': data, 'logfire.json_schema': json.dumps(schema)},
)
span.end()
```

alexmojaki:
Try `json.dumps(data)`.

alexmojaki:
Sorry @sbhrwlr I'm realising I misunderstood you. If the schema is nonempty, then it should always try to parse the string as JSON, and go with the parsed value if that succeeds. So e.g. `"null"` with the schema `{"type": "object"}` should be parsed to `None` and displayed as such. It's only after parsing JSON that the schema should be questioned more.

Adarshthakur-850:
In short:
To fix it, **always parse the string as JSON first** (e.g. `"null"` → `None`).
Then, **validate the parsed value against the schema**.
Update the schema to allow `null` if needed:

```json
{ "type": ["object", "null"] }
```

**Never validate the raw string before JSON parsing.**

Adarshthakur-850:
In short:
To fix it, **always parse the string as JSON first** (e.g. `"null"` → `None`).
Then, **validate the parsed value against the schema**.
Update the schema to allow `null` if needed:

```json
{ "type": ["object", "null"] }
```

**Never validate the raw string before JSON parsing.**

alexmojaki:
No, don't 'validate', and definitely don't update the schema. Just use the schema where it makes sense as a loose guide for how to display the value.

sbhrwlr:
@alexmojaki In cases where the parsed JSON value is a list or dict, but the schema specifies a mismatched type (e.g., 'object' vs 'array'), should we still attempt to interpret the schema for display purposes, or bypass schema logic entirely in such mismatches?

alexmojaki:
I don't even know how you would use an object schema to interpret the contents of an array anyway. You can stop using the schema the moment it becomes unclear.

Really the primary goals are: (1) preserve existing behaviour where it seems sensible and (2) prevent errors.

sbhrwlr:
https://github.com/pydantic/logfire/pull/1289 added changes here @alexmojaki

PR #1265, opened 2025-07-21
min_level

alexmojaki:


logfire/_internal/config.py
```diff
@@ -320,6 +322,14 @@ def configure(  # noqa: D417
 
             Defaults to `True` if and only if the Python version is at least 3.11.
 
+        min_level:
+            Minimum log level for logs and spans to be created. By default, all logs and spans are created.
+            For example, set to 'info' to only create logs with level 'info' or higher, thus filtering out debug logs.
+            For spans, this only applies when `_level` is explicitly specified in `logfire.span`.
+            Setting the level after will be ignored by this.
```
dmontagu:
```suggestion
            Changing the level of a span _after_ it is created will be ignored by this.
```

logfire/_internal/config.py
```diff
@@ -320,6 +322,14 @@ def configure(  # noqa: D417
...
+            If a span is not created, this has no effect on the current active span, or on logs/spans created inside the
+            filtered `logfire.span` context manager.
+            If set to `None`, uses the `LOGFIRE_MIN_LEVEL` environment variable.
```
dmontagu:
```suggestion
            If set to `None`, uses the `LOGFIRE_MIN_LEVEL` environment variable; if that is not set, there is no minimum level.
```

logfire/_internal/config.py
```diff
@@ -514,6 +525,9 @@ class _LogfireConfigData:
     sampling: SamplingOptions
     """Sampling options."""
 
+    min_level: int
```
dmontagu, line 528: Given this must be an integer I guess we should be clear that 0 (or whatever the min value we want is) means it will affect no spans

logfire/_internal/integrations/fastapi.py
```diff
@@ -294,8 +294,11 @@ async def run_endpoint_function(
                 # both `http.method` and `http.route`.
                 **{'http.route': request.scope['route'].path},
                 **stack_info,
-                _level='debug',
             )
+            # This span is level debug mostly for legacy reasons.
+            # We set the level after starting so that it's not excluded by the min_level setting,
+            # since this span is requested explicitly.
+            extra_span.set_level('debug')
```
dmontagu: Is this really the desired behavior? I'm thinking if you are opting into having this span, and have min level info (rather than debug), maybe we should just make this span be info level? It feels weird to me that I can set min level to be info, and yet still get debug spans coming through as long as they are the _right_ debug spans..

logfire/_internal/config.py
alexmojaki: This isn't user facing.

logfire/_internal/integrations/fastapi.py
alexmojaki: I don't know the right answer. I was just hesitant about changing the level of the span.

alexmojaki:
Closes https://linear.app/pydantic/issue/PYD-1365/how-to-guide-not-logging-all-traces

docs/guides/onboarding-checklist/add-manual-tracing.md
```diff
@@ -270,3 +262,44 @@ with logfire.span('Doing a thing') as span:
...
+   That makes this line a complete no-op.
+
+creates, logs, and sends the `root` span and the `info message` log, with the log being a direct child of the span.
+
+To set the minimum level for console logging only (`info` by default):
+
+```python
+import logfire
+
+logfire.configure(console=logfire.ConsoleOptions(min_log_level='debug'))
```
dmontagu, line 304: Feels unfortunate that this is `min_log_level` and the other is `min_level` but I guess perhaps it makes it easier to distinguish..

dmontagu: and I guess not worth a breaking change anyway

dmontagu (approved):

PR #1273, opened 2025-07-25
Create devcontainer.json

goodgollyholly:


Kludex:
I don't think this is very useful without defining the setup properly, but also, no one from our team uses it, so it will be hard to maintain the devcontainer. Besides... It's easier to just do `make install` in this repository.

PR #1287, opened 2025-08-01
Remove text that older mypy versions don't support

metanivek:
On older versions of `mypy`, I get this error:

```
<project-dir>/.venv/lib/python3.10/site-packages/logfire/_internal/utils.py:182: error: Invalid "type: ignore" comment  [syntax]
```


alexmojaki:
Thanks!

Issue #989, opened 2025-04-10
`logfire_api.LogfireSpan` should support `ReadableSpan` attributes


alexmojaki:
https://github.com/pydantic/logfire/pull/983 should be generalised. The real `LogfireSpan` is a subclass of `ReadableSpan`. The fake one shouldn't be but should delegate to `ReadableSpan(name='')` (the other arguments have defaults).

RichardMarto:
Hey @alexmojaki I’ve took a shot on that, any feedback is welcomed.

PR #1288, opened 2025-08-03
Closes logfire_api.LogfireSpan should support ReadableSpan attributes #989

RichardMarto:


tests/test_logfire_api.py
```diff
@@ -69,6 +69,19 @@ def test_runtime(logfire_api_factory: Callable[[], ModuleType], module_name: str
     with logfire_api.span('test span') as span:
         assert isinstance(span, logfire_api.LogfireSpan)
         span.set_attribute('foo', 'bar')
+        assert hasattr(span, 'context')
```
alexmojaki: These tests would have passed before this change. Assert the actual values.

logfire-api/logfire_api/__init__.py
```diff
@@ -19,8 +20,11 @@
         def configure(*args, **kwargs): ...
 
         class LogfireSpan:
+            def __init__(self, *args, **kwargs):
+                self._readable_span = ReadableSpan(name='')
+
             def __getattr__(self, attr):
-                return MagicMock()  # pragma: no cover
+                return self._readable_span.attributes.get(attr) # pragma: no cover
```
alexmojaki:
```suggestion
                return getattr(self._readable_span, attr, MagicMock())
```

You're confusing OTel attributes with Python attributes. And we still want to cover all cases.

tests/test_logfire_api.py
RichardMarto: Hmmm, not sure I got it, as this test is parametrize, the actual values would change depending if its the fake or the actual concrete class, should I add a new test just for the fake one?

alexmojaki: a new test or just an if statement here

RichardMarto: sorry about the delay, hope its better now.
I’ve ended up removing the test for `instrumentation_info` as it is deprecated and testing it fails some jobs.

tests/test_logfire_api.py
```diff
@@ -69,6 +71,21 @@ def test_runtime(logfire_api_factory: Callable[[], ModuleType], module_name: str
     with logfire_api.span('test span') as span:
         assert isinstance(span, logfire_api.LogfireSpan)
         span.set_attribute('foo', 'bar')
+
+        if module_name == 'logfire_api.':
+            assert getattr(span, 'context') is None
+            assert getattr(span, 'name') == ''
```
alexmojaki:
```suggestion
            assert span.context is None
            assert span.name == ''
```

etc

RichardMarto: gotcha, just a sec

alexmojaki:
Thanks!

logfire-api/logfire_api/__init__.py
```diff
@@ -48,8 +52,7 @@ def message(self, message: str): ...  # pragma: no cover
             def is_recording(self) -> bool:  # pragma: no cover
                 return False
 
-            @property
-            def context(self): ...  # pragma: no cover
```
DouweM, line 52: Why was `context` removed here? This reverted https://github.com/pydantic/logfire/pull/983, reintroducing https://github.com/pydantic/pydantic-ai/issues/1399

cc @alexmojaki

DouweM: Judging by the test, it was working in this PR, but then https://github.com/pydantic/logfire/pull/1310 reverted some of this but not `context` (and removed the test), and now it's broken again 😅

RichardMarto: Hmmm, weird, I thought my `__getattr__` implementation would replace that, any work needed?

DouweM: @RichardMarto I think this code was OK, but then that PR I linked effectively reverted it

RichardMarto: Yeah, I guess by delegating the mocks properties to `ReadableSpan` added it as a dependency somewhere it shouldn’t be, anyway, I’m happy to help.

alexmojaki: ugh, sorry @DouweM , I'd forgotten about all that and in my PR was thinking "surely it's not important for attributes to be None", I just focused on things with more interesting types. I guess it actually is important for things to be None.

Also the whole idea of https://github.com/pydantic/logfire/issues/989 was clearly flawed in the first place.

alexmojaki: https://github.com/pydantic/logfire/pull/1317

PR #1289, opened 2025-08-03
added safe json loads for str types values

sbhrwlr:


sbhrwlr:
@alexmojaki please have a look.

alexmojaki:
This still fails on a slight modification of the original repro:

```python
import json

import logfire

tracer = logfire.DEFAULT_LOGFIRE_INSTANCE.config.get_tracer_provider().get_tracer('tracer')
schema = {'type': 'object', 'properties': {'foo': {'type': 'object'}}}
span = tracer.start_span(
    'foo',
    attributes={'foo': 3, 'logfire.json_schema': json.dumps(schema)},
)
span.end()
```

logfire/_internal/exporters/console.py
```diff
@@ -251,6 +251,18 @@ def _details_parts(self, span: Record, indent_str: str) -> TextParts:
         else:
             return []
 
+    def is_schema_valid(self, schema: dict[str, Any], value: Any) -> bool:
+        """Check if the schema is valid for the value."""
+        schema_type = schema.get('type') if schema else None
+        if schema_type:
+            if value is None:
+                return False
+            elif isinstance(value, dict) and schema_type != 'object':
```
alexmojaki: i think `if schema_type == 'object' and not isinstance(value, dict):` would work better

sbhrwlr: makes sense I've made the changes

logfire/_internal/exporters/console.py
```diff
@@ -251,6 +251,18 @@ def _details_parts(self, span: Record, indent_str: str) -> TextParts:
         else:
             return []
 
+    def is_schema_valid(self, schema: dict[str, Any], value: Any) -> bool:
+        """Check if the schema is valid for the value."""
+        schema_type = schema.get('type') if schema else None
+        if schema_type:
+            if value is None:
+                return False
```
alexmojaki:
```suggestion
```

you shouldn't need this now.

please test https://github.com/pydantic/logfire/pull/1289#issuecomment-3150200122

alexmojaki:
Anything wrong with this simpler approach? https://github.com/pydantic/logfire/pull/1294

sbhrwlr:
@alexmojaki The logic is placed well, but the deep nesting of if-else blocks could impact readability. We should consider using helper methods to simplify it and that will make this modular as well.

alexmojaki:
Thanks for looking. Yes, it would be great if that could be refactored, but that PR is not the problem.

PR #1294, opened 2025-08-04
Handle wrong JSON schema in console exporter

alexmojaki:


alexmojaki:
Thanks @sbhrwlr for the initial work!

PR #1310, opened 2025-08-12
Fix logfire_api requiring OpenTelemetry SDK

alexmojaki:
Closes https://github.com/pydantic/pydantic-ai/issues/2516


Viicos:
Thanks @alexmojaki, would it make sense to have a test that importing `logfire_api` succeeds in a fresh environment, without any dependencies installed to prevent such issues in the future?

alexmojaki:
That would be ideal, yes.

PR #1337, opened 2025-08-20
fix: handle optional args in logfire.instrument

### LINKED ISSUE #1333: Extracting optional args results in a warning
### Description

When I use either `extract_args` or specify  in the message e.g. `Calling method {optional1=}` I get an error:
```log
FormattingFailedWarning: 
    Ensure you are either:
      (1) passing an f-string directly, with inspect_arguments enabled and working, or
      (2) passing a literal `str.format`-style template, not a preformatted string.
    See https://logfire.pydantic.dev/docs/guides/onboarding-checklist/add-manual-tracing/#messages-and-span-names.
    The problem was: The field {optional1} is not defined.
```

Here's an example, where I explicitly want just `optional1` and not `self`:
```python
@logfire.instrument("Calling myfunc {optional1=}")
async def myfunc(self, optional1: str | None = None) -> None:
```

The expected behaviour is that the value of `optional1` is output and not a `FormattingFailedWarning` even if `optional1` wasn't passed, in which case it should use the provided default, in this case `None`.

The same behaviour happens if using `extract_args`.

Related enhancement to [allow excluding args from extraction](https://github.com/pydantic/logfire/issues/1332).

```TOML
logfire="4.3.3"
platform="macOS-15.6-arm64-arm-64bit"
python="3.12.9 (main, Mar 17 2025, 21:36:21) [Clang 20.1.0 ]"
[related_packages]
requests="2.32.4"
pydantic="2.11.7"
fastapi="0.116.1"
openai="1.99.9"
protobuf="5.29.5"
rich="14.1.0"
executing="2.2.0"
opentelemetry-api="1.36.0"
opentelemetry-exporter-otlp-proto-common="1.36.0"
opentelemetry-exporter-otlp-proto-http="1.36.0"
opentelemetry-instrumentation="0.57b0"
opentelemetry-instrumentation-httpx="0.57b0"
opentelemetry-instrumentation-system-metrics="0.57b0"
opentelemetry-proto="1.36.0"
opentelemetry-sdk="1.36.0"
opentelemetry-semantic-conventions="0.57b0"
opentelemetry-util-http="0.57b0"
```

stevenh:
Allow optional arguments to be specified in both the `msg_template` and the `extract_args` parameter of the `logfire.instrument` decorator.

Fixes #1333


logfire/_internal/instrument.py
```diff
@@ -63,19 +63,28 @@ def decorator(func: Callable[P, R]) -> Callable[P, R]:
         attributes = get_attributes(func, msg_template, tags)
         open_span = get_open_span(logfire, attributes, span_name, extract_args, func)
 
+        def open_span_defaults(
+            func: Callable[P, R], *func_args: P.args, **func_kwargs: P.kwargs
+        ) -> AbstractContextManager[Any, bool | None]:
+            """Open a span with default arguments bound from the function."""
+            sig = inspect.signature(func)
+            bound = sig.bind(*func_args, **func_kwargs)
+            bound.apply_defaults()
```
alexmojaki: all of this needs to happen inside `open_span`. in general that should mostly mean just adding `bound.apply_defaults()` in some places.

stevenh: Thanks for the pointer updated.

PR #1355, opened 2025-08-26
Add experimental `exception_callback` configuration

alexmojaki:
Also fixes some minor issues in how exceptions were being recorded


logfire/_internal/main.py
```diff
@@ -755,7 +760,7 @@ def log(
                 if isinstance(exc_info, tuple):
                     exc_info = exc_info[1]
                 if isinstance(exc_info, BaseException):
-                    record_exception(span, exc_info)
+                    span.record_exception(exc_info)
```
alexmojaki, line 767: This fixes the fact that `record_exception` was still being called twice in the stack in this case.

logfire/_internal/main.py
```diff
@@ -2393,11 +2398,7 @@ def record_exception(
         if not self._span.is_recording():
             return
 
-        span = self._span
-        while isinstance(span, _LogfireWrappedSpan):
-            span = span.span
-        record_exception(
-            span,
+        self._span.record_exception(
```
alexmojaki, line 2466: This ensures that we use our test timestamp generator so tests don't need IsInt

logfire/_internal/config.py
```diff
@@ -182,6 +183,11 @@ class AdvancedOptions:
     log_record_processors: Sequence[LogRecordProcessor] = ()
     """Configuration for OpenTelemetry logging. This is experimental and may be removed."""
 
+    exception_callback: ExceptionCallback | None = None
+    """Callback function that is called when an exception is recorded on a span.
+
+    This is experimental and may be removed."""
```
dmontagu:
```suggestion
    This is experimental and may be modified or removed."""
```

Issue #1361, opened 2025-09-01
FastAPI exception handlers: properly handled exceptions logged as errors in Logfire


renaudcepre:
### Question

# FastAPI exception handlers: properly handled domain exceptions logged as errors in Logfire

## Summary
When using Logfire's `instrument_fastapi()`, exceptions that are properly handled by FastAPI exception handlers are still logged as errors in Logfire, even though they result in valid HTTP responses (404, 403, etc.). This pollutes error logs with normal business logic cases.

## Expected Behavior
Exceptions that are handled by FastAPI exception handlers and converted to proper HTTP responses (4xx status codes) should NOT be logged as errors in Logfire. They should either not be logged as exceptions at all

## Actual Behavior
All exceptions are logged as errors in Logfire, regardless of whether they're handled by exception handlers.

## Code Example

**Exception Handler Setup:**
```python
from fastapi import FastAPI
from fastapi.responses import JSONResponse
from generic_rag.domain.exceptions.base import DomainError

@app.exception_handler(DomainError)
async def domain_error_handler(request: Request, exc: DomainError) -> Response:
    return JSONResponse(status_code=404, content={"detail": str(exc)})

# Logfire instrumentation
logfire.instrument_fastapi(app, capture_headers=True)
```

**Route that raises handled exception:**
```python
@router.get("/users/{user_id}")
def get_user(user_id: UUID) -> User:
    # This raises UserNotFoundError which inherits from DomainError
    return domain.get_user_for_company(user_id, api_user)
```

## Result in Logfire
```json
{
  "is_exception": true,
  "exception_type": "generic_rag.domain.exceptions.users.UserNotFoundError", 
  "otel_status_code": "ERROR",
  "http_response_status_code": 404
}
```

<img width="251" height="30" alt="Image" src="https://github.com/user-attachments/assets/fa2d39f1-1e75-425c-a9db-980fe0df9418" />

The HTTP response is correctly 404, but Logfire still logs it as an error.

## Workaround
Currently, the only workaround is to wrap every route in try/catch, which contradicts the use of error handler:

```python
@router.get("/users/{user_id}")  
def get_user(user_id: UUID) -> User:
    try:
        return domain.get_user_for_company(user_id, api_user)
    except DomainError as exc:
        # Handle manually to avoid Logfire logging as error
        return JSONResponse(status_code=404, content={"detail": str(exc)})
```

With this approach:
```json
{
  "is_exception": false,
  "otel_status_code": "UNSET", 
[...]

PR #1363, opened 2025-09-02
fastapi record handled exceptions

renaudcepre:


dmontagu:
Closing as superseded: #1355 landed `logfire.types.ExceptionCallback`, which is the supported way to stop recording exceptions that your handlers deal with — see the discussion in #1361. Thanks for kicking this off.

PR #1381, opened 2025-09-07
Add ensure_ascii=False to json.dumps calls

qiangxinglin:


alexmojaki:
Hi, what's the motivation for this?

qiangxinglin:
@alexmojaki For non-English words, logfire frontend can correctly rendered it. But when the traces too old and past its retention period, importing them to other observability platforms (like [Signoz](https://signoz.io/)) would render the words in \uXXXX\uYYYY form.

alexmojaki:
Makes sense, thanks!

Issue #1324, opened 2025-08-15
Unable to import `propagate` from logfire_api


smthngslv:
### Description

Hi, while writing custom integration faced with problem:

```python
from logfire_api.propagate import get_context, attach_context
```

or 

```python
import logfire_api.propagate
```

are failed with 

```
ModuleNotFoundError: No module named 'logfire_api.propagate'
```

However,

```
import logfire_api as logfire
```

works well. But does not allow to access `propagate` module anyway:

```
import logfire_api as logfire
print(logfire.propagate)
```

fails with

```
AttributeError: module 'logfire_api' has no attribute 'propagate'
```

If I install `logfire` itself - works well.

---

Python: 3.13.5/3.12.11
Logfire API: 4.3.3
Steps to reproduce:

```
uvx --from logfire_api==4.3.3 python
```

and then just

```
from logfire_api.propagate import get_context, attach_context
```
---

Thanks!

### Python, Logfire & OS Versions, related packages (not required)

```TOML
>>> import logfire_api as logfire
>>> logfire.logfire_info()
'logfire_info() is not implement by logfire-api'
```

alexmojaki:
We should just export `get_context, attach_context` from `logfire` directly, i.e. the root `__init__.py`.

dhruv-ahuja:
Can I pick this up, if that's okay? @alexmojaki

alexmojaki:
Done, currently releasing

PR #1385, opened 2025-09-08
Export Context Propagate APIs through Logfire Interface

dhruv-ahuja:
This PR addresses feedback provided in the issue #1324 by exporting `propagate.py`'s `get_context` and `attach_context` APIs through the Logfire module interface.


alexmojaki:
Thanks! Please also update docs where possible, e.g. https://logfire.pydantic.dev/docs/how-to-guides/distributed-tracing/#manual-context-propagation. The old usage is still correct but it'd be nice to show the new simpler usage.

docs/how-to-guides/distributed-tracing.md
```diff
@@ -5,7 +5,7 @@
 **Logfire** provides a thin wrapper around the OpenTelemetry context propagation API to make manual distributed tracing easier. You shouldn't usually need to do this yourself, but it demonstrates the concept nicely. Here's an example:
 
 ```python
-from logfire.propagate import attach_context, get_context
+from logfire import attach_context, get_context
```
alexmojaki: i meant remove this line and rely only on `import logfire`

dhruv-ahuja: Got it, I wasn't sure whether we wanted to keep direct function imports in this case, I've fixed this.

Is it okay to assume that in such cases where we don't have lot of imports, we always want to access APIs through the `logfire` package?

alexmojaki: I think it's very nice to always be able to use `import logfire; logfire.<something>`, it's simple and consistent. `from logfire import configure, info, error, warn, span` wouldn't generally work well. So I want the docs to stick to `import logfire` wherever possible.

docs/how-to-guides/sampling.md
```diff
@@ -246,10 +246,10 @@ uvicorn.run(app)
 ```
 
 A workaround is to explicitly put the new spans in their own trace using [
-`attach_context`][logfire.propagate.attach_context]:
+`attach_context`][logfire.attach_context]:
 
 ```python
-from logfire.propagate import attach_context
+from logfire import attach_context
```
alexmojaki: Please change it to `import logfire` everywhere

dhruv-ahuja:
@alexmojaki I have cross-checked and ensured all code examples and documentation is using `import logfire` everywhere.
Please check now and let me know if anything else is needed.

Also once done, please let me know what issue I can pick up next.

logfire-api/logfire_api/__init__.py
```diff
@@ -6,6 +6,8 @@
 from typing import Any, ContextManager, Literal, TYPE_CHECKING, Sequence
 from unittest.mock import MagicMock
 
+from logfire.propagate import ContextCarrier
```
alexmojaki: The point of logfire-api is that logfire isn't installed.

dhruv-ahuja: Understood, had missed that - have gone through the readme now.

Would you like me to type hint the function return as `Mapping[str, Any]` or define the type alias inline?

We would have to define the type alias at the root/global level as it is not being recognized as a valid type under the `except ImportError:` code block.

So it'd look like:
```py
ContextCarrier = Mapping[str, Any]

try:
    logfire_module = importlib.import_module('logfire')
    sys.modules[__name__] = logfire_module
...
```

alexmojaki: I don't think a type alias is needed here. not even sure if it needs a type hint at all.

dhruv-ahuja: Okay, I have simplified the code. Please check now.

logfire-api/logfire_api/__init__.py
```diff
@@ -303,8 +302,7 @@ def get_baggage(*args, **kwargs) -> dict[str, str]:...
         def set_baggage(*args, **kwargs) -> ContextManager[None]:
             return nullcontext()
 
-        def get_context(*args, **kwargs) -> ContextCarrier:
-            return {}
+        def get_context(*args, **kwargs) -> dict[str, Any]: ...
```
alexmojaki: Needs to `return {}`. The other functions here which are typed as returning dicts but don't are also wrong and should be fixed.

dhruv-ahuja: Done

alexmojaki:
Thanks! If you'd like to do more take a look at https://github.com/pydantic/logfire/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22

PR #1396, opened 2025-09-10
Fix Root Model Not Rendering As Expected

### LINKED ISSUE #1107: Pydantic RootModel not rendered correctly

dhruv-ahuja:
This PR fixes #1107.
Brings over core logic introduced in #1108 PR with additional required changes to make schema non-trivial and ensure apt formatter parsing.

Logs for `RootModel`s encapsulating string literals:
```
uvp ../rootmodel.py

RootModel[NoneType](root=None)
21:17:00.213 MyRootModel [NoneType]
             │ ../rootmodel.py:48 info
             │ r1=None
             │ r2=None

RootModel[int](root=-9991)
21:17:00.233 MyRootModel [int]
             │ ../rootmodel.py:48 info
             │ r1=-9991
             │ r2=-9991

RootModel[bool](root=False)
21:17:00.235 MyRootModel [bool]
             │ ../rootmodel.py:48 info
             │ r1=False
             │ r2=False

RootModel[float](root=2.0)
21:17:00.236 MyRootModel [float]
             │ ../rootmodel.py:48 info
             │ r1=2.0
             │ r2=2.0

RootModel[str](root='raw_str')
21:17:00.237 MyRootModel [str]
             │ ../rootmodel.py:48 info
             │ r1='raw_str'
             │ r2='raw_str'

```


dhruv-ahuja:
@alexmojaki I was thinking of something like this for the problem discussed in #1108, where `r2` schema is trivial, meaning it won't be formatted and data literals in this case will remain in a string form, eg. `'2.0'` or `'"something"'`.

If this is the correct approach, how can I test it out locally if I register `string` data type in `JsonArgsValueFormatter` class?


image

alexmojaki:
```python
from pydantic import RootModel

import logfire

MyRootModel = RootModel[str]

r = MyRootModel('test')
logfire.info('MyRootModel', r1=r, r2=r.root)
#              │ r1='"test"'  # this is bad
#              │ r2='test'
```

logfire/_internal/json_encoder.py
```diff
@@ -240,7 +248,7 @@ def encoder_by_type() -> dict[type[Any], EncoderFunction]:
 
 def to_json_value(o: Any, seen: set[int]) -> JsonValue:
     try:
-        if isinstance(o, (int, float, str, bool, type(None))):
+        if isinstance(o, (str, int, bool, float, type(None))):
```
dhruv-ahuja: undo this order change

logfire/_internal/json_encoder.py
```diff
@@ -55,6 +55,13 @@ def _to_str(o: Any, _seen: set[int]) -> str:
     return str(o)
 
 
+# def _str_encoder(o: Any, _seen: set[int]) -> str:
```
dhruv-ahuja: remove this dead code

dhruv-ahuja:
@alexmojaki can you please check my changes thus far?
Got the main logic working, however facing an issue with IsJson values being converted to normal dictionaries, like [here in CI test](https://github.com/pydantic/logfire/actions/runs/17651495061/job/50163449005?pr=1396#step:9:415).

Please provide pointers on how to deal with this. Meanwhile will try to fix this.
Also please ignore print statements and other testing code, will cleanup at the end.

alexmojaki:
No we only want to change the schema for strings that were inside RootModels since they get an extra layer of JSON encoding that we can't really avoid.

logfire/_internal/json_schema.py
```diff
@@ -286,6 +287,16 @@ def _pydantic_model_schema(obj: Any, seen: set[int]) -> JsonDict:
     import pydantic
 
     assert isinstance(obj, pydantic.BaseModel)
+
+    # generate the schema for the actual wrapped object in the model
+    if isinstance(obj, pydantic.RootModel):
+        # return a complex schema to ensure JSON parsing for strings inside RootModel since they get an
+        # extra layer of JSON encoding
+        if obj.root.__class__ is str:  # type: ignore
+            return {'type': 'string', 'x-python-datatype': 'string'}
```
alexmojaki: Please also check what happens with ints and None

dhruv-ahuja:
@alexmojaki I have made the required changes and updated description with Logfire dashboard screenshots, please check.

edit: added `string` datatype to `json_formatter.py` as I assume it was most likely failing on the client side, encountered `unsupported datatype` error using the `ConsoleSpanExporter`.

logfire/_internal/json_schema.py
dhruv-ahuja: Okay, let me do that and get back to you

dhruv-ahuja: Tested using this script, added a print statement to json_schema.py as well:

```py
import json

from pydantic import BaseModel, RootModel
import logfire

logfire.configure(
    console=logfire.ConsoleOptions(verbose=True),
    send_to_logfire=False,
)


r_instance = RootModel[None](root=None)
# r_instance = RootModel[int](root=-9991)
# r_instance = RootModel[bool](root=False)
# r_instance = RootModel[float](root=2.0)
# r_instance = RootModel[str](root="raw_str")
print(repr(r_instance))
```


float:
```
RootModel[float](root=2.0)
schema gen for root model; obj=root=2.0 root=2.0 class=<class 'float'>
17:40:06.389 MyRootModel
             │ ../rootmodel.py:62 info
             │ r1='2.0'
             │ r2=2.0
```

bool:
```
RootModel[bool](root=False)
schema gen for root model; obj=root=False root=False class=<class 'bool'>
17:41:05.456 MyRootModel
             │ ../rootmodel.py:62 info
             │ r1='false'
             │ r2=False
```

integer:
```
RootModel[int](root=-9991)
schema gen for root model; obj=root=-9991 root=-9991 class=<class 'int'>
17:41:34.994 MyRootModel
             │ ../rootmodel.py:62 info
             │ r1='-9991'
             │ r2=-9991
```

None:
```
RootModel[NoneType](root=None)
schema gen for root model; obj=root=None root=None class=<class 'NoneType'>
17:41:53.463 MyRootModel
             │ ../rootmodel.py:62 info
             │ r1=None
             │ r2=None
```

alexmojaki: right, that looks bad for numbers and bools, they look like strings. they need similar treatment here.

alexmojaki:
```python
from pydantic import RootModel

import logfire

MyRootModel = RootModel[str]

r = MyRootModel('test')
logfire.info('MyRootModel', r1=r, r2=r.root)
#              │ r1=test
#              │ r2='test'
```

I expect

```
              │ r1='test'
              │ r2='test'
```

logfire/_internal/json_schema.py
dhruv-ahuja: I have fixed this and updated description, added test for the types as well. Please check.

logfire/_internal/json_encoder.py
```diff
@@ -155,6 +155,17 @@ def _pydantic_model_encoder(o: Any, seen: set[int]) -> JsonValue:
...
+
+    if isinstance(o, pydantic.RootModel):
+        root = o.root  # type: ignore
+        if isinstance(root, (str, int, bool, float, type(None))):
+            if root is None:
+                return root
+            elif isinstance(root, (int, bool, float)):
+                return str(root)
+            # add '' characters around the raw string literal to ensure its rendered properly by consumers upon parsing
+            return f"'{root}'"
```
alexmojaki: No, this would be done in the formatter just for console logging, not in the data

logfire/_internal/json_encoder.py
```diff
@@ -155,6 +155,17 @@ def _pydantic_model_encoder(o: Any, seen: set[int]) -> JsonValue:
...
 
     assert isinstance(o, pydantic.BaseModel)
+
+    if isinstance(o, pydantic.RootModel):
+        root = o.root  # type: ignore
+        if isinstance(root, (str, int, bool, float, type(None))):
+            if root is None:
+                return root
+            elif isinstance(root, (int, bool, float)):
+                return str(root)
```
alexmojaki: no, i don't want to convert these things to strings

logfire/_internal/json_encoder.py
dhruv-ahuja: Okay, so we keep the data as is without any conversion, and we only convert the data for logging to console?
So we aren't doing any actual data type manipulation in this case, is that right?

And the code here would just become 
```py
if isinstance(o, pydantic.RootModel):
    root = o.root  # type: ignore
    if isinstance(root, (str, int, bool, float, type(None))):
        return root
```

alexmojaki: i think we always return root, and we make sure that there's a schema if needed

logfire/_internal/json_formatter.py
```diff
@@ -204,6 +205,36 @@ def _format_bytes(self, _indent_current: int, value: Any, schema: JSONSchema | N
...
+        """
+        cls = schema and schema.get('title')
+        output = f'{cls}({value})' if cls else value
+
+        # check whether the given value is bool or number (as is case when handling primitives in RootModel)
+        try:
+            float(output)
+            is_valid_string = False
+        except ValueError:
+            is_valid_string = output.lower() not in ('true', 'false')
```
alexmojaki: no, put info in the schema and use that

logfire/_internal/json_formatter.py
```diff
@@ -204,6 +205,36 @@ def _format_bytes(self, _indent_current: int, value: Any, schema: JSONSchema | N
...
+
+        # add '' characters around the string literal to ensure its rendered properly by consumers upon parsing
+        if is_valid_string:
+            output = f"'{output}'"
+        else:
+            output = safe_repr(value)
```
alexmojaki:
```suggestion
        output = repr(value)
```

repr will add quotes. no exceptions are expected here.

logfire/_internal/json_schema.py
```diff
@@ -282,10 +284,29 @@ def _exception_schema(obj: Exception, _seen: set[int]) -> JsonDict:
...
+
+    root_class = obj.root.__class__  # type: ignore
+
+    # handle None separately as it has dedicated type
+    if root_class is type(None):
+        return {'type': 'null'}
+    # return a complex schema to ensure JSON parsing for strings inside RootModel since they get an
+    # extra layer of JSON encoding
+    elif root_class in (str, float, bool, int):
+        return {'type': 'string', 'x-python-datatype': 'string'}
```
alexmojaki: `'type': 'string'` for an int is weird.

dhruv-ahuja:
@alexmojaki how can I run PR-specific coverage locally? tried the command mentioned here but it runs for all the files
https://github.com/pydantic/logfire/actions/runs/17774267051/job/50518115247?pr=1396

alexmojaki:
I don't know of any way to get coverage without running everything

logfire/_internal/json_formatter.py
dhruv-ahuja: not needed as we have separated numeric value `type`

logfire/_internal/json_schema.py
alexmojaki: same for bool

dhruv-ahuja: registered new types for number values and added tests as well, only caveat is they're being represented as strings right now, but data is being logged in the same manner in console

```
00:03:58.094 MyRootModel [int]
             │ ../rootmodel.py:48 info
             │ r1=-9991
             │ r2=-9991


RootModel[float](root=2.9)
00:03:58.096 MyRootModel [float]
             │ ../rootmodel.py:48 info
             │ r1=2.9
             │ r2=2.9
```

tests for reference (see the return is in string format):
```
        pytest.param(
            -999,
            {'type': 'integer'},
            '-999',
            id='integer',
        ),
        pytest.param(
            -999.125,
            {'type': 'number'},
            '-999.125',
            id='number',
        ),
```

alexmojaki: also i don't know why you're comparing types directly instead of using `isinstance`

dhruv-ahuja: I can register a separate datatype if needed, but would it be required - since idea is we only need to handle the type formatting inside `RootModel`s and we only need to represent them as complex types to render correctly.

alexmojaki: IIUC them being stored as strings is pretty much inevitable

dhruv-ahuja: I didn't bother with boolean since I thought it is more "in-line" with string as compared to numbers.
Let me add separate type for that and use `isinstance`

dhruv-ahuja: addressed feedback

logfire/_internal/json_schema.py
```diff
@@ -282,10 +284,32 @@ def _exception_schema(obj: Exception, _seen: set[int]) -> JsonDict:
...
+    assert isinstance(obj, pydantic.RootModel)
+
+    root = obj.root  # type: ignore
+
+    if isinstance(root, type(None)):
+        return {'type': 'null'}
+    # return a complex schema to ensure JSON parsing for simple objects inside RootModel since they get an
+    # extra layer of JSON encoding
+    elif isinstance(root, (str, bool, int, float)):
+        type_ = {'str': 'string', 'int': 'integer', 'bool': 'boolean', 'float': 'number'}[root.__class__.__name__]
```
alexmojaki: this will raise an exception for a subclass of `str`

logfire/_internal/json_schema.py
```diff
@@ -282,10 +284,32 @@ def _exception_schema(obj: Exception, _seen: set[int]) -> JsonDict:
...
+        return {
+            'type': type_,
+            'x-python-datatype': 'string' if type_ == 'string' else None,
```
alexmojaki: does `x-python-datatype` achieve anything here?

dhruv-ahuja: Yeah, we are using the datatype handler in formatter to ensure we add quotes to string when logging

alexmojaki:
Please add tests for the console exporter, the JSON formatting tests aren't actually very robust.

logfire/_internal/json_schema.py
dhruv-ahuja: Yes it does,
does the following code look okay to handle this scenario?
```
    elif isinstance(root, (str, bool, int, float)):
        type_key = root_class = root.__class__.__name__
        if root_class != 'str' and isinstance(root, str):
            type_key = 'str'

        type_ = {'str': 'string', 'int': 'integer', 'bool': 'boolean', 'float': 'number'}[type_key]
        return {
            'type': type_,
            'x-python-datatype': 'string' if type_ == 'string' else None,
            'title': root_class if root_class != 'str' else None,
        }
```

Output would be something like this, where `Color` is a `StrEnum`
```
RootModel[Color](root=<Color.RED: 'red'>)
00:56:53.070 MyRootModel [Color]
             │ ../rootmodel.py:59 info
             │ r1=Color(red)
             │ r2=Color('red')
```

alexmojaki: if isinstance(root, str): type_ = 'string'
elif isinstance(root, int): type_ = 'integer'
...

all of these can be subclassed

alexmojaki: correction: check bool before int, because bool is a subclass of int

dhruv-ahuja: Done, added `number` type as well, to ensure objects like `IntEnum` are represented properly in the logs.

dhruv-ahuja:
@alexmojaki please validate whether the core logic is fine now, meanwhile I'll start looking into the console exporter tests.
Should the tests be the same kind as the json formatter or is there anything else I need to check also?

alexmojaki:
Add tests to `test_console_exporter.py` in the style of the others there.

tests/test_console_exporter.py
```diff
@@ -1019,3 +1019,65 @@ def test_console_exporter_list_data_with_object_schema_mismatch(capsys: pytest.C
...
+            IsStr(),
+            '│ with_model=Model(',
+            "│                name='with_model',",
+            '│            )',
+            "│ with_str=str('with_str')",
+            "│ with_str_inner='with_str'",
+            '│ with_int=int(-150)',
+            '│ with_int_inner=-150',
+            '│ with_float=float(2.0)',
+            '│ with_float_inner=2.0',
```
alexmojaki: we don't want these titles in the normal case

dhruv-ahuja: Fixed, thank you for catching it

dhruv-ahuja:
> Add tests to `test_console_exporter.py` in the style of the others there.

Done, added test for `StrEnum` and `IntEnum` objects as well to properly cover those scenarios

tests/test_json_args_formatting.py
```diff
@@ -85,6 +85,30 @@
 }""",
             id='complex_key_types',
         ),
+        pytest.param(
+            'test string',
+            {'type': 'string', 'x-python-datatype': 'string'},
```
alexmojaki: 'string' and 'number' are not python types, so `'x-python-datatype': 'string'` doesn't make sense. I still don't know why `x-python-datatype` is needed here at all, but if it is, it should be the name of the actual class, i.e. str/int/float.

Also please double check how this is being rendered in the UI.

tests/test_json_args.py
```diff
@@ -1380,6 +1382,93 @@ def to_dict(self):
...
+        with_bool_inner=root_with_bool.root,
+        with_none=root_with_none,
+        with_none_inner=root_with_none.root,
+        with_color=root_with_color,
+        with_color_inner=root_with_color.root,
+        with_order=root_with_order,
+        with_order_inner=root_with_order.root,
+    )
+
+    assert exporter.exported_spans_as_dict() == [
```
alexmojaki:
```suggestion
    assert exporter.exported_spans_as_dict(parse_json_attributes=True) == [
```

tests/test_json_args_formatting.py
dhruv-ahuja: I picked up the type naming convention from this code block but thought of keeping consistent datatype for int and float to avoid registering multiple types, what pattern should I follow?

```py
def _enum_schema(obj: Enum, seen: set[int]) -> JsonDict:
    enum_values = [e.value for e in obj.__class__]
    enum_types = set(type(value).__name__ for value in enum_values)
    if all(t in {'str', 'int', 'bool', 'float'} for t in enum_types):
        type_ = {'str': 'string', 'int': 'integer', 'bool': 'boolean', 'float': 'number'}[enum_types.pop()] # <- this
```

Here are the links to the relevant spans in dashboard:
- [str](https://logfire-us.pydantic.dev/public-trace/a9ceddef-bc02-45b8-b5aa-7ef5a7eb5bc7?spanId=24aeafeb4731734a) -- currently the python level repr is bit unnatural without quotes but I think that will be handled with the current PR

rest all are fine:
- [None](https://logfire-us.pydantic.dev/public-trace/69f6f31c-a670-4abb-aa4b-595eb354675b?spanId=5833a9a6d5ffee06)
- [int](https://logfire-us.pydantic.dev/public-trace/da8895f1-8513-4479-9cd2-de48f0e15359?spanId=04d430fe9492913e)
- [bool](https://logfire-us.pydantic.dev/public-trace/3768fe7a-88cb-481b-bbbc-800f6065c425?spanId=bad863862b016283)
- [float](https://logfire-us.pydantic.dev/public-trace/8a083ccf-069e-4294-ae1b-ef7d02375e8a?spanId=81e906059d1568d6)

we need the dedicated datatype so we can log strings with quotes, and also to handle subclasses representations properly
image

image

alexmojaki: but that function returns `type': type_` and `'x-python-datatype': 'Enum'`, not `'x-python-datatype': type_`

alexmojaki: those are private links, i can't view them. 

> I think that will be handled with the current PR

what do you mean, aren't you looking at what happens with the current PR?

dhruv-ahuja: > those are private links, i can't view them.

That's strange, I have created the links with a 1 week duration

image

dhruv-ahuja: > those are private links, i can't view them.

My bad, I had miscopied the `str` link, I've fixed that now. All of them are now apt public links.

> what do you mean, aren't you looking at what happens with the current PR?

I meant the RootModel string representation in python mode doesn't have quotes, even though we are passing a valid string object with the type schema. Same with subclasses of int and str.

str json schema: 
`"attributes_json_schema": "{\"type\":\"object\",\"properties\":{\"r1\":{\"type\":\"string\",\"x-python-datatype\":\"string\"},\"r2\":{}}}"`

str UI visuals:
image

dhruv-ahuja: > but that function returns type': type_ and 'x-python-datatype': 'Enum', not 'x-python-datatype': type_

Right, that makes sense. I have made this change.

dhruv-ahuja: Are the changes okay now?

alexmojaki (approved):
Thanks for your patience!

Issue #1390, opened 2025-09-10
Allow logging to stderr


kbruegge:
### Description

Hey there,

I might have missed it, but I would love to be able to log to `stderr` in the console. 
Currently I have to disable logging so I can pipe the `stdout` for further processing. Or maybe there is a smarter workaround? 

```logfire.configure(console = False) # maybe add options here. Or even allow passing a rich.console instance?```  

Thanks for the good work. Love the project 👍

kbruegge:
I see there is already a `ConsoleOptions` object which I can pass [docs](https://logfire.pydantic.dev/docs/reference/api/logfire/#logfire.ConsoleOptions). But it seems like I can't change the output to stderr either.

alexmojaki:
Yes, a field should be added to ConsoleOptions. SimpleConsoleSpanExporter and its subclasses already have an `output` param that isn't being used.

Mildly related, the exporters should really just accept a ConsoleOptions instead of each param separately.

pipinstalled:
@alexmojaki  can I work on this if it's free?

alexmojaki:
@pipinstalled yes please, not sure if @Lonercode didn't notice that their PR didn't pass CI.

PR #1397, opened 2025-09-10
Fix: Add output argument to ConsoleOptions in config.py

Lonercode:
This PR fixes an issue with the ConsoleOptions configuration where the output argument was missing from the config, causing errors when trying to redirect logs. The change adds the output argument. A proposed fix for [issue 1390](https://github.com/pydantic/logfire/issues/1390)

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -14,7 +14,7 @@ from contextlib import suppress
 from dataclasses import dataclass, field
 from pathlib import Path
 from threading import RLock, Thread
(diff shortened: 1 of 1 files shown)


alexmojaki:
This is the right start, but it doesn't actually do anything yet.

Lonercode:
Thanks for the feedback :) I guess you mean that instead of just adding the output argument, I should make it so that it redirects the logs to` sys.stdout` or `sys.stderr` .

alexmojaki:
Yes

Issue #1227, opened 2025-07-06
Support request and response body capture in aiohttp client instrumentation


amardeep:
### Description

httpx instrumentation supports capturing request and response body. This is missing from aiohttp client instrumentation.

Google's genai library switches to aiohttp if it finds aiohttp python package installed - for eg. due to another dependency eg. litellm. As `logfire.instrument_aiohttp_client()` doesn't support request/response capture, it is limited in its utility. Please add feature parity with httpx instrumentation for aiohttp too.

adtyavrdhn:
It doesn't seem like Otel supports these captures as of now looking at the instrumentation implementation of AioHttpClientInstrumentor.

alexmojaki:
The otel HTTPXClientInstrumentor doesn't really support it either, we jump through hoops like this:

```python

    def capture_body_if_text(self, attr_name: str = 'http.response.body.text'):
        def hook(span: LogfireSpan):
            try:
                # response.text uses errors='replace' under the hood.
                # We rely on decoding errors to guess when the response is not text.
                text = self.response.content.decode(self.response.encoding or 'utf-8')
            except (UnicodeDecodeError, LookupError):
                return
            self.capture_text_as_json(span, attr_name=attr_name, text=text)

        self.on_response_read(hook)

    @cached_property
    def response(self) -> httpx.Response:
        frame = inspect.currentframe().f_back.f_back  # type: ignore
        while frame:  # pragma: no branch
            response = frame.f_locals.get('response')
            frame = frame.f_back
            if isinstance(response, httpx.Response):
                return response
        raise RuntimeError('Could not find the response object')  # pragma: no cover

    def on_response_read(self, hook: Callable[[LogfireSpan], None]):
        if self.is_async:

            async def aread(original_aread: Callable[[], Awaitable[bytes]]) -> bytes:
                with self.attach_original_span_context(), self.logfire_instance.span('Reading response body') as span:
                    content = await original_aread()
                    hook(span)
                return content

            self.wrap_response_aread(aread)
        else:

            def read(original_read: Callable[[], bytes]) -> bytes:
                with self.attach_original_span_context(), self.logfire_instance.span('Reading response body') as span:
                    content = original_read()
                    hook(span)
                return content

            self.wrap_response_read(read)

    def wrap_response_read(self, hook: Callable[[Callable[[], bytes]], bytes]):
        response = self.response
        original_read = response.read

        @functools.wraps(original_read)
        def read() -> bytes:
            try:
[...]

PR #1403, opened 2025-09-15
Aiohttp request response capture

adtyavrdhn:
## Adds header and response body capture for #1227 

Response body capture is not implemented yet, not entirely sure on what the best way to approach that would be.

I tried to keep the implementation similar to that of httpx

I have to move the tests to use pytest_aiohttp instead and use TestClient and TestServer, I've raised this to get feedback on the implementation so far.


adtyavrdhn:
Hi @alexmojaki 

There are a couple of things I am yet to iron out but I would appreciate it if you could take a glance at the implementation so far.

alexmojaki:
this is a lot to take a glance at, do you have specific questions?

adtyavrdhn:
I’m capturing both the request and response headers from TraceRequestEndParams, since the traceparent header wasn’t present in TraceRequestStartParams. My only concern is that this means I’m capturing the request headers within the capture_headers call of LogfireAioHttpResponseInfo.

alexmojaki:
it might actually help to split capturing headers, request body, and response body into 2 or 3 PRs to make each more reviewable

alexmojaki:
yes it seems weird for `LogfireAioHttpResponseInfo.capture_headers` to capture request headers. I think it's fine to not capture the traceparent header.

adtyavrdhn:
Yeah, I'll split this one into two(headers and response body capture). It doesn't have request capture as of now.

adtyavrdhn:
I'll close this after I merge the individual PRs, keeping it around for reference.

adtyavrdhn:
Closing this, individual PRs got merged :)

PR #1409, opened 2025-09-17
Aiohttp response capture

adtyavrdhn:
Adds Response Body Capture for #1227 

Introduces capture_all and capture_response_body to Aiohttp Client Instrumentation

### DIFF
--- a/logfire/_internal/integrations/aiohttp_client.py
+++ b/logfire/_internal/integrations/aiohttp_client.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Callable, Literal
+import functools
+from typing import TYPE_CHECKING, Any, Callable, Literal, cast
 
 import attr
 from aiohttp.client_reqrep import ClientResponse
 from aiohttp.tracing import TraceRequestEndParams, TraceRequestExceptionParams, TraceRequestStartParams
-from opentelemetry.trace import Span
+from opentelemetry.trace import NonRecordingSpan, Span, use_span
 from yarl import URL
 
 try:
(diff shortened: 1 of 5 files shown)


alexmojaki:
I don't like the idea of having `capture_all` when it doesn't capture the request body. It may mislead users, especially if they're familiar with `capture_all` for httpx, and there's less reason to have it when it's only replacing one flag anyway.

adtyavrdhn:
Yeah I was thinking the same thing, I'll roll it back and add it when I can add support for request body capture as well. Looks weird in the documentation as well.

logfire/_internal/integrations/aiohttp_client.py
```diff
@@ -81,6 +84,39 @@ def capture_headers(self):
...
+    def capture_body_if_text(self, attr_name: str = 'http.response.body.text') -> None:
+        response = self.response
+        if response is None:
+            return
+
+        original_read = response.read
+
+        @functools.wraps(original_read)
+        async def read() -> bytes:
+            if getattr(response, '__logfire_body_captured', False):
```
alexmojaki: I think you could just set an attribute on `self` instead of the response

tests/otel_integrations/test_aiohttp_client.py
```diff
@@ -288,6 +288,197 @@ async def test_aiohttp_client_exception_handling_with_hooks(exporter: TestExport
...
+
+        async with aiohttp.test_utils.TestServer(test_app) as server:
+            await server.start_server()
+
+            logfire.instrument_aiohttp_client(capture_headers=True, capture_response_body=True)
+
+            async with aiohttp.ClientSession() as session:
+                async with session.get(f'http://localhost:{server.port}/body') as response:  # type: ignore
+                    await response.json()
+                    await response.read()
```
alexmojaki: please assert what these return

logfire/_internal/integrations/aiohttp_client.py
adtyavrdhn: Oh yeah, I can remove frozen now that I am not inheriting from TraceRequestEndParams, I'll do that.

tests/otel_integrations/test_aiohttp_client.py
adtyavrdhn: Done, we see the response in the snapshot as well though so what's the reason to add another assert here?

alexmojaki: to make sure that the method still works properly for the user and the instrumentation isn't interfering

adtyavrdhn: Gotcha, thanks :))

PR #1419, opened 2025-09-22
feat: access spans created by instrument decorator

oliviaBahr:
I find myself often wanting to interact with spans created by the instrument decorator.

I've done two implementations: one adding an optional param to the decorator, the other using a context var on the logfire instance.
I prefer the context var impl because the usage is nicer but I'm not sure about the overhead.


alexmojaki:
1. A function called `logfire.current_span()` can't just work for `instrument`, it at least needs to work for `logfire.span`, it should probably return something useful for *all* OTel spans. We've wanted to implement a function like this, with this name, for a while but haven't had time. It would wrap arbitrary spans while allowing things like setting complex attributes.
2. This absolutely can't depend on whether `logfire.current_span` appears somewhere in the body of the decorated function.

dmontagu:
Closing based on Alex's feedback and that it's stale, though would be open to alternatives that address the feedback

PR #1423, opened 2025-09-23
Trigger `auth` command from `prompt`

Kludex:


logfire/_internal/cli/auth.py
```diff
@@ -0,0 +1,75 @@
...
+            except ValueError:
+                selected_region = -1
+        logfire_url = list(REGIONS.values())[selected_region - 1]['base_url']
+
+    device_code, frontend_auth_url = request_device_code(args._session, logfire_url)
+    frontend_host = urlparse(frontend_auth_url).netloc
+
+    # We are not using the `prompt` parameter from `input` here because we want to write to stderr.
+    sys.stderr.write(f'Press Enter to open {frontend_host} in your browser...\n')
+    input()
```
Kludex, line 60: This is the only lines changed here.

logfire/_internal/cli/auth.py
Kludex, line 1: I moved this parse method because I wanted to import from `prompt.py`.

PR #1436, opened 2025-09-26
Add prompt-toolkit as dependency

Kludex:
We need to decide a bit of UX.

1. Do we erase the messages after selecting a project?

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -54,6 +54,8 @@ from opentelemetry.sdk.trace import SpanProcessor, SynchronousMultiSpanProcessor
 from opentelemetry.sdk.trace.export import BatchSpanProcessor, SimpleSpanProcessor
 from opentelemetry.sdk.trace.id_generator import IdGenerator
 from opentelemetry.sdk.trace.sampling import ParentBasedTraceIdRatio, Sampler
+from prompt_toolkit.shortcuts import choice
+from prompt_toolkit.styles import Style
 from rich.console import Console
 from rich.prompt import Confirm, Prompt
 from typing_extensions import Self, Unpack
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -53,6 +53,7 @@ dependencies = [
     "typing-extensions >= 4.1.0",
     "tomli >= 2.0.1; python_version < '3.11'",
     "executing >= 2.0.1",
+    "prompt-toolkit >= 3.0.52",
```
samuelcolvin, line 56: can we say `>=3`?

logfire/_internal/config.py
```diff
@@ -1457,21 +1459,23 @@ def use_existing_project(
...
-                for index, item in enumerate(filtered_projects)
-            }
-            project_choices_str = '\n'.join(
-                [f'{index}. {item[0]}/{item[1]}' for index, item in project_choices.items()]
-            )
-            selected_project_key = Prompt.ask(
-                f'Please select one of the following projects by number:\n{project_choices_str}\n',
-                choices=list(project_choices.keys()),
-                default='1',
+            org_options = sorted({project['organization_name'] for project in projects})
```
samuelcolvin, line 1464: we shouldn't ask this if there's only one org.

samuelcolvin (commented):
This is already a improvement on what we have now.

But I've pushed a demo of what I'd like, feel free to delete my commit, and maybe that has to wait for `logfire init`, but we do need to improve the API response to include default org and org created date, or at least created order.

dmontagu:
Closing this draft — as discussed, the UX here wants to be settled alongside `logfire init` rather than bolted on, and the branch has gone stale.

PR #1457, opened 2025-10-03
Record response when streaming OpenAI Responses API

stevenh:
Implement streaming state for OpenAI Responses API ensuring that the reported span contains full details of the model response.

Fix VCR sensitive header removal, using lower case headers to correctly match requests and adding a before_record_response to process response headers.


tests/conftest.py
```diff
@@ -186,6 +186,43 @@ def multiple_credentials(tmp_path: Path) -> Path:
...
+    Returns:
+        The scrubbed response dictionary with sensitive headers removed.
+    """
+    headers: dict[str, list[str]] = response.get('headers', {})
+    new_headers: dict[str, list[str]] = {}
+    for k, v in headers.items():
+        if k.lower() not in SENSITIVE_HEADERS:
+            new_headers[k] = v
+
+    response['headers'] = new_headers
```
alexmojaki:
```suggestion
    response['headers'] = {k: v for k, v in response.get('headers', {}).items() if k.lower() not in SENSITIVE_HEADERS}
```

tests/conftest.py
```diff
@@ -186,6 +186,43 @@ def multiple_credentials(tmp_path: Path) -> Path:
...
+
+    return response
+
+
 @pytest.fixture(scope='module')
-def vcr_config():
-    return {'filter_headers': ['authorization', 'cookie', 'Set-Cookie', 'x-goog-api-key']}
+def vcr_config() -> dict[str, Any]:
+    return {
+        'filter_headers': SENSITIVE_HEADERS,
```
alexmojaki, line 220: Did this not work on its own? it seemed to work for `authorization`. was it just about using lowercase for `Set-Cookie`?

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -57,12 +58,14 @@ def get_endpoint_config(options: FinalRequestOptions) -> EndpointConfig:
         return EndpointConfig(
             message_template='Responses API with {gen_ai.request.model!r}',
             span_data={
+                'request_data': json_data,
```
alexmojaki: This is essentially part of the 'old' format, I don't think we want it here (at least not in full) since it will duplicate the input messages here and in `events`.

tests/otel_integrations/test_openai.py
```diff
@@ -240,7 +240,7 @@ def request_handler(request: httpx.Request) -> httpx.Response:
     elif request.url == 'https://api.openai.com/v1/completions':
         json_body = json.loads(request.content)
         if json_body.get('stream'):
-            chunks = [
+            completion_chunks = [
```
alexmojaki, line 243: Is this change useful? if not please revert for a cleaner diff

tests/otel_integrations/test_openai.py
```diff
@@ -1419,6 +1421,208 @@ def test_completions(instrumented_client: openai.Client, exporter: TestExporter)
...
+def test_responses_stream(exporter: TestExporter) -> None:
+    client = openai.Client()
+    logfire.instrument_openai(client)
+    with client.responses.stream(
+        model='gpt-4.1',
+        input='What is four plus five?',
+    ) as stream:
+        for event in stream:
+            if event.type == 'response.output_text.delta':
+                print(event.delta, end='', flush=True)
```
alexmojaki: no printing pls

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -119,6 +122,21 @@ def get_response_data(self) -> Any:
...
 
+class OpenaiResponsesStreamState(StreamState):
+    def __init__(self):
+        self._state = ResponseStreamState(input_tools=openai.omit, text_format=openai.omit)
+
+    def record_chunk(self, chunk: Any) -> None:
+        self._state.handle_event(chunk)
+
+    def get_response_data(self) -> Any:
+        response = self._state._completed_response  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
stevenh: No way around this.

logfire/_internal/integrations/llm_providers/openai.py
alexmojaki: oh, i'm guessing you need the request/response_data pair in the UI. can you use request_data when it's streaming, `events` otherwise? or, maybe slightly more complicated, see how `events` is handled currently in the non-streaming case.

tests/conftest.py
stevenh: It will remove authorize as it's lowercase already, but not Set-Cookie as the values it compares against are lowercase. Also required to be lowercase for the in check in scrub_headers

alexmojaki: and just changing to `set-cookie` doesn't work?

logfire/_internal/integrations/llm_providers/openai.py
stevenh: Yer it fails with key errors if request_data is not present in record_streaming log info due to:
```
'streaming response from {request_data[model]!r} took {duration:.2f}s',
```

tests/otel_integrations/test_openai.py
stevenh: Prevents warnings due to incompatible types.

logfire/_internal/integrations/llm_providers/openai.py
alexmojaki: right, so let's not also have `events` in the streaming case.

tests/conftest.py
stevenh: Nope VCR `filter_headers` is only applied to [requests](https://github.com/kevin1024/vcrpy/blob/ac70eaa17fccfe453f153aa26ef67c86e4a07d35/vcr/config.py#L186) not [responses](https://github.com/kevin1024/vcrpy/blob/ac70eaa17fccfe453f153aa26ef67c86e4a07d35/vcr/config.py#L163), which is why the [existing cassettes still have Set-Cookie](https://github.com/pydantic/logfire/blob/main/tests/otel_integrations/cassettes/test_openai_agents/test_chat_completions.yaml#L63) in them. You can also see that [request headers](https://github.com/pydantic/logfire/blob/main/tests/otel_integrations/cassettes/test_openai_agents/test_chat_completions.yaml#L5) in the cassettes are all lower case but the responses are mixed.

logfire/_internal/integrations/llm_providers/openai.py
alexmojaki: no please don't change it in the non-streaming case

stevenh: To confirm my understanding, you want to remove the event's data when logging the span in `record_streaming`, preventing duplicate information being sent?

alexmojaki: don't even include `events` in the first place when it's streaming

stevenh: I think I follow you now.

I was missing that you could check in `get_endpoint_config` if the request was streaming. I had looked at `FinalRequestOptions` wasn't there so had assumed it wasn't possible, but just found it in `json_data`, so now believe your intent was to check that and build `span_data` conditionally, which is what I've now done.

Also made `request_data` conditional, based on your earlier comment but let me know if you want it always present for consistency.

alexmojaki:
Thanks!

Issue #972, opened 2025-03-31
Celery needs to be instrumented on client side for distributed tracing


alexmojaki:
Demo:

```python
import logfire
from celery import Celery
from celery.signals import worker_init
from fastapi import FastAPI

@worker_init.connect()
def init_worker(*args, **kwargs):
    logfire.configure(service_name="worker")
    logfire.instrument_celery()

app = Celery("tasks", broker="redis://localhost:6379/0")

@app.task
def add(x: int, y: int):
    return x + y

fapp = FastAPI()

logfire.configure()
logfire.instrument_fastapi(fapp)
########### Important:
logfire.instrument_celery()


@fapp.get("/hello")
async def hello(name: str):
    add.delay(42, 50)
    return {"message": f"hello {name}"}


if __name__ == "__main__":
    import uvicorn

    uvicorn.run(fapp)
```

The top-level `logfire.instrument_celery()` is required for the context propagation, based on https://github.com/open-telemetry/opentelemetry-python-contrib/issues/1002#issuecomment-1275273134

This should be documented at least, maybe fixed in some other way, e.g. in OTel.

dhruv-ahuja:
Should this be a warning in the existing documentation, or maybe a separate section where we extend the existing celery example to showcase distributed tracing example?

Code snippet below, in which case user can run the same tasks.py file.
```py
import logfire
from celery import Celery
from celery.signals import worker_init


# celery worker configuration
@worker_init.connect()
def init_worker(*args, **kwargs):
    logfire.configure(service_name='worker')
    logfire.instrument_celery()
    print('logfire instrumented')


app = Celery('worker', broker='redis://@localhost:6379/0')


@app.task
def add(x: int, y: int):
    return x + y


if __name__ == '__main__':
    print('running as a client app')
    logfire.configure(service_name='client-app')
    logfire.instrument_celery()
    add.delay(42, 50)
```

<img width="1205" height="285" alt="Image" src="https://github.com/user-attachments/assets/539c91c5-4022-4030-b24c-9e3c659016a1" />

alexmojaki:
A separate section sounds good. It should set `distributed_tracing=True` in the worker to prevent the warning.

dhruv-ahuja:
created [PR](https://github.com/pydantic/logfire/pull/1464) for the same

PR #1464, opened 2025-10-05
Document Celery's Distributed Tracing Behaviour

dhruv-ahuja:
Creates a dedicated section in the Celery documentation covering the distributed tracing scenario, as discussed in #972.

### DIFF
--- a/docs/integrations/event-streams/celery.md
+++ b/docs/integrations/event-streams/celery.md
@@ -106,6 +106,40 @@ To run the beat, you can use the following command:
 celery -A tasks beat --loglevel=info
 ```
 
+## Distributed Tracing
+
+You must configure Celery instrumentation across client and worker processes, to ensure worker-generated spans are linked to the parent spans in the caller process.
+
+Let's extend the original example to showcase a distributed tracing setup, including both worker and client code with separate entrypoints. You can run the worker process with `celery -A tasks worker --loglevel=info`, and the client by executing `python tasks.py` from a separate terminal session:
+
+```py title="tasks.py" hl_lines="6-11 19-22"
+import logfire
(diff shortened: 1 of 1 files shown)


docs/integrations/event-streams/celery.md
```diff
@@ -106,6 +106,40 @@ To run the beat, you can use the following command:
...
+app = Celery("tasks", broker="redis://localhost:6379/0")
+
+
+@app.task
+def add(x: int, y: int):
+    return x + y
+
+
+if __name__ == "__main__":  # (2)!
+    logfire.configure(service_name="client-app")
```
alexmojaki, line 135:
```suggestion
    logfire.configure(service_name="client")
```

Then this can fit in a screenshot nicely, which should be added

docs/integrations/event-streams/celery.md
```diff
@@ -106,6 +106,40 @@ To run the beat, you can use the following command:
...
+    logfire.instrument_celery()
+    add.delay(42, 50)
+```
+
+1. Mark the configuration as running in a distributed tracing environment. This is only required for the worker.
```
alexmojaki, line 140: Explain in more detail and link to docs.

docs/integrations/event-streams/celery.md
```diff
@@ -106,6 +106,40 @@ To run the beat, you can use the following command:
...
+2. Running the file directly will simulate a client application that calls the Celery task.
+
 The keyword arguments of [`logfire.instrument_celery()`][logfire.Logfire.instrument_celery] are passed to the
```
alexmojaki, line 143: Move this to before the installation, it doesn't fit here.

PR #1470, opened 2025-10-07
fix: be more explicit with truncation

ivo-1:


logfire/_internal/utils.py
```diff
@@ -106,7 +106,7 @@ def safe_repr(obj: Any) -> str:
 Truncatable = TypeVar('Truncatable', str, bytes, 'list[Any]', 'tuple[Any, ...]')
 
 
-def truncate_string(s: str, *, max_length: int, middle: str = '...') -> str:
+def truncate_string(s: str, *, max_length: int, middle: str = '[...truncated by logfire...]') -> str:
```
alexmojaki, line 109: See first part of https://github.com/pydantic/logfire/issues/1471#issuecomment-3377604476. I don't think this is good in the case of message formatting, but it makes sense for baggage. Or maybe it should be the default and message formatting should explicitly use `...` or even the single unicode character.

ivo-1: I see. Do I understand you correctly that the proposed changes don't really make sense given they are only in the SDK and not the backend? Or do you think it still makes sense to do this at least for baggage (and in the python SDK)?

alexmojaki: Right, basically I don't want such a long `middle` except when `max_length` is quite high, which in the SDK means baggage only IMO.

Issue #1493, opened 2025-10-17
Add _new_context argument to logfire.instrument


njz-cvm:
### Description

Take the following ASGI Quart application as an example, using asgi_instrumentation.

```py

@logfire.instrument("Tracking Visit")
async def track_user():
  async with db:
    await db.record_event()

@app.route("/")
async def index():
  current_app.add_background_task(track_user)
  return await render_template("index.html")

```

When this code is run and someone visits the index, a span will be created for the route. Then, a child span will be created of the background task. This does not make much sense, though there is a solution. To get around this, you can use this work around.

```py

async def track_user():
  with logfire.attach_context({}), logfire.span("Tracking Visit")
    async with db:
      await db.record_event()

@app.route("/")
async def index():
  current_app.add_background_task(track_user)
  return await render_template("index.html")

```

While this code works, it removes any association with the span that created it, looks messy, and adds an indentation. I propose the following solution.

```py

@logfire.instrument("Tracking Visit", _new_context = True)
async def track_user():
  async with db:
    await db.record_event()

@app.route("/")
async def index():
  current_app.add_background_task(track_user)
  return await render_template("index.html")

```

In this approach, the parent span is disassociated from track_user, but it could add data to the span referencing back to it. While someone could re-implement logfire.instrumentation to add this feature, an official keyword in logfire would be preferred. At the moment the Logfire website does not have a way to associate spans as "siblings" (to my knowledge). But if support were added in the future, the _new_context implementation could be modified without changing the logfire.instrument API.

alexmojaki:
> At the moment the Logfire website does not have a way to associate spans as "siblings" (to my knowledge).

It does, it's called span links, here's an example:

```python
import logfire

logfire.configure()

with logfire.span('span1') as span1:
    ...

with logfire.span('span2', _links=[(span1.get_span_context(), None)]) as span2:
    ...
```

Here's what the UI shows:

<img width="944" height="71" alt="Image" src="https://github.com/user-attachments/assets/d1d66a1d-e9ba-4331-8e63-6181c9659dd5" />

These were implemented in https://github.com/pydantic/logfire/pull/587, where I said "Documentation can happen after dogfooding." We never actually got around to using them ourselves and use cases rarely come up. So we haven't tried to improve the API which is currently quite unfriendly, or document it. But it works.

[...]

PR #1499, opened 2025-10-24
Add `new_trace` parameter to `logfire.instrument`

njz-cvm:
As per #1493, it's current a bit awkward to remove context from instrumentation. This PR adds the "new_context" keyword to ensure designated function calls do not inherit from parent spans.

Here is an example from the new test:
```py
def context_factory():
    return {'new_contex': 123}

@tagged.instrument('hello-world {a=}', record_return=True, new_context=context_factory)
def hello_world(a: int) -> str:
    return f'hello {a}'

@tagged.instrument('parent', record_return=True)
def parent() -> str:
    return hello_world(5)
```

Alternatively, new_context can be set to `True`, which is the same as `new_context=dict`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
           enable-cache: true
 
       - run: uv sync --python 3.12 --frozen --group docs
-      - run: uv pip install --upgrade mkdocs-material mkdocstrings-python griffe==0.48.0
+      - run: uv pip install --upgrade mkdocs-material mkdocstrings-python 'mkdocstrings<1' griffe==0.48.0
         env:
           UV_EXTRA_INDEX_URL: ${{ secrets.UV_EXTRA_INDEX_URL }}
       - run: |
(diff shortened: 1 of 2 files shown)


tests/test_logfire.py
```diff
@@ -748,6 +748,213 @@ def hello_world(a: int) -> str:
...
+            },
+        ]
+    )
+
+
+def test_instrument_new_context(exporter: TestExporter) -> None:
+    tagged = logfire.with_tags('test_instrument')
+
+    def context_factory():
+        return {'new_contex': 123}
```
alexmojaki: We didn't discuss this context factory API, I'm not sure if I like it or understand the point. It's certainly not being tested here, `new_contex` doesn't do anything or appear in the exported spans.

njz-cvm: I didn't mention it in the original issue because it wasn't something I'd considered. But when I realized I'd just be using attach_context behind the scenes, it seemed preferable to allow users the option to provide input for attach_context.

If you're against it, it's not that useful for me and I don't mind cutting it. But if it's worth keeping I'll add better testing for it.

njz-cvm:
> this feature isn't really compelling enough without the span link

I'll look at adding it, just trying to determine the best approach.

alexmojaki:
If you'd still like me to review this, please ensure that all checks are passing.

njz-cvm:
Sorry, I didn't have time to finish the required tests. I have changed the API to be what we previously discussed, with new_context being simply True / False. I agree with you that this makes the most sense.

alexmojaki:
@njz-cvm does this look good to you?

njz-cvm:
Looks great to me, this is a much cleaner implementation.

PR #1513, opened 2025-10-31
Document base url flag for CLI auth

bruno-espino:


docs/reference/cli.md
```diff
@@ -21,7 +21,15 @@ logfire auth
 ```
 
 You will be prompted to select a [data region](./data-regions.md) (EU or US). To specify this
-via the cli instead of interactively, use `logfire --region eu auth` or `logfire --region eu auth`
+via the cli instead of interactively, use `logfire --region eu auth` or `logfire --region us auth`
+
+### Authenticating with Self-Hosted Instances
```
alexmojaki: This doesn't work as a header, it looks like it applies to the existing content below "Then you will be given the option..."

Maybe `!!! note`

PR #1546, opened 2025-11-24
Docs: Import getLogger in logging configuration example

hanlee55:
This pull request makes a small update to the logging integration documentation.

The change adds an import for `getLogger` alongside `dictConfig` in the Python example, ensuring users have the necessary import for standard logging usage.


alexmojaki:
Thanks!

PR #1548, opened 2025-11-25
Add managed variables

alexmojaki:


logfire/_internal/config.py
```diff
@@ -67,6 +67,9 @@
 
 from ..propagate import NoExtractTraceContextPropagator, WarnOnExtractTraceContextPropagator
 from ..types import ExceptionCallback
+from ..variables.config import VariablesConfig
```
alexmojaki: just noting that i think this will need to be moved to make pydantic and whatever other deps not required

pyproject.toml
```diff
@@ -80,6 +80,7 @@ sqlite3 = ["opentelemetry-instrumentation-sqlite3 >= 0.42b0"]
 aws-lambda = ["opentelemetry-instrumentation-aws-lambda >= 0.42b0"]
 google-genai = ["opentelemetry-instrumentation-google-genai >= 0.4b0"]
 litellm = ["openinference-instrumentation-litellm >= 0"]
+variables = ["pydantic>=2.0.0"]  # TODO: Change minimum pydantic version?
```
alexmojaki: and openfeature-sdk, right?

logfire/_internal/config.py
```diff
@@ -258,6 +261,44 @@ class CodeSource:
     """
 
 
+@dataclass(init=False)
+class VariablesOptions:
```
alexmojaki, line 324: having both VariablesOptions and VariablesConfig is worrying

logfire/_internal/config.py
```diff
@@ -282,6 +323,7 @@ def configure(  # noqa: D417
     min_level: int | LevelName | None = None,
     add_baggage_to_attributes: bool = True,
     code_source: CodeSource | None = None,
+    variables: VariablesOptions | None = None,
```
alexmojaki, line 362: i think this should be kept under AdvancedOptions and marked as experimental to begin with

logfire/_internal/config.py
```diff
@@ -258,6 +261,44 @@ class CodeSource:
     """
 
 
+@dataclass(init=False)
+class VariablesOptions:
+    """Configuration of metrics."""
```
```suggestion
    """Configuration of variables."""
```

logfire/_internal/main.py
```diff
@@ -2331,6 +2338,9 @@ def shutdown(self, timeout_millis: int = 30_000, flush: bool = True) -> bool:  #
         self._meter_provider.shutdown(remaining)
         return (start - time()) < timeout_millis
 
+    def var(self, name: str, default: T | ResolveFunction[T], type_: type[T] | None = None) -> Variable[T]:
```
alexmojaki: noting that we will probably want to think about this name before releasing.

alexmojaki: this doesn't typecheck:

```python
from typing import TypeVar

T = TypeVar('T')

def var(default: T, type_: type[T]) -> T:
    ...

v = var('x', str | int)
```

alexmojaki: i think the param should just be `type` and we ignore linting errors on our side so that user code looks nice. `logfire.var(..., type=str)` won't trigger lint errors for users.

alexmojaki: i think we want `foo = logfire.var(default=42)` to just work without necessarily having to specify `name='foo'`

logfire/variables/config.py
```diff
@@ -0,0 +1,265 @@
...
+        | ValueMatchesRegex
+        | ValueDoesNotMatchRegex
+        | KeyIsPresent
+        | KeyIsNotPresent,
+        Discriminator('kind'),
+    ],
+)
+
+
+VariantKey = str
```
alexmojaki: should these be NewTypes?

logfire/variables/config.py
```diff
@@ -0,0 +1,265 @@
...
+
+_VariablesConfigAdapter = TypeAdapter(VariablesConfig)
+
+
+def _matches_all_conditions(conditions: list[Condition], attributes: Mapping[str, Any]) -> bool:
+    """Check if all conditions match the provided attributes."""
+    for condition in conditions:
+        if not condition.matches(attributes):
+            return False
+    return True
```
alexmojaki, line 502: @copilot refactor this with `all`

logfire/variables/providers/local.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import Any
+
+from logfire.variables.config import VariablesConfig
+
+__all__ = ['LogfireLocalProvider']
+
+from logfire.variables.providers.abstract import VariableProvider, VariableResolutionDetails
+
+
+# TODO: Need to create a LogfireRemoteProvider
```
alexmojaki:
```suggestion
```

logfire/variables/providers/remote.py
```diff
@@ -0,0 +1,162 @@
...
+            return VariableResolutionDetails(value=None, _reason='unrecognized_variable')
+
+        variant = variable_config.resolve_variant(targeting_key, attributes)
+        if variant is None:
+            return VariableResolutionDetails(value=None, _reason='resolved')
+        else:
+            return VariableResolutionDetails(value=variant.serialized_value, variant=variant.key, _reason='resolved')
+
+    def shutdown(self):
+        # TODO: When should this be called? If never, should we delete this code?
```
alexmojaki: `logfire.shutdown` and `logfire.configure` should call this.

logfire/variables/remote.py
```diff
@@ -0,0 +1,162 @@
...
+        self._pid = os.getpid()
+
+    def _worker(self):
+        while not self._shutdown:
+            self._worker_awaken.wait(self._polling_interval.total_seconds())
+            if self._shutdown:
+                break
+            # TODO: Ideally we'd be able to terminate while the following request was going even if it takes a while
+            #   Note it's far more reasonable to terminate this worker thread "gracelessly" than an OTel exporter's.
+            #   Is there anything similar to an anyio CancelScope we can use here?
```
alexmojaki: i don't see any reason to try and quickly cancel things in `_worker`. it's a daemon thread so it won't hold up process shutdown.

logfire/variables/remote.py
```diff
@@ -0,0 +1,162 @@
...
+            self.refresh()
+            self._worker_awaken.clear()
+
+    def _get_raw(self, endpoint: str, params: dict[str, Any] | None = None) -> Response:
+        # TODO: Should we try to unify this and `_get` with LogfireClient in some way? Are they even necessary?
```
alexmojaki: the separation in LogfireClient helps dedupe code. since there's only one call to _get/_get_raw here and the behaviour is different (warn instead of error) might as well keep these separate for now.

pyproject.toml
dmontagu: no I removed that dependency

logfire/_internal/config.py
dmontagu: I agree, I'm assuming you understand the difference between them, do you have a suggestion for alternative naming?

logfire/_internal/config.py
dmontagu: sure

logfire/_internal/config.py
dmontagu: lol

logfire/_internal/main.py
dmontagu: Yeah this was Samuel's (and maybe Douwe's?) preference, I didn't love it, but I also don't really care. I'm fine with other names, lmk if you have ideas/preferences.

dmontagu: I'm happy to not handle unions well until there is better support for TypeForms; we could use the `pydantic_graph.beta.util.TypeExpression` trick here as a way to get support for unions. I think it might also work to do `types: list[type[T]]` like we do for output_type in `pydantic_ai.Agent`, I think type checking will infer `str | int` if you pass `[str, int]`:

```python
from typing import TypeVar

T = TypeVar('T')

def var(default: T, types: list[type[T]]) -> T:
    ...

v = var('x', [str, int])
reveal_type(v)
#> scratch.py:9:13 - information: Type of "v" is "str | int"
```

Not sure how you feel about any of the above.

logfire/variables/config.py
dmontagu: 🤷‍♂️ I see pros and cons, as is it's still useful for documentation. I generally don't like needing to add newtypes as an end user, if end users never pass these into things from python then I guess it's okay to use newtypes but I feel like it's not unreasonable to start with it just being strings.

logfire/_internal/main.py
dmontagu: That seems difficult to make work reliably across refactors etc.; given the name _needs_ to match on the provider side (i.e., wherever you manage the config, probably Logfire), I'd be a bit nervous to do that "magically".

Do you have a mechanism we could use to infer the name like that? E.g., if you can explain to me how to use stack frame traversal/source code introspection to make this work I'd be interested. I know we do it for agents/graphs etc. on first access but that just feels a little fragile to me, e.g. if you import the variable with a different name in a different file, it will actually break the usage. I'd be more okay with it if it was guaranteed to have the name that was used at the place where the function was called, I just haven't seen a mechanism we could use to ensure that synchronization.

I think if we can be confident of inferring the name correctly from the source  code, we _can_ separately ensure the names match known variable names and have valid schemas with config validation, we already need to do something like that for the sake of creating the variables in the UI (I can explain, but I'll spare you in this comment), but as far as I'm aware the hard part is reliably inferring the name properly.

logfire/variables/remote.py
dmontagu: Okay, I'm concluding that the point is we don't feel a need to actually join the thread at all? The main issue I see with that is potential issues in pytest with resources not getting cleaned up properly but it might be fine, I'll try to not do any joining and we can revisit this if it causes problems.

logfire/variables/remote.py
dmontagu: makes sense

pyproject.toml
alexmojaki: oh, it's still in the dev deps below. so are we using openfeature or not?

logfire/_internal/main.py
```diff
@@ -2329,8 +2336,25 @@ def shutdown(self, timeout_millis: int = 30_000, flush: bool = True) -> bool:  #
         if not remaining:  # pragma: no cover
             return False
         self._meter_provider.shutdown(remaining)
+
+        remaining = max(0, timeout_millis - (time() - start))
+        if not remaining:  # pragma: no cover
+            return False
+
+        self.config.get_variable_provider().shutdown()
```
alexmojaki: maybe this should happen before the rest if we don't expect it to actually take any time

logfire/_internal/main.py
```diff
@@ -2329,8 +2336,25 @@ def shutdown(self, timeout_millis: int = 30_000, flush: bool = True) -> bool:  #
...
+
         return (start - time()) < timeout_millis
 
+    def var(self, *, name: str, default: T | ResolveFunction[T], type: type[T] | Sequence[type[T]]) -> Variable[T]:
+        from logfire.variables.variable import Variable
+
+        tp: type[T]
+        if isinstance(type, Sequence):
+            tp = Union[tuple(type)]  # pyright: ignore[reportAssignmentType]
```
alexmojaki: does this somehow pass type checking for the user?

logfire/variables/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+
+    raise AttributeError(f'module {__name__!r} has no attribute {name!r}')
+
+
+# from logfire.variables.local import LocalVariableProvider
+# from logfire.variables.remote import LogfireRemoteVariableProvider
+# from logfire.variables.config import VariablesConfig
+
+
+# from logfire.variables import LocalVariableProvider, LogfireRemoteVariableProvider, VariablesConfig
```
alexmojaki:
```suggestion
```

logfire/variables/variable.py
```diff
@@ -0,0 +1,180 @@
...
+        current = _VARIABLE_OVERRIDES.get() or {}
+        token = _VARIABLE_OVERRIDES.set({**current, self.name: value})
+        try:
+            yield
+        finally:
+            _VARIABLE_OVERRIDES.reset(token)
+
+    async def refresh(self, force: bool = False):
+        """Asynchronously refresh the variable."""
+        await asyncio.to_thread(self.logfire_instance.config.get_variable_provider().refresh, force=force)
```
alexmojaki: anyio?

logfire/variables/remote.py
```diff
@@ -0,0 +1,201 @@
...
+            name='LogfireRemoteProvider',
+            target=self._worker,
+            daemon=True,
+        )
+        self._worker_thread.start()
+        self._pid = os.getpid()
+
+    def _worker(self):
+        while not self._shutdown:
+            self._worker_awaken.wait(self._polling_interval.total_seconds())
```
alexmojaki, line 246: i think this should be moved to the end of the loop so that polling starts immediately

logfire/_internal/main.py
dmontagu: yes, at least with pyright, sequences of types get turned into their union, you can try it out. I can add an assert_type test for it.

logfire/variables/variable.py
dmontagu: I think anyio isn't a dependency; should I use a utility that imports a similar function from anyio if anyio is installed and otherwise falls back to asyncio?

logfire/variables/remote.py
dmontagu: makes sense, I'll make that change.

logfire/variables/variable.py
alexmojaki: sure. i guess it's not worth making it a dependency of the extra just for this. does nothing else here use asyncio?

also

```suggestion
        await asyncio.to_thread(self.refresh_sync, force=force)
```

logfire/variables/remote.py
dmontagu: (Note to self: my main fear was that this would cause an issue with testing.)

logfire/_internal/config.py
dmontagu: I think after playing with this we mutually decided to just keep it at the top level as it made certain things in the implementation simpler. Let me know if I am misremembering.

logfire/_internal/main.py
dmontagu: I added support for a sequence of types (which will be converted to a union by pyright and perhaps other typecheckers?); I'll note we can presumably get rid of this with a deprecationwarning when TypeForm lands.

logfire-api/logfire_api/__init__.pyi
alexmojaki, line 1: there's no need to generate stubs during a PR. it happens during release. i'd actually kind like these to be reverted to make the PR more manageable.

tests/test_variables.py
```diff
@@ -0,0 +1,1702 @@
...
+        condition = ValueEquals(attribute='plan', value='enterprise')
+        assert condition.matches({'plan': 'free'}) is False
+
+    def test_no_match_when_missing(self):
+        condition = ValueEquals(attribute='plan', value='enterprise')
+        assert condition.matches({}) is False
+
+    def test_kind_discriminator(self):
+        condition = ValueEquals(attribute='plan', value='enterprise')
+        assert condition.kind == 'value-equals'
```
alexmojaki, line 59: This reads like the LLM knows it's being paid by the token

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+    """The regex pattern the attribute value must not match."""
+    kind: Literal['value-does-not-match-regex'] = 'value-does-not-match-regex'
+    """Discriminator field for condition type."""
+
+    def matches(self, attributes: Mapping[str, Any]) -> bool:
+        """Check if the attribute value does not match the regex pattern."""
+        value = attributes.get(self.attribute)
+        if not isinstance(value, str):
+            return False
+        return bool(re.search(self.pattern, value))
```
alexmojaki:
```suggestion
        return not re.search(self.pattern, value)
```

tests/test_variables.py
```diff
@@ -0,0 +1,1702 @@
...
+    def test_select_variant_full_probability(self):
+        rollout = Rollout(variants={'v1': 1.0})
+        for i in range(10):
+            assert rollout.select_variant(f'user{i}') == 'v1'
+
+    def test_select_variant_without_seed(self):
+        rollout = Rollout(variants={'v1': 0.5, 'v2': 0.5})
+        # Without seed, still works but isn't deterministic
+        result = rollout.select_variant(None)
+        assert result in {'v1', 'v2', None}
```
alexmojaki:
```suggestion
        assert result in {'v1', 'v2'}
```

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+        | KeyIsNotPresent,
+        Discriminator('kind'),
+    ],
+)
+
+
+VariantKey = str
+VariableName = str
+
+# TODO: Do we need to make the following dataclasses into pydantic dataclasses or BaseModels so the validators run when
```
alexmojaki: yes, why not?

tests/test_variables.py
```diff
@@ -0,0 +1,1702 @@
...
+class TestRolloutOverride:
+    def test_basic_override(self):
+        override = RolloutOverride(
+            conditions=[ValueEquals(attribute='plan', value='enterprise')],
+            rollout=Rollout(variants={'premium': 1.0}),
+        )
+        assert len(override.conditions) == 1
+        assert override.rollout.variants == {'premium': 1.0}
+
+    def test_multiple_conditions(self):
```
alexmojaki: needs to test behaviour

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+        population: list[VariantKey | None] = []
+        weights: list[float] = []
+        for k, v in self.variants.items():
+            population.append(k)
+            weights.append(v)
+
+        p_code_default = 1 - sum(weights)
+        if p_code_default > 0:
+            population.append(None)
+            weights.append(p_code_default)
```
alexmojaki: this should be done in `__init__`.

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+    """A single stage in a scheduled rollout sequence.
+
+    Rollout schedules progress through stages sequentially, with each stage having its own
+    duration, rollout configuration, and optional conditional overrides. This allows for
+    gradual rollouts where traffic percentages can increase over time.
+
+    Example: A three-stage rollout might have:
+    - Stage 1: 5% of traffic for 1 hour (canary)
+    - Stage 2: 25% of traffic for 4 hours (early adopters)
+    - Stage 3: 100% of traffic (full rollout)
```
alexmojaki: i like this example better than the one in RolloutSchedule. but it makes sense for all the explanations of a schedule to be in RolloutSchedule, not here.

i don't like the idea of having to define many stages if i want a gradual rollout e.g. increase by 5% every hour. i'd rather say "increase linearly from x% to y% starting at time T over N hours".

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+    """The sequence of rollout stages to progress through.
+
+    Stages are processed in order. The active stage is determined by comparing
+    the current time against start_at and the cumulative durations of previous stages.
+    """
+    # TODO: Need to add rollback condition support (possibly only in backend?)
+    #   Note: we could add this client side using the logfire query client if the token has read capability.
+    #   However, this should maybe be discouraged if it's viable to run health check queries server-side.
+    #   We could expose a `health_check` field that contains one (or more?) SQL queries, which would either be
+    #   evaluated client side or server side. However, I don't love the
```
alexmojaki: same

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+    - Stage 1: 5% of traffic for 1 hour (canary)
+    - Stage 2: 25% of traffic for 4 hours (early adopters)
+    - Stage 3: 100% of traffic (full rollout)
+    """
+
+    duration: timedelta
+    """Duration to remain in this stage before progressing to the next.
+
+    Once a stage's duration has elapsed, the schedule automatically advances to the
+    next stage. If this is the final stage and its duration has elapsed, the schedule
```
alexmojaki: so does the final stage need a duration at all?

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+    The schedule is considered active when `start_at` is set and is in the past.
+    Once all stages have completed (i.e., current time exceeds start_at plus the
+    sum of all stage durations), the base rollout and overrides from the parent
+    VariableConfig are used.
+    """
+
+    start_at: datetime | None
+    """The datetime when this schedule becomes active.
+
+    If None, the schedule is inactive and the base rollout is used.
```
alexmojaki: is that useful?

logfire/variables/config.py
```diff
@@ -0,0 +1,533 @@
...
+    """
+
+    duration: timedelta
+    """Duration to remain in this stage before progressing to the next.
+
+    Once a stage's duration has elapsed, the schedule automatically advances to the
+    next stage. If this is the final stage and its duration has elapsed, the schedule
+    is considered complete.
+
+    Note: Automated rollback based on error rates is only supported server-side and should
```
alexmojaki: automated rollback of any kind sounds like quite an ambitious thing to implement. why are we even implementing scheduled stages right now at all? can we just get something basic out first?

logfire/variables/config.py
dmontagu: 🤦‍♂️

logfire/variables/config.py
dmontagu: but do you really mind if that happens in UI rather than in python struct? Doing it in UI means we can have a clean configuration option but can change the logic etc. and not require ambitious changes, doing it in struct means we need to support all the potential patterns we'd like for shorthand configuration. Happy to discuss more but I'm not convinced that a linear rollout is right most of the time, or that end users (in aggregate) won't end up needing flexibility.

logfire/variables/config.py
dmontagu: the point was that there is a meaning to the value even if it's not currently used. We can remove/comment-out all this code for now, probably makes sense to discuss how we want to handle this synchronously.

logfire/variables/config.py
dmontagu: do you mean None or having a way to specify the start time? It's useful to specify the start time — e.g. if you want the feature to go live tomorrow morning once the team is ready to deal with issues.

I added the ability to set it to `None` as a way to store the schedule on the config before you've decided precisely when you want it to go live. You could achieve this by just setting it to some time far in the future but I was also thinking this could play nice with automated rollback — if there's a failure, we update the config to have no start time here (and maybe have some other error field populated). I'm 100% open to tweaks to this.

logfire/variables/config.py
dmontagu: I think I explained this to you elsewhere, but the point was I wanted to convince myself that we weren't going to feel the need to add/modify the config shape to support these features if/when we wanted to ship them in the future. I'm comfortable removing and/or commenting out fields/code related to this for now.

logfire/variables/config.py
dmontagu: yeah I think I've convinced myself that whole set of behavior can live entirely server side / in CRUD for us, and you'd use IaC if you wanted to manage it in code. And that IaC config doesn't need to live here. Maybe that would be the right way to handle all of this schedule stuff, I think we can remove it from the PR for now.

logfire/variables/abstract.py
```diff
@@ -11,10 +12,28 @@
 T_co = TypeVar('T_co', covariant=True)
 
 
+# TODO: Should we rename this to `ResolvedVariable`?
```
alexmojaki: sounds reasonable

logfire/variables/abstract.py
```diff
@@ -32,6 +51,28 @@ class VariableResolutionDetails(Generic[T_co]):
...
+    def __enter__(self):
+        self._exit_stack.__enter__()
+
+        import logfire
+
+        # TODO:
+        #  * Should we "nest" the value into a 'logfire.variables' key, rather than separate keys for each variable?
+        #  * Is there a better value to use here over `<code_default>` when the variant is None?
+        #  * Should either of the above be configurable?
+        #  * Should we _require_ you to enter the context to get the value of a variable?
```
alexmojaki: - No, baggage works better with simple small values
- `<default>`?
- nah, can be done later anyway
- no, too inconvenient

logfire/variables/variable.py
```diff
@@ -116,21 +116,8 @@ def refresh_sync(self, force: bool = False):
...
-            targeting_key: Optional key for deterministic variant selection (e.g., user ID).
-            attributes: Optional attributes for condition-based targeting rules.
-
-        Returns:
-            The resolved value of the variable.
-        """
-        return (self.get_details(targeting_key, attributes)).value
-
-    def get_details(
+    # TODO: Should we rename this to `resolve`?
```
alexmojaki: get() sounds better to me

logfire/variables/variable.py
```diff
@@ -143,35 +130,64 @@ def get_details(
             A VariableResolutionDetails object containing the resolved value, selected variant,
             and any errors that occurred.
         """
+        # TODO:
+        #  * Should we include the serialized value as an attribute here?
+        #  * Should we _not_ include the deserialized value as an attribute here?
```
alexmojaki: no, yes. deserialized value seems easily better, except when deserialization fails. but the parsed json is still good even when it fails validation, so really that just means maybe logging the raw string if it's invalid json which is probably an extreme edge case.

logfire/variables/variable.py
```diff
@@ -143,35 +130,64 @@ def get_details(
...
-        # TODO: How much of the following code should be in the try: except:?
+        with self.logfire_instance.span(
+            'Resolve variable {name}',
+            name=self.name,
+            targeting_key=targeting_key,
+            attributes=merged_attributes,
+        ) as span:
+            result = self._resolve(targeting_key, merged_attributes)
+            span.set_attributes({
+                'name': result.name,
```
alexmojaki: is this different from `self.name`?

logfire/variables/variable.py
```diff
@@ -0,0 +1,218 @@
...
+            name: Unique name identifying this variable.
+            default: Default value to use when no configuration is found, or a function
+                that computes the default based on targeting_key and attributes.
+            type: The expected type of this variable's values, used for validation.
+            logfire_instance: The Logfire instance this variable is associated with. Used to determine config, etc.
+        """
+        self.name = name
+        self.default = default
+
+        self.logfire_instance = logfire_instance
```
alexmojaki:
```suggestion
        self.logfire_instance = logfire_instance.with_settings(custom_scope_suffix='variables')
```

logfire/variables/variable.py
```diff
@@ -0,0 +1,218 @@
...
+        #  * Should we _not_ include the deserialized value as an attribute here?
+        #  * Should we remove the `attributes` attribute of the span below?
+        #  * Should we remove the _merged_ part of the `attributes` attribute below?
+        #  * Should we _rename_ the `attributes` attribute below?
+        #  * Should some/all of the above be configurable? On a logfire/variable/call-site-specific basis?
+
+        merged_attributes = self._get_merged_attributes(attributes)
+
+        with self.logfire_instance.span(
+            'Resolve variable {name}',
```
alexmojaki:
```suggestion
        # Include the variable name directly here to make the span name more useful,
        # it'll still be low cardinality. This also prevents it from being scrubbed from the message.
        # Don't inline the f-string to avoid f-string magic.
        span_name = f'Resolve variable {self.name}'
        with self.logfire_instance.span(
            span_name,
```

logfire/variables/variable.py
```diff
@@ -0,0 +1,218 @@
...
+        self.logfire_instance.config.get_variable_provider().refresh(force=force)
+
+    # TODO: Should we rename this to `resolve`?
+    def get(
+        self, targeting_key: str | None = None, attributes: Mapping[str, Any] | None = None
+    ) -> VariableResolutionDetails[T]:
+        """Resolve the variable and return full details including variant and any errors.
+
+        Args:
+            targeting_key: Optional key for deterministic variant selection (e.g., user ID).
```
alexmojaki, line 195: I think we agreed a good default targeting_key would be trace_id

logfire/variables/variable.py
```diff
@@ -0,0 +1,218 @@
...
+        Returns:
+            A VariableResolutionDetails object containing the resolved value, selected variant,
+            and any errors that occurred.
+        """
+        # TODO:
+        #  * Should we include the serialized value as an attribute here?
+        #  * Should we _not_ include the deserialized value as an attribute here?
+        #  * Should we remove the `attributes` attribute of the span below?
+        #  * Should we remove the _merged_ part of the `attributes` attribute below?
+        #  * Should we _rename_ the `attributes` attribute below?
```
alexmojaki: i think it's fine as is

logfire/variables/variable.py
```diff
@@ -0,0 +1,218 @@
...
+        #  * Should some/all of the above be configurable? On a logfire/variable/call-site-specific basis?
```
alexmojaki: i think the most important part is that it should be possible to globally disable creating these spans entirely, or have them off by default and allow globally enabling them.

logfire/variables/config.py
alexmojaki: it sounds to me like there isn't a meaning to the value of the duration in the final stage

logfire/variables/config.py
alexmojaki: deferring this discussion since we're not doing rollout for now, right?

logfire/variables/variable.py
dmontagu: yeah good idea, I think we talked about that before but I forgot.

logfire/_internal/config.py
```diff
@@ -311,6 +311,17 @@ class RemoteVariablesConfig:
     """Whether the remote variables should be fetched before first resolving a value."""
     polling_interval: timedelta | float = timedelta(seconds=30)
     """The time interval for polling for updates to the variables config."""
+    api_token: str | None = None
+    """API token for accessing the variables endpoint.
+
+    If not provided, will be loaded from LOGFIRE_API_TOKEN environment variable.
+    This token should have the 'project:read_variables' scope.
```
alexmojaki: it's not used for writing variables?

logfire/_internal/config.py
```diff
@@ -311,6 +311,17 @@ class RemoteVariablesConfig:
...
+    """
+    base_url: str | None = None
+    """Base URL for the Logfire API.
+
+    If not provided, will be inferred from the API token or logfire token.
```
```suggestion
    If not provided, will be inferred from the API token or write token.
```

logfire/_internal/config_params.py
```diff
@@ -59,6 +59,8 @@ class _DefaultCallback:
 """Minimum log level for logs and spans to be created. By default, all logs and spans are created."""
 TOKEN = ConfigParam(env_vars=['LOGFIRE_TOKEN'])
 """Token for the Logfire API."""
+API_TOKEN = ConfigParam(env_vars=['LOGFIRE_API_TOKEN'])
+"""API token for Logfire API access (used for managed variables and other /v1 endpoints)."""
```
alexmojaki: I think more needs to be said to distinguish these, maybe emphasising that `LOGFIRE_TOKEN` is usually what users want.

logfire/variables/config.py
```diff
@@ -225,7 +225,7 @@ class Variant:
     # format: Literal['json', 'yaml']  # TODO: Consider supporting yaml, and not just JSON; allows comments and better formatting
     description: str | None = None
     """Optional human-readable description of this variant."""
-    version: str | None = None  # TODO: should this be required?
+    version: int | None = None  # TODO: should this be required? should this be `str`?
```
alexmojaki: `str` is more flexible, it can contain a number if that's all we need, but maybe we'll want a hash or uuid for some reason

logfire/variables/push.py
```diff
@@ -241,71 +297,163 @@ def _create_variable(
...
-    project: str,
+    client: _VariablesClient,
     change: VariableChange,
 ) -> None:
     """Update an existing variable's schema on the server."""
     server_id = change.server_id
     if not server_id:
         # Need to look up the variable by name to get its ID
-        var_data = client.get_variable_by_name(organization, project, change.name)
+        var_data = client.get_variable_by_name(change.name)
```
alexmojaki: it looks like this is the only place this endpoint is used? is it worth it?

logfire/variables/push.py
```diff
@@ -241,71 +297,163 @@ def _create_variable(
...
+    """Project information for variable operations."""
+
+    project_id: str
+    organization_name: str
+    project_name: str
+    project_url: str
+    logfire_api_url: str
+
+
+def _select_project(
```
alexmojaki: this looks like it's duplicating CLI logic, where there's already tech debt

logfire/variables/push.py
```diff
@@ -241,71 +297,163 @@ def _create_variable(
...
+            return None
+    except (ValueError, EOFError, KeyboardInterrupt):
+        print('\nAborted.')
+        return None
+
+
+def _get_variables_client_from_token(
+    data_dir: str | None = None,
+) -> tuple[_VariablesClient, str] | None:
+    """Try to create a variables client from LOGFIRE_TOKEN in credentials file.
```
alexmojaki: LOGFIRE_TOKEN is an env var, not something in a file. it looks like this intentionally doesn't use an env var.

logfire/variables/push.py
```diff
@@ -241,71 +297,163 @@ def _create_variable(
...
+            )
+        else:
+            print('\033[31mInvalid selection.\033[0m', file=sys.stderr)
+            return None
+    except (ValueError, EOFError, KeyboardInterrupt):
+        print('\nAborted.')
+        return None
+
+
+def _get_variables_client_from_token(
```
alexmojaki:
```suggestion
def _get_variables_client_from_creds_file(
```

or `from_write_token`

logfire/_internal/config.py
alexmojaki: oh god, is the user token used for that? are we involving 3 tokens?

logfire/variables/push.py
```diff
@@ -241,71 +297,163 @@ def _create_variable(
...
+        client = LogfireClient.from_url(base_url)
+    except Exception as e:
+        print(f'\033[31mError: Not logged in or invalid user token: {e}\033[0m', file=sys.stderr)
+        return None
+
+    # Get list of writable projects
+    try:
+        projects = client.get_user_projects()
+    except Exception as e:
+        print(f'\033[31mError fetching projects: {e}\033[0m', file=sys.stderr)
```
alexmojaki: doesn't `e` already contain `Error retrieving list of projects`?

logfire/variables/push.py
```diff
@@ -241,71 +297,163 @@ def _create_variable(
...
+    """
+    from logfire._internal.client import LogfireClient
+
+    try:
+        client = LogfireClient.from_url(base_url)
+    except Exception as e:
+        print(f'\033[31mError: Not logged in or invalid user token: {e}\033[0m', file=sys.stderr)
+        return None
+
+    # Get list of writable projects
```
alexmojaki: we're not restricted to writable projects though are we?

logfire/variables/push.py
```diff
@@ -323,11 +471,18 @@ def push_variables(
     - Updates JSON schemas for existing variables if they've changed
     - Warns about existing variants that are incompatible with new schemas
 
+    Authentication priority:
+    1. If LOGFIRE_TOKEN is found in the data_dir credentials file, uses that token
```
alexmojaki: LOGFIRE_TOKEN is an env var, not found in the file

logfire/variables/push.py
```diff
@@ -359,28 +513,30 @@ def push_variables(
...
+    if result is not None:
+        client, project_url = result
+        print(f'Using project token for: {project_url}')
+    else:
+        # Fall back to user token auth
+        result = _get_variables_client_from_user_token(yes=yes)
+        if result is None:
+            print(
+                '\033[31mError: No valid credentials found. Either:\n'
+                '  - Run your application with LOGFIRE_TOKEN set, or\n'
```
alexmojaki: Setting LOGFIRE_TOKEN doesn't look like it'll do anything

logfire/variables/push.py
```diff
@@ -533,26 +697,30 @@ def validate_variables(
...
-    # Create client with user auth
-    try:
-        client = LogfireClient.from_url(None)
-    except Exception as e:
-        print('\033[31mError: Failed to authenticate. Run `logfire auth` first.\033[0m', file=sys.stderr)
-        print(f'\033[31m{e}\033[0m', file=sys.stderr)
-        return False
+    # Try to get a variables client
+    # Priority: 1. LOGFIRE_TOKEN from credentials file, 2. User token with project selection
+    result = _get_variables_client_from_token(data_dir)
```
alexmojaki: this code looks very duplicated

logfire/variables/push.py
```diff
@@ -0,0 +1,771 @@
...
+            logfire.push_variables()
+
+            # Or push specific variables only
+            logfire.push_variables([feature_enabled])
+        ```
+    """
+    import logfire as logfire_module
+
+    if variables is None:
+        variables = logfire_module.DEFAULT_LOGFIRE_INSTANCE.get_variables()
```
alexmojaki:
```suggestion
        variables = logfire_module.get_variables()
```

logfire/variables/push.py
```diff
@@ -0,0 +1,771 @@
...
+                lines.append(f'  \033[31m✗ {error.variable_name} (variant: {error.variant_key})\033[0m')
+                # Format the error message, indenting each line
+                error_lines = str(error.error).split('\n')
+                for line in error_lines[:5]:  # Limit to first 5 lines
+                    lines.append(f'      {line}')
+                if len(error_lines) > 5:
+                    lines.append(f'      ... ({len(error_lines) - 5} more lines)')
+
+    if report.variables_not_on_server:
+        lines.append('\n\033[33m=== Variables Not Found on Server ===\033[0m')
```
alexmojaki: please use named constants instead of `\033[33m` etc, i don't know what color this is

logfire/variables/push.py
```diff
@@ -0,0 +1,771 @@
...
+    if not diff.has_changes:
+        print('\n\033[32mNo changes needed. Server is up to date.\033[0m')
+        return False
+
+    # Check for incompatible variants in strict mode
+    if strict:
+        has_incompatible = any(c.incompatible_variants for c in diff.changes if c.change_type == 'update_schema')
+        if has_incompatible:
+            print(
+                '\n\033[31mError: Some existing variants are incompatible with the new schema. '
```
alexmojaki: Seems like it should inform the user even without --strict

logfire/variables/push.py
```diff
@@ -0,0 +1,771 @@
...
+
+
+@dataclass
+class VariableChange:
+    """Represents a change to be made to a variable."""
+
+    name: str
+    change_type: str  # 'create', 'update_schema', 'no_change'
+    local_schema: dict[str, Any] | None = None
+    server_schema: dict[str, Any] | None = None
```
alexmojaki: Can we get some proper typing and pydantic validation instead of Any?

logfire/variables/push.py
```diff
@@ -0,0 +1,771 @@
...
+    changes: list[VariableChange]
+    orphaned_server_variables: list[str]  # Variables on server not in local code
+
+    @property
+    def has_changes(self) -> bool:
+        """Return True if there are any changes to apply."""
+        return any(c.change_type != 'no_change' for c in self.changes)
+
+
+def _get_json_schema(variable: Variable[Any]) -> dict[str, Any]:
```
alexmojaki: Should all these `Variable[Any]`s be `Variable[object]`?

tests/test_push_variables.py
```diff
@@ -0,0 +1,433 @@
...
+                'json_schema': {'type': 'boolean'},
+                'variants': {},
+            },
+        }
+    }
+
+    diff = _compute_diff([var], server_config)
+
+    assert 'orphan-feature' in diff.orphaned_server_variables
+    assert 'my-feature' not in diff.orphaned_server_variables
```
alexmojaki: Please replace as many assertions as possible with inline-snapshot

logfire/variables/variable.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Args:
+            name: Unique name identifying this variable.
+            default: Default value to use when no configuration is found, or a function
+                that computes the default based on targeting_key and attributes.
+            type: The expected type of this variable's values, used for validation.
+            logfire_instance: The Logfire instance this variable is associated with. Used to determine config, etc.
+        """
+        self.name = name
+        self.default = default
```
alexmojaki, line 119:
```suggestion
        self.default = default
        self.description = description
        self.value_type = type
```

logfire/variables/variable.py
dmontagu: Sorry, so what are you suggesting we include as attributes? Based on the "no, yes" I would assume nothing, but the way you've written the answer sounds like you do want something..

logfire/variables/variable.py
dmontagu: I added an "aliases" mechanism to migrate variable names if preferred, so you can request the value of a variable using a now-replaced value of the name, and this will have the value of the "current" name rather than the alias you may have used in the code when declaring the value (which is the value of `self.name`). I'm not sure if the aliases functionality is necessary or even a good idea, can change it if you want.

logfire/variables/config.py
dmontagu: More runtime overhead, and I'd also be concerned about import time building the schemas, maybe it's negligible though. Also deepens the dependence on pydantic, maybe it's fine but if we wanted any version of supporting this _without_ using pydantic it would become harder to achieve. (Obviously right now that's not even a goal..)

logfire/_internal/config.py
dmontagu: I have added support for all of the above in the corresponding platform PR, but I'm okay keeping things simpler in the SDK for now. I'd be okay just requiring the use of a project-scoped API token for everything right now.

dmontagu: This docstring only mentions `project:read_variables` scope because, as part of this struct, that is the only scope needed. But for other usage of the API token other scopes may be relevant. (This struct is not involved with variable writing; I believe that happens in some other API calls that don't make use of this struct.)

dmontagu: Yes, you can see the logic in e.g. `logfire.variables.push.push_variables`. I'll update that to _allow_ passing an API token manually but if not provided it is read from environment (similar to logfire.configure).

logfire/_internal/config_params.py
dmontagu: I've updated to:
```python
TOKEN = ConfigParam(env_vars=['LOGFIRE_TOKEN'])
"""Token for sending application telemetry data to Logfire, also known as a "write token"."""
API_TOKEN = ConfigParam(env_vars=['LOGFIRE_API_TOKEN'])
"""API token for Logfire API access (used for managed variables and other public APIs)."""
```
But not sure if this is good enough.. Also not sure if we should more seriously consider renaming things..

logfire/variables/config.py
dmontagu: I think it's probably reasonable to have some or all of:
* date it was updated at
* numeric incrementing version number
* content hash

Keeping it structured like above makes it unambiguous what should be done when creating new versions and doesn't require a user to choose a versioning scheme. I think allowing a user-defined string will just add a bunch of up-front cognitive burden for the user with minimal benefit.

That said, I'm not sure how much of the above is worth keeping and also not sure how it should be exposed in the structs / web API. Right now I just have the numeric incrementing version number but I'm willing to change this. Let's discuss next week.

logfire/variables/push.py
dmontagu: removed

logfire/variables/push.py
dmontagu: removed

logfire/variables/push.py
dmontagu: removed

logfire/variables/push.py
dmontagu: if we had a type for a JSON schema I'd be more of a fan of that, I don't really want to introduce one in this PR, I feel like it's something the pydantic library itself should introduce support for

logfire/variables/push.py
dmontagu: we can, but does it matter?

logfire/_internal/config.py
dmontagu: I've refactored this now so that `push_variables` is a method on the logfire instance, and uses the variable provider. So I think this comment is now mostly irrelevant.

logfire/_internal/main.py
```diff
@@ -2336,24 +2344,148 @@ def shutdown(self, timeout_millis: int = 30_000, flush: bool = True) -> bool:  #
...
             return False
-        self._meter_provider.shutdown(remaining)
+
         return (start - time()) < timeout_millis
 
+    def var(
+        self,
+        name: str,
+        *,
+        type: type[T] | Sequence[type[T]],
```
Viicos: Is it worth _not_ supporting the sequence type until PEP 747 (type forms) gets accepted? Presumably, once we have this PEP we just need users to provide a union type?

I'm mentioning this because the PEP got submitted to the steering council a while ago, and it's only a matter of days until they have a meeting to decide on the fate of this PEP.

logfire/variables/variable.py
```diff
@@ -0,0 +1,231 @@
...
+            description: Optional human-readable description of what this variable controls.
+            logfire_instance: The Logfire instance this variable is associated with. Used to determine config, etc.
+        """
+        self.name = name
+        self.value_type = type
+        self.default = default
+        self.description = description
+
+        self.logfire_instance = logfire_instance.with_settings(custom_scope_suffix='variables')
+        self.type_adapter = TypeAdapter[T](type)
```
Viicos: Not sure how variables are being used, but maybe it's worth deferring the build? (with the `defer_build` config).

dmontagu: Is there potential for issues with resolving forward refs that eager building avoids? I want to prevent those sorts of issues

docs/reference/advanced/managed-variables.md
```diff
@@ -0,0 +1,763 @@
...
+# Define your variables
+agent_config = logfire.var(
+    name='support-agent-config',
+    type=AgentConfig,
+    default=AgentConfig(...),
+)
+```
+
+!!! note "API Token Required"
+    Remote variables require an API token with the `project:read_variables` scope. This is different from the write token (`LOGFIRE_TOKEN`) used to send traces and logs. Set it via the `LOGFIRE_API_TOKEN` environment variable or pass it directly to `RemoteVariablesConfig(api_token=...)`.
```
alexmojaki: is creating these documented?

logfire/variables/abstract.py
```diff
@@ -0,0 +1,906 @@
...
+            strict: If True, fail if any existing variants are incompatible with new schemas.
+
+        Returns:
+            True if changes were applied (or would be applied in dry_run mode), False otherwise.
+        """
+        if not variables:
+            print('No variables to push. Create variables using logfire.var() first.')
+            return False
+
+        print(f'Using provider: {type(self).__name__}')
```
alexmojaki: printing `Using provider: LogfireRemoteVariableProvider` in the typical case seems weird. do we need this at all?

logfire/variables/abstract.py
```diff
@@ -0,0 +1,906 @@
...
+            message = f'Some changes will result in variants incompatible with the new schema.\n{incompatible_changes}'
+            if strict:
+                print(
+                    f'\n{ANSI_RED}Error: {message}\nRemove --strict flag to proceed anyway.{ANSI_RESET}',
+                    file=sys.stderr,
+                )
+                return False
+            else:
+                print(
+                    f'\n{ANSI_YELLOW}Warning: {message}\n',
```
alexmojaki: missing reset at the end here. probably best to make some util functions like `print(f'{color}{message}{ANSI_RESET}, file=sys.stderr)` to make this mistake harder to make

logfire/variables/variable.py
```diff
@@ -0,0 +1,231 @@
...
+
+        Returns:
+            A ResolvedVariable object containing the resolved value, selected variant,
+            and any errors that occurred.
+        """
+        # TODO:
+        #  * Should we include the serialized value as an attribute here?
+        #  * Should we _not_ include the deserialized value as an attribute here?
+        #  * Should some/all of the above be configurable? On a logfire/variable/call-site-specific basis?
+        #   TODO: Alex says "i think the most important part is that it should be possible to globally disable creating these spans entirely, or have them off by default and allow globally enabling them."
```
alexmojaki: why not keep the gh comment unresolved?

logfire/variables/variable.py
alexmojaki: include the most deserialized value available, ideally something pydantic validated, otherwise json

logfire/variables/variable.py
Viicos: It shouldn't affect anything, the `TypeAdapter` eagerly fetches the parent namespace in both cases. I'll note that you can in fact adapt the `_parent_depth` argument to actually use the namespace where the variable was defined (but this can get complex, as I assume users are required to use the `var()` function instead of the `Variable` class directly, so there's extra depth to be taken into account).

However, with Python 3.14 now available, I would avoid trying to cope with string forward references as much as possible. At some point, they will be something of the past.

logfire/variables/config.py
dmontagu: changed them to be pydantic models on Marcelo's suggestion

logfire/variables/variable.py
dmontagu: I think for now let's ignore this, it's going to be rare, if anyone complains about build time or other issues we can defer the builds though. I'd rather have the app be (barely) slow than buggy, and given it only affects users I think it's okay

docs/reference/advanced/managed-variables.md
dmontagu: not sure, I guess we also need to figure out what we want to do with docs given this will be behind a feature flag anyway..

logfire/_internal/main.py
dmontagu: I prefer `managed_variable` or `variable` over `var`. @Kludex more strongly prefers `managed_variable` over both of those.

logfire/__init__.py
```diff
@@ -85,6 +93,16 @@
...
 
+# Variables
+var = DEFAULT_LOGFIRE_INSTANCE.var
+get_variables = DEFAULT_LOGFIRE_INSTANCE.get_variables
+push_variables = DEFAULT_LOGFIRE_INSTANCE.push_variables
+push_variable_types = DEFAULT_LOGFIRE_INSTANCE.push_variable_types
+validate_variables = DEFAULT_LOGFIRE_INSTANCE.validate_variables
+sync_config = DEFAULT_LOGFIRE_INSTANCE.sync_config
+pull_config = DEFAULT_LOGFIRE_INSTANCE.pull_config
+generate_config = DEFAULT_LOGFIRE_INSTANCE.generate_config
```
alexmojaki: these names sound too generic and somehow related to logfire.configure. they should all have `variables` in their name. maybe even rename all these methods (except `var`) to start with `variables_`, e.g. `logfire.variables_get`, `logfire.variables_push`, etc, a bit like the `instrument_` and `metric_` 'namespaces'. it's a bit more autocomplete friendly. it would also be possible to implement `logfire.variables.get` etc in a way that works with both the modules and Logfire instances.

alexmojaki: that point aside, there's inconsistency in the verbs used: get vs pull and sync vs push. sync is especially unclear.

logfire/variables/config.py
```diff
@@ -0,0 +1,545 @@
...
+                rollout=Rollout(variants={}),
+                overrides=[],
+                json_schema=json_schema,
+                example=example,
+            )
+            variable_configs[variable.name] = config
+
+        return VariablesConfig(variables=variable_configs)
+
+    def merge(self, other: VariablesConfig) -> VariablesConfig:
```
alexmojaki, line 473: is this used?

docs/reference/advanced/managed-variables.md
```diff
@@ -0,0 +1,1256 @@
...
+
+# Define your variables
+agent_config = logfire.var(name='agent_config', type=AgentConfig, default=AgentConfig(...))
+feature_flag = logfire.var(name='feature-enabled', type=bool, default=False)
+
+# Generate a config with name, schema, and example for each variable
+config = logfire.generate_config()
+
+# Save to a YAML file (JSON also supported)
+config.write('variables.yaml')
```
alexmojaki: this method doesn't seem to exist

dmontagu:
I'm going to open a new PR, this has gotten too long

PR #1552, opened 2025-11-29
feat(config): allow custom Views in MetricOptions

cyberksh:
I'm using the FastAPI integration which exposes metrics like [http.server.response.size](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-asgi/src/opentelemetry/instrumentation/asgi/__init__.py#L641) which I want to drop. 

The only way to do this seems to be to add a custom View and setting `DropAggregation`:

```python
View(
    instrument_name="http.server.response.size",
    aggregation=DropAggregation()
)
```

### DIFF
--- a/docs/join-slack/index.html
+++ b/docs/join-slack/index.html
@@ -5,7 +5,7 @@
     <meta charset="utf-8">
     <title>Redirecting to slack...</title>
     <script>
-      location.href="https://join.slack.com/t/pydanticlogfire/shared_invite/zt-30nu5u4c0-kKihQCzbaFF7_NQa_dxMUQ"
+      location.href="https://join.slack.com/t/pydanticlogfire/shared_invite/zt-3kkfnpjsl-DCt9R8GJs4IY3Aej9tIBYg"
     </script>
     <meta http-equiv="refresh" content="0; url=https://join.slack.com/t/pydanticlogfire/shared_invite/zt-30nu5u4c0-kKihQCzbaFF7_NQa_dxMUQ">
 </head>
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,7 +180,7 @@ dev = [
     "opentelemetry-instrumentation-google-genai >= 0.4b0",
     "google-genai >= 0",
     "openinference-instrumentation-litellm >= 0",
-    "litellm >= 0",
+    "litellm != 1.80.9",
     "pip >= 0",
 ]
 docs = [
@@ -197,7 +197,7 @@ docs = [
 
 [tool.inline-snapshot]
 default-flags = ["disable"]
-format-command="ruff format --stdin-filename {filename}"
+format-command = "ruff format --stdin-filename {filename}"
 
 [tool.uv.sources]
 logfire-api = { workspace = true }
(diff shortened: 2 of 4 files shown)


logfire/_internal/config.py
```diff
@@ -230,6 +248,9 @@ class MetricsOptions:
     collect_in_spans: bool = False
     """Experimental setting to add up the values of counter and histogram metrics in active spans."""
 
+    views: Sequence[View] = DEFAULT_VIEWS
+    """Sequence of views to be enabled for metric collection"""
```
alexmojaki: It should be easy to keep the default views, something like this: `MetricsOptions(views=MetricsOptions.DEFAULT_VIEWS + [...])`

Please also expand this docstring and add tests.

cyberksh: Done

logfire/_internal/config.py
```diff
@@ -224,12 +242,36 @@ class PydanticPlugin:
 class MetricsOptions:
     """Configuration of metrics."""
 
+    DEFAULT_VIEWS: ClassVar[Sequence[View]] = DEFAULT_VIEWS
```
alexmojaki: please move the definition here, no need for the global constant

tests/test_metrics.py
```diff
@@ -530,3 +530,43 @@ def test_reconfigure(caplog: pytest.LogCaptureFixture):
...
+    )
+
+    # Extend DEFAULT_VIEWS with a custom view
+    extended_views = [*logfire.MetricsOptions.DEFAULT_VIEWS, custom_view]
+    assert len(extended_views) == 3
+    assert extended_views[-1] is custom_view
+
+    # Verify we can create MetricsOptions with extended views
+    options = logfire.MetricsOptions(views=extended_views)
+    assert len(options.views) == 3
```
alexmojaki: Please test the actual effect of replacing/extending views by emitting metrics

tests/test_metrics.py
```diff
@@ -530,3 +530,43 @@ def test_reconfigure(caplog: pytest.LogCaptureFixture):
...
+
+    # Each element should be a View
+    for view in logfire.MetricsOptions.DEFAULT_VIEWS:
+        assert isinstance(view, View)
+
+
+def test_metrics_options_extend_default_views() -> None:
+    """Test that users can extend DEFAULT_VIEWS with their own views."""
+    from opentelemetry.sdk.metrics import Counter
+    from opentelemetry.sdk.metrics.view import View
```
alexmojaki: please keep these imports at the top level

logfire/_internal/config.py
```diff
@@ -224,12 +224,52 @@ class PydanticPlugin:
...
+      to reduce cardinality.
+    """
+
     additional_readers: Sequence[MetricReader] = ()
     """Sequence of metric readers to be used in addition to the default which exports metrics to Logfire's API."""
 
     collect_in_spans: bool = False
     """Experimental setting to add up the values of counter and histogram metrics in active spans."""
 
+    views: Sequence[View] = field(default_factory=lambda: MetricsOptions.DEFAULT_VIEWS)
```
alexmojaki, line 264: why the change to use default_factory?

cyberksh:
```python
line 264, in MetricsOptions
    views: Sequence[View] = MetricsOptions.DEFAULT_VIEWS
                            ^^^^^^^^^^^^^^
NameError: name 'MetricsOptions' is not defined
```

Doesn't work otherwise.

alexmojaki:
Thanks!

PR #1561, opened 2025-12-04
Support kwargs in ProxyLogger emit()

qianl15:
This PR fixes an issue where `logger.emit()` throws `TypeError: ProxyLogger.emit() got an unexpected keyword argument 'timestamp'` when it uses the new interface from OpenTelemetry 1.38.0. The related breaking change is from https://github.com/open-telemetry/opentelemetry-python/pull/4737

The fix is to add an overload for the `emit()` function to accept kwargs as well as `LogRecord`.


alexmojaki:
Thanks!

PR #1578, opened 2025-12-16
Update why.md to fix formatting

strawgate:
It looks like the grid cards were being used to create a "-> Read More" CTA on the Why Logfire page.

It doesn't look the cleanest but not bad.
Screenshot 2025-12-15 at 10 53 13 PM

It seems like maybe the grid cards expect each card to be its own line in the file. The `Pydantic Integration` card having two lines of text pushes the "-> Read More" CTA to the left side breaking the formatting:
Screenshot 2025-12-15 at 10 51 43 PM


alexmojaki:
Thanks!

PR #1588, opened 2025-12-19
Enforce working examples in docstrings and documentation

### LINKED ISSUE #1587: Add pytest-examples to verify documentation examples
Add pytest-examples (perhaps first with just linting/formatting?) to detect drift between examples in docs and reality

From https://github.com/pydantic/logfire/pull/1582#issuecomment-3674336941

strawgate:
Require formatted and runnable examples in docstrings and documentation, and bring the documentation and code up to compliance with these new requirements. Closes #1587 

Uses `pytest-examples` to ensure that examples in docstrings and documentation are properly formatted and runnable with assertions on any `print` statements in the example. Adds `pytest-timeout` to the dev dependencies which is used to prevent example tests from running forever if they hit blocking code. 

There are ~300 newly passing tests and ~140 skipped tests (examples that are not checked for formatting or if they are runnable).

### DIFF
--- a/docs/integrations/pytest.md
+++ b/docs/integrations/pytest.md
@@ -114,8 +114,7 @@ When running tests with `--logfire`, any instrumented library calls create spans
 
 ### Example: Custom Spans in Tests
 
-```python
-# test_api.py
+```python title="test_api.py"
 import logfire
 
 
@@ -142,7 +141,7 @@ pytest: my-project
 
 ### Example: Logging During Tests
 
-```python
+```python skip="true" skip-reason="incomplete"
 import logfire
 
 def test_operation():
@@ -187,7 +186,7 @@ After running your tests with `--logfire`, you can view the traces in the Logfir
 
 If you were previously using a manual tracing pattern in `conftest.py`:
 
-```python
+```python skip="true" skip-reason="incomplete"
 # Old pattern - no longer needed
 @pytest.hookimpl(hookwrapper=True)
 def pytest_runtest_protocol(item):


docs/guides/onboarding-checklist/add-metrics.md
```diff
@@ -39,11 +40,7 @@ To create a counter metric, use the [`logfire.metric_counter`][logfire.Logfire.m
 ```py
 import logfire
 
-counter = logfire.metric_counter(
-    'exceptions',
-    unit='1',  # (1)!
-    description='Number of exceptions caught'
```
alexmojaki, line 49: Add a comma at the end here to keep the args on separate lines, especially so that `# (1)!` stays on the `unit` arg. same for others below.

docs/how-to-guides/query-api.md
```diff
@@ -195,20 +197,20 @@ LIMIT 1
...
 if response.status_code == 200:
-    print("Query Successful!")
+    print('Query Successful!')
     print(response.json())
 else:
-    print(f"Failed to execute query. Status code: {response.status_code}")
+    print(f'Failed to execute query. Status code: {response.status_code}')
+    #> Failed to execute query. Status code: 401
     print(response.text)
+    #> {"detail":"Invalid token"}
```
alexmojaki:
```suggestion
    print(response.text)
```

this is being skipped, right?

docs/integrations/http-clients/aiohttp.md
```diff
@@ -54,85 +54,95 @@ The `logfire.instrument_aiohttp_client()` method accepts various parameters to c
...
-import logfire
 from aiohttp.tracing import TraceRequestEndParams, TraceRequestExceptionParams
 from opentelemetry.trace import Span
-from typing import Union
+
+import logfire
 
 
-def capture_response_headers(span: Span, response: Union[TraceRequestEndParams, TraceRequestExceptionParams]):
+def capture_response_headers(span: Span, response: TraceRequestEndParams | TraceRequestExceptionParams):
```
alexmojaki:
```suggestion
def capture_response_headers(span: Span, response: 'TraceRequestEndParams | TraceRequestExceptionParams'):
```

then skip the future import?

docs/reference/sql.md
```diff
@@ -124,7 +124,9 @@ logfire.configure()
 with logfire.span('foo') as span:
     trace_id = span.get_span_context().trace_id
     print(trace_id)
+    #> 2135173042243855739684686441807112708
     print(format_trace_id(trace_id))
+    #> 019b38289f5dfbf64cd7de68d37c6a04
 ```
 
 will print something like:
```
alexmojaki: Let's not have both this and the comments above

docs/integrations/web-frameworks/wsgi.md
```diff
@@ -17,22 +17,24 @@ Install `logfire` with the `wsgi` extra:
...
+    return [b'Hello World']
+
 
 app = logfire.instrument_wsgi(app)
 
-with make_server("", 8000, app) as httpd:
-    print("Serving on port 8000...")
+with make_server('', 8000, app) as httpd:
+    print('Serving on port 8000...')
+    #> Serving on port 8000...
```
alexmojaki:
```suggestion
```

tests/test_docs.py
```diff
@@ -0,0 +1,73 @@
...
+ruff_ignore = [
+    'D101',  # ignore missing docstring in public classes
+    'D102',  # ignore missing docstring in public methods
+    'D103',  # ignore missing docstring in public functions
+]
+
+
+# Override the autouse fixtures from conftest.py to prevent them from
+# interfering with doc examples. Doc examples call their own logfire.configure()
+# and we don't want them to pollute the test exporter state.
```
alexmojaki: I don't see how the test exporter state would get polluted. are you sure this fixture is needed?

strawgate:
Hey sorry, should have left this in draft, I'll pick this back up and get it ready to be merged

docs/how-to-guides/query-api.md
strawgate: It is yeah, wasn't sure if we wanted the output to remain in the skipped tests or not but happy to remove

docs/integrations/http-clients/aiohttp.md
strawgate: That is a lot better

tests/test_docs.py
strawgate: I was having issues with some of the tests and global state and tried this, it didnt make a difference, this is definitely not needed.

strawgate:
@alexmojaki thanks for the review, I think this is looking better now.

docs/integrations/llms/openai.md
```diff
@@ -9,8 +9,9 @@ We support instrumenting both the [standard OpenAI SDK](https://github.com/opena
 
 **Logfire** supports instrumenting calls to OpenAI with the [`logfire.instrument_openai()`][logfire.Logfire.instrument_openai] method, for example:
 
-```python hl_lines="7"
+```python hl_lines="7" skip-run="true" skip-reason="external-connection"
```
alexmojaki: See https://github.com/pydantic/logfire/issues/1639

You've added a line which means that the hl_lines 7 is inaccurate now. Unfortunately, either that issue has to be solved or the whole pull request needs to be checked for all uses of hl_lines to make sure that they haven't gone stale.

strawgate: I'll review all uses of hl_lines in the docs (there are only 20) and report back

strawgate: There are actually a bunch of line highlighting issues on main so I can address any caused by this PR and fix the other ones on main
Screenshot 2026-01-19 at 12 31 12 PM
Screenshot 2026-01-19 at 12 31 27 PM

strawgate:
@alexmojaki i did an audit of the 20 uses of hl_lines in the docs, resolved all issues I could find (whether caused by this PR or currently mishighlighted on main), and manually verified the correct lines are highlighted in the served docs with this PR

PR #1589, opened 2025-12-19
Add API client to SDK (which uses API tokens)

dmontagu:
Wanting to get the patterns for this sorted out now so that we can use it for managed variables.

Not sure if there are big picture differences in how we want this functionality exposed, but I figured it was worth opening as a PR now so we can iterate without that all happening in the managed variables PR.


dmontagu:
I'm also imagining that we can roll the API querying currently exposed in the read client into this.

PR #1595, opened 2025-12-26
Adds Aiohttp request body capture.

### LINKED ISSUE #1227: Support request and response body capture in aiohttp client instrumentation

adtyavrdhn:
### closes #1227  

Final piece of the aiohttp client captures, was able to get to it after a while :)


adtyavrdhn:
@alexmojaki I was able to get to it finally. I tried to use the same frame walking trick you used in httpx.

logfire/_internal/integrations/aiohttp_client.py
```diff
@@ -69,6 +108,45 @@ class LogfireAioHttpRequestInfo(TraceRequestStartParams, LogfireClientInfoMixin)
     def capture_headers(self):
         capture_request_or_response_headers(self.span, self.headers, 'request')
 
+    def capture_body_if_text(self, attr_name: str = 'http.request.body.text'):
```
alexmojaki, line 112:
```suggestion
    @handle_internal_errors
    def capture_body_if_text(self, attr_name: str = 'http.request.body.text'):
```

alexmojaki:
Awesome, thanks! The docs need updating, especially https://logfire.pydantic.dev/docs/integrations/http-clients/aiohttp/#configuration, but that can be a separate PR.

adtyavrdhn:
I can add it in this PR by tomorrow :)

Issue #1614, opened 2026-01-09
Only test latest pydantic in weekly CI


alexmojaki:
We currently install pydantic from the latest main branch on github in development by default. This requires building pydantic core when installing dependencies which is slow. We should only do this in the weekly CI job (`.github/workflows/weekly_deps_test.yml`), while normal development just loads the latest version on pypi.

SoulSniper-V2:
Hi, I can handle this! I'll update the workflow to only use the GitHub main branch build in the weekly job and default to PyPI for standard dev.

strawgate:
It looks like maybe this was already resolved?

Interestingly, in https://github.com/pydantic/logfire/commit/5126c16e769bb9677db7dd2d832149dde859921f which points to PR 1615 which doesn't appear to actually exist...

alexmojaki:
Really weird, seems like the PR got deleted?

PR #1615, opened 2026-01-09
optimize: default to stable pydantic in dev, keep git build for weekly CI

SoulSniper-V2:
Addresses #1614.

Switches the default `dev` dependency to stable `pydantic` to speed up local installs. The `weekly_deps_test` workflow is updated to explicitly install `pydantic @ git+...` when testing against `main`, ensuring we still catch regressions.


alexmojaki:
Thanks!

Issue #1586, opened 2025-12-19
Migrate OpenAI and Anthropic instrumentations to new OpenTelemetry semantic conventions


alexmojaki:
`instrument_openai`, `instrument_openai_agents`, and `instrument_anthropic` currently use a messy mix of legacy custom data formats. None use the semantic conventions in https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/ . We want to fix all that.

Eventually we want to only produce new-style data and not keep any legacy code around. This will require at least one major version bump. Ideally only one, but that might be tricky. It could help to have a branch into which several PRs are made.

Ideally we would be using `opentelemetry-python-contrib` libraries to reduce the maintenance load and avoid reinventing the wheel, but they seem to move very slowly. Here's the state they're currently in:

- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4042 was recently opened and instruments the OpenAI responses API, which I requested some 8 months ago.
- [opentelemetry-instrumentation-openai-agents-v2](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-openai-agents-v2) could deal with `instrument_openai_agents` but I haven't looked at how well it works.
- https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4034 adds basic anthropic instrumentation, right now that library is just a boilerplate skeleton.

I think simultaneously updating behaviour and switching to these libraries is too difficult, especially as it would require making major contributions to those libraries. I've been waiting for them to catch up (e.g. instrument the Responses API) but it's been too slow. But we should still aim to use them eventually. So we should check what those libraries produce and try to align with them to avoid future breaking changes. If the contrib libraries don't follow semconv, an issue should be opened there. If there's no semconv for some particular bit of data, we should follow whatever the library does.

TODO:

- [x] Use `parse_json_attributes=True` in all calls to `exporter.exported_spans_as_dict` in related tests. This can be a trivial PR on its own that doesn't touch behaviour but will make future PRs cleaner.
- [x] Migrate attributes of `instrument_openai` for chat completions and `instrument_anthropic`, being done in https://github.com/pydantic/logfire/pull/1580 . Drop `request_data` and `response_data` entirely in favour of `gen_ai.*`.
[...]

PR #1621, opened 2026-01-09
Update GenAI span format to OTel semantic conventions

jimilp7:
- Span names now use '{operation} {model}' format (e.g., 'chat gpt-4o-mini')
  - Removed deprecated gen_ai.system attribute from OpenAI responses
  - Spans now use SpanKind.CLIENT instead of INTERNAL


alexmojaki:
Marking as draft since this contains breaking changes, will come back to this later

dmontagu:
Closing — the breaking semconv migration is tracked as a whole in #1586, and that's where this work will be picked up rather than from this branch. Thanks for the early push on it.

PR #1626, opened 2026-01-13
fix: pass LOGFIRE_BASE_URL to MCP configurations for self-hosted instances

mythral:
## Summary

- Fixes MCP server configuration to include `LOGFIRE_BASE_URL` when using `--base-url` flag
- Without this fix, self-hosted Logfire instances couldn't use the `prompt` command with `--claude`, `--codex`, or `--opencode` flags because the MCP server would default to the standard Logfire URL

## Problem

When running:
```bash
logfire --base-url https://logfire.example.com prompt \
  --project org/project issue:123 --claude
```

The CLI correctly used the custom URL for API calls (authentication, token creation, prompt retrieval), but when configuring the MCP server, only `LOGFIRE_READ_TOKEN` was passed. The `LOGFIRE_BASE_URL` environment variable was not included, causing the MCP server to connect to the wrong Logfire instance.

## Changes

- Updated `configure_claude()` to include `-e LOGFIRE_BASE_URL=<url>` in the claude mcp add command
- Updated `configure_codex()` to include `LOGFIRE_BASE_URL` in the TOML configuration
- Updated `configure_opencode()` to include `LOGFIRE_BASE_URL` in the JSON configuration
- Added tests to verify base URL is passed correctly for all three integrations

## Test plan

- [x] All existing prompt-related tests pass
- [x] New tests verify `LOGFIRE_BASE_URL` is included in Claude, Codex, and OpenCode configurations
- [x] Format and lint checks pass

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1857,10 +1857,15 @@ def test_parse_prompt_claude_no_mcp(
 ) -> None:
     monkeypatch.setattr(shutil, 'which', lambda x: True)  # type: ignore
 
-    def logfire_mcp_installed(_: list[str]) -> bytes:
-        return b'not installed'
+    captured_commands: list[list[str]] = []
 
-    monkeypatch.setattr(subprocess, 'check_output', logfire_mcp_installed)
+    def mock_check_output(cmd: list[str]) -> bytes:
+        captured_commands.append(cmd)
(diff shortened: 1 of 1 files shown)


mythral:
FYI: I updated this with tests for fully validating the schema format's against the docs from the various providers.

PR #1634, opened 2026-01-17
Make OTel endpoint configurable at runtime

len-foss:
Some customer environments do not have the ability to set any custom environment variables (it's bad, but environment setups are typically very hard to change). 
They rely on runtime configuration only.
Moreover, they have strong privacy requirements that go beyond simple GDPR. 
Therefore, just relying on logfire is a problem; the self-hosted solution is behind enterprise paywall, so it's not an easily available option, especially to test on a staging environment.

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -326,6 +326,7 @@ def configure(
     add_baggage_to_attributes: bool = True,
     code_source: CodeSource | None = None,
     distributed_tracing: bool | None = None,
+    otel_endpoint: str | None = None,
     advanced: AdvancedOptions | None = None,
     **deprecated_kwargs: Unpack[DeprecatedKwargs],
 ) -> Logfire:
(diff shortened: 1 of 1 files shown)


strawgate:
You should be able to use `os.environ` to change the environment vars visible to your process at runtime without actually changing the system or how the process is invoked.

```
os.environ["OTEL_EXPORTER_OTLP_ENDPOINT"] = "https://my-logs-are-on-fire.com:443"

logfire.configure(...)
```

As for making this configurable in code -- I imagine you wouldn't just want to expose an otlp_endpoint setting on the top-level configure call -- you'd want the entire exporter / otlp to be configurable with code as you may be proxying OTLP, only providing a traces endpoint, etc.

len-foss:
@strawgate Thank you for your reply. 

And for your patience, as I'm really not familiar with OTel stuff, but I figured it should be as simple as that from: 
https://logfire.pydantic.dev/docs/how-to-guides/alternative-backends/#example-with-jaeger

But I think this other link is much clearer:
https://ai.pydantic.dev/logfire/#alternative-observability-backends

It's possible to make a more complete example using the docker-compose at https://github.com/livingdocsIO/monitoring as basis, to have everything running easily and test all features. 
This was to go for a staging environment, so I was looking for a quick solution... I see why logfire is more convenient.

alexmojaki:
```python
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk.trace.export import BatchSpanProcessor
import logfire

exporter = OTLPSpanExporter(endpoint='http://localhost:4318/v1/traces')
span_processor = BatchSpanProcessor(exporter)
logfire.configure(additional_span_processors=[span_processor])
```

PR #1636, opened 2026-01-18
fix: anthropic response data should include tool call id

### LINKED ISSUE #1635: Anthropic response with tool calls is missing the tool call id
### Description

When instrumenting the Anthropic SDK and using tool calls, the tool call id is missing on the response data. The request data contains the correct tool call id (when sending tool call results back). As far as I can see, `block.id` would have to be included here: https://github.com/pydantic/logfire/blob/906a34fb5656fae55a58c88c89112ded107ef8fe/logfire/_internal/integrations/llm_providers/anthropic.py#L125

```TOML
4.19.0
```

mathislucka:
Fixes #1635


alexmojaki:
Thanks!

PR #1593, opened 2025-12-24
Improve OTel GenAI semantic convention compliance for LLM instrumentations

jimilp7:
## Summary

  This PR improves OTel GenAI Semantic Convention compliance across LLM instrumentations.

  ## What's New

  ### 1. OpenAI & Anthropic Instrumentation Improvements

  Enhanced existing instrumentations to better comply with [OTel GenAI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/):

  | Attribute | Before | After |
  |-----------|--------|-------|
  | SpanKind | `INTERNAL` | `CLIENT` |
  | SpanName | `"Chat Completion with {model!r}"` | `"chat gpt-4o-mini"` |
  | `gen_ai.system` | `"openai"` (deprecated) | Removed |
  | `gen_ai.provider.name` | Missing | `"openai"` / `"anthropic"` |
  | `server.address` | Missing | `"api.openai.com"` |
  | `gen_ai.request.*` | Missing | `max_tokens`, `temperature`, `top_p`, etc. |
  | `gen_ai.tool.definitions` | Missing | Tool schemas when provided |

  ### 2. LangChain/LangGraph Instrumentation (New)

  Adds `logfire.instrument_langchain()` to capture tool definitions from LangChain agents.

  **Why:** LangSmith's OTEL integration does NOT include tool definitions - they're sent via their proprietary API. This instrumentation patches LangChain's callback system to capture them.

  **Usage:**
  ```python
  import logfire
  from langgraph.prebuilt import create_react_agent

  logfire.configure()
  logfire.instrument_langchain()

  agent = create_react_agent(model, tools)
  agent.invoke({"messages": [...]})
```

  Captures:
  - gen_ai.tool.definitions - Tool schemas (name, description, parameters)
  - gen_ai.input.messages / gen_ai.output.messages - OTel format
  - gen_ai.system_instructions - Extracted from system messages
  - Token usage and model information

  Related

  - Addresses #1586 (OpenTelemetry Semantic Conventions Migration)
  - Builds on #1580 and #1590

PR #1650, opened 2026-01-23
feat: Add response api to gen_ai.input.messages and gen_ai.output.messages

brightsparc:


alexmojaki:
There are now big merge conflicts because I just merged https://github.com/pydantic/logfire/pull/1652, that was step 1 of https://github.com/pydantic/logfire/issues/1586

Similar to https://github.com/pydantic/logfire/pull/1593#issuecomment-3724741449, please move changes which only add small attributes to a separate PR, these can be easily merged into main without breaking anything or significantly increasing span size. The messages format is a whole thing on its own.

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -179,6 +209,230 @@ def get_endpoint_config(options: FinalRequestOptions) -> EndpointConfig:
...
+    """
+    input_messages: list[dict[str, Any]] = []
+    system_instructions: list[dict[str, Any]] = []
+
+    for msg in messages:
+        role = msg.get('role', 'unknown')
+        content = msg.get('content')
+
+        if role == 'system':
+            # System messages go to system_instructions
```
alexmojaki: No, system_instructions should only be the dedicated `instructions` param, which i think just doesn't exist for chat completions. See https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-system-instructions

> [9] gen_ai.system_instructions: This attribute SHOULD be used when the corresponding provider or API allows to provide system instructions or messages separately from the chat history.

> Instructions that are part of the chat history SHOULD be recorded in gen_ai.input.messages attribute instead.

brightsparc:
> There are now big merge conflicts because I just merged #1652, that was step 1 of #1586
> 
> Similar to [#1593 (comment)](https://github.com/pydantic/logfire/pull/1593#issuecomment-3724741449), please move changes which only add small attributes to a separate PR, these can be easily merged into main without breaking anything or significantly increasing span size. The messages format is a whole thing on its own.

Okay, will look into this thanks.

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -179,6 +209,230 @@ def get_endpoint_config(options: FinalRequestOptions) -> EndpointConfig:
...
+                    'id': tool_call_id,
+                    'response': content,
+                }
+            )
+
+        input_messages.append(
+            {
+                'role': role,
+                'parts': parts,
+                **({'name': msg.get('name')} if msg.get('name') else {}),
```
alexmojaki: what is this?

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -179,6 +209,230 @@ def get_endpoint_config(options: FinalRequestOptions) -> EndpointConfig:
...
+
+    return input_messages, system_instructions
+
+
+def _convert_content_part(part: dict[str, Any] | str) -> dict[str, Any]:
+    """Convert a single content part to semconv format."""
+    if isinstance(part, str):
+        return {'type': 'text', 'content': part}  # pragma: no cover
+
+    part_type = part.get('type', 'text')
```
alexmojaki: should the default type be `unknown`?

brightsparc: Updated this, and cleaned up to use TypedDict

logfire/_internal/integrations/llm_providers/openai.py
brightsparc: Clean this up to use Types similar to [PyDanticAI](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/_otel_messages.py)

logfire/_internal/integrations/llm_providers/openai.py
brightsparc: Good point, fixed. this.

tests/otel_integrations/test_openai.py
```diff
@@ -698,6 +839,450 @@ def test_sync_chat_completions_with_stop_string(instrumented_client: openai.Clie
...
+                                        'type': 'object',
+                                        'title': 'CompletionUsage',
+                                        'x-python-datatype': 'PydanticModel',
+                                    },
+                                },
+                            },
+                            'gen_ai.output.messages': {'type': 'array'},
+                            'gen_ai.response.finish_reasons': {'type': 'array'},
+                        },
+                    },
```
brightsparc, line 1225: @alexmojaki is there a reason why the json_schema doesn't include object type for things like string, integer, boolean, number?  We could update to include these types? (see below) 

```
def create_json_schema(obj: Any, seen: set[int]) -> JsonDict:
    """Create a JSON Schema from the given object.

    Args:
        obj: The object to create the JSON Schema from.
        seen: A set of object IDs that have already been processed.

    Returns:
        The JSON Schema.
    """
    if obj is None:
        return {'type': 'null'}

    try:
        # cover common types first before calling `type_to_schema` to avoid the overhead of imports if not necessary
        obj_type = obj.__class__

        # TODO: add these types?
        if obj_type is str:
            return {'type': 'string'}
        elif obj_type is int:
            return {'type': 'integer'}
        elif obj_type is bool:
            return {'type': 'boolean'}
        elif obj_type is float:
            return {'type': 'number'}

        if id(obj) in seen:
            return {}
```

alexmojaki:
@brightsparc these changes look great but are you still going to split the PR in two?

brightsparc:
> @brightsparc these changes look great but are you still going to split the PR in two?

This PR adds `input.messages` and `output.messages` for chat completions and responses.  I could probably split this for chat vs responses, but I don't think it would reduce the number of changes to split by attributes as they touch the same files?

I'm looking to make sure test coverage is 100% so this should mitigate most concerns?

logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -128,6 +277,24 @@ def on_response(response: ResponseT, span: LogfireSpan) -> ResponseT:
...
+        span.set_attribute(RESPONSE_ID, response.id)
+
+        # Add token usage
+        if response.usage:
+            span.set_attribute(INPUT_TOKENS, response.usage.input_tokens)
+            span.set_attribute(OUTPUT_TOKENS, response.usage.output_tokens)
+
+        # Add finish reason
+        if response.stop_reason:
+            span.set_attribute(RESPONSE_FINISH_REASONS, [response.stop_reason])
```
alexmojaki, line 292: This is the kind of thing I'm talking about when I say:

> Similar to https://github.com/pydantic/logfire/pull/1593#issuecomment-3724741449, please move changes which only add small attributes to a separate PR, these can be easily merged into main without breaking anything or significantly increasing span size. The messages format is a whole thing on its own.

brightsparc: Right, I will work on breaking this up into 3 PRs.
1. Scalar values
2. Input Messages
3. Output Messages

Once I have these I will link to this PR and it can be closed.

alexmojaki: 2 and 3 can stay together

brightsparc: The 3 PRS are

- #1657 
- #1658
- #1659

I suggest we land them in this order, as the input/output do overlap a bit and may need to rebase.

brightsparc: Right, I missed this message, will combine 3 into 2

brightsparc: Closing in favour of #1657 and #1666

PR #1656, opened 2026-01-23
Configure logfire locally for pytest

alexmojaki:


docs/integrations/pytest.md
```diff
@@ -115,21 +118,21 @@ When running tests with `--logfire`, any instrumented library calls create spans
...
         assert user['name'] == 'Test User'
 
-    with logfire.span('verify user'):
+    with logfire_instance.span('verify user'):
         # Verify the user was created correctly
         assert user['id'] == 123
 ```
 
+!!! tip "Use `logfire_instance` fixture"
+    When creating spans within tests, prefer using the injected `logfire_instance` fixture instead of the global `logfire` object for fine control of which spans are send to Logfire.
```
alexmojaki: I don't think this clearly explains what the fixture does. Also `logfire_instance` is a very generic name, can we do better?

docs/integrations/pytest.md
```diff
@@ -5,7 +5,10 @@ integration: logfire
 ---
 # Pytest
 
-Logfire includes a built-in pytest plugin that enables distributed tracing for your test suite. The plugin creates spans for test sessions and individual tests, allowing you to see exactly what happens during test execution.
+!!! tip "Running under Pytest... 🧪"
+    Only Pytest spans are send by default. Please refer to the [Testing Logfire Instrumentation](../reference/advanced/testing.md) guide for details on testing application instrumentation.
```
alexmojaki: I thought we were going to change the default when running with the plugin?

docs/integrations/pytest.md
```diff
@@ -5,7 +5,10 @@ integration: logfire
...
+
+Logfire includes a built-in Pytest plugin that enables distributed tracing for your test suite. The plugin creates spans for test sessions and individual tests, allowing you to see exactly what happens during test execution.
```
alexmojaki: This isn't related to distributed tracing

docs/integrations/pytest.md
```diff
@@ -142,17 +145,32 @@ pytest: my-project
...
-    logfire.info("Operation completed", result=result)
+    logfire_instance.info("Operation completed", result=result)
     assert result == expected
 ```
 
 All log messages appear as spans nested under the test span.
 
+## Application Code Spans
+
+You need to explicitly enable sending spans from your application code during tests, since the default behavior is to only send Pytest spans. To do this, configure Logfire in your application code with `send_to_logfire=True` or `send_to_logfire='if-token-present'`:
```
alexmojaki:
```suggestion
You need to explicitly enable sending spans from your application code during tests, since the default behavior when running with Pytest is to only send Pytest spans. To do this, configure Logfire in your application code with `send_to_logfire=True` or `send_to_logfire='if-token-present'`:
```

docs/reference/advanced/testing.md
```diff
@@ -1,11 +1,12 @@
...
-title: Testing with Logfire
+title: Testing Logfire Instrumentation
 description: Logfire makes it very easy to test emitted logs and spans using the utilities in the logfire.testing module.
 ---
-# Testing with Logfire
+# Testing Logfire Instrumentation
 
 !!! tip "Running under Pytest... 🧪"
     When running your test suite under Pytest, we set [`send_to_logfire=False`][logfire.configure(send_to_logfire)] by default.
+    Please refer to the [Pytest instrumentation](../../integrations/pytest.md) guide for details on instrumenting your tests.
```
alexmojaki: This doesn't really clarify the difference between the two pages.

alexmojaki:
Thanks!

PR #1664, opened 2026-01-27
test: use @customize to improve the snapshot generation

15r10nk:


15r10nk:
@alexmojaki here is another inline-snapshot update PR.

PR #1668, opened 2026-01-28
Fix Responses API streaming error handling to not mask original errors

jordanhboxer:
When streaming fails (e.g., API connection lost), the `response.completed` event is never received. Previously, `get_response_data()` would raise a `RuntimeError` in this case, which would mask the original streaming error. Now it returns `None` and `get_attributes()` handles this gracefully, allowing the original error to propagate.


alexmojaki:
Thanks!

PR #1676, opened 2026-01-30
fix: don't log @validate_call function exceptions as validation errors

### LINKED ISSUE #1516: Pydantic plugin records errors in validated functions as a validation error
### Description

When using logfire's pydantic auto instrumentation errors and the `@validate_call` decorator unexpected logging takes place.

Example:

```python
@validate_call
def my_test_function(value: int) -> None:
    raise RuntimeError

my_test_func(1)
```

Results in the log message of:
```
Validation on call raised RuntimeError
```

## Expectations
I would expect the above snippet to not log a validation error at all as the arguments did actually validate successfully. 

## Workaround
Talking in slack I was given this work around:

```python
@validate_call(config=ConfigDict(plugin_settings={'logfire': {'record': 'off'}}))
```

I believe this work around has a few failure modes:
1. A bit verbose to have to be setting manually in every instance we'd need to use it to avoid excessive logging
2. Very easy to forget to add in proactively
3. Would also disable logging of -actual- validation errors unless they were captured elsewhere.

## Assertion
I believe it's not intuitive for this to be captured as a validation error.   

Digging into the implementation I found similar blocks to this: 

```python
                span = self._on_enter(input_data)
                try:
                    result = validator(input_data, *args, **kwargs)
                except ValidationError as error:
                    self._on_error_span(span, error)
                    self._count_validation(success=False)
                    raise
                except Exception as exception:
                    self._on_exception_span(span, exception)
                    self._count_validation(success=False)
                    raise
                else:
                    self._on_success(span, result)
                    self._count_validation(success=True)
                    return result
```

When combined that the `validator()` call, in this case, implicitly calls the wrapped function it's understandable how we got here.

I'd suggest either in general or when the schema is "on call" the blanket exception logging is skipped. 

```TOML
logfire="4.14.2"
platform="macOS-26.0.1-arm64-arm-64bit"
python="3.11.8 (main, Feb 20 2024, 15:59:26) [Clang 15.0.0 (clang-1500.1.0.2.5)]"
[related_packages]
requests="2.32.5"
pydantic="2.12.3"
fastapi="0.120.0"
protobuf="5.29.5"
rich="14.2.0"
executing="2.2.1"
opentelemetry-api="1.38.0"
opentelemetry-exporter-otlp-proto-common="1.38.0"
opentelemetry-exporter-otlp-proto-http="1.38.0"
opentelemetry-instrumentation="0.59b0"
opentelemetry-instrumentation-httpx="0.59b0"
opentelemetry-instrumentation-system-metrics="0.59b0"
opentelemetry-proto="1.38.0"
opentelemetry-sdk="1.38.0"
opentelemetry-semantic-conventions="0.59b0"
opentelemetry-util-http="0.59b0"
```

wingding12:
## Summary

Fixes #1516

When using `@validate_call` with the pydantic plugin, exceptions raised inside the decorated function were incorrectly being logged as validation errors. This is misleading because the arguments validated successfully - the exception came from the function's execution, not from validation.

## Changes

This fix changes the behavior so that for `@validate_call`:
- `ValidationError` exceptions are still logged as validation failures  
- Non-`ValidationError` exceptions are treated as successful validation (since the arguments validated correctly before the function ran)

### Implementation

- Added `schema_kind` parameter to `_ValidateWrapper` to identify when we're dealing with `@validate_call`
- Modified exception handling in all three recording modes (`all`, `failure`, `metrics`) to check if `schema_kind == 'validate_call'`
- For `validate_call`, non-`ValidationError` exceptions now:
  - End the span with success status
  - Count as successful validation in metrics
  - Don't log exception details

## Limitations

Due to Pydantic's current design, we cannot distinguish between exceptions raised by validators and exceptions raised by the decorated function - both are raised without being wrapped in `ValidationError` for `@validate_call`.

This fix takes the conservative approach of not logging any non-`ValidationError` exceptions as validation errors for `@validate_call`. The proper fix would be in Pydantic itself, to separate argument validation from function execution (as noted by @Viicos in the issue).

## Test Plan

Added 4 new tests:
- `test_validate_call_exception_not_logged_as_validation_error` - Function exceptions are not logged as validation errors
- `test_validate_call_exception_record_failure_mode` - Function exceptions are not logged in failure mode
- `test_validate_call_validation_error_still_logged` - Actual validation errors are still logged correctly
- `test_validate_call_validator_exception_not_logged_as_validation_error` - Documents the limitation that validator exceptions are also not logged

All existing tests pass.

### DIFF
--- a/logfire/integrations/pydantic.py
+++ b/logfire/integrations/pydantic.py
@@ -75,6 +75,7 @@ class _ValidateWrapper:
         'schema_name',
         '_record',
         '_logfire',
+        '_schema_kind',
     )
 
     def __init__(
@@ -85,6 +86,7 @@ class _ValidateWrapper:
         _plugin_settings: PluginSettings | dict[str, Any],
         schema_type_path: SchemaTypePath,
         record: Literal['all', 'failure', 'metrics'],
+        schema_kind: SchemaKind,
     ) -> None:
         self.validation_method = validation_method
 
@@ -92,6 +94,7 @@ class _ValidateWrapper:
         # that are currently exposed by the plugin to potentially configure the validation handlers.
         self.schema_name = get_schema_name(schema)
         self._record = record
+        self._schema_kind = schema_kind
 
         self._logfire = logfire.DEFAULT_LOGFIRE_INSTANCE
         # trace_sample_rate = _plugin_settings.get('logfire', {}).get('trace_sample_rate')
@@ -132,6 +135,15 @@ class _ValidateWrapper:
                     self._count_validation(success=False)
                     raise
                 except Exception as exception:
+                    # For validate_call, non-ValidationError exceptions are likely from the decorated
+                    # function's execution, not from validation itself. We should not record these
+                    # as validation errors since the arguments validated successfully.
+                    # See https://github.com/pydantic/logfire/issues/1516
+                    if self._schema_kind == 'validate_call':
+                        # End the span successfully since validation passed
+                        self._on_success(span, None)
+                        self._count_validation(success=True)
+                        raise
                     self._on_exception_span(span, exception)
                     self._count_validation(success=False)
                     raise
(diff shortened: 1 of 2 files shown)


alexmojaki:
@wingding12 you've opened 3 PRs like this, I'm guessing using AI. If you want me to review them, please get CI passing and mark the PR as ready for review.

dmontagu:
Closing as stale — CI is still failing and this is still a draft six months after the review note. Happy to reopen if you pick it back up; #1516 stays open to track the underlying behaviour.

PR #1697, opened 2026-02-09
feat: add browser proxy helpers: `forward_export_request` and `logfire_proxy`

### LINKED ISSUE #1640: Make it easier to proxy from the browser to the logfire backend
Add some Python functions as in https://logfire.pydantic.dev/docs/integrations/javascript/browser/ which take care of forwarding OTel requests.

AlanPonnachan:
Closes #1640

### Description
This PR adds experimental helper functions to simplify setting up a backend proxy for browser monitoring. 

**Problem:**
To send traces/logs from a browser (frontend) to Logfire, the browser needs to authenticate. However, exposing the Logfire Write Token in frontend code is insecure. The recommended solution is to send data to the backend, which then forwards it to Logfire with the token injected.

**Solution (Updated):**
This PR introduces two new functions in `logfire.experimental.forwarding` to make this pattern trivial to implement while allowing maximum flexibility for production security:

1. `forward_export_request(...)`: A low-level function that takes request details (method, path, headers, body), strips sensitive client headers, injects the server's configured Logfire token, and securely forwards it to the Logfire API.
2. `logfire_proxy`: An asynchronous Starlette/FastAPI handler. By exposing this as a standalone handler rather than a baked-in method, developers can mount it to their own routes. This makes it easy to add FastAPI dependencies (like custom authentication, CORS, or rate-limiting) to protect the endpoint.

### Usage Example
**FastAPI**
```python
from fastapi import FastAPI
import logfire
from logfire.experimental.forwarding import logfire_proxy

app = FastAPI()
logfire.configure()

# Mount the proxy handler. The browser OTel SDK can now send traces to /logfire-proxy/v1/traces
# Note: You can easily add dependencies here for authentication: dependencies=[Depends(verify_session)]
app.add_route('/logfire-proxy/{path:path}', logfire_proxy, methods=['POST'])
```

### Implementation Details
- **Security:** `forward_export_request` strictly validates that the method is `POST` and the target path starts with `/v1/traces`, `/v1/logs`, or `/v1/metrics`. It also normalizes paths to prevent directory traversal attacks.
- **Headers:** Hop-by-hop headers (like `Host`, `Content-Length`) and sensitive headers (like client-side `Authorization` or `Cookie`) are stripped before forwarding.
- **DoS Mitigation:** The `logfire_proxy` handler enforces a configurable body size limit (default 50MB) checking both the `Content-Length` header early and the actual body read late.
- **Observability:** The outbound proxy request is wrapped in `logfire.suppress_instrumentation()` so the proxy doesn't generate infinite traces of its own exported telemetry.

### Changes
- **New Module:** Created `logfire/experimental/forwarding.py` to encapsulate the forwarding logic and ASGI handler, keeping the core `Logfire` class clean.
- **Tests:** Added `tests/test_experimental_forwarding.py` covering path traversal, limits, method restrictions, explicit configs.

### DIFF
--- a/docs/how-to-guides/mcp-server.md
+++ b/docs/how-to-guides/mcp-server.md
@@ -131,6 +131,26 @@ Create a `.zed/settings.json` file in your project's root directory:
 
 ---
 
+## Sandboxed Environments
+
+If browser-based authentication is not available (e.g. in sandboxed environments), generate an API key with at least the `project:read` scope from your organization or project settings, then use it as a Bearer token:
+
+```json
+{
+  "mcpServers": {
+    "logfire": {
+      "type": "http",
+      "url": "https://logfire-us.pydantic.dev/mcp",
+      "headers": {
+        "Authorization": "Bearer <your-logfire-api-key>"
+      }
+    }
+  }
+}
+```
+
+---
+
 ## Running Locally (Deprecated)
 
 !!! warning


AlanPonnachan:
@alexmojaki comments are resolved and all tests are passing.

logfire/_internal/integrations/forwarding.py
alexmojaki, line 1: please move this to a module `logfire.experimental.forwarding` and expose the `ForwardRequestResponse` type in `__all__` from there

logfire/_internal/main.py
```diff
@@ -297,6 +298,30 @@ def _instrument_span_with_args(
             log_internal_error()
             return NoopSpan()  # type: ignore
 
+    def forward_request(
```
alexmojaki: this name is not very clear in isolation. i think `forward_export_request` is better. similar for the type(s).

logfire/_internal/main.py
```diff
@@ -1112,6 +1137,56 @@ def instrument_pydantic_ai(
             **kwargs,
         )
 
+    def instrument_fastapi_proxy(
```
alexmojaki: can't have a name starting with `instrument_` that isn't about instrumenting something.

logfire/_internal/main.py
```diff
@@ -1112,6 +1137,56 @@ def instrument_pydantic_ai(
...
+    ) -> None:
+        """Instrument a FastAPI app to proxy requests to Logfire.
+
+        This adds a route to the app that forwards requests to the Logfire API.
+        This is useful for proxying requests from a browser to Logfire,
+        to avoid exposing your write token in the browser.
+
+        **Security Note**: This endpoint is unauthenticated. Any client capable of reaching this
+        endpoint can send telemetry data to your Logfire project. In production, ensure you
+        have appropriate protections in place (e.g. CORS policies, rate limiting, or upstream authentication).
```
alexmojaki: it's not clear how i would use this in production unless all the relevant protections exist as middleware. in our own app for example, authentication uses dependency injection, i.e. a parameter in endpoint functions. i don't see how we'd be able to use this.

i'm not sure what the best solution looks like. for starters it's worth noting that the generic forward_export_request is already enough to make its own PR and this could be figured out separately.

one step could be to expose the `logfire_proxy` function, which users could easily call in their own custom route that may have e.g. authentication dependency parameters.

it may also be possible to save users from having to write all of `app.add_route(f'{prefix}/{{path:path}}', logfire_proxy, methods=methods, include_in_schema=False)` with some fancy design, but that might be overengineering.

logfire/_internal/main.py
```diff
@@ -1112,6 +1137,56 @@ def instrument_pydantic_ai(
...
+
+        Args:
+            app: The FastAPI app to instrument.
```
alexmojaki: looks like this could also work for starlette

logfire/_internal/integrations/forwarding.py
```diff
@@ -0,0 +1,115 @@
...
+    if user_agent_key:
+        original_ua = new_headers.pop(user_agent_key)
+        new_headers['User-Agent'] = f'logfire-proxy/{VERSION} {original_ua}'
+    else:
+        new_headers['User-Agent'] = f'logfire-proxy/{VERSION}'
+
+    new_headers['Authorization'] = token
+
+    try:
+        response = requests.request(
```
alexmojaki: ideally this would use a session created in logfire.configure, which in particular would handle retrying. let that be a followup.

alexmojaki: wrap in `with logfire.suppress_instrumentation()` so that `logfire.instrument_requests()` doesn't log these.

logfire/_internal/integrations/forwarding.py
```diff
@@ -0,0 +1,115 @@
...
+            method=method,
+            url=url,
+            headers=new_headers,
+            data=body,
```
alexmojaki: possible followup: gzip compression

logfire/_internal/main.py
```diff
@@ -1112,6 +1137,56 @@ def instrument_pydantic_ai(
...
+
+            path = request.path_params['path']
+
+            # Performance: Run synchronous requests call in a thread pool to avoid blocking the event loop
+            response = await run_in_threadpool(
+                self.forward_request, method=request.method, path=path, headers=request.headers, body=body
+            )
+            return Response(content=response.content, status_code=response.status_code, headers=dict(response.headers))
+
+        methods = ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS', 'HEAD']
```
alexmojaki: i think only POST actually applies, why allow all methods?

AlanPonnachan: yes it should only POST

logfire/experimental/forwarding.py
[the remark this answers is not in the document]
AlanPonnachan: Updated

.github/workflows/main.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           enable-cache: true
 
       - run: uv sync --python 3.12 --frozen --group docs
-      - run: uv pip install --upgrade mkdocs-material mkdocstrings-python 'mkdocstrings<1' griffe==0.48.0
+      - run: uv pip install --upgrade mkdocs-material 'mkdocstrings-python<1.16' 'mkdocstrings<1' griffe==0.48.0
```
alexmojaki: why?

AlanPonnachan: I added that because the `ci/docs` workflow is currently failing with `ImportError: cannot import name 'DocstringSectionTypeAliases' from 'griffe'`. It looks like a recent release of `mkdocstrings-python` broke compatibility with the pinned `griffe==0.48.0`. 

May be I can revert back the changes and open a separate PR to fix the docs CI?

logfire/experimental/forwarding.py
```diff
@@ -0,0 +1,186 @@
...
+def forward_export_request(
+    method: str,
+    path: str,
+    headers: Mapping[str, str],
+    body: bytes | None,
+    *,
+    config: LogfireConfig | None = None,
+) -> ForwardExportRequestResponse:
+    """Forward an export request to the Logfire API."""
+    import requests
```
alexmojaki: move to top

logfire/experimental/forwarding.py
```diff
@@ -0,0 +1,186 @@
...
+
+    if method.upper() != 'POST':
```
alexmojaki: i don't think there's a point in having a `method` param, rather just always `requests.post`. the check in `logfire_proxy` is fine.

logfire/experimental/forwarding.py
```diff
@@ -0,0 +1,186 @@
...
+    content: bytes
+
+
+def forward_export_request(
+    method: str,
+    path: str,
+    headers: Mapping[str, str],
+    body: bytes | None,
+    *,
+    config: LogfireConfig | None = None,
```
alexmojaki: rather pass a logfire instance, no need to make the user write `.config`

logfire/experimental/forwarding.py
```diff
@@ -0,0 +1,186 @@
...
+    headers: Mapping[str, str]
+    content: bytes
+
+
+def forward_export_request(
+    method: str,
+    path: str,
+    headers: Mapping[str, str],
+    body: bytes | None,
+    *,
```
alexmojaki: let's make everything keyword only

logfire/experimental/forwarding.py
```diff
@@ -0,0 +1,186 @@
...
+
+    if not path.startswith('/'):
+        path = '/' + path
+
+    # Security: Normalize path to prevent directory traversal attacks (e.g. /v1/traces/../secret)
+    # We unquote first to handle percent-encoded traversals (e.g. %2e%2e) that might bypass normalization
+    path = posixpath.normpath(unquote(path))
+
+    allowed_prefixes = ('/v1/traces', '/v1/logs', '/v1/metrics')
+    if not any(path == prefix or path.startswith(prefix + '/') for prefix in allowed_prefixes):
```
alexmojaki: do we need to support `/v1/traces/foo`?

logfire/experimental/forwarding.py
```diff
@@ -0,0 +1,186 @@
...
+        return ForwardExportRequestResponse(
+            status_code=400,
+            headers={'Content-Type': 'text/plain'},
+            content=b'Invalid path: must start with /v1/traces, /v1/logs, or /v1/metrics',
+        )
+
+    token = config.token
+    if isinstance(token, list):
+        # Proxying only supports the first configured project/token
```
alexmojaki, line 59: this should be documented

logfire/experimental/forwarding.py
```diff
@@ -0,0 +1,186 @@
...
+    Returns:
+        A Starlette/FastAPI Response object.
+    """
+    from starlette.concurrency import run_in_threadpool
+    from starlette.responses import Response
+
+    if request.method.upper() != 'POST':
+        return Response(status_code=405, content='Method Not Allowed')
+
+    # DoS Prevention: Check Content-Length before reading body
```
alexmojaki: is this really protecting us from DoS? it allows content-length to be omitted entirely or to just be a lie. even if it required that content-length was present and accurate to accept the request, it would still have to read the whole body before it could check that.

.github/workflows/main.yml
alexmojaki: I think this was fixed in main

logfire/experimental/forwarding.py
AlanPonnachan: good catch!

.github/workflows/main.yml
AlanPonnachan: I reverted the change to latest main version.
With this change ci/docs is failing

logfire/experimental/forwarding.py
AlanPonnachan: Valid point!
I tried to make a solution with a two-layered defense to completely protect the server's memory:

-  kept the Content-Length check purely as a lightweight optimization to instantly reject overtly large, honest requests before reading a single byte.
-  Instead of using await request.body() , we will read the body in chunks. If a malicious client omits or lies about the Content-Length, we immediately abort the connection the millisecond the buffered data crosses max_body_size, preventing memory exhaustion.

.github/workflows/main.yml
alexmojaki: hmmm, might be a difference with external fork PRs. see https://mkdocstrings.github.io/python/changelog/#203-2026-02-20, try <2.0.3

logfire/experimental/forwarding.py
alexmojaki: looks great. please just update this comment since it's about honest requests.

logfire/experimental/forwarding.py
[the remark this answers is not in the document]
AlanPonnachan: This is intentional.

AlanPonnachan:
@alexmojaki

alexmojaki:
Thanks! Please also write some documentation.

PR #1700, opened 2026-02-10
Add version parameter to LLM instrumentation

dmontagu:
## Summary

This builds on #1666 by adding a `version: Literal[1, 'latest'] = 1` parameter to `instrument_openai()` and `instrument_anthropic()`. The core issue with #1666 as-is was that semconv message attributes were always emitted alongside the existing `request_data`/`response_data` attributes, leading to data duplication.

This PR resolves that by letting users opt in:

- **`version=1` (default)**: Preserves existing behavior — `request_data`, `response_data`, `events` attributes are emitted as before. Fully backwards compatible.
- **`version='latest'`**: Uses OTel Gen AI semantic convention attributes (`gen_ai.input.messages`, `gen_ai.output.messages`, `gen_ai.system_instructions`) instead. Old duplicative attributes are dropped.

In the next major version after this, we'll add `version=2` as an option, and that will be a snapshot of the "latest" logic at the time of that major version release, and we'll make that the new default. (And maybe deprecate version=1 at that time). And we'll keep the `version='latest'` option and make any further changes under that version.

## Approach

- Uses `functools.partial` to bind the `version` kwarg at the call site in `main.py`, so the generic `llm_provider.py` framework requires no changes at all.
- Adds `*Latest` stream state subclasses that override `get_attributes()` to emit semconv-format data.
- Keeps a minimal `request_data: {'model': model}` for streaming template compatibility.

## Changes

- `logfire/_internal/main.py` — Added `version` parameter to both methods, using `partial()` to bind it
- `logfire/_internal/integrations/llm_providers/openai.py` — Version-aware `get_endpoint_config` and `on_response`, plus `*Latest` stream state classes and semconv conversion helpers
- `logfire/_internal/integrations/llm_providers/anthropic.py` — Same pattern
- `logfire-api/logfire_api/_internal/main.pyi` — Updated stubs
- Tests — Added focused `version='latest'` tests for OpenAI (6 tests) and Anthropic (3 tests)

## Test plan

- [x] All existing tests pass with default `version=1` (no behavior change)
- [x] New `version='latest'` tests verify semconv attributes are present and old attributes are absent
- [x] Pre-commit hooks pass (ruff, pyright)

### DIFF
--- a/logfire/_internal/integrations/llm_providers/anthropic.py
+++ b/logfire/_internal/integrations/llm_providers/anthropic.py
@@ -141,7 +141,7 @@ def convert_anthropic_messages_to_semconv(
         if content is not None:
             if isinstance(content, str):
                 parts.append({'type': 'text', 'content': content})
-            elif isinstance(content, list):
+            elif isinstance(content, list):  # pragma: no branch
                 for part in cast('list[dict[str, Any] | str]', content):
                     parts.append(_convert_anthropic_content_part(part))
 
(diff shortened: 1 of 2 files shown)


dmontagu:
Superseded by a new PR from branch dm/semconv-version-param-v2, rebased on top of PR #1666.

PR #1703, opened 2026-02-10
Add rules extracted from past PR reviews to AGENTS.md

DouweM:
@alexmojaki I extracted these rules from around 500 PR comments you made since 2025-01-01 using a `braindump` CLI I built that I'll be open sourcing soon, which I also used for  https://github.com/pydantic/pydantic-ai/pull/4214. Having all those things there that previously lived in my head has made AI codegen and review a lot more effective.

It also has the ability to get the provenance for each of these generated rules, to understand which PRs/comments it came from. Happy to show you tomorrow.

Very curious to hear if you agree with all of these or if `braindump` misinterpreted things!

(Note that I renamed CLAUDE.md to AGENTS.md and made CLAUDE.md a symlink, as is apparently the custom these days, so the top of the file shows up in the diff unnecessarily)


alexmojaki:
thanks. yes i'd like to see the provenance. also the docs build is failing, probably because it expects `docs/AGENTS.md` to be in `mkdocs.yml`, which of course it shouldn't.

docs/AGENTS.md
```diff
@@ -0,0 +1,22 @@
+<!-- braindump: rules extracted from PR review patterns -->
```
alexmojaki, line 1: Is this helpful?

DouweM:
Now that https://github.com/pydantic/braindump is public, I suggest you run it yourself! You can trace the provenance of each rule using `braindump lookup`.

PR #1704, opened 2026-02-11
Fix reST double backticks in db_api docstrings

dmontagu:
## Summary
- Replace reStructuredText-style double backticks (` `` `) with Markdown single backticks in `logfire/db_api.py` docstrings

Spotted by @Viicos in https://github.com/pydantic/logfire/pull/1692.


logfire/db_api.py
```diff
@@ -29,7 +29,7 @@
     from logfire.experimental.query_client import ColumnDetails
```
Viicos, line 30:
```suggestion

# TODO: make use of PEP 661 sentinels once accepted.
```

PR #1706, opened 2026-02-11
Add version-specific tests for LLM semconv attributes

sheldor1510:
## Summary
- Adds 10 new tests verifying that `version='latest'` and `version=1` produce distinct span attributes across OpenAI and Anthropic LLM integrations
- Covers streaming chat completions, non-streaming completions, streaming completions, and embeddings for OpenAI
- Covers streaming messages for Anthropic
- Each test pair verifies that `version='latest'` emits `gen_ai.input.messages`/`gen_ai.output.messages` without `response_data`, while `version=1` emits `request_data`/`response_data` without semconv message attributes

## Test plan
- [x] All 10 new tests pass
- [x] All existing tests continue to pass (48/48 non-VCR tests)
- [x] Linting passes (`make lint`)
- [x] Type checking passes (`make typecheck`)

### DIFF
--- a/tests/otel_integrations/test_openai.py
+++ b/tests/otel_integrations/test_openai.py
@@ -3072,10 +3072,10 @@ def test_dont_suppress_httpx(exporter: TestExporter) -> None:
                                         'http.status_code': 200,
                                         'net.peer.name': 'api.openai.com',
                                     },
-                                    'total': IsNumeric,
+                                    'total': IsNumeric(),
                                 }
                             ],
-                            'total': IsNumeric,
+                            'total': IsNumeric(),
                         },
                         'http.client.request.duration': {
                             'details': [


tests/otel_integrations/test_anthropic.py
```diff
@@ -1137,31 +1137,31 @@ def test_sync_messages_version_latest(exporter: TestExporter) -> None:
...
+                            'parts': [{'type': 'text', 'content': 'Four plus five equals nine.'}],
                             'finish_reason': 'end_turn',
                         }
                     ],
-                    'gen_ai.response.model': 'claude-3-haiku-20240307',
-                    'gen_ai.response.id': 'test_id',
-                    'gen_ai.usage.input_tokens': 2,
-                    'gen_ai.usage.output_tokens': 3,
+                    'gen_ai.response.model': 'claude-sonnet-4-20250514',
+                    'gen_ai.response.id': 'msg_01VnWgU8ykFH7dMLRVcigZtG',
```
brightsparc: Please use IsStr() for variables that will change.

tests/otel_integrations/test_openai.py
```diff
@@ -2027,27 +2027,31 @@ def test_sync_chat_completions_version_latest(exporter: TestExporter) -> None:
...
                     'logfire.span_type': 'span',
                     'gen_ai.system': 'openai',
-                    'gen_ai.response.model': 'gpt-4',
-                    'operation.cost': 0.00012,
-                    'gen_ai.response.id': 'test_id',
-                    'gen_ai.usage.input_tokens': 2,
-                    'gen_ai.usage.output_tokens': 1,
+                    'gen_ai.response.model': 'gpt-4.1-2025-04-14',
+                    'operation.cost': 9.4e-05,
+                    'gen_ai.response.id': 'chatcmpl-D8Djn5gMxH9y2Kqd7fsCM9KGlQSqs',
```
brightsparc: Please use `IsStr()`

brightsparc (commented):
Try running the recording a couple of times so that you use IsStr() / IsNum() to avoid dirty variables changing each execution.

tests/otel_integrations/test_anthropic.py
```diff
@@ -1137,31 +1137,31 @@ def test_sync_messages_version_latest(exporter: TestExporter) -> None:
...
                     'logfire.msg_template': 'Message with {request_data[model]!r}',
-                    'logfire.msg': "Message with 'claude-3-haiku-20240307'",
+                    'logfire.msg': "Message with 'claude-sonnet-4-20250514'",
                     'logfire.tags': ('LLM',),
                     'logfire.span_type': 'span',
                     'gen_ai.output.messages': [
                         {
                             'role': 'assistant',
-                            'parts': [{'type': 'text', 'content': 'Nine'}],
+                            'parts': [{'type': 'text', 'content': IsStr()}],
```
brightsparc: This doesn't feel like an instance where we should be expecting the value to change?

dmontagu:
I suggest removing the `# pragma: no branch`es from the non-test code related to this since I think they are probably no longer necessary in most/all cases, if any still are you can always re-add them where appropriate.

dmontagu:
I see the lack of pragmas means we drop a bunch of coverage, that seems to indicate to me that the tests aren't hitting those code paths in just-version-1 and just-version-'latest' modes, or the branch wouldn't be necessary. (Unless the failing ones are failing other ways). I think it's fine to re-add the pragmas if you think these tests are useful and just merge this, you don't need to feel obligated to achieve coverage of all those branches, though I do see some value in it if you wanted to cover those.

sheldor1510:
> I see the lack of pragmas means we drop a bunch of coverage, that seems to indicate to me that the tests aren't hitting those code paths in just-version-1 and just-version-'latest' modes, or the branch wouldn't be necessary. (Unless the failing ones are failing other ways). I think it's fine to re-add the pragmas if you think these tests are useful and just merge this, you don't need to feel obligated to achieve coverage of all those branches, though I do see some value in it if you wanted to cover those.

@dmontagu I added more tests to achieve coverage. Let me know if this looks good and if it follows best practice.

alexmojaki:
Thanks!

PR #1711, opened 2026-02-14
Add experimental datasets package

dmontagu:
## Summary

- Add `tags: list[str] | None = None` parameter to all 12 SDK client methods (6 sync + 6 async): `create_case`, `update_case`, `list_cases`, `add_case`, `add_cases`, and `import_cases`
- Restructure datasets documentation from a single `datasets.md` page into 5 subpages: overview, web UI guide, SDK guide, running evaluations, and SDK reference
- Add experimental feature flag admonition to each docs page
- Add collapsible SDK equivalent examples in the web UI docs

## Test plan

- [ ] Verify SDK import: `python -c "from logfire.experimental.datasets import LogfireDatasetsClient"`
- [ ] Verify docs build: `mkdocs build` with no warnings
- [ ] Test tags parameter works end-to-end against a running backend

### DIFF
--- a/tests/test_datasets_client.py
+++ b/tests/test_datasets_client.py
@@ -18,6 +18,7 @@ try:
 except (ImportError, AttributeError):
     pytest.skip('pydantic_evals not compatible with this environment', allow_module_level=True)
 
+import logfire.experimental.datasets as datasets_module
 from logfire.experimental.api_client import (
     AsyncLogfireAPIClient,
     CaseNotFoundError,
@@ -232,6 +233,12 @@ class TestSerializeCase:
         result = _serialize_case(case)
         assert result == {'inputs': {'question': 'hi'}, 'evaluators': []}
 
+    def test_none_evaluators(self):
+        case: Case[MyInput, MyOutput, Any] = Case(inputs=MyInput(question='hi'))
+        case.evaluators = None  # type: ignore[assignment]
+        result = _serialize_case(case)
+        assert result == {'inputs': {'question': 'hi'}}
+
     def test_full_case(self):
         @dataclass
         class MyEval:
@@ -287,6 +294,16 @@ class TestImportPydanticEvals:
         with patch.dict('sys.modules', {'pydantic_evals': None}):
             with pytest.raises(ImportError, match='pydantic-evals is required'):
                 _import_pydantic_evals()
+
+
+class TestDatasetsModuleReexports:
+    def test_reexports(self):
+        """Verify that logfire.experimental.datasets re-exports from api_client."""
+        assert datasets_module.LogfireAPIClient is LogfireAPIClient
+        assert datasets_module.AsyncLogfireAPIClient is AsyncLogfireAPIClient
+        assert datasets_module.DatasetNotFoundError is DatasetNotFoundError
+        assert datasets_module.CaseNotFoundError is CaseNotFoundError
+        assert datasets_module.DatasetApiError is DatasetApiError
 
 
 class TestValidateDatasetName:


logfire/experimental/api_client.py
```diff
@@ -0,0 +1,1075 @@
...
+        self.timeout = timeout
+        headers = client_kwargs.pop('headers', {})
+        headers['authorization'] = f'Bearer {api_key}'
+        self.client: T = client(timeout=timeout, base_url=base_url, headers=headers, **client_kwargs)
+
+    def _handle_response(self, response: Response) -> Any:
+        """Handle API response, raising appropriate errors."""
+        if response.status_code == 404:
+            detail = response.json() if response.content else 'Not found'
+            raise DatasetNotFoundError(detail)
```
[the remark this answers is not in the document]
dmontagu: Fixed — `_handle_response` now accepts `is_case_endpoint` parameter. Case-specific endpoints (get_case, update_case, delete_case) pass `is_case_endpoint=True`, and the handler checks the response text for "case" to raise `CaseNotFoundError` instead of `DatasetNotFoundError`.

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1075 @@
...
+        elif hasattr(evaluator, '__dataclass_fields__'):
+            # Dataclass - get field values
+            from dataclasses import asdict
+
+            arguments = asdict(evaluator)
+        else:
+            arguments = None
+
+        # Use None if no arguments, otherwise use the dict
+        if arguments and len(arguments) == 0:
```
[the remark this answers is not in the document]
dmontagu: Fixed — changed to `if arguments is not None and len(arguments) == 0:`.

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1075 @@
...
+            )
+        if metadata is not None:
+            data['metadata'] = _serialize_value(metadata) if not isinstance(metadata, dict) else metadata
+        if evaluators is not None:
+            data['evaluators'] = _serialize_evaluators(evaluators)
+        if tags is not None:
+            data['tags'] = tags
+
+        response = self.client.put(f'/v1/datasets/{dataset_id_or_name}/cases/{case_id}/', json=data)
+        return self._handle_response(response)
```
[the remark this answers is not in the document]
dmontagu: Fixed — added an `_UNSET` sentinel as default for nullable fields in `update_dataset` (description, guidance) and `update_case` (name, expected_output, metadata, evaluators, tags). Users can now pass `None` explicitly to clear these fields. The backend has also been updated to support this via `model_fields_set`.

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,1075 @@
...
+        Args:
+            id_or_name: The dataset ID (UUID) or name.
+
+        Returns:
+            Dataset details including schemas and metadata.
+
+        Raises:
+            DatasetNotFoundError: If the dataset does not exist.
+        """
+        response = self.client.get(f'/v1/datasets/{id_or_name}/')
```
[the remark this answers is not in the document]
dmontagu: Fixed — all URL path segments from user input (dataset names, case IDs) are now URL-encoded via `urllib.parse.quote(value, safe="")` to prevent path traversal with special characters.

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1100 @@
...
+        if cases and hasattr(cases[0], 'inputs'):
+            # These are Case objects
+            serialized_cases = [_serialize_case(case) for case in cases]  # type: ignore[arg-type]
+        else:
+            # Already dicts
+            serialized_cases: list[dict[str, Any]] = list(cases)  # type: ignore[arg-type]
+
+        if tags is not None:
+            for case_data in serialized_cases:
+                case_data['tags'] = tags
```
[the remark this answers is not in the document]
dmontagu: Fixed — changed `list(cases)` to `[dict(c) for c in cases]` to shallow-copy dicts before adding tags. Added a test (`test_import_cases_dicts_not_mutated`) verifying the caller's original dicts are not mutated.

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,1100 @@
...
+        headers['authorization'] = f'Bearer {api_key}'
+        self.client: T = client(timeout=timeout, base_url=base_url, headers=headers, **client_kwargs)
+
+    def _handle_response(self, response: Response, *, is_case_endpoint: bool = False) -> Any:
+        """Handle API response, raising appropriate errors."""
+        if response.status_code == 404:
+            detail = response.json() if response.content else 'Not found'
+            if is_case_endpoint and 'case' in response.text.lower():
+                raise CaseNotFoundError(detail)
+            raise DatasetNotFoundError(detail)
```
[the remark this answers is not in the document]
dmontagu: Acknowledged — the `is_case_endpoint` heuristic is intentionally simple since the backend consistently includes "case" in case-specific 404 messages. If we ever need more robustness, we can switch to a structured error code in the JSON body, but this works well for now.

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,1101 @@
...
+            data['output_schema'] = _type_to_schema(output_type)
+        if metadata_type is not None:
+            data['metadata_schema'] = _type_to_schema(metadata_type)
+        if guidance is not _UNSET:
+            data['guidance'] = guidance
+        if ai_managed_guidance is not None:
+            data['ai_managed_guidance'] = ai_managed_guidance
+
+        response = self.client.put(f'/v1/datasets/{_quote(id_or_name)}/', json=data)
+        return self._handle_response(response)
```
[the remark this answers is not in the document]
dmontagu: Good observation. This is intentional — schemas are derived from Python types, so there's no natural way to express "clear this schema" via the type-based API. Users who need to remove a schema can use the underlying httpx client directly (`client.client.put(...)`). Not worth adding complexity for this edge case right now.

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,1101 @@
...
+    def _handle_response(self, response: Response, *, is_case_endpoint: bool = False) -> Any:
+        """Handle API response, raising appropriate errors."""
+        if response.status_code == 404:
+            detail = response.json() if response.content else 'Not found'
+            if is_case_endpoint and 'case' in response.text.lower():
+                raise CaseNotFoundError(detail)
+            raise DatasetNotFoundError(detail)
+        if response.status_code >= 400:
+            detail = response.json() if response.content else response.text
+            raise DatasetApiError(response.status_code, detail)
```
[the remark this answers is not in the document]
dmontagu: Fixed — `_handle_response` now wraps `response.json()` in a try/except, falling back to `response.text` for non-JSON error bodies (e.g. HTML from load balancers). Also extracted the JSON parsing into a `_parse_response_detail` helper.

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1101 @@
...
+        self.timeout = timeout
+        headers = client_kwargs.pop('headers', {})
+        headers['authorization'] = f'Bearer {api_key}'
+        self.client: T = client(timeout=timeout, base_url=base_url, headers=headers, **client_kwargs)
+
+    def _handle_response(self, response: Response, *, is_case_endpoint: bool = False) -> Any:
+        """Handle API response, raising appropriate errors."""
+        if response.status_code == 404:
+            detail = response.json() if response.content else 'Not found'
+            if is_case_endpoint and 'case' in response.text.lower():
```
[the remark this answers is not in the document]
dmontagu: Fixed — the 404 classification now checks the parsed JSON dict's `detail` field (`detail.get('detail', '')`) instead of raw `response.text`, so dataset names containing 'case' won't cause misclassification.

tests/test_datasets_client.py
```diff
@@ -0,0 +1,1304 @@
...
+    return client
+
+
+def make_async_client(
+    responses: dict[tuple[str, str], httpx.Response | None] | None = None,
+) -> AsyncLogfireDatasetsClient:
+    transport = make_mock_transport(responses)
+    client = AsyncLogfireDatasetsClient(api_key='test-key', base_url='https://test.logfire.dev')
+    client.client = httpx.AsyncClient(transport=transport, base_url='https://test.logfire.dev')
+    return client
```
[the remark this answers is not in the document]
dmontagu: should refactor _BaseLogfireDatasetsClient to accept a client rather than client type and kwargs for better type-safety on the kwargs

docs/guides/web-ui/datasets/sdk.md
```diff
@@ -0,0 +1,222 @@
+---
+title: "SDK Guide"
+description: "Manage evaluation datasets programmatically with the Logfire Python SDK."
+---
+
+# SDK Guide
+
+!!! warning "Experimental Feature"
+
+    Managed datasets is an experimental feature currently gated behind a feature flag. Reach out to us on [Slack](https://logfire.pydantic.dev/docs/join-slack/) or [contact us](../../../help.md) to learn how to enable it for your project.
```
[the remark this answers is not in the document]
dmontagu: "Managed datasets" is the name of the feature, and it is one feature. This is a classic _human_ mistake in grammar, bad copilot.

docs/guides/web-ui/datasets/evaluations.md
```diff
@@ -0,0 +1,118 @@
+---
+title: "Running Evaluations"
+description: "Export managed datasets and run evaluations with pydantic-evals."
+---
+
+# Running Evaluations
+
+!!! warning "Experimental Feature"
+
+    Managed datasets is an experimental feature currently gated behind a feature flag. Reach out to us on [Slack](https://logfire.pydantic.dev/docs/join-slack/) or [contact us](../../../help.md) to learn how to enable it for your project.
```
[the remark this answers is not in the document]
dmontagu: wrong again

docs/guides/web-ui/datasets/sdk-reference.md
```diff
@@ -0,0 +1,41 @@
+---
+title: "SDK Reference"
+description: "Complete method and exception reference for the Logfire datasets SDK."
+---
+
+# SDK Reference
+
+!!! warning "Experimental Feature"
+
+    Managed datasets is an experimental feature currently gated behind a feature flag. Reach out to us on [Slack](https://logfire.pydantic.dev/docs/join-slack/) or [contact us](../../../help.md) to learn how to enable it for your project.
```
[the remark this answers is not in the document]
dmontagu: wrong again

logfire/experimental/datasets/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+        from pydantic_evals import Dataset
+
+        dataset: Dataset[MyInput, MyOutput, None] = client.export_dataset(
+            'my-evaluation-dataset',
+            input_type=MyInput,
+            output_type=MyOutput,
+        )
+
+        # Run evaluations
+        report = await dataset.evaluate(my_task)
```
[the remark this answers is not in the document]
dmontagu: yep

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1117 @@
...
+            # Export with types and custom evaluators
+            dataset: Dataset[MyInput, MyOutput, None] = client.export_dataset(
+                'my-dataset',
+                input_type=MyInput,
+                output_type=MyOutput,
+                custom_evaluator_types=[MyCustomEvaluator],
+            )
+
+            # Use with evaluations
+            report = await dataset.evaluate(my_task)
```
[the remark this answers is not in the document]
dmontagu: can drop the asyncio.run(main()) but otherwise, yes

docs/guides/web-ui/datasets/index.md
```diff
@@ -0,0 +1,42 @@
+---
+title: "Managed Datasets: Build and Manage Evaluation Test Cases"
+description: "Create, manage, and export typed evaluation datasets in Pydantic Logfire. Integrate with pydantic-evals to run evaluations against your AI systems."
+---
+
+# Managed Datasets
+
+!!! warning "Experimental Feature"
+
+    Managed datasets is an experimental feature currently gated behind a feature flag. Reach out to us on [Slack](https://logfire.pydantic.dev/docs/join-slack/) or [contact us](../../../help.md) to learn how to enable it for your project.
```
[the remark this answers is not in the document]
dmontagu: wrong

docs/guides/web-ui/datasets/ui.md
```diff
@@ -0,0 +1,168 @@
+---
+title: "Web UI Guide"
+description: "Create and manage evaluation datasets through the Logfire web interface."
+---
+
+# Web UI Guide
+
+!!! warning "Experimental Feature"
+
+    Managed datasets is an experimental feature currently gated behind a feature flag. Reach out to us on [Slack](https://logfire.pydantic.dev/docs/join-slack/) or [contact us](../../../help.md) to learn how to enable it for your project.
```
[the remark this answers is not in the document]
dmontagu: wrong

pyproject.toml
```diff
@@ -83,6 +83,7 @@ aws-lambda = ["opentelemetry-instrumentation-aws-lambda >= 0.42b0"]
 google-genai = ["opentelemetry-instrumentation-google-genai >= 0.4b0"]
 litellm = ["openinference-instrumentation-litellm >= 0"]
 dspy = ["openinference-instrumentation-dspy >= 0"]
+datasets = ["httpx", "pydantic-evals>=1.0.0; python_version >= '3.10'"]
```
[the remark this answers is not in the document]
dmontagu: yep, not sure what the min version should be though

tests/test_datasets_client.py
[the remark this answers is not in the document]
dmontagu: Fixed — refactored `_BaseLogfireDatasetsClient` to accept a pre-configured client instance. `make_client()` and `make_async_client()` now pass `client=httpx.Client(transport=transport, ...)` directly instead of creating then overwriting.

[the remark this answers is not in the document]
dmontagu: Fixed — `_BaseLogfireDatasetsClient` now accepts a `client: T` instance directly. Both sync and async constructors accept an optional `client=` keyword argument, eliminating the need for `**client_kwargs: Any`. Test helpers now pass the mock-transport client directly.

logfire/experimental/datasets/__init__.py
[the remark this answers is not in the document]
dmontagu: Fixed — wrapped the `await` call in an `async def run_eval()` function.

logfire/experimental/datasets/client.py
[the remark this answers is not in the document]
dmontagu: Fixed — wrapped the `await` call in an `async def run_eval()` function (without `asyncio.run()`).

pyproject.toml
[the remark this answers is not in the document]
dmontagu: Fixed — added `httpx>=0.27.2` lower bound to the datasets extra.

dmontagu:
Probably should have a method like `logfire.datasets_client()` to get the `LogfireDatasetsClient`.

Maybe it should be `logfire.api_client()` or `logfire.client().datasets` to be more general...

@alexmojaki we should discuss on Monday

mkdocs.yml
```diff
@@ -106,6 +106,12 @@ nav:
               - LLM Panels: guides/web-ui/llm-panels.md
               - Dashboards: guides/web-ui/dashboards.md
               - Evals: guides/web-ui/evals.md
+              - Datasets:
+                  - guides/web-ui/datasets/index.md
```
alexmojaki:
```suggestion
                  - Datasets: guides/web-ui/datasets/index.md
```

mkdocs.yml
```diff
@@ -393,6 +399,7 @@ plugins:
         "why-logfire/python-centric.md": "why.md"
         "why-logfire/sql.md": "why.md"
         "get-started/traces.md": "concepts.md"
+        "guides/web-ui/datasets.md": "guides/web-ui/datasets/index.md"
```
alexmojaki:
```suggestion
```

pyproject.toml
```diff
@@ -83,6 +83,7 @@ aws-lambda = ["opentelemetry-instrumentation-aws-lambda >= 0.42b0"]
 google-genai = ["opentelemetry-instrumentation-google-genai >= 0.4b0"]
 litellm = ["openinference-instrumentation-litellm >= 0"]
 dspy = ["openinference-instrumentation-dspy >= 0"]
+datasets = ["httpx>=0.27.2", "pydantic-evals>=1.0.0; python_version >= '3.10'"]
```
alexmojaki: This says 'install pydantic-evals for python >= 3.10', not 'datasets requires python >= 3.10' which seems to be the intention based on the docs

docs/guides/web-ui/datasets/sdk.md
```diff
@@ -0,0 +1,222 @@
...
+
+```python skip="true" skip-reason="external-connection"
+with LogfireDatasetsClient(api_key='your-api-key') as client:
+    ...
+```
+
+!!! note "API key scopes"
+    The API key must have the `project:read_datasets` scope to read datasets, and `project:write_datasets` to create, update, or delete datasets and cases. You can create API keys with these scopes under **Settings > API Keys** in the Logfire UI.
+
+The `base_url` is automatically inferred from the API key. You can override it if needed (e.g., for local development):
```
alexmojaki:
```suggestion
The `base_url` is automatically inferred from the API key. You can override it if needed (e.g. for self-hosters):
```

alexmojaki:
> Probably should have a method like `logfire.datasets_client()` to get the `LogfireDatasetsClient`.
> 
> Maybe it should be `logfire.api_client()` or `logfire.client().datasets` to be more general...
> 
> @alexmojaki we should discuss on Monday

Seems like this can always be added later, the way it currently works is nicely tucked away in `experimental`.

docs/guides/web-ui/datasets/ui.md
```diff
@@ -0,0 +1,168 @@
...
+
+This guide covers creating and managing datasets through the Logfire web interface. For programmatic access, see the [SDK Guide](sdk.md).
+
+## Navigate to Datasets
+
+Open the **Datasets** page from the left-hand navigation in your project. This shows all datasets in the project along with their case counts and descriptions.
```
alexmojaki: "left-hand navigation" isn't the default yet

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1138 @@
...
+InputsT = TypeVar('InputsT')
+OutputT = TypeVar('OutputT')
+MetadataT = TypeVar('MetadataT')
+
+_UNSET: Any = object()
+"""Sentinel to distinguish 'not provided' from explicit None."""
+
+
+def _quote(value: str) -> str:
+    """URL-encode a path segment to prevent path traversal with special characters."""
```
alexmojaki: this seems bad, like we should either be using a URL param or restricting characters allowed in dataset names

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1138 @@
...
+            data['ai_managed_guidance'] = ai_managed_guidance
+
+        response = self.client.post('/v1/datasets/', json=data)
+        return self._handle_response(response)
+
+    def update_dataset(
+        self,
+        id_or_name: str,
+        *,
+        name: str | None = _UNSET,
```
alexmojaki: shouldn't be nullable

tests/test_datasets_client.py
```diff
@@ -0,0 +1,1311 @@
...
+        {
+            'name': 'test-case',
+            'inputs': {'question': 'What is 2+2?'},
+            'expected_output': {'answer': '4'},
+        }
+    ],
+}
+
+
+def make_mock_transport(responses: dict[tuple[str, str], httpx.Response | None] | None = None) -> httpx.MockTransport:
```
alexmojaki, line 86: prefer vcr if possible

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,1138 @@
...
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from types import TracebackType
+from typing import TYPE_CHECKING, Any, Generic, TypeVar, cast, overload
+from urllib.parse import quote as _url_quote
+
+from pydantic import TypeAdapter
```
alexmojaki, line 56: need to add pydantic to the extra

docs/guides/web-ui/datasets/sdk.md
```diff
@@ -0,0 +1,222 @@
...
+            expected_output=AnswerOutput(answer='Au', confidence=0.98),
+            metadata=CaseMetadata(category='science', difficulty='medium'),
+        ),
+    ],
+    tags=['batch-import'],
+)
+```
+
+!!! tip "Referencing datasets by name or ID"
+    All dataset operations accept either the dataset's UUID or its name. Using the name (e.g., `'qa-golden-set'`) is more readable, while the UUID is guaranteed to be unique across renames.
```
alexmojaki: Neither dataset nor case IDs are visible in the UI

docs/evaluate/datasets/ui.md
```diff
@@ -0,0 +1,168 @@
...
+    See [Adding Cases](sdk.md#adding-cases) in the SDK Guide for more options.
+
+## Adding Cases from Traces
+
+You can create test cases directly from production data:
+
+1. Open **Live View** and find a trace or span that represents a good test case.
+2. Click the **Add to Dataset** button on the span.
+3. Select the target dataset and review the extracted inputs and outputs.
+4. Optionally edit the values before saving.
```
alexmojaki, line 102: The schema isn't validated before the case is saved

alexmojaki: When I try editing the case in the datasets UI and save I get this:

Image

which is fine, but the error should be in red, and the `/:` is not clear to a user

docs/evaluate/datasets/sdk-reference.md
```diff
@@ -0,0 +1,41 @@
...
+| `create_dataset(name, *, input_type, output_type, metadata_type, description, guidance)` | Create a new dataset. Types are converted to JSON schemas automatically. |
+| `update_dataset(id_or_name, *, name, input_type, output_type, metadata_type, description)` | Update a dataset's metadata or schemas. |
+| `delete_dataset(id_or_name)` | Delete a dataset and all its cases. |
+| `list_cases(dataset_id_or_name, *, tags)` | List all cases in a dataset. Optionally filter by tags. |
+| `get_case(dataset_id_or_name, case_id)` | Get a specific case. |
+| `add_case(dataset_id_or_name, case, *, tags)` | Add a single `pydantic_evals.Case` to a dataset. Optionally assign tags. |
+| `add_cases(dataset_id_or_name, cases, *, tags)` | Add multiple `pydantic_evals.Case` objects in bulk. Optionally assign tags to all cases. |
+| `create_case(dataset_id_or_name, inputs, *, name, expected_output, metadata, evaluators, source_trace_id, source_span_id, tags)` | Create a case from raw values. |
+| `update_case(dataset_id_or_name, case_id, *, name, inputs, expected_output, metadata, evaluators, tags)` | Update an existing case. |
+| `delete_case(dataset_id_or_name, case_id)` | Delete a case. |
```
alexmojaki, line 27: Just making threads here for general review, not related really to this line. When I delete a case in the UI, it doesn't immediately disappear from the list.

docs/evaluate/datasets/sdk-reference.md
```diff
@@ -0,0 +1,41 @@
...
+| `export_dataset(id_or_name, input_type, output_type, metadata_type)` | Export as a typed `pydantic_evals.Dataset`. |
```
alexmojaki, line 28: These files are identical

Image

docs/evaluate/datasets/sdk.md
```diff
@@ -0,0 +1,222 @@
...
+
+
+@dataclass
+class AnswerOutput:
+    answer: str
+    confidence: float
+
+
+@dataclass
+class CaseMetadata:
```
alexmojaki, line 87: Weird that metadata doesn't show in the list of cases in a dataset

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,1138 @@
...
+            return data
+
+        # Convert to typed Dataset using pydantic-evals
+        Dataset, _ = _import_pydantic_evals()
+
+        # Create a properly typed Dataset class
+        typed_dataset_cls: type[Dataset[InputsT, OutputT, MetadataT]] = Dataset[input_type, output_type, metadata_type]  # type: ignore
+        return typed_dataset_cls.from_dict(data, custom_evaluator_types=list(custom_evaluator_types))  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+
+    def import_cases(
```
alexmojaki: What's the difference between importing cases and adding them?

docs/guides/web-ui/datasets/sdk-reference.md
```diff
@@ -0,0 +1,41 @@
...
+| `delete_dataset(id_or_name)` | Delete a dataset and all its cases. |
+| `list_cases(dataset_id_or_name, *, tags)` | List all cases in a dataset. Optionally filter by tags. |
+| `get_case(dataset_id_or_name, case_id)` | Get a specific case. |
+| `add_case(dataset_id_or_name, case, *, tags)` | Add a single `pydantic_evals.Case` to a dataset. Optionally assign tags. |
+| `add_cases(dataset_id_or_name, cases, *, tags)` | Add multiple `pydantic_evals.Case` objects in bulk. Optionally assign tags to all cases. |
+| `create_case(dataset_id_or_name, inputs, *, name, expected_output, metadata, evaluators, source_trace_id, source_span_id, tags)` | Create a case from raw values. |
+| `update_case(dataset_id_or_name, case_id, *, name, inputs, expected_output, metadata, evaluators, tags)` | Update an existing case. |
+| `delete_case(dataset_id_or_name, case_id)` | Delete a case. |
+| `export_dataset(id_or_name, input_type, output_type, metadata_type)` | Export as a typed `pydantic_evals.Dataset`. |
+| `import_cases(dataset_id_or_name, cases, *, tags)` | Bulk import cases (as `Case` objects or dicts). Optionally assign tags to all cases. |
```
alexmojaki: in the UI, if I export a dataset, and then import the resulting file, it treats the whole dataset as a single case. exporting and importing should be symmetrical. also it clearly doesn't apply validation on import.

mkdocs.yml
dmontagu: This is a mkdocs nav change — will apply separately.

mkdocs.yml
dmontagu: Will apply separately along with the other mkdocs.yml suggestion.

pyproject.toml
dmontagu: Good point — the environment marker only controls whether pydantic-evals gets installed, it does not enforce a minimum Python version for the extra itself. This is a pip/uv limitation: extras cannot declare their own `requires-python`. The docs note about Python 3.10+ is accurate as a user-facing instruction, but the packaging does not enforce it. Not sure if there is a better way to handle this.

Leaving unresolved so you can comment, but I don't currently intend to change anything.

docs/guides/web-ui/datasets/sdk.md
dmontagu: Fixed — updated to "for self-hosters".

docs/guides/web-ui/datasets/ui.md
dmontagu: Fixed — removed "from the left-hand navigation".

logfire/experimental/datasets/client.py
dmontagu: Fixed — went with restricting characters. Removed `_quote()` and added client-side name validation via regex (`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`). Also opened a platform PR (pydantic/platform#17528) to add the same regex as a Pydantic `Field(pattern=...)` constraint on the backend. Leaving this unresolved so you can comment if you don't like this solution.

logfire/experimental/datasets/client.py
dmontagu: Fixed — changed from `str | None = _UNSET` to `str = _UNSET`.

logfire/experimental/api_client.py
dmontagu: Fixed — added `pydantic>=2` to the datasets extra in pyproject.toml.

docs/guides/web-ui/datasets/sdk.md
dmontagu: Fixed — reworked the tip to recommend using names (which are more readable) and mention that the UUID is available in the `create_dataset()` response if needed.

docs/evaluate/datasets/ui.md
dmontagu: Addressed in pydantic/platform#17528 — adds server-side JSON schema validation on case create/update.

docs/evaluate/datasets/sdk-reference.md
dmontagu: Addressed in pydantic/platform#17528 — adds query invalidation on case delete so the UI updates immediately.

docs/evaluate/datasets/sdk-reference.md
dmontagu: Addressed in pydantic/platform#17528 — export now supports a `format` query param (json vs pydantic-evals), and import handles the pydantic-evals `{name, cases, evaluators}` format.

docs/evaluate/datasets/sdk.md
dmontagu: Addressed in pydantic/platform#17528 — metadata column added to the cases table in the UI.

logfire/experimental/datasets/client.py
dmontagu: There was no meaningful difference — consolidated `import_cases` into `add_cases`. The `add_cases` method now handles both `Case` objects and plain dicts, and always uses the `/cases/bulk/` endpoint. Also removed `add_case` and `create_case` (callers just use `add_cases` with a single-element list).

docs/guides/web-ui/datasets/sdk-reference.md
dmontagu: Addressed in pydantic/platform#17528 — export now supports a `format` query param (json vs pydantic-evals), and import handles the pydantic-evals `{name, cases, evaluators}` format.

tests/test_datasets_client.py
dmontagu: Noted — deferred to a follow-up. Documented in `plans/datasets-future-work.md` under "VCR Tests".

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,962 @@
...
+            else case.expected_output  # pyright: ignore[reportUnknownMemberType]
+        )
+
+    if case.metadata is not None:
+        data['metadata'] = _serialize_value(case.metadata) if not isinstance(case.metadata, dict) else case.metadata  # pyright: ignore[reportUnknownMemberType]
+
+    if case.evaluators:
+        data['evaluators'] = _serialize_evaluators(case.evaluators)
+
+    return data
```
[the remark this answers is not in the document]
dmontagu: This is intentional — pydantic-evals `Case` objects do not have a `tags` field. Tags are a dataset-management concept handled at the API level via the `tags` parameter on `add_cases`, not a property of individual Case objects.

docs/evaluate/datasets/ui.md
alexmojaki: and the second comment?

logfire/experimental/datasets/client.py
```diff
@@ -0,0 +1,962 @@
...
+        if output_type is not None:
+            data['output_schema'] = _type_to_schema(output_type)
+        if metadata_type is not None:
+            data['metadata_schema'] = _type_to_schema(metadata_type)
+        if guidance is not _UNSET:
+            data['guidance'] = guidance
+        if ai_managed_guidance is not None:
+            data['ai_managed_guidance'] = ai_managed_guidance
+
+        response = self.client.put(f'/v1/datasets/{id_or_name}/', json=data)
```
alexmojaki: i think technically this should be patch, not put

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,962 @@
...
+
+        response = self.client.post('/v1/datasets/', json=data)
+        return self._handle_response(response)
+
+    def update_dataset(
+        self,
+        id_or_name: str,
+        *,
+        name: str = _UNSET,
+        input_type: type[Any] | None = None,
```
alexmojaki, line 423: I find it weird that all the schemas are optional, especially the input schema, considering inputs themselves are required

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,962 @@
...
+        output_type: type[Any] | None = None,
```
alexmojaki, line 424: i can update a schema to something that no longer matches the cases

logfire/experimental/api_client.py
```diff
@@ -0,0 +1,962 @@
...
+            raise DatasetNotFoundError(detail)
+        if response.status_code >= 400:
+            detail = _parse_response_detail(response, default=response.text)
+            raise DatasetApiError(response.status_code, detail)
+        if response.status_code == 204:
+            return None
+        return response.json()
+
+
+class LogfireAPIClient(_BaseLogfireAPIClient[Client]):
```
alexmojaki, line 239: probably future work: dedupe the sync and async clients

docs/evaluate/datasets/ui.md
dmontagu: Good point — the second comment (error should be in red, `/:` is unclear) is a platform UI issue. I've noted it for the platform-side follow-up.

logfire/experimental/datasets/client.py
dmontagu: Fixed in 9461352c — changed to PATCH for both `update_dataset` and `update_case` (sync and async). The backend will also need to be updated to accept PATCH.

logfire/experimental/api_client.py
dmontagu: No schema is equivalent to `{}` (accept anything), so it doesn't make a practical difference — inputs are still required, they just aren't validated against a schema. We could require it, but I think it's fine as-is.

logfire/experimental/api_client.py
dmontagu: Agreed this should be resolved at some point, but I don't want it to block the initial release. We can add backend validation that checks existing cases against the new schema as a follow-up.

PR #1715, opened 2026-02-17
Bypass inline-snapshot import overhead in CI

dmontagu:
## Summary

- Add `tests/_inline_snapshot.py` shim that provides lightweight stubs for `snapshot`, `Is`, `customize_repr`, `raises`, and `warns` when `--inline-snapshot`/`--snap` flags are not passed to pytest
- Replace all 60+ direct imports of `inline_snapshot` across test files with imports from the shim
- The shim detects snapshot flags in `sys.argv` and falls through to the real `inline_snapshot` library when snapshot updates are needed

This avoids the expensive `inline_snapshot` import (AST rewriting, source parsing, etc.) during normal CI test runs where snapshots are only being compared, not updated. The stubs compare values directly and emit warnings on mismatch instead of silently passing.

Modeled after https://github.com/pydantic/pydantic-ai/pull/4346.

## Test plan

- [x] Verified `tests/test_logfire.py` passes with the shim (105 passed, 1 skipped)
- [x] Verified `tests/test_auth.py`, `tests/test_baggage.py`, `tests/test_metrics.py`, `tests/test_tail_sampling.py`, `tests/test_configure.py` pass (exercises `raises`, `warns`, `Is` stubs)
- [x] Verified `tests/exporters/` tests pass
- [x] Verified no direct `inline_snapshot` imports remain outside the shim
- [x] Pre-commit hooks pass (ruff, pyright)

### DIFF
--- a/tests/_inline_snapshot.py
+++ b/tests/_inline_snapshot.py
@@ -1,0 +1,105 @@
+"""Wrapper for inline_snapshot that uses lightweight stubs by default.
+
+inline_snapshot is expensive to import and has heavy startup overhead (AST rewriting, etc.)
+that significantly slows pytest startup.
+
+When no --inline-snapshot flag is passed to pytest, we use lightweight stubs:
+- snapshot(value) returns a proxy that compares using the value, warning on mismatch
+- snapshot() with no args raises an error directing you to use --inline-snapshot=create
+- Is(value) compares using the underlying value
+- customize_repr is a no-op decorator
+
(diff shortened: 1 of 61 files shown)


dmontagu:
Closing — the inline-snapshot shim only showed ~5-10% improvement on logfire (unlike pydantic-ai where xdist amplifies the import overhead). Not worth the maintenance cost of 62 changed files for a modest gain. The Depot runners PR (#1714) provides a bigger win.

PR #1719, opened 2026-02-19
Add Azure AI Inference integration

Kludex:
## Summary
- Add `instrument_azure_ai_inference()` to instrument `ChatCompletionsClient.complete()` and `EmbeddingsClient.embed()` from the `azure-ai-inference` SDK
- Custom monkey-patching approach (Azure SDK doesn't use `httpx` `.request()` pattern)
- Supports sync/async clients, streaming, tool calls, and both v1/latest semconv versions
- 11 tests covering chat, streaming, embeddings, tool calls, uninstrumentation, and message conversion

## Test plan
- [x] `uv run pytest tests/otel_integrations/test_azure_ai_inference.py` - 11/11 passing
- [x] `uv run pytest tests/test_logfire_api.py` - logfire-api stub verified
- [x] `make lint` - clean
- [x] `make typecheck` - zero errors
- [x] `uv run mkdocs build --no-strict` - docs build successfully


alexmojaki:
@Kludex we don't want the old 'semconv'

tests/otel_integrations/test_azure_ai_inference.py
```diff
@@ -0,0 +1,566 @@
...
+                        },
+                        {
+                            'path': ['attributes', 'response_data', 'combined_chunk_content'],
+                            'matched_substring': 'secret',
+                        },
+                    ],
+                },
+            },
+        ]
+    )
```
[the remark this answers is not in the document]
alexmojaki: comparisons to snapshots aren't disabled, they just won't update by default.

a span processor automatically fills in the response model from the request model as a fallback which explains this.

alexmojaki:
Regarding streaming, we also don't want to continue the old pattern with a separate log. Copy from https://github.com/pydantic/logfire/issues/1586:

> Use a single span for streaming, instead of a second log with the response. Unfortunately this will sometimes lead to odd behaviour when reading the stream is interrupted, but the OTel SIG decided this edge case was acceptable, as noted in https://github.com/open-telemetry/semantic-conventions-genai/issues/6#issuecomment-2659124349 . See also the note there about implementing `__del__` to minimise such problems, at least when GC happens nicely...

logfire/_internal/main.py
```diff
@@ -1389,6 +1389,90 @@ def instrument_anthropic(
             is_async_client,
         )
 
+    def instrument_azure_ai_inference(
+        self,
+        azure_ai_inference_client: Any = None,
```
Kludex: This should use the proper type, as the other methods - should be in the `TYPE_CHECKING` above.

We should also have this noted in the `CLAUDE.md`.

logfire/_internal/main.py
```diff
@@ -1389,6 +1389,90 @@ def instrument_anthropic(
...
+            try:
+                from azure.ai.inference.aio import (
+                    ChatCompletionsClient as AsyncChatCompletionsClient,
+                    EmbeddingsClient as AsyncEmbeddingsClient,
+                )
+
+                clients_to_instrument.extend([AsyncChatCompletionsClient, AsyncEmbeddingsClient])
+            except ImportError:  # pragma: no cover
+                pass
+            azure_ai_inference_client = clients_to_instrument
```
Kludex: This import check should be handled the same way as the other instrumentations: in the modules of the instrumentions.

logfire/_internal/main.py
```diff
@@ -1389,6 +1389,90 @@ def instrument_anthropic(
...
+        )
+
+        logfire.configure()
+        logfire.instrument_azure_ai_inference(client)
+
+        response = client.complete(
+            model='gpt-4',
+            messages=[{'role': 'user', 'content': 'What is four plus five?'}],
+        )
+        print(response.choices[0].message.content)
```
Kludex, line 1444: This should be padding 4 characeters >>.

dmontagu:
Closing — per the review, this needs rebuilding against the new GenAI semantic conventions (single span for streaming, no legacy semconv), which is a rewrite rather than a revision of this branch. Tracked in #1586.

PR #1731, opened 2026-02-24
Add variable composition and Handlebars template rendering

dmontagu:
## Summary

- **Variable composition**: Managed variable values can reference other managed variables with `@{variable_name}@` syntax. Composition happens during resolution, before deserialization, and supports nested references, dotted access like `@{brand.tagline}@`, escaping with `\@{...}@`, cycle/depth protection, and `composed_from` metadata on resolved variables and spans.
- **Template rendering**: Variable values can contain Handlebars `{{placeholder}}` templates rendered with runtime inputs. Rendering uses `pydantic-handlebars` and preserves `{{...}}` placeholders while `@{...}@` composition is expanded first.
- **`TemplateVariable[T, InputsT]` API**: `logfire.template_var()` adds a single-step flow where `get(inputs)` resolves the variable, expands `@{...}@` references, renders `{{...}}` placeholders, and deserializes to the declared type.
- **Explicit render path for regular variables**: `logfire.var(..., template_inputs=Inputs)` records the template input schema and lets callers resolve first, then call `resolved.render(inputs)` when they want rendering.
- **Validation and sync support**: Template input schemas are included in variable config/sync, and validation checks composed values for undeclared template fields plus missing references and reference cycles.
- **Docs and demo**: Adds managed-variable docs for templates/composition and a runnable demo covering composition, structured variables, template inputs, and composition-time conditionals.

## Syntax

Composition references use `@{...}@` to avoid conflicting with prompt/runtime templating languages:

```text
@{safety_rules}@
@{brand.tagline}@
@{#if beta_enabled}@Try beta features@{else}@Welcome@{/if}@
```

Runtime template inputs continue to use normal Handlebars syntax:

```text
Hello {{user_name}}. @{safety_rules}@
```

Resolution order:

```text
resolve managed variable
→ expand @{other_variable}@ references
→ render {{runtime_input}} placeholders
→ deserialize/validate to the declared type
```

## Main modules

| Module | Purpose |
|--------|---------|
| `logfire/variables/composition.py` | Reference discovery and expansion with cycle/depth handling |
| `logfire/variables/reference_syntax.py` | `@{...}@` rendering shim that preserves runtime `{{...}}` placeholders |
| `logfire/variables/template_validation.py` | Template field validation and composition cycle detection |
| `logfire/variables/variable.py` | Shared resolution pipeline plus `Variable` and `TemplateVariable` |

## Test plan

- [x] `uv run pytest tests/test_variable_composition.py tests/test_variable_templates.py tests/test_template_validation.py tests/test_variables.py -q`
- [x] `uv run pre-commit run --files docs/reference/advanced/managed-variables/index.md docs/reference/advanced/managed-variables/templates-and-composition.md examples/python/variable_composition_demo.py logfire/_internal/main.py logfire/variables/abstract.py logfire/variables/composition.py logfire/variables/reference_syntax.py logfire/variables/template_validation.py logfire/variables/variable.py tests/test_template_validation.py tests/test_variable_composition.py tests/test_variable_templates.py`
- [x] `uv run mkdocs build --no-strict`


petyosi:
@cubic-dev-ai

petyosi:
@cubic-dev-ai

petyosi:
@cubic-dev-ai

petyosi:
@cubic-dev-ai

logfire/_internal/main.py
```diff
@@ -2603,27 +2636,133 @@ def var(
...
+        return variable
+
+    @overload
+    def template_var(
+        self,
+        name: str,
+        *,
+        type: type[T],
+        default: T | ResolveFunction[T],
+        inputs_type: type[dict[Any, Any]],
```
alexmojaki: Why the overload? I see if I remove it then `inputs_type=dict` on its own fails in a test, but should that be supported without type arguments? with this overload, `inputs_type=dict[str, int]` doesn't work as expected, it gets turned into `dict[Any, Any]`

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,267 @@
...
+
+
+prompt = logfire.var('prompt', type=str, default='Hello {{user_name}}')
+
+with prompt.get() as resolved:
+    rendered = resolved.render(Inputs(user_name='Alice'))
+    print(rendered)  # "Hello Alice"
+```
+
+This is useful when you want the flexibility to render templates on some code paths but not others.
```
alexmojaki: Why do we want `template_var` then? To enforce using `get()` correctly?

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,267 @@
...
+    default=AgentConfig(
+        instructions='You are helping {{user_name}}, a {{tier}} customer.',
+        model='openai:gpt-4o-mini',
+        temperature=0.7,
+    ),
+    inputs_type=UserContext,
+)
+```
+
+```python skip="true"
```
alexmojaki:
```suggestion

```

logfire/_internal/main.py
```diff
@@ -2603,27 +2636,133 @@ def var(
...
+        self,
+        name: str,
+        *,
+        type: type[T],
+        default: T | ResolveFunction[T],
+        inputs_type: type[InputsT],
+        description: str | None = None,
+    ) -> TemplateVariable[T, InputsT]: ...
+
+    def template_var(
```
alexmojaki: as discussed on slack, this should raise an error when the extra isn't installed.

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,267 @@
...
+
+For example, if your `inputs_type` declares `user_name: str` and `is_premium: bool`, but a version value references `{{unknown_field}}`, the validation will flag this as an error.
+
+## Variable Composition {#variable-composition}
+
+**Composition** lets a variable's value reference other variables using `@{variable_name}@` syntax. When the variable is resolved, `@{ref}@` references are expanded by looking up the referenced variable and substituting its value.
+
+This is useful for building values from reusable fragments:
+
+```python skip="true"
```
alexmojaki: Why is this skipped?

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,267 @@
...
+The `@{}@` syntax supports a small Handlebars-compatible subset for composing variables. It supports simple references, dotted field reads, and block helpers that branch or iterate over a top-level referenced variable:
+
+| Syntax | Description |
+|--------|-------------|
+| `@{variable_name}@` | Insert a variable's value |
+| `@{variable.field}@` | Access a nested field |
+| `@{#if variable}@...@{else}@...@{/if}@` | Conditional on whether a variable is set |
+| `@{#each items}@...@{/each}@` | Iterate over a list variable |
+
+Block helper conditions and iterables must be top-level variable names. Use `@{#if user}@...@{user.active}@...@{/if}@` rather than `@{#if user.active}@`.
```
alexmojaki, line 162: This seems like an unfortunate restriction. Why don't `{{` and `@{` support the same things? Why don't they use the same implementation?

tests/test_variable_composition.py
```diff
@@ -0,0 +1,817 @@
...
+        )
+        expanded, composed = expand_references('"@{config.prompt}@ using @{config.model}@"', 'my_var', resolve_fn)
+        assert json.loads(expanded) == 'Hello Alice using gpt-4'
+        assert len(composed) == 1
+        assert composed[0].name == 'config'
+        assert composed[0].value == '{"prompt": "Hello Alice", "model": "gpt-4"}'
+        assert len(composed[0].composed_from) == 1
+        assert composed[0].composed_from[0].name == 'name'
+
+    def test_cycle_detection(self):
```
alexmojaki, line 143: Why is this the chosen behavior? A cycle silently "succeeds" — `result.exception` is `None`, `_reason` is `'resolved'`, and the leftover `@{a}@` ends up in the final string. The error is only visible if a caller walks into `composed_from[*].composed_from[*].error`, which most won't.

Two reasonable alternatives:
1. Surface the cycle: set `_reason='other_error'` and populate `exception` on the top-level `ResolvedVariable` (matching what happens for `VariableCompositionError` in `_expand_and_deserialize`).
2. Fall back to the code default on cycle (also matching the depth-exceeded behavior).

Was the silent-pass-through chosen deliberately (e.g. so a partial composition can still be useful), or is it an artifact of the cycle being caught at the inner recursion level? A comment in the code or test would help.

logfire/variables/composition.py
```diff
@@ -0,0 +1,365 @@
...
+        VariableCompositionCycleError: If a circular reference is detected.
+    """
+    if _depth > MAX_COMPOSITION_DEPTH:
+        raise VariableCompositionError(
+            f'Maximum composition depth ({MAX_COMPOSITION_DEPTH}) exceeded '
+            f"while expanding '{variable_name}'. This likely indicates a circular reference."
+        )
+
+    if variable_name in _visited:
+        raise VariableCompositionCycleError(f'Circular reference detected: {" -> ".join(_visited)} -> {variable_name}')
```
alexmojaki, line 129: `_visited` is a `frozenset`, so this `' -> '.join(_visited)` produces a non-deterministic path. For a mutual cycle `b -> a -> b`, the message can come out as `a -> b -> b`, which is misleading (it implies `b` referenced itself directly). The cycle tests only do `'Circular reference' in error`, so this slips through.

Could be fixed by tracking `_visited` as a tuple (or any order-preserving structure) and adding a test that asserts the exact path string.

docs/reference/advanced/managed-variables/templates-and-composition.md
alexmojaki: Also, can we make `render` typesafe by adding an optional template inputs type to `var`?

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,267 @@
...
+
+# Resolution: compose @{tone_instructions}@ first, then render {{user_name}} and {{language}}
+with chat_prompt.get(ChatInputs(user_name='Alice', language='French')) as resolved:
+    print(resolved.value)
+    # "You are helping Alice. Respond in French. Be friendly and concise."
+```
+
+### Cycle Detection
+
+The system detects circular references at write time. If variable A references `@{B}@` and variable B references `@{A}@`, pushing this configuration will produce an error. This prevents infinite loops during resolution.
```
alexmojaki: pushing the configuration only shows a warning, there doesn't seem to be any error at any point.

logfire/_internal/main.py
petyosi: Fixed in 94d3c408: `template_var()` and `var(..., template_inputs=...)` now check for `pydantic-handlebars` at definition time and raise the existing dependency error if `logfire[variables]` is missing.

docs/reference/advanced/managed-variables/templates-and-composition.md
petyosi: Fixed in 67e785e7: merged the split example back into one code block.

logfire/variables/composition.py
petyosi: Fixed in 67e785e7: `_visited` is now an ordered tuple instead of a `frozenset`, and the cycle tests assert exact path strings.

docs/reference/advanced/managed-variables/templates-and-composition.md
petyosi: Fixed in 67e785e7: changed the docs to say `variables_validate()` reports cycles and `variables_push(strict=True)` fails, instead of saying push always errors.

docs/reference/advanced/managed-variables/templates-and-composition.md
petyosi: Agreed that full parity would be nicer, but leaving that out of this pass. Composition has extra pre-render requirements (dependency discovery, recursive resolution, cycle/depth checks, unresolved-reference preservation, and `composed_from` metadata), so making `@{...}@` fully match `{{...}}` likely needs AST-level dependency extraction rather than more regex parsing. The smaller follow-up would be supporting dotted paths in block helper conditions.

docs/reference/advanced/managed-variables/templates-and-composition.md
alexmojaki: this doesn't give me errors:

```python
agent_prompt = logfire.var(
    'agent_prompt3',
    type=str,
    default='You are a helpful assistant3. @{agent_prompt3}@',
)

logfire.variables_push(strict=True)
```

I see `print(logfire.variables_validate())` shows the warning: `ValidationReport(errors=[], variables_checked=1, variables_not_on_server=[], description_differences=[], reference_warnings=['Reference cycle detected: agent_prompt3 -> agent_prompt3'])`

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,267 @@
...
+        temperature=0.7,
+    ),
+    inputs_type=UserContext,
+)
+
+with agent_config.get(UserContext(user_name='Alice', tier='premium')) as resolved:
+    print(resolved.value.instructions)  # "You are helping Alice, a premium customer."
+    #> You are helping Alice, a premium customer.
+    print(resolved.value.model)         # "openai:gpt-4o-mini" (unchanged)
+    #> openai:gpt-4o-mini
```
alexmojaki:
```suggestion
    print(resolved.value.instructions)
    #> You are helping Alice, a premium customer.
    print(resolved.value.model)
    #> openai:gpt-4o-mini
```

logfire/variables/variable.py
```diff
@@ -278,46 +256,21 @@ def _resolve(
...
                     )
                 # Label not found - fall through to default resolution
 
             serialized_result = provider.get_serialized_value(self.name, targeting_key, attributes)
 
             if serialized_result.value is None:
                 default = self._get_default(targeting_key, attributes)
+                if render_fn is not None:
+                    default = self._render_default(default, render_fn)
                 return _with_value(serialized_result, default)
```
alexmojaki: Case 1: composition is skipped entirely when the parent variable falls back to its code default.

If `provider.get_serialized_value(self.name, ...)` returns `None` (variable not registered in the provider, or registered but no label/rollout resolves), we go straight to `_get_default()` and return it. `_expand_and_deserialize` is never called, so any `@{ref}@` syntax inside the code default surfaces verbatim in `result.value`.

This is what makes the docs example misleading when run without a configured provider — the example just calls `agent_prompt.get()` with only code defaults and shows `@{safety_rules}@` expanded, but in reality the literal text is returned. The example block is marked `skip="true"` so pytest-examples doesn't catch the divergence.

There's an explicit test at `tests/test_variable_composition.py:769` (`test_no_value_no_composition`) that locks this in. Is the behavior intentional, or should code defaults also flow through `_expand_and_deserialize`? If intentional, the docs should say so loud and clear.

logfire/variables/variable.py
```diff
@@ -327,6 +280,109 @@ def _resolve(
...
+        if has_references(serialized_value):
+
+            def resolve_ref(ref_name: str) -> tuple[str | None, str | None, int | None, str]:
+                ref_resolved = provider.get_serialized_value(ref_name, targeting_key, attributes)
+                return (
+                    ref_resolved.value,
+                    ref_resolved.label,
+                    ref_resolved.version,
+                    ref_resolved._reason,  # pyright: ignore[reportPrivateUsage]
+                )
```
alexmojaki: Case 2: this `resolve_ref` only consults the provider — it never falls back to the referenced variable's code default.

If `x` is registered (`logfire.var('x', default='X', type=str)`) but `provider.get_serialized_value('x', ...)` returns `None` (empty labels / rollout doesn't pick anything), the composition layer treats `@{x}@` exactly the same as a reference to a totally unknown variable: `ref_serialized is None` at `composition.py:152`, and the literal `@{x}@` is left in the output. `x`'s code default is never used.

That's surprising and arguably worse than Case 1 — the user has gone to the trouble of registering both variables and providing a default for `x`, but the default isn't honored when `x` is referenced from another variable.

Fix would be to maintain a `name -> Variable` registry on `LogfireConfig` (something likely already exists for instrumentation) and have `resolve_ref` fall back to the registered variable's default-serialization when the provider returns `None`. That preserves the symmetry "`x.get()` gives you the default" ↔ "`@{x}@` gives you the default."

Also: no test covers this. `tests/test_variable_composition.py:609` (`test_nonexistent_reference_left_unexpanded`) is the closest, but `nonexistent` has no `VariableConfig` at all — distinct from "registered with code default, provider returns None."

tests/test_variable_composition.py
```diff
@@ -0,0 +1,815 @@
...
+        """When variable resolves to None (code default), no composition happens."""
+        variables_config = VariablesConfig(
+            variables={
+                'main': VariableConfig(
+                    name='main',
+                    json_schema={'type': 'string'},
+                    labels={},
+                    rollout=Rollout(labels={}),
+                    overrides=[],
+                ),
```
alexmojaki, line 821: The test still passes when these lines are removed, i.e. `variables={}`. Maybe that should be an additional separate test case. If not, these lines seem redundant.

docs/reference/advanced/managed-variables/templates-and-composition.md
alexmojaki: I spent some time prototyping the render-pass alternative on `worktree-poc-render-discovery` (see [`logfire/variables/discovery_render.py`](https://github.com/pydantic/logfire/blob/worktree-poc-render-discovery/logfire/variables/discovery_render.py)). Not AST-level — just the existing renderer with a tracking context.

**Approach.** Hand the pydantic-handlebars renderer a `dict` subclass whose `.get()` / `__getitem__` records every top-level lookup. Override `if`/`unless`/`each`/`with` to render both branches (so refs in `else` aren't missed) and have `each`/`with` pass the parent tracker as the iterated context (so inner top-level refs still hit the same tracker).

**What it catches that the current regex doesn't:**

| Input | Regex | Render |
|---|---|---|
| `@{#if user.active}@` | `[]` | `['user']` |
| `@{#each user.posts}@` | `[]` | `['user']` |
| `@{lookup obj key}@` | `[]` | `['obj', 'key']` |
| `@{#if a}@x@{else if b}@y@{/if}@` | `['a']` | `['a', 'b']` |
| `@{#each items}@@{../top}@@{/each}@` | `['items']` | `['items', 'top']` |
| `@{ user }@` (whitespace) | `[]` | `['user']` |
| `@{#if (eq a b)}@` | `[]` | `['a', 'b']` * |

\* only with `extra_helpers=True` (or a dummy `eq` registration). Without it, the renderer raises on unknown-helper before the subexpression's arg paths are evaluated.

So the "dotted block headers" gap isn't the only one — mustache helper calls, parent paths, `else if` chains, hash args, and whitespace are all silently missed too. Most matter more in theory than in practice, but `else if` and dotted block headers seem common enough that users would hit them.

**Caveats found while building it:**

1. `HandlebarsEnvironment.render()` calls `to_jsonable_python(ctx)` on the context (`_environment.py:28`), which flattens any dict subclass to a plain dict and drops the side-effecting `.get`. The POC bypasses this with `env._compile_fn(template)(context)` — private API. A real implementation would want pydantic-handlebars to grow a "pass context through unchanged" mode, or we live with the dependency.
2. Parse errors are a regression: the regex still extracts refs from malformed templates because it's just text scanning; the renderer bails on parse failure. Matters for the validation path.
3. Subexpressions only work if every helper users might write is registered (or `extra_helpers=True`).
4. Render order ≠ first-occurrence order. Same set, different list — doesn't seem to matter for current consumers.

POC branch: `worktree-poc-render-discovery` (commit `a1a27435`). Comparison harness is in `ignoreme/poc_compare.py` (gitignored).

logfire/variables/variable.py
```diff
@@ -327,6 +288,124 @@ def _resolve(
...
+                    default = self._get_default(targeting_key, attributes)
+                    return ResolvedVariable(
+                        name=self.name,
+                        value=default,
+                        exception=VariableCompositionError(composition_error),
+                        _reason='other_error',
+                        label=serialized_result.label,
+                        version=serialized_result.version,
+                        composed_from=composed,
+                    )
```
alexmojaki, line 389: The new `_first_composition_error` fallback is all-or-nothing: as soon as *any* nested composition error is found anywhere in the tree, the parent's entire composed value is discarded and `result.value` becomes the code default — even if other top-level references in the same parent resolved successfully.

Concrete repro (runs against this branch):

```
# parent = '@{good_a}@ + @{bad_b}@'
#   good_a -> 'GOOD'        (resolves cleanly)
#   bad_b -> '@{bad_c}@'    cycles with bad_c
#
# Result:
# value:     'PARENT_FALLBACK'  ← code default, not 'GOOD + @{bad_b}@'
# reason:    other_error
# exception: VariableCompositionError('Circular reference detected: parent -> bad_b -> bad_c -> bad_b')
# composed_from:
#   @{good_a}@ -> value='GOOD'   ← computed correctly, then thrown away
#   @{bad_b}@ -> value='@{bad_b}@'
#     @{bad_c}@ -> value='@{bad_b}@'
#       @{bad_b}@ -> ERROR: Circular reference...
```

`good_a`'s successful expansion is right there in `composed_from`, and the renderer even produced the half-rendered `'GOOD + @{bad_b}@'` along the way — but the caller sees only `'PARENT_FALLBACK'`. A single misconfigured variable can blank out unrelated successful expansions in any parent that references it.

Also not covered by tests. `test_cycle_falls_back_gracefully` and `test_composition_exception_falls_back` both have only one reference in the parent, so neither exercises the surprising "successful siblings get discarded" half of this behavior. The tests prove "error surfaces and falls back", not "good work is also lost."

Two questions:

1. Is this the intended trade-off? Arguments for: a partially-rendered value with stray `@{x}@` literals could be more misleading than a clean fallback. Arguments against: operators who rely on composition for prompts/configs will see unrelated parents go dark whenever one referenced variable misbehaves — a noisy failure mode.
2. If yes, please add a test that pins it down explicitly (parent with both a resolving and a cycling ref; assert both `result.value == default` and `composed_from[good_one].value == 'GOOD'`) and a comment in `_first_composition_error` explaining the choice. Right now the behavior is implicit and easy to break later by accident.

logfire/variables/variable.py
```diff
@@ -398,10 +700,20 @@ def _with_value(details: ResolvedVariable[Any], new_value: T_co) -> ResolvedVari
...
 
 
+def _first_composition_error(composed: list[ComposedReference]) -> str | None:
+    """Return the first nested composition error, if any."""
+    for ref in composed:
+        if ref.error is not None:
+            return ref.error
+        if nested_error := _first_composition_error(ref.composed_from):
+            return nested_error
+    return None
```
alexmojaki, line 778: Two flavors of "failed reference" are now treated very differently, and the asymmetry isn't documented or obviously intentional:

| Failure mode | Where it lives | What this function does | Caller effect |
|---|---|---|---|
| Cycle, depth limit, non-JSON value | `ComposedReference.error` (string) | Returns the message → caller falls back to code default with `_reason='other_error'` | All-or-nothing fail |
| Unresolved (unknown) variable / `value is None` from provider | `ComposedReference.reason='unrecognized_variable'`, `error=None` | Skipped — `ref.error is None` | Literal `@{x}@` left in the rendered value, `_reason='resolved'` |

So if a referenced variable is *missing*, you silently get `'Hello @{x}@'` and resolution looks successful. If the same variable is *cyclic*, you get the parent's code default + an exception. Hard to predict from a user's standpoint, and easy to construe as a bug either way (e.g. "why does my unknown-variable case not fall back?").

Worth picking one model and applying it consistently — either:
1. Unresolved references also surface as errors (set `error=` at composition.py:152-163 and let `_first_composition_error` pick them up), or
2. Cycles also leave their literal `@{x}@` in place rather than nuking the whole result.

Or, if the current split is intentional (e.g. "missing" is a normal operational state during rollout, "cycle" is a misconfiguration), say so in a comment here.

logfire/variables/variable.py
```diff
@@ -327,6 +288,124 @@ def _resolve(
...
+                    serialized_value,
+                    self.name,
+                    resolve_ref,
+                )
+                if composition_error := _first_composition_error(composed):
+                    default = self._get_default(targeting_key, attributes)
+                    return ResolvedVariable(
+                        name=self.name,
+                        value=default,
+                        exception=VariableCompositionError(composition_error),
```
alexmojaki, line 384: Minor: this wraps the error message string in a fresh `VariableCompositionError(...)`, losing the original subclass. `expand_references` raises specific subclasses (`VariableCompositionCycleError`, depth-limit `VariableCompositionError`, etc.), but by the time a nested error reaches here it's been flattened to `str(e)` inside `ComposedReference.error` and re-wrapped here as the base class.

Caller-side consequence: `isinstance(result.exception, VariableCompositionCycleError)` will never match for the normal (nested) cycle path, only for the rare direct-cycle path through the `except VariableCompositionError as e` branch below at line 352 — where the original `e` is preserved. Two paths to the same conceptual outcome, two different exception types.

If callers are expected to switch on cycle-vs-depth-vs-other, this should preserve the subclass — either by storing the original exception (not just `str(e)`) on `ComposedReference`, or by re-detecting based on the message at this site. If callers are *not* expected to switch on subclass, then the `VariableCompositionCycleError` distinction is doing nothing useful in the public surface and could be flattened to just `VariableCompositionError` everywhere.

logfire/variables/variable.py
```diff
@@ -335,6 +414,44 @@ def _get_default(
...
 
+    def _get_serialized_default(
+        self, targeting_key: str | None = None, merged_attributes: Mapping[str, Any] | None = None
+    ) -> str | None:
+        """Return the code default serialized as JSON, or None if serialization fails."""
+        try:
+            default = self._get_default(targeting_key, merged_attributes)
+            return self.type_adapter.dump_json(default).decode('utf-8')
+        except (ValueError, TypeError, RuntimeError):
+            return None
```
alexmojaki, line 480: This silently swallows three broad exception classes and returns `None` — no log, no warning, no metric. `None` then propagates two different ways depending on caller:

- In `_resolve_serialized_default` (just below): `if serialized_default is None: return None`, which makes `_resolve` fall through to `_with_value(serialized_result, self._get_default(...))` — the *unserialized* default. So a default that fails JSON serialization is still handed back to the user via the non-composition path, and they get the un-composed value. Two different fallback paths for the same variable depending on whether composition is needed.

- In `resolve_ref` (line 325): `None` means "no code default", so `@{x}@` is left as a literal. From the caller's perspective, indistinguishable from "variable not registered."

Concerns:

1. **Too broad a net.** `ValueError` and `TypeError` are reasonable for `dump_json` failure. `RuntimeError` is unusual and catches a lot of unrelated programmer errors — would prefer to see what motivated including it (a specific reproducer in tests?).
2. **Silent.** If a user's default genuinely can't be serialized, this is almost always a bug they want to know about. Currently they'd see `@{x}@` left as literal, scratch their head, and have no breadcrumb. At minimum a `logger.warning` with the variable name and the original exception. Better: re-raise (since this is a programmer error, not a runtime condition), or surface via `exception` on the `ResolvedVariable` like other failure modes.
3. **Inconsistent with the rest of the file.** The composition layer already has a "surface errors" path (`_first_composition_error`). A failed code-default serialization is conceptually similar — "composition couldn't proceed" — but it doesn't flow through that path, just disappears.

If this is defensive code for a case that's never expected to fire, a comment explaining what's actually being defended against (and why it can't just raise) would help future readers.

logfire/variables/abstract.py
```diff
@@ -224,8 +390,8 @@ def has_errors(self) -> bool:
 
     @property
     def is_valid(self) -> bool:
-        """Return False if there are any validation errors or any variables not defined in the (possibly remote) config."""
-        return len(self.errors) == 0 and len(self.variables_not_on_server) == 0
+        """Return False if there are validation errors, missing variables, or reference warnings."""
+        return len(self.errors) == 0 and len(self.variables_not_on_server) == 0 and len(self.reference_warnings) == 0
```
alexmojaki: Naming nit: `reference_warnings` is a misnomer if its presence makes `is_valid` return False and causes `push(strict=True)` to abort with `Error: Reference warnings found` (line 1307). A field called "warnings" that fatally blocks an operation isn't a warning — it's an error.

Pick one model:

1. **They're errors.** Rename to `reference_errors` (or merge into `errors`). `is_valid == False` and `push(strict=True)` failing are then consistent with the name.
2. **They're genuinely warnings.** Drop them from `is_valid` and from `push(strict=True)`'s blocker check. Show them in the diff output as informational; let `strict=False` keep behaving as today (push proceeds). `strict=True` semantically means "halt on errors," not "halt on anything I labeled noteworthy."
3. **Mixed: usually warnings, but escalated under strict.** Keep the name but document this explicitly in the `strict` docstring and don't include them in `is_valid`. The "strict promotes warnings to errors" model is well-known (Python's `-W error`, gcc `-Werror`) — but it shouldn't leak into the data class `is_valid` property, which should reflect a single notion of validity.

Currently the code straddles models 1 and 3 inconsistently: `is_valid` treats them as errors (model 1), but the user-facing string says "Reference warnings found" (model 3). Worth resolving before this becomes public API.

logfire/variables/variable.py
```diff
@@ -327,6 +288,124 @@ def _resolve(
...
+
+        # Expand @{references}@ if any are present
+        if has_references(serialized_value):
+
+            def resolve_ref(ref_name: str) -> tuple[str | None, str | None, int | None, str]:
+                ref_resolved = provider.get_serialized_value(ref_name, targeting_key, attributes)
+                if ref_resolved.value is None and (ref_variable := self._variable_registry.get(ref_name)) is not None:
+                    ref_default = ref_variable._get_serialized_default(targeting_key, attributes)
+                    if ref_default is not None:
+                        return (ref_default, None, None, 'code_default')
```
alexmojaki: This introduces a new `'code_default'` reason that's observable on `ComposedReference.reason` when a child ref falls back to its registered code default. Useful — but the same fact is *not* observable at the top level when the **parent itself** falls back to its code default.

Lean repro:

```python
# parent_a is registered in provider with empty labels/rollout (so provider returns None).
# parent_a has a plain code default with no @{}@:
parent_a = logfire.var('parent_a', type=str, default='hello (default)')
a = parent_a.get()
assert a.value == 'hello (default)'
assert a._reason == 'resolved'   # no signal that the code default was used

# parent_b is in provider, references @{child}@. child is registered with a code default but
# not in provider.
parent_b = logfire.var('parent_b', type=str, default='IRRELEVANT')
logfire.var('child', type=str, default='CHILD-DEFAULT')
b = parent_b.get()
assert b.value == 'hello CHILD-DEFAULT'
assert b.composed_from[0].reason == 'code_default'   # clear signal child used default
```

In (B) you can tell programmatically that a code default was used. In (A) you can't — `_reason` echoes the provider's outcome whether or not the code default was needed. Callers wanting to detect "parent used default" have to compare `value` against the literal default themselves.

To be fair: the lack of a top-level signal is pre-existing — that code path at line 270 (`_with_value(serialized_result, self._get_default(...))`) has always echoed the provider's `_reason`. This PR introduces the *child* signal, which is what makes the asymmetry newly visible.

Two ways to make this symmetric:

1. **Add the signal at the top level too.** When `_resolve_serialized_default` (or the simple-default fallback at line 270) runs, set `_reason = 'code_default'` on the top-level `ResolvedVariable`. The override at line 451 (`if result._reason == 'resolved': result._reason = serialized_result._reason`) would need to be reconsidered.
2. **Drop the child signal.** Don't bake `code_default` into `ComposedReference.reason`; just leave it as `'resolved'` or `'unrecognized_variable'`. Callers don't get to introspect either case.

(1) is more useful since "did this variable use its registered default vs. the operator-configured value?" is a legitimate observability question for both parents and children. Worth deciding before `'code_default'` becomes part of the public surface.

logfire/variables/composition.py
```diff
@@ -0,0 +1,365 @@
...
+    name: str
+    """Name of the referenced variable."""
+    value: str | None
+    """Expanded raw string value, or None if unresolved."""
+    label: str | None
+    """Label of the referenced variable's resolution."""
+    version: int | None
+    """Version of the referenced variable's resolution."""
+    reason: str
+    """Resolution reason (e.g., 'resolved', 'unrecognized_variable')."""
```
alexmojaki: Naming inconsistency with the parent type: `ResolvedVariable._reason` (`abstract.py:99`) is `_`-prefixed and carries the comment *"we might eventually make this public, but I didn't want to yet"*. The new `ComposedReference.reason` here is plain and public, answering that deferred question one way for children without revisiting the parent.

Result: two fields representing the same concept on closely-related types, one private, one public — and the enums are already drifting:

| Reason value | `ResolvedVariable._reason` | `ComposedReference.reason` |
|---|---|---|
| `'resolved'` | yes | yes |
| `'unrecognized_variable'` | yes | yes |
| `'context_override'` | yes | no |
| `'missing_config'` | yes | no |
| `'validation_error'` | yes | no |
| `'other_error'` | yes | no |
| `'no_provider'` | yes | no |
| `'code_default'` | no | **yes** (new in this PR) |

The asymmetric `'code_default'` is the one I flagged in the parent-vs-child observability comment (#discussion_r3220057944). The naming inconsistency is the structural version of the same issue.

Clean options:

1. **Promote `_reason` to public**: rename `ResolvedVariable._reason` → `reason`, extract the enum into a shared `Literal[...]` type alias used by both classes, and keep the two enums in sync going forward.
2. **Demote `ComposedReference.reason` to private**: rename to `_reason` here. Preserves the original "deferred" stance but means any further additions to either enum need to stay private.

(1) is more useful now that `code_default` is a signal callers will reasonably want to switch on. Doing it now is much cheaper than after `ComposedReference` ships.

logfire/variables/variable.py
```diff
@@ -327,6 +288,124 @@ def _resolve(
...
+                return (
+                    ref_resolved.value,
+                    ref_resolved.label,
+                    ref_resolved.version,
+                    ref_resolved._reason,  # pyright: ignore[reportPrivateUsage]
+                )
```
alexmojaki: **Bug: `variable.override()` does not propagate through composition.**

`var.override(...)` is honoured when you call `var.get()` directly (handled at line 244 inside `_resolve`), but a sibling variable whose value contains `@{var}@` will *not* see the overridden value — composition silently uses the provider value (or the code default) instead. So the override is inconsistent with itself: `x.get()` and the `x` substitution inside `parent.get()` can return different values in the same execution.

Repro:

```python
# Provider has x='PROVIDER_X' and parent='hello @{x}@'.
x = logfire.var('x', type=str, default='IRRELEVANT')
parent = logfire.var('parent', type=str, default='IRRELEVANT')

assert x.get().value == 'PROVIDER_X'
assert parent.get().value == 'hello PROVIDER_X'

with x.override('OVERRIDDEN_X'):
    assert x.get().value == 'OVERRIDDEN_X'          # direct .get() respects override
    p = parent.get()
    assert p.value == 'hello PROVIDER_X'            # ← @{x}@ ignores the override
    assert p.composed_from[0].reason == 'resolved'  # not 'context_override'
```

The missing `'context_override'` value on `ComposedReference.reason`'s enum tipped this off — it can never appear in practice because `resolve_ref` here never produces it.

**Why this happens, and the broader pattern.**

`resolve_ref` is a partial reimplementation of `_resolve`'s lookup priority, and the parts that got dropped are silently load-bearing:

| What `_resolve` does (line 243-274) | What `resolve_ref` does (here) |
|---|---|
| 1. Check `_VARIABLE_OVERRIDES` | — (missing → this bug) |
| 2. Call provider | 1. Call provider |
| 3. Fall back to code default | 2. Fall back to code default via `_variable_registry` |
| 4. Span / baggage instrumentation | — (probably also wanted) |

The layering itself is fine: composition operates on raw JSON strings on purpose, so the logic can be shared with the backend OFREP endpoint (per `composition.py`'s module docstring). The accidental part is that the SDK's wiring of `resolve_fn` got hand-rolled here as a subset of `_resolve`, rather than calling a shared lookup helper.

**Architectural fix.** Extract a single SDK-side helper — e.g. `LogfireConfig.resolve_serialized(name, targeting_key, attributes) -> (value, label, version, reason)` — that encodes the full priority (override → provider → registered code default), then have both `_resolve` (for its own variable) and `resolve_ref` (for child refs) route through it. The composition module's API stays unchanged; the SDK side just stops drifting.

If you do that, several of the other open threads collapse as side effects:

- `'context_override'` naturally becomes a possible `ComposedReference.reason` value (no enum drift to manage).
- The `_reason` enum split between parent (`ResolvedVariable._reason`) and child (`ComposedReference.reason`) stops drifting, because both reasons originate from the same helper. (discussion_r3220113401)
- The "no top-level signal that the parent used its own code default" asymmetry goes away, because the parent's own `_reason` is derived through the same logic as child reasons. (discussion_r3220057944)

So most of the recent observability threads are downstream of "`resolve_ref` reinvented `_resolve` incompletely." Fixing that one place is probably less code than addressing the symptoms individually.

docs/reference/advanced/managed-variables/templates-and-composition.md
alexmojaki: Discussed with David: let's just remove render, this use case isn't worth it. If we wanted this use case later, it'd be better to add a method to template_vars to get without rendering.

logfire/_internal/main.py
alexmojaki: Resolution: remove the (dict) overload and corresponding test, require parametrising dict in usage.

docs/reference/advanced/managed-variables/templates-and-composition.md
alexmojaki: Discussion with David: either way, @{}@ should support all the syntax, using some proper handlebars parsing thing instead of regex. Could be the method in the comment above, or maybe a dedicated helper in handlebars which returns names used by a template.

logfire/variables/variable.py
alexmojaki: Decision: keep falling back to code default as is, but also emit a warning. The idea is that the code default is expected to be reliable and unlikely to end up with unrendered syntax. In the future we can make this more likely by emitting a warning when you create a variable whose default can't be rendered purely using defaults of other existing local variables, recursively.

logfire/variables/variable.py
alexmojaki: Decision: variable composition should use the full _resolve path as much as possible. it should respect overrides and create baggage. however it shouldn't create an extra span.

logfire/variables/composition.py
alexmojaki: Decision: make both public

logfire/variables/variable.py
alexmojaki: decision: unresolved vars are like other errors, as in https://github.com/pydantic/logfire/pull/1731#discussion_r3221396041, emit a warning and fallback

logfire/variables/variable.py
alexmojaki: need to check if the other comments make this outdated

logfire/variables/variable.py
alexmojaki: decision: top-level reason should have code_default when provider explicitly says to use code default

logfire/_internal/main.py
dmontagu: Done in c068526c — dropped the `dict` overload (and its test). Callers now parametrise (`dict[str, str]`, etc.) or use a Pydantic model.

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Done in 8a3a1a74 — removed both `ResolvedVariable.render()` and `template_inputs=` on `logfire.var()`. Template rendering now lives exclusively in `template_var()`. If we want "resolve without rendering" later we'll add that as a method on `TemplateVariable` as you suggested.

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Deferring this to a follow-up PR per discussion with David. Filing it as a tracked issue so we don't lose the work-already-done — your `worktree-poc-render-discovery` prototype is the obvious starting point.

tests/test_variable_composition.py
dmontagu: Addressed in fd0c69fd — composition errors (cycles, depth, render failures, unresolved refs) now emit a `RuntimeWarning` when we fall back to the code default, so it's no longer silent. The cycle still surfaces via `result.exception` (`VariableCompositionError`) and `composed_from[...].error` as before. Added `test_cycle_falls_back_gracefully` to lock in the warn+fallback behavior.

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Verified — `test_push_variables_strict_fails_with_reference_errors` in `tests/test_push_variables.py` covers exactly the cycle-in-code-default case you described, and confirms `push_variables(strict=True)` returns False and writes nothing. Renamed the field from `reference_warnings` to `reference_errors` in 551b74fb so the name matches the fact that it blocks strict push and makes `is_valid` False.

logfire/variables/variable.py
dmontagu: Addressed across two commits. db1f8e14 routes child references through a shared lookup that consults overrides + provider + the referenced variable's registered code default. 3b2cb6b7 surfaces `reason='code_default'` at the top level when the parent itself falls back, so the misleading docs example you noticed now reflects reality. The skip on that example block is the standard pytest-examples skip used across the docs for snippets that call `logfire.configure()` more than once — not because the behavior would diverge.

logfire/variables/variable.py
dmontagu: Addressed in db1f8e14 — `resolve_ref` now follows the full lookup priority (override → provider → registered code default) by looking the referenced variable up in `self._variable_registry` and reusing its `_get_serialized_default`. The new test `test_reference_falls_back_to_registered_code_default` pins down the registered-with-default-but-provider-returns-None case you flagged.

tests/test_variable_composition.py
dmontagu: The test is now parametrised over both cases (`register_main=False` with `variables={}` and `register_main=True` with `_make_variables_config(main=None)`), so each branch covers a distinct scenario.

logfire/variables/variable.py
dmontagu: Addressed in fd0c69fd. Per the huddle decision, kept the "fall back to code default" behavior but added a `RuntimeWarning` so the misconfiguration is loud instead of silent. The siblings-discarded trade-off is now visible at runtime; we can layer the validate-at-creation-time check on top later, as discussed.

logfire/variables/variable.py
dmontagu: Addressed in fd0c69fd. Unresolved references are now treated the same as other composition errors: `composition.py` sets `ComposedReference.error` for them, `_first_composition_error` picks them up, and the SDK emits a warning and falls back to the code default. New test `test_nonexistent_reference_falls_back_with_warning` covers it.

logfire/variables/variable.py
dmontagu: Leaving as-is for this PR. With the warn-and-fallback change in fd0c69fd, callers learn about the failure at runtime via the warning regardless of which `VariableCompositionError` subclass is set on `result.exception`, so the subclass-distinction asymmetry is much less load-bearing. If we want callers to branch on cycle vs depth-vs-other later, the right fix is to propagate the original exception via `ComposedReference` rather than just `str(e)` — but that touches the composition module's public surface, which I'd rather do in a follow-up.

logfire/variables/variable.py
dmontagu: Leaving as-is for this PR. With db1f8e14 + fd0c69fd in place, a serialization failure on a referenced variable's code default now propagates as an unresolved reference, which surfaces as a composition error (warn + fallback at the parent level) rather than disappearing silently. That covers the worst-case path you flagged. I'd rather not broaden the catch in a follow-up commit without a concrete reproducer for the `RuntimeError` branch you questioned — happy to revisit if one comes up.

logfire/variables/abstract.py
dmontagu: Renamed to `reference_errors` in 551b74fb. Per discussion with David, model (1) was the right call: the field already makes `is_valid` False and blocks `push(strict=True)`, so the name should say "error." Also re-coloured the formatted-report output to red instead of yellow and renamed the helper / tests for consistency.

logfire/variables/variable.py
dmontagu: Done in 3b2cb6b7. `ResolvedVariable.reason` is now `'code_default'` whenever the parent's code default is used, mirroring the existing child-level signal. Added `test_top_level_reason_is_code_default_when_provider_has_no_value` to lock it in.

logfire/variables/composition.py
dmontagu: Done in 3b2cb6b7 — both fields are now public. `_reason` was renamed to `reason` on `ResolvedVariable`, and both `ResolvedVariable.reason` and `ComposedReference.reason` are now typed with a shared `ResolutionReason` Literal alias exported from `logfire.variables.composition`. `'code_default'` is now a valid value for both (also addressing the parent/child asymmetry from #discussion_r3220057944).

logfire/variables/variable.py
dmontagu: Addressed in db1f8e14 — `resolve_ref` now checks `_VARIABLE_OVERRIDES` first, then the provider, then the referenced variable's registered code default. So `@{x}@` inside a parent variable sees the same value that `x.get()` would. Added `test_override_propagates_through_composition` covering exactly the `x.override('OVERRIDDEN_X')` case you described, including asserting `composed_from[0].reason == 'context_override'`. The fully-shared `_resolve_serialized` helper is a cleaner refactor but I scoped this to the bug fix; happy to follow up.

logfire/variables/__init__.py
```diff
@@ -13,6 +13,11 @@
     VariableNotFoundError,
     VariableWriteError,
 )
+from logfire.variables.composition import (
+    ComposedReference,
+    VariableCompositionCycleError,
+    VariableCompositionError,
+)
```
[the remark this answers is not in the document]
dmontagu: Good catch — addressed in b2a2f90c. Added `ComposedReference`, `ResolutionReason`, `VariableCompositionError`, and `VariableCompositionCycleError` to `__all__` (and imported `ResolutionReason`, which was missing entirely). These are all part of the public surface — `ComposedReference` is the element type of `composed_from` and `ResolutionReason` is the type of `reason`.

docs/reference/advanced/managed-variables/index.md
```diff
@@ -112,6 +113,43 @@ With managed variables, you can iterate safely in production:
...
+
+
+prompt = logfire.template_var(
+    'system_prompt',
+    type=str,
+    default='Hello {{user_name}}!{{#if is_premium}} Welcome back, valued member.{{/if}}',
+    inputs_type=PromptInputs,
+)
+
+with prompt.get(PromptInputs(user_name='Alice', is_premium=True)) as resolved:
```
alexmojaki, line 147: This line raises LogfireNotConfiguredWarning in this snippet. We could maybe make it not do that if we think using vars before/without configuring logfire is reasonable, otherwise `logfire.configure()` should be added.

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,244 @@
...
+    'system_prompt',
+    type=str,
+    default='Hello {{user_name}}! Welcome to our service.',
+    inputs_type=PromptInputs,
+)
+```
+
+When you call `.get()`, you pass an instance of the inputs type. The SDK renders all `{{placeholder}}` expressions in the resolved value before returning:
+
+```python skip="true"
```
alexmojaki: make it one full snippet

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,244 @@
...
+1. **Resolve** — fetch the serialized value from the provider (or use the code default)
+2. **Compose** — expand any `@{variable_name}@` references (see [Composition](#variable-composition) below)
+3. **Render** — render `{{placeholder}}` Handlebars templates using the provided inputs
+4. **Deserialize** — validate and deserialize to the variable's type
+
+### Template Variables Parameters
+
+`logfire.template_var()` accepts the same parameters as `logfire.var()` plus:
+
+| Parameter | Description |
```
alexmojaki: this being a table seems weird

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,244 @@
...
+| `{{obj.nested}}` | Dot-notation access |
+| `{{#if field}}...{{/if}}` | Conditional block |
+| `{{#unless field}}...{{/unless}}` | Inverse conditional |
+| `{{#each items}}...{{/each}}` | Iterate over a list |
+| `{{#with obj}}...{{/with}}` | Change context |
+| `{{! comment }}` | Comment (not rendered) |
+
+**Example with conditionals:**
+
+```python skip="true"
```
alexmojaki: fold this into the first example

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,244 @@
...
+
+class PromptInputs(BaseModel):
+    user_name: str
+    is_premium: bool = False
+
+
+prompt = logfire.template_var(
+    'system_prompt',
+    type=str,
+    default='Hello {{user_name}}! Welcome to our service.',
```
alexmojaki: if i change this to `{{user_name2}}`, it just renders `Hello ! Welcome to our service.`. There's no warning or anything. I don't think that's the behaviour we settled on, especially the lack of warning.

alexmojaki: By contrast I get a warning in the composition case: RuntimeWarning: Variable 'agent_prompt' composition failed; falling back to code default: Referenced variable 'safetyd_rules' could not be resolved.

logfire/variables/variable.py
alexmojaki: > The fully-shared _resolve_serialized helper is a cleaner refactor but I scoped this to the bug fix; happy to follow up.

@dmontagu was that your choice or AI?

dmontagu: AI, I'll ask it to do the cleaner refactor :'(

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Done in e7ab01d7 — the `template_var` definition and the `with .get()` usage are now one runnable snippet (no `skip="true"`).

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Done in e7ab01d7 — replaced the one-row parameters table with prose.

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Done in e7ab01d7 — folded the conditional example (`{{#if is_premium}}`) into the first snippet, which now shows two `.get()` calls covering both branches.

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: The skip was unnecessary — composition doesn't require `pydantic-handlebars`. Removed in e7ab01d7; the example is now runnable via `test_docs`.

docs/reference/advanced/managed-variables/index.md
dmontagu: Added the missing `logfire.configure()` in e7ab01d7 so the snippet runs without the warning; `skip="true"` is gone.

logfire/variables/variable.py
dmontagu: Leaving as-is for this PR — the warn-and-fallback path emits the failure message regardless of which `VariableCompositionError` subclass is set on `result.exception`, so the subclass distinction is much less load-bearing now. If we want callers to branch on cycle-vs-depth-vs-other later, the fix is to propagate the original exception via `ComposedReference` rather than just `str(e)`, which touches the composition module's public surface and belongs in a follow-up.

logfire/variables/variable.py
dmontagu: Did the cleaner refactor in 85dcbadf: extracted `_BaseVariable._lookup_serialized`, which encodes the override → provider → registered-code-default priority once, and routed both `_resolve` (for `self.name`) and the composition expander's `resolve_ref` (for child `@{ref}@` lookups) through it. The two paths can no longer drift, and `'context_override'` now naturally appears on `ComposedReference.reason` for child refs.

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Tracking the follow-up in #1950. Your `worktree-poc-render-discovery` POC is referenced there as the starting point — the render-pass discovery approach (vs regex) is what would unlock full `@{...}@` syntax parity, and it overlaps with the render-time warning work that issue also covers.

docs/reference/advanced/managed-variables/templates-and-composition.md
dmontagu: Agreed the asymmetry vs the composition warning isn't right. Deferring the render-time strict-mode work to a follow-up so this PR can land — tracked in #1950, which captures the open design decisions (strict flag in `pydantic-handlebars`, default policy, fallback chain, misrender substitution behavior, push-/pull-time validation).

dmontagu:
Reopened with a cleaner two-commit history in #1951 (one commit for code, one for docs/demo). End state is byte-identical to this branch — verified by `git diff feature/variable-composition..feature/variable-composition-clean` being empty. This PR will be closed once #1951 is reviewed and merged.

Issue #923, opened 2025-03-12
Sending traceback with Django Ninja


Lodimup:
### Question

Is it possible to send traceback along with the log? I am using Django.

This is the log I want to send along with with the log
```
Simulate unhandled exception
Traceback (most recent call last):
  File "/workspaces/gateway/.venv/lib/python3.12/site-packages/ninja/operation.py", line 120, in run
    result = self.view_func(request, **values)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/gateway/app/appcore/routes/sample.py", line 135, in get_simulate_unhandled_exception
    raise Exception("Simulate unhandled exception")
Exception: Simulate unhandled exception
```

Lodimup:
Thank you! I couldn't get it to work with the current setup, will investigate more based on your suggestions.

alexmojaki:
Reopening because this should probably be incorporated into the SDK.

Also note that instead of `logfire.exception('Unhandled exception')` you could try:

```python
from opentelemetry.trace import get_current_span

get_current_span().record_exception(exc)
```

which should put the exception directly on the request span instead of creating a child log. Does that work better?

Br1an67:
I'd like to work on this. The approach patches `NinjaAPI.on_exception` to record exceptions on the current OpenTelemetry span before Django Ninja handles them, matching how the OTel Django middleware would capture exceptions in plain Django views.

PR: #1746

PR #1746, opened 2026-03-01
feat: add Django Ninja exception recording to instrument_django()

Br1an67:
## Summary

Integrates Django Ninja exception recording into `instrument_django()` via a new `instrument_ninja=True` parameter (enabled by default).

Django Ninja catches exceptions before they propagate to Django's middleware, preventing OpenTelemetry's Django instrumentation from recording them. This patch hooks `NinjaAPI.on_exception` at the **class level** to record exceptions on the current span.

Closes https://github.com/pydantic/logfire/issues/923

## Changes

- **`logfire/_internal/integrations/django.py`**: Added `_instrument_django_ninja()` which patches `NinjaAPI.on_exception` on the class. Silently skipped if `django-ninja` is not installed. Uses `_LOGFIRE_INSTRUMENTED` constant for double-instrumentation guard.
- **`logfire/_internal/main.py`**: Added `instrument_ninja: bool = True` parameter to `instrument_django()`. Removed separate `instrument_django_ninja()` method.
- **Type stubs**: Updated `instrument_django` signature with `instrument_ninja` parameter.
- **Tests**: 7 tests covering good route, handled errors, unhandled errors, double-instrumentation guard, and `head_sample_rate=0` branches.
- Removed `django-ninja` optional extra (users instrumenting Django Ninja are expected to have it installed).

## Key Design Decisions

- Patches at the **class level** so it works without arguments (for `logfire run`)
- Correctly sets `escaped=True` when Django Ninja re-raises unhandled exceptions, `escaped=False` when handled
- No separate API method — integrated into the existing `instrument_django()` flow

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,6 +111,7 @@ dev = [
     "fastapi != 0.124.3, != 0.124.4",
     "Flask >= 3.0.3",
     "django >= 4.2.16",
+    "django-ninja >= 1.0.0",
     "dirty-equals >= 0.8.0",
     "pytest >= 8.3.4",
     "pytest-django >= 4.6.0",
@@ -312,6 +313,8 @@ filterwarnings = [
     "ignore:Call to 'get_connection' function with deprecated usage of input argument/s.*:DeprecationWarning",
     # OpenTelemetry event logger stuff being used by Pydantic AI
     "ignore:You should use `\\w*(L|l)ogger\\w*` instead. Deprecated since version 1.39.0 and will be removed in a future release.:DeprecationWarning",
+    # django-ninja uses deprecated asyncio.iscoroutinefunction on Python 3.14+
+    "ignore:'asyncio.iscoroutinefunction' is deprecated.*:DeprecationWarning",
 ]
 DJANGO_SETTINGS_MODULE = "tests.otel_integrations.django_test_project.django_test_site.settings"
 
--- a/logfire-api/logfire_api/_internal/main.pyi
+++ b/logfire-api/logfire_api/_internal/main.pyi
@@ -698,7 +698,7 @@ class Logfire:
         Args:
             **kwargs: Additional keyword arguments to pass to the OpenTelemetry `instrument` method, for future compatibility.
         """
-    def instrument_django(self, capture_headers: bool = False, is_sql_commentor_enabled: bool | None = None, request_hook: Callable[[trace_api.Span, HttpRequest], None] | None = None, response_hook: Callable[[trace_api.Span, HttpRequest, HttpResponse], None] | None = None, excluded_urls: str | None = None, **kwargs: Any) -> None:
+    def instrument_django(self, capture_headers: bool = False, is_sql_commentor_enabled: bool | None = None, request_hook: Callable[[trace_api.Span, HttpRequest], None] | None = None, response_hook: Callable[[trace_api.Span, HttpRequest, HttpResponse], None] | None = None, excluded_urls: str | None = None, instrument_ninja: bool = True, **kwargs: Any) -> None:
         """Instrument `django` so that spans are automatically created for each web request.
 
         Uses the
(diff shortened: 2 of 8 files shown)


logfire/_internal/integrations/django_ninja.py
```diff
@@ -0,0 +1,53 @@
...
+    on the current span before Django Ninja handles them.
+
+    See the `Logfire.instrument_django_ninja` method for details.
+    """
+    if getattr(api.on_exception, _LOGFIRE_INSTRUMENTED, False):
+        return
+
+    from opentelemetry.trace import get_current_span
+
+    original_on_exception = api.on_exception
```
alexmojaki: 1. We need this to work without requiring passing arguments, for the `logfire run` command. If you think it's useful to be able to specify `api` instead of instrumenting everything globally, then that can be an option, but it's not essential. In this case, that means patching `NinjaAPI.on_exception` on the class.
2. Please make this part of `instrument_django` instead of adding another method and requiring users to call both.
3. Please mention this in the django docs.

logfire/_internal/integrations/django_ninja.py
```diff
@@ -0,0 +1,53 @@
...
+    which prevents OpenTelemetry's Django instrumentation from recording them.
+    This function patches the API's `on_exception` method to record exceptions
+    on the current span before Django Ninja handles them.
+
+    See the `Logfire.instrument_django_ninja` method for details.
+    """
+    if getattr(api.on_exception, _LOGFIRE_INSTRUMENTED, False):
+        return
+
+    from opentelemetry.trace import get_current_span
```
alexmojaki: move to top

logfire/_internal/integrations/django_ninja.py
```diff
@@ -0,0 +1,53 @@
...
+            response = original_on_exception(request, exc)
+        except Exception:
+            if span.is_recording():  # pragma: no branch
+                span.record_exception(exc, escaped=True)
+            raise
+        if span.is_recording():  # pragma: no branch
+            span.record_exception(exc, escaped=False)
+        return response
+
+    patched_on_exception._logfire_instrumented = True  # type: ignore[attr-defined]
```
alexmojaki: use constant for consistency

logfire/_internal/integrations/django_ninja.py
```diff
@@ -0,0 +1,53 @@
...
+    from opentelemetry.trace import get_current_span
+
+    original_on_exception = api.on_exception
+
+    def patched_on_exception(request: HttpRequest, exc: Exception) -> HttpResponse:
+        span = get_current_span()
+        try:
+            response = original_on_exception(request, exc)
+        except Exception:
+            if span.is_recording():  # pragma: no branch
```
alexmojaki: test these branches by setting the head sample rate to 0

pyproject.toml
```diff
@@ -66,6 +66,7 @@ aiohttp-client = ["opentelemetry-instrumentation-aiohttp-client >= 0.42b0"]
 aiohttp-server = ["opentelemetry-instrumentation-aiohttp-server >= 0.55b0"]
 celery = ["opentelemetry-instrumentation-celery >= 0.42b0"]
 django = ["opentelemetry-instrumentation-django >= 0.42b0", "opentelemetry-instrumentation-asgi >= 0.42b0"]
+django-ninja = ["django-ninja >= 1.0.0"]
```
alexmojaki: this extra isn't needed, we assume that users instrumenting a package have that package. same goes for catching the import error.

Br1an67:
Thanks for the detailed review! I've restructured the PR completely based on your feedback:

1. **No separate method** — ninja patching is now part of `instrument_django(instrument_ninja=True)` (enabled by default)
2. **Class-level patching** — patches `NinjaAPI.on_exception` on the class, works without arguments for `logfire run`
3. **Imports at top level** — `get_current_span` imported at module top
4. **`_LOGFIRE_INSTRUMENTED` constant** used for double-instrumentation guard
5. **head_sample_rate=0 tests** added to verify `is_recording()` branches
6. **Removed `django-ninja` extra** and ImportError catch — silently skips if not installed

Deleted the separate `instrument_django_ninja()` method, its API stubs, and public exports entirely.

PR #1755, opened 2026-03-05
sso docs

ChristopherGS:
Explain the SSO setup for Logfire enterprise cloud customers

### DIFF
--- a/docs/how-to-guides/sso-setup.md
+++ b/docs/how-to-guides/sso-setup.md
@@ -1,11 +1,11 @@
 ---
 title: "SSO Setup (Enterprise Cloud)"
-description: "Step-by-step guide to configure Single Sign-On (SSO) for Logfire Enterprise Cloud. Supports Okta, Azure Entra ID, Auth0, Google Workspace, and any OIDC-compatible provider."
+description: "Step-by-step guide to configure Single Sign-On (SSO) for Logfire Enterprise Cloud. Supports Okta, Microsoft Azure Entra ID, Keycloak, and any OIDC-compatible provider."
 ---
 
 # SSO Setup
 
-Logfire Enterprise Cloud supports Single Sign-On (SSO) via any [OIDC-compatible](https://openid.net/developers/how-connect-works/) identity provider, including Okta, Microsoft Azure Entra ID, Auth0, Google Workspace, AWS Cognito, Keycloak, and more. Under the hood, Logfire uses [Dex](https://github.com/dexidp/dex), an open-source OIDC gateway that integrates with virtually any IdP.
+Logfire Enterprise Cloud supports Single Sign-On (SSO) via [OIDC-compatible](https://openid.net/developers/how-connect-works/) identity providers, including Okta, Microsoft Azure Entra ID, and Keycloak. Under the hood, Logfire uses [Dex](https://github.com/dexidp/dex), an open-source OIDC gateway.
 
 This guide uses **Microsoft Azure Entra ID** as an example, but the general steps — registering an OIDC app, obtaining a Client ID, Client Secret, and Issuer URL, then connecting it in Logfire — apply to any supported provider.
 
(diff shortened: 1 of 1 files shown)


docs/how-to-guides/sso-setup.md
```diff
@@ -0,0 +1,142 @@
+---
+title: "SSO Setup (Enterprise Cloud)"
+description: "Step-by-step guide to configure Single Sign-On (SSO) for Logfire Enterprise Cloud. Supports Okta, Azure Entra ID, Auth0, Google Workspace, and any OIDC-compatible provider."
+---
+
+# SSO Setup
+
+Logfire Enterprise Cloud supports Single Sign-On (SSO) via any [OIDC-compatible](https://openid.net/developers/how-connect-works/) identity provider, including Okta, Microsoft Azure Entra ID, Auth0, Google Workspace, AWS Cognito, Keycloak, and more. Under the hood, Logfire uses [Dex](https://github.com/dexidp/dex), an open-source OIDC gateway that integrates with virtually any IdP.
```
ddanielcruzz: I'm not sure about this 
> Auth0, Google Workspace, AWS Cognito,

In the UI we have only Okta, Azure and Keycloak

Image

I remember Azure config require something special so that's why we had to signal it through the UI, but I could be wrong, do you remember @bruno-espino

bruno-espino: Yes it required some override on the dex settings if I'm not mistaken.

The self-hosted version supports more providers becuase you can configure Dex however you want

docs/how-to-guides/sso-setup.md
```diff
@@ -0,0 +1,142 @@
...
+3. Click **Continue with Entra ID** and verify you can log in successfully with your corporate credentials.
+
+---
+
+## Step 8: Invite Team Members
+
+1. Go to your Enterprise Cloud organization in Logfire.
+2. Navigate to **Settings** → **Invite Members**.
+3. Create an invite link (set it to never expire for convenience if you plan to share it in internal documentation).
+4. Share both the **invite link** and the **SSO login URL** with your team.
```
ddanielcruzz: I made it so the invite link for enterprise cloud redirects to their sso login url if the user is not authenticated

docs/how-to-guides/sso-setup.md
```diff
@@ -0,0 +1,142 @@
...
+
+---
+
+## Managing Existing Authentication Providers
+
+During the transition, existing login methods (e.g., Google, GitHub) remain active, so current users are not disrupted.
```
ddanielcruzz, line 112: I think here it's missing steps on how to link accounts of users  who signed up and joined the org when the OIDC provider was not setup. 

If I'm not mistaken they have to go the org after the OIDC is enabled and in account connections, connect their account. I think @bruno-espino fixed this

docs/how-to-guides/sso-setup.md
```diff
@@ -0,0 +1,142 @@
...
+
+Once your team has successfully migrated to Entra ID SSO:
+
+- You can **disconnect** individual login methods from **Organization Settings** → **Identity Providers**.
+- Advise team members to use the SSO login URL going forward. If other providers are still enabled, users may inadvertently log in with their personal accounts instead.
+
+!!! warning "Existing Users and Email Addresses"
+    Users who previously signed up with a different email (e.g., a personal Gmail) will appear with that email in Logfire, even after connecting Entra ID. To have team members appear under their corporate email addresses:
+
+    1. Have the user **join via the invite link** using the SSO login URL — this creates a new Logfire account linked to their corporate identity.
```
ddanielcruzz: Users can now update their email address if they go to `<url>/settings/emails`
Screenshot 2026-03-05 at 12 27 50 p m

docs/how-to-guides/sso-setup.md
ddanielcruzz: @ChristopherGS if we want this to be true (support many providers) we need to remove the type from the dropdown. I created a Linear issue with more context https://linear.app/pydantic/issue/PYD-3345/replace-oidc-provider-type-dropdown-with-granular-dex-flag-controls

Let me know if you want me to work on this.

docs/how-to-guides/sso-setup.md
bruno-espino: That's right.

If a user has logged in with Github, and now his org sets up Azure, before logging it with Azure directly he needs to:
- Login with Github
- Go to the org settings
- Connect with the new Azure provider at the Account connections tab

Then he will be able to login with both Github and Azure.

PR #1784, opened 2026-03-19
Remove Rich link styling from project URL output

### LINKED ISSUE #1267: Google Colab/IPython cell output conflicts on logfire import
### Problem Statement
We have previously run some python scripts via Google Colab notebooks, usually using the shell command `!uv run my_script.py` . After instrumenting the codebase with logfire, there is some unexpected behavior blocking cell output.

Logfire configures properly within the script and the script itself executes perfectly. Spans and instrumented functions, auto-instrumented objects, etc are all properly logged and sent to the dashboard. However, the cell output shows `Logfire Project URL: ____ (blank)` in the output, rendered as expecting input but no change in behavior upon typing anything in. After this appears, everything further is blocked from appearing in the cell output, including print statements, Python logging records, or Logfire logging records (manual and instrumented).

Behavior of the same script is normal locally and in a container on a production server.

### Investigation
- Cell/script output is normal prior to `import logfire` and importing is what initiates the erroneous behavior. 
- Write token was used to configure, but behavior is the same whether supplied via env variable or logfire.configure(). The blank project URL printout appears prior to configuration regardless, but the write token is working fine as the script executes.
- When `!logfire auth` is called, logfire logging statements, print statements, etc behave as expected when manually typed into notebook cells.
- When `!logfire auth` is called and then the script is run via shell command, the script file itself behaves properly, but the first imported file that contains a logfire import triggers the erroneous behavior
- As mentioned, behavior with the same script is normal locally and on a production server, as well as when run via the bash terminal supplied with the colab notebook.
- Tried all documented logfire.configure() settings and environment variables, which seem not to do anything because the problem occurs on logfire import
- Tried some Colab and IPython settings to reduce rendering of outputs, no luck there

### Workarounds

1. `uv run my_script.py &> output.txt` writes the expected output to the file
2. Executing using the `%%bash` or `%%sh` cell magics prints the expected output, albeit dumped at the end of execution rather than printed real-time

### Expected Behavior
1. Importing logfire and using logfire-instrumented code does not cause any erroneous behavior that blocks real-time print statements, logging records, or logfire console outputs. If need be, environment variables, or config/credential files are supplied prior to script execution.
2. Alternatively, a flag or environment variable is set that disables interactive outputs (and raises exceptions if disabled when CLI commands are called that rely on them)

Thanks! Enjoying logfire so far in general and looking forward to the future development

```TOML
Behavior exhibited on 3.21.1 and 3.25.0
```

ameenalkhaldi:
## Summary

Remove the Rich `[link=...]` markup from the project URL console output, replacing it with plain `[cyan]` styling.

## Problem

Rich's `[link=...]` markup emits OSC 8 terminal hyperlink escape sequences. These escape sequences break output in Google Colab and IPython environments — particularly when the project URL is printed from a background thread during `logfire.configure()`, the escape sequences interfere with Colab's output handling and can swallow subsequent `print()` output entirely.

See the detailed investigation in #1267.

## Fix

- Replace `[link={url} cyan]{url}[/link]` with `[cyan]{url}[/cyan]` in `print_token_summary()`
- Remove the now-unnecessary custom `Theme` and `Style` imports from `_print_summary()` (they were only used to customize the link color)

The URL remains visible and cyan-colored in terminals that support Rich markup, but no longer emits hyperlink escape sequences.

## Test plan

- [x] All existing tests in `test_configure.py` and `test_cli.py` pass (162/162) — they assert on the rendered text via `capsys`, which is unchanged
- [x] Full non-Docker test suite passes (1262 passed)
- [x] `ruff format`, `ruff check`, and `pyright` all pass
- [x] All pre-commit hooks pass

Fixes #1267

---
## Summary by cubic
Remove Rich hyperlink markup from the project URL output. Use plain cyan text to avoid OSC 8 sequences that break Google Colab and IPython output.

- **Bug Fixes**
  - Use `[cyan]{url}[/cyan]` instead of `[link={url} cyan]{url}[/link]` in `print_token_summary()`.
  - Remove custom `Theme`/`Style`; use default `Console`.
  - URL stays visible and colored; notebook output is no longer swallowed.

<sup>Written for commit 17be3c864b4867f8cf41c3d486ff430f81ed8551. Summary will update on new commits.</sup>

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -1843,19 +1843,15 @@ class LogfireCredentials:
         """Print a summary of the existing project."""
         if self.project_url:  # pragma: no branch
             _print_summary(
-                f'[bold]Logfire[/bold] project URL: [link={self.project_url} cyan]{self.project_url}[/link]',
+                f'[bold]Logfire[/bold] project URL: [cyan]{self.project_url}[/cyan]',
                 min_content_width=len(self.project_url),
             )
 
 
 def _print_summary(message: str, min_content_width: int) -> None:
     from rich.console import Console
-    from rich.style import Style
-    from rich.theme import Theme
 
-    # customise the link color since the default `blue` is too dark for me to read.
-    custom_theme = Theme({'markdown.link_url': Style(color='cyan')})
-    console = Console(stderr=True, theme=custom_theme)
+    console = Console(stderr=True)
     if console.width < min_content_width + 4:  # pragma: no cover
         console.width = min_content_width + 4
     console.print(message)


alexmojaki:
Thanks!

PR #1788, opened 2026-03-21
fix: correct false import suggestions

### LINKED ISSUE #1296: logfire run always suggests requests, sqlite3, urllib
Understandable, because they're part of the standard library, but still confusing because people usually won't be using them. 

@Kludex Should we check if the script contains `import <package>` and only include them then? In multi-file projects, people are even less likely to use them.

```py
# script.py
print('Hello, world!')
```

```bash
uvx logfire run script.py
```

```

╭──────────────────────────────────────────────────────────────── Logfire Summary ────────────────────────────────────────────────────────────────╮
│                                                                                                                                                 │
│  Your instrumentation checklist:                                                                                                                │
│                                                                                                                                                 │
│  ☐ requests (need to install opentelemetry-instrumentation-requests)                                                                            │
│  ☐ sqlite3 (need to install opentelemetry-instrumentation-sqlite3)                                                                              │
│  ☐ urllib (need to install opentelemetry-instrumentation-urllib)                                                                                │
│                                                                                                                                                 │
│                                                                                                                                                 │
│  To install all recommended packages at once, run:                                                                                              │
│                                                                                                                                                 │
│  uv add opentelemetry-instrumentation-requests opentelemetry-instrumentation-sqlite3 opentelemetry-instrumentation-urllib                       │
│                                                                                                                                                 │
│  ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────  │
│                                                                                                                                                 │
│  To hide this summary box, use: logfire run --no-summary.                                                                                       │
│                                                                                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

GopalGB:
## Summary

Fixes #1296 — `logfire run` always suggested `requests`, `sqlite3`, and `urllib` for instrumentation, even for a simple `print('Hello, world!')` script.

**Root cause:**
- `sqlite3` and `urllib` were always recommended because they were in `STANDARD_LIBRARY_PACKAGES`, which bypassed the "is this package installed?" check
- `requests` was recommended because it's a transitive dependency of opentelemetry, so it appeared as "installed" even though the user's code doesn't use it

**Fix:**
- Added `find_script_imports()` which uses Python's `ast` module to extract top-level imports from the user's script/module
- Renamed `STANDARD_LIBRARY_PACKAGES` to `ALWAYS_AVAILABLE_PACKAGES` and added `requests` to it
- For packages in `ALWAYS_AVAILABLE_PACKAGES`, only recommend them if the user's code actually imports them
- When no script/module context is available (e.g. `logfire inspect`), behavior is unchanged — all available packages are still recommended

**Before:**
```
$ logfire run script.py  # script.py: print('Hello, world!')
☐ requests (need to install opentelemetry-instrumentation-requests)
☐ sqlite3 (need to install opentelemetry-instrumentation-sqlite3)
☐ urllib (need to install opentelemetry-instrumentation-urllib)
```

**After:**
```
$ logfire run script.py  # script.py: print('Hello, world!')
# No false suggestions
```

## Test plan

- [ ] Existing tests pass (backward compatible — `find_recommended_instrumentations_to_install` without `script_imports` param behaves identically)
- [ ] `logfire run script.py` with `print('Hello')` shows no suggestions for requests/sqlite3/urllib
- [ ] `logfire run script.py` with `import sqlite3` correctly suggests sqlite3 instrumentation
- [ ] `logfire inspect` still shows all available recommendations (no script context = no filtering)
- [ ] `logfire run -m module_name` correctly analyzes module imports

---
## Summary by cubic
Fixes #1296 by stopping false import suggestions in `logfire run`. We now analyze the script or module entrypoint imports, so `requests`, `sqlite3`, and `urllib` are only suggested when used.

- **Bug Fixes**
  - Added `find_script_imports()` (AST) to read imports from a script or `-m` module; prefers `__main__.py` over `__init__.py` and skips filtering if source can’t be read.
  - Replaced `STANDARD_LIBRARY_PACKAGES` with `ALWAYS_AVAILABLE_PACKAGES` (adds `requests`); only recommend these if actually imported by the user’s code.
  - Moved `sys.path.insert(0, os.getcwd())` before import analysis so local `-m` modules resolve correctly.
  - Threaded script context through: `collect_instrumentation_context()` and `find_recommended_instrumentations_to_install()` now accept `script_imports`; behavior unchanged when no script/module is provided.

<sup>Written for commit 3dcf344b318012f9feaa6d44c54bd27f29322a1e. Summary will update on new commits.</sup>

### DIFF
--- a/logfire-api/logfire_api/_internal/cli/run.pyi
+++ b/logfire-api/logfire_api/_internal/cli/run.pyi
@@ -6,7 +6,7 @@ from dataclasses import dataclass
 from rich.console import Console
 from rich.text import Text
 
-STANDARD_LIBRARY_PACKAGES: Incomplete
+ALWAYS_AVAILABLE_PACKAGES: Incomplete
 OTEL_INSTRUMENTATION_MAP: Incomplete
 
 @dataclass
@@ -27,13 +27,16 @@ def instrument_packages(installed_otel_packages: set[str], instrument_pkg_map: d
     Returns a list of packages that were successfully instrumented.
     """
 def instrument_package(import_name: str): ...
-def find_recommended_instrumentations_to_install(instrument_pkg_map: dict[str, str], installed_otel_pkgs: set[str], installed_pkgs: set[str]) -> set[tuple[str, str]]:
+def find_recommended_instrumentations_to_install(instrument_pkg_map: dict[str, str], installed_otel_pkgs: set[str], installed_pkgs: set[str], script_imports: set[str] | None = None) -> set[tuple[str, str]]:
     """Determine which OpenTelemetry instrumentation packages are recommended for installation.
 
     Args:
         instrument_pkg_map: Mapping of instrumentation package names to the packages they instrument.
         installed_otel_pkgs: Set of already installed instrumentation package names.
         installed_pkgs: Set of all installed package names.
+        script_imports: Set of top-level module names imported by the user's script/module.
+            Used to filter out false suggestions for packages that are always available
+            (stdlib or transitive deps) but not actually used.
 
     Returns:
         Set of tuples where each tuple is (instrumentation_package, target_package) that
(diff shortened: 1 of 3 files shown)


GopalGB:
Thanks for the thorough review @cubic-dev-ai! Both issues were valid — pushed fixes in 3dcf344:

**Fix 1 (P2: sys.path ordering):** Moved `sys.path.insert(0, os.getcwd())` before `collect_instrumentation_context()` so that `-m` local modules can be resolved by `find_spec` during import analysis.

**Fix 2 (P2: __init__.py vs __main__.py):** Added logic to prefer `__main__.py` over `__init__.py` when analyzing packages, since `runpy.run_module()` executes `__main__.py` at runtime.

Both changes preserve backward compatibility — `script_imports=None` still skips filtering for `logfire inspect`.

GopalGB:
Force-pushed: rebased onto latest `main` and re-authored commits under `GopalGB <67310594+GopalGB@users.noreply.github.com>` for consistent attribution.

The remaining `coverage` / `codecov/patch` failure is on lines in `logfire/_internal/cli/run.py` covering the new `-m` module-resolution branch and the `__main__.py` vs `__init__.py` selection. Adding tests for those paths requires running the CLI subprocess against fixture packages with various layouts; happy to add them if @cubic-dev-ai's P1 (CWD-prepend ordering) is the right design — would value @alexmojaki's input on whether the cwd insertion should be scoped via try/finally to just the `find_spec` analysis pass before I write the test fixtures.

alexmojaki:
Sorry, I don't want to hide the recommendations just because the packages aren't imported in the script. Added a comment to the issue: https://github.com/pydantic/logfire/issues/1296#issuecomment-4429924014

Issue #1297, opened 2025-08-05
logfire run suggests adding `opentelemetry-instrumentation-` packages instead of dependency groups


DouweM:
Instead of this:

```

╭──────────────────────────────────────────────────────────────── Logfire Summary ────────────────────────────────────────────────────────────────╮
│                                                                                                                                                 │
│  Your instrumentation checklist:                                                                                                                │
│                                                                                                                                                 │
│  ☐ requests (need to install opentelemetry-instrumentation-requests)                                                                            │
│  ☐ sqlite3 (need to install opentelemetry-instrumentation-sqlite3)                                                                              │
│  ☐ urllib (need to install opentelemetry-instrumentation-urllib)                                                                                │
│                                                                                                                                                 │
│                                                                                                                                                 │
│  To install all recommended packages at once, run:                                                                                              │
│                                                                                                                                                 │
│  uv add opentelemetry-instrumentation-requests opentelemetry-instrumentation-sqlite3 opentelemetry-instrumentation-urllib                       │
│                                                                                                                                                 │
│  ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────  │
│                                                                                                                                                 │
│  To hide this summary box, use: logfire run --no-summary.                                                                                       │
│                                                                                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I'd like to see `uv add logfire[requests,sqlite3,urllib]` for those packages we have dependency groups for. I think we should have dependency groups for all instrumentation packages we recommend.

If we detect we were run using `uvx logfire run` (can we?), it would be good to suggest re-running with `uvx --from 'logfire[requests,sqlite3,urllib]' logfire run`.

If we detect we're run with `uv run --with logfire logfire run` (can we?), we should suggest `uv run --with 'logfire[requests,sqlite3,urllib]' logfire run`.

If we can't detect exactly how we're run, as is likely, it may be worth listing that they should install the `logfire[requests,sqlite3,urllib]` package and then suggest all three methods.

RichardMarto:
Well, since the 'logfire run' one was closed, I guess I can take that one.

PR #1833, opened 2026-04-05
CLI: Use logfire[extras] for instrumentation suggestions and improve environment-aware tips

jalpatel11:
This PR improves the recommendations provided by `logfire run` when missing instrumentations are detected. Instead of suggesting individual OpenTelemetry instrumentation packages, the CLI now recommends installing the corresponding Logfire extras (e.g., `logfire[requests]` instead of `opentelemetry-instrumentation-requests`). This ensures users are using the dependencies Pydantic Logfire is tested against and simplifies dependency management.

I have also added logic to detect the execution context (such as `uvx` or `uv run --with`) and suggest the appropriate re-run commands. This makes it easier for users in ephemeral environments to get their scripts running with the necessary dependencies.

### Technical Implementation

1.  **OTel to Logfire Extra Mapping**: Added a mapping `OTEL_PACKAGE_TO_LOGFIRE_EXTRA` in `logfire/_internal/cli/run.py` to translate detected OTel packages into Logfire extras.
2.  **Environment Detection**:
    -   Used `os.environ.get('UV') == '1'` and `sys.executable` path analysis to detect if `logfire` is running as a `uv` tool (`uvx`) or via an ephemeral `uv run`.
    -   Updated `_full_install_command` to generate context-specific suggestions using `--from` and multiple `--with` flags for these environments.
3.  **Command Formatting**: Integrated `shlex.quote` for all generated commands to ensure that brackets in `logfire[...]` syntax are correctly escaped for shells like `zsh`.
4.  **urllib Support**: Added the `urllib` extra to `pyproject.toml` and updated the instrumentation maps in `run.py` to ensure it is recommended correctly as `logfire[urllib]`.
5.  **Multi-method Suggestions**: For non-ephemeral environments where the installation method is unknown, the CLI now suggests both `uv add` and `pip install`.

---

### Verification

All 91 CLI tests in `tests/test_cli.py` pass, including 5 new tests covering:
-   `uvx` and `uv run` heuristic detection.
-   Standalone OTel package handling with multiple `--with` flags.
-   Quoted extras syntax in generated output.
-   `logfire[urllib]` and `logfire[sqlite3]` mapping verification.


jalpatel11:
Closes #1297

dmontagu:
Closing in favour of #2082, which is the version of this under active review for #1297 and has worked through the review comments. Thanks for taking a run at it.

PR #1856, opened 2026-04-10
fix: replace assert with proper exception in query client

glitch-ux:
## Summary

- Replace `assert response.status_code == 200` with an explicit `if` check raising `QueryExecutionError` in `_BaseLogfireQueryClient.handle_response_errors`
- The `assert` is stripped when Python runs with `-O` (optimize flag), causing unexpected HTTP status codes (401, 403, 500, 503, etc.) to silently pass through instead of raising an error — downstream code then fails with a confusing `JSONDecodeError` or similar

## Test plan

- [ ] Existing VCR-based query client tests continue to pass (they exercise the 200 success path through `handle_response_errors`)
- [ ] Verify with `python -O -c "assert False"` that asserts are indeed stripped, confirming the bug exists
- [ ] Manually confirm that a non-200/400/422 response now raises `QueryExecutionError` with a clear message

### DIFF
--- a/logfire/experimental/query_client.py
+++ b/logfire/experimental/query_client.py
@@ -109,7 +109,7 @@ class _BaseLogfireQueryClient(Generic[T]):
         if response.status_code == 422:  # pragma: no cover
             raise QueryRequestError(response.json())
         if response.status_code != 200:
-            raise QueryExecutionError(f'Unexpected response status code: {response.status_code}')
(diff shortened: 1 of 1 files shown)


dmontagu:
Closing as superseded by #2166, which just landed the same assert-to-exception fix in `handle_response_errors` along with distinct error types and test coverage. Thanks for flagging the `python -O` hazard here.

PR #1862, opened 2026-04-14
Do row→column transformation client-side in query client

adriangb:
## Summary
- `logfire.experimental.query_client` now always requests `json_rows=true` from the backend for JSON queries, shifting the row→column buffering cost off the shared server and onto the caller.
- `query_json` (sync + async) now delegates to `query_json_rows` and runs the transformation locally via a new module-level `_rows_to_columns` helper that mirrors the server implementation.
- `query_json_rows` is unchanged in behavior; `query_csv` / `query_arrow` still hit the same URLs (`json_rows` is only injected for `application/json` accept).

## Test plan
- [x] `uv run pytest tests/test_query_client.py` (14/14 pass with the cassettes updated for the new URL/payload of the first interaction)
- [x] `uv run pyright logfire/experimental/query_client.py`
- [x] `uv run ruff check logfire/experimental/query_client.py`

### DIFF
--- a/logfire/experimental/query_client.py
+++ b/logfire/experimental/query_client.py
@@ -74,10 +74,7 @@ class RowQueryResults(TypedDict):
 
 def _rows_to_columns(result: RowQueryResults) -> QueryResults:
     """Convert a row-oriented JSON query result to a column-oriented one."""
-    columns_by_name: dict[str, ColumnData] = {
-        col['name']: {'name': col['name'], 'datatype': col['datatype'], 'nullable': col['nullable'], 'values': []}
-        for col in result['columns']
-    }
+    columns_by_name: dict[str, ColumnData] = {col['name']: {**col, 'values': []} for col in result['columns']}
     for row in result['rows']:
         for key, value in row.items():
             columns_by_name[key]['values'].append(value)
@@ -102,8 +99,11 @@ class _BaseLogfireQueryClient(Generic[T]):
         min_timestamp: datetime | None = None,
         max_timestamp: datetime | None = None,
         limit: int | None = None,
+        accept: str = 'application/json',
     ) -> dict[str, str]:
         params: dict[str, str] = {'sql': sql}
+        if accept == 'application/json':
+            params['json_rows'] = 'true'
         if limit is not None:
             params['limit'] = str(limit)
         if min_timestamp:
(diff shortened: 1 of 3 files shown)


logfire/experimental/query_client.py
```diff
@@ -385,9 +389,10 @@ async def _query(
         min_timestamp: datetime | None = None,
         max_timestamp: datetime | None = None,
         limit: int | None = None,
-        row_oriented: bool = False,
     ) -> Response:
-        params = self.build_query_params(sql, min_timestamp, max_timestamp, limit, row_oriented)
+        params = self.build_query_params(sql, min_timestamp, max_timestamp, limit)
+        if accept == 'application/json':
+            params['json_rows'] = 'true'
```
alexmojaki:
```suggestion
        params = self.build_query_params(sql, min_timestamp, max_timestamp, limit, accept)
```

and adjust build_query_params, to keep things nice and dry like before

logfire/experimental/query_client.py
```diff
@@ -72,6 +72,18 @@ class RowQueryResults(TypedDict):
     rows: list[dict[str, Any]]
 
 
+def _rows_to_columns(result: RowQueryResults) -> QueryResults:
+    """Convert a row-oriented JSON query result to a column-oriented one."""
+    columns_by_name: dict[str, ColumnData] = {
+        col['name']: {'name': col['name'], 'datatype': col['datatype'], 'nullable': col['nullable'], 'values': []}
```
alexmojaki:
```suggestion
        col['name']: {**col, 'values': []}
```

logfire/experimental/query_client.py
```diff
@@ -90,13 +99,13 @@ def build_query_params(
         min_timestamp: datetime | None = None,
         max_timestamp: datetime | None = None,
         limit: int | None = None,
-        row_oriented: bool = False,
+        accept: str = 'application/json',
```
alexmojaki: having this default seems wrong if json_rows=true is a problem when `accept` is not json. if it's not a problem then we could just set `json_rows=true` all the time and not check `accept` at all.

alexmojaki (approved):

adriangb:
@alexmojaki mind if we make a release after this? We should probably smoke test it too.

PR #1871, opened 2026-04-18
Add level param to @logfire.instrument()

### LINKED ISSUE #1452: Support for log level specification in the instrument decorator
### Description

It would be great if the logfire.instrument decorator supported the _level argument in the same way that logfire.span does.

ie the below should behave the same

`
import logfire

@logfire.instrument('Running foo', _level='debug')
def foo_1():
    logfire.debug('Inside foo 1')

def foo_2():
    with logfire.span('Running foo', _level='debug'):
        logfire.debug('Inside foo 2')

if __name__ == "__main__":
    foo_1()
    foo_2()
`

imp-joshi:
Closes #1452.

Logfire's span APIs (`logfire.span`, `logfire.debug`, `logfire.warn`, etc.) all let you record a span with a log level attached, which `min_level` then filters against. `@logfire.instrument()` was the only one that didn't. This PR closes that gap.

## What changed

You can now do:

```python
@logfire.instrument("my span", level="warn")
def my_func(): ...
```

and it behaves exactly like tagging a manual `with logfire.span("my span"):` block at warn level. The span gets the warn-level attributes, and if your `min_level` is above warn, the span is suppressed entirely.

The `level=None` default is completely untouched, no overhead on existing code.

<details>
<summary>Implementation note</summary>

`log_level_attributes(level)` runs once at decoration time and merges into the attribute dict. The `level_num < config.min_level` check runs at each call so it picks up any `logfire.configure()` that happens after decoration; if the level is below `min_level`, `open_span` returns a `NoopSpan` and skips the work.

</details>

## NoopSpan._span (pre-existing bug, fixed here)

When `open_span` returns a `NoopSpan` (now an intentional code path, not just an error case), `record_return=True` would crash trying to call `.is_recording()` on a lambda returned by `__getattr__`. Added `_span` as an explicit property returning `trace_api.INVALID_SPAN` so `set_user_attributes_on_raw_span` exits cleanly.

## Tests

- `test_instrument_with_level` - warn level, verifies `logfire.level_num: 13` on the span
- `test_instrument_level_filtered` - debug with `min_level="info"`, nothing exported
- `test_instrument_level_filtered_record_return` - same but with `record_return=True` (exercises the NoopSpan fix)
- `test_instrument_with_level_async` - async path
- `test_instrument_with_level_and_extract_args` - level attrs and function args coexist
- `test_instrument_level_filtered_extract_args` - filtering with `extract_args=True`
- `test_instrument_level_filtered_extract_args_iterable` - filtering with `extract_args=('x',)`

### DIFF
--- a/logfire-api/logfire_api/__init__.py
+++ b/logfire-api/logfire_api/__init__.py
@@ -131,7 +131,7 @@ except ImportError:
 
             def install_auto_tracing(self, *args, **kwargs) -> None: ...
 
-            def instrument(self, *args, _level=None, **kwargs):
+            def instrument(self, *args, **kwargs):
                 def decorator(func):
                     return func
 
@@ -322,5 +322,5 @@ except ImportError:
         def get_context(*args, **kwargs) -> dict[str, Any]:
             return {}
 
-        def attach_context(*args, **kwargs)-> ContextManager[None]:
+        def attach_context(*args, **kwargs) -> ContextManager[None]:
             return nullcontext()
--- a/logfire-api/logfire_api/__init__.py
+++ b/logfire-api/logfire_api/__init__.py
@@ -131,7 +131,7 @@ except ImportError:
 
             def install_auto_tracing(self, *args, **kwargs) -> None: ...
 
-            def instrument(self, *args, **kwargs):
+            def instrument(self, *args, _level=None, **kwargs):
                 def decorator(func):
                     return func
 
--- a/logfire/_internal/instrument.py
+++ b/logfire/_internal/instrument.py
@@ -58,7 +58,7 @@ def instrument(
     record_return: bool,
     allow_generator: bool,
     new_trace: bool,
-    _level: LevelName | int | None = None,
+    level: LevelName | int | None = None,
 ) -> Callable[[Callable[P, R]], Callable[P, R]]:
     from .main import set_user_attributes_on_raw_span
 
@@ -70,7 +70,7 @@ def instrument(
             )
 
         attributes = get_attributes(func, msg_template, tags)
-        open_span = get_open_span(logfire, attributes, span_name, extract_args, func, new_trace, _level=_level)
+        open_span = get_open_span(logfire, attributes, span_name, extract_args, func, new_trace, level=level)
 
         if inspect.isgeneratorfunction(func):
             if not allow_generator:
@@ -129,15 +129,15 @@ def get_open_span(
     extract_args: bool | Iterable[str],
     func: Callable[P, R],
     new_trace: bool,
-    _level: LevelName | int | None = None,
+    level: LevelName | int | None = None,
 ) -> Callable[P, AbstractContextManager[Any]]:
     final_span_name: str = span_name or attributes[ATTRIBUTES_MESSAGE_TEMPLATE_KEY]  # pyright: ignore[reportAssignmentType]
 
     level_num: int | None = None
-    if _level is not None:
-        _level_attrs = log_level_attributes(_level)
-        level_num = int(_level_attrs[ATTRIBUTES_LOG_LEVEL_NUM_KEY])
-        attributes = {**attributes, **_level_attrs}
+    if level is not None:
+        level_attrs = log_level_attributes(level)
+        level_num = int(level_attrs[ATTRIBUTES_LOG_LEVEL_NUM_KEY])
+        attributes = {**attributes, **level_attrs}
 
     def get_logfire():
         # This avoids having a `logfire` closure variable, which would make the instrumented
--- a/logfire/_internal/instrument.py
+++ b/logfire/_internal/instrument.py
@@ -131,6 +131,8 @@ def get_open_span(
     new_trace: bool,
     level: LevelName | int | None = None,
 ) -> Callable[P, AbstractContextManager[Any]]:
+    from .main import NoopSpan
+
     final_span_name: str = span_name or attributes[ATTRIBUTES_MESSAGE_TEMPLATE_KEY]  # pyright: ignore[reportAssignmentType]
 
     level_num: int | None = None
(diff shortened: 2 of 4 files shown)


logfire-api/logfire_api/__init__.py
```diff
@@ -131,7 +131,7 @@ def log_slow_async_callbacks(self, *args, **kwargs) -> None:  # pragma: no cover
 
             def install_auto_tracing(self, *args, **kwargs) -> None: ...
 
-            def instrument(self, *args, **kwargs):
+            def instrument(self, *args, _level=None, **kwargs):
```
alexmojaki:
```suggestion
            def instrument(self, *args, **kwargs):
```

logfire/_internal/main.py
```diff
@@ -608,6 +608,7 @@ def instrument(
         record_return: bool = False,
         allow_generator: bool = False,
         new_trace: bool = False,
+        _level: LevelName | int | None = None,
```
alexmojaki: just call it `level`, the underscore prefix is to avoid conflict with kwargs in `span`.

logfire/_internal/main.py
```diff
@@ -3083,6 +3097,11 @@ def message(self) -> str:  # pragma: no cover
     def message(self, message: str):
         pass
 
+    @property
+    def _span(self) -> Any:
+        # set_user_attributes_on_raw_span checks is_recording() first; INVALID_SPAN returns False
```
alexmojaki: This comment doesn't make sense to me because is_recording is defined below and returns false.

logfire/_internal/instrument.py
```diff
@@ -156,13 +170,21 @@ def extra_span_kwargs() -> dict[str, Any]:
 
     # This is the fast case for when there are no arguments to extract
     def open_span(*_: P.args, **__: P.kwargs):  # pyright: ignore[reportRedeclaration]
+        if level_num is not None and level_num < get_logfire().config.min_level:
```
alexmojaki, line 175: please precompute get_logfire().config.min_level

logfire/_internal/instrument.py
```diff
@@ -190,6 +212,10 @@ def open_span(*func_args: P.args, **func_kwargs: P.kwargs):
         if extract_args_final:  # check that there are still arguments to extract
 
             def open_span(*func_args: P.args, **func_kwargs: P.kwargs):
+                if level_num is not None and level_num < get_logfire().config.min_level:
+                    from .main import NoopSpan
```
alexmojaki: please import this at the top of get_open_span

logfire/_internal/instrument.py
alexmojaki: actually the whole open_span function can be precomputed and returned early when the level is below min_level

logfire/_internal/instrument.py
```diff
@@ -139,6 +155,13 @@ def get_logfire():
         def get_logfire():
             return logfire
 
+    if level_num is not None and level_num < get_logfire().config.min_level:
+
+        def open_span(*_: P.args, **__: P.kwargs):  # pyright: ignore[reportRedeclaration]
+            return NoopSpan()
+
+        return open_span
```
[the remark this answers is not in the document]
imp-joshi: Decoration-time evaluation is intentional here, per alexmojaki's review: he asked to [precompute `config.min_level`](https://github.com/pydantic/logfire/pull/1871#discussion_r3123790335) and [return `open_span` early when the level is below it](https://github.com/pydantic/logfire/pull/1871#discussion_r3123827247). The configure-before-decorate ordering is the expected usage pattern.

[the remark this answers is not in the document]
alexmojaki: sorry, devin is right, decorating before configuring should absolutely work as expected, my bad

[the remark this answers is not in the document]
imp-joshi: yeah, got a bit confused about this earlier too. went with the precompute route since you'd asked for it, but reverting it back now.

logfire-api/logfire_api/__init__.py
imp-joshi: fixed in [f9e9e2a](https://github.com/pydantic/logfire/pull/1871/commits/f9e9e2ad63b5616b7c65aecf77e7481498d56841)

logfire/_internal/main.py
imp-joshi: fixed in [f9e9e2a](https://github.com/pydantic/logfire/pull/1871/commits/f9e9e2ad63b5616b7c65aecf77e7481498d56841)

logfire/_internal/main.py
imp-joshi: dropped the comment in [449b8cf](https://github.com/pydantic/logfire/pull/1871/commits/449b8cf86b0c36c3b2b03a6449d1bb6839f7e0c4)

logfire/_internal/instrument.py
imp-joshi: first implemented in [449b8cf](https://github.com/pydantic/logfire/pull/1871/commits/449b8cf86b0c36c3b2b03a6449d1bb6839f7e0c4). Then reverted in the [e217c6d](https://github.com/pydantic/logfire/pull/1871/commits/e217c6d3cd1cc0d1d5ffb6f57fd6624fd685e44c) due to the concern raised by [devin](https://github.com/pydantic/logfire/pull/1871#discussion_r3138740002)

logfire/_internal/instrument.py
imp-joshi: done in [449b8cf](https://github.com/pydantic/logfire/pull/1871/commits/449b8cf86b0c36c3b2b03a6449d1bb6839f7e0c4)

alexmojaki:
Thanks!

PR #1879, opened 2026-04-23
Datasets SDK: upload dataset-level + report-level evaluators

dmontagu:
## Summary

The hosted Datasets API now persists `Dataset.evaluators` and `Dataset.report_evaluators` (companion platform PR https://github.com/pydantic/platform/pull/20399 adds the columns + routes), so the SDK no longer has to reject them. This PR plumbs both fields through `push_dataset`, `create_dataset` / `update_dataset`, and `get_dataset`.

- Drops `_validate_push_dataset` (and the test that relied on it).
- `create_dataset` / `update_dataset` (sync + async) accept new `evaluators` / `report_evaluators` kwargs and forward them to the `/v1/datasets` endpoints.
- `_build_push_dataset_kwargs` always sends serialized `evaluators` / `report_evaluators` (empty list if none) so a subsequent push that drops them locally also clears the hosted side. New unit test covers this; existing test that asserted rejection becomes one that asserts upload.
- `get_dataset` (sync + async) gains `custom_report_evaluator_types` and threads it into `Dataset.from_dict`, mirroring `Dataset.from_file`'s signature.
- Docs (`docs/evaluate/datasets/sdk.md`) replace the "not supported yet" callout with a description of the round-trip behavior plus the new `custom_report_evaluator_types` parameter.
- Stub for `logfire_api` regenerated for the affected file.

## Test plan
- [x] `make lint` passes.
- [x] `make typecheck` passes (0 errors).
- [x] `uv run pytest tests/test_datasets_client.py` — all 101 tests pass, including the new `test_push_dataset_uploads_dataset_level_evaluators` and updated `test_build_push_dataset_kwargs_minimal`.
- [ ] After the platform PR ships and the SDK is released: re-run `platform/src/demos/logfire_demo/datasets_sdk_demo.py` to confirm `push_dataset` round-trips both fields and `get_dataset(custom_report_evaluator_types=[...])` deserializes the report evaluators.

### DIFF
--- a/docs/evaluate/datasets/sdk.md
+++ b/docs/evaluate/datasets/sdk.md
@@ -144,8 +144,6 @@ with LogfireAPIClient(api_key='your-api-key') as client:
     )
     ```
 
-    Both case-level and dataset-level evaluators can also be edited from the Logfire UI; report-level evaluators are managed from the dataset edit page.
-
 ## Manual Dataset Management
 
 If you need lower-level control, the SDK still exposes `create_dataset(...)`, `add_cases(...)`, `update_dataset(...)`, and the other primitives directly.
--- a/docs/evaluate/datasets/sdk.md
+++ b/docs/evaluate/datasets/sdk.md
@@ -127,9 +127,24 @@ with LogfireAPIClient(api_key='your-api-key') as client:
 - it uploads all cases through the existing import/upsert API
 - it uses `on_case_conflict='update'` by default, so named cases are updated on repeat pushes
 
-!!! note "Dataset-level evaluators are not uploaded yet"
+!!! note "Round-tripping evaluators"
 
-    `push_dataset(...)` uploads case-level evaluators with their cases, but it currently rejects dataset-level `evaluators` and `report_evaluators` because hosted datasets do not store them yet. Case-level evaluators are also not yet surfaced in the Logfire UI, so they round-trip through `get_dataset(..., custom_evaluator_types=[...])` but won't show up when browsing cases in the web app. We're working on this!
+    `push_dataset(...)` uploads case-level evaluators with their cases, plus dataset-level `evaluators` and `report_evaluators` from the `Dataset` itself. Each push overwrites the hosted values, so removing an evaluator locally and re-pushing also clears it on the server.
+
+    To deserialize the hosted values back into typed instances, pass the same custom types you would to `Dataset.from_file(...)`:
+
+    ```python skip="true" skip-reason="external-connection"
+    dataset = client.get_dataset(
+        'qa-golden-set',
+        QuestionInput,
+        AnswerOutput,
+        CaseMetadata,
+        custom_evaluator_types=[MyEvaluator],
+        custom_report_evaluator_types=[MyReportEvaluator],
+    )
+    ```
+
+    Both case-level and dataset-level evaluators can also be edited from the Logfire UI; report-level evaluators are managed from the dataset edit page.
 
 ## Manual Dataset Management
 
--- a/logfire/experimental/api_client.py
+++ b/logfire/experimental/api_client.py
@@ -66,9 +66,9 @@ except ImportError as e:  # pragma: no cover
 
 if TYPE_CHECKING:
     from pydantic_evals import Case, Dataset
-    from pydantic_evals.evaluators import Evaluator
+    from pydantic_evals.evaluators import Evaluator, ReportEvaluator
 else:
-    Case = Dataset = Evaluator = Any  # type: ignore[assignment]
+    Case = Dataset = Evaluator = ReportEvaluator = Any  # type: ignore[assignment]
 
 DEFAULT_TIMEOUT = Timeout(30.0)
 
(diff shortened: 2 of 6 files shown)


logfire/experimental/api_client.py
```diff
@@ -461,6 +474,8 @@ def update_dataset(
         output_type: type[Any] | None = None,
         metadata_type: type[Any] | None = None,
         description: str | None = _UNSET,
+        evaluators: list[dict[str, Any]] | None = _UNSET,
+        report_evaluators: list[dict[str, Any]] | None = _UNSET,
```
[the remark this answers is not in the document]
dmontagu: Fixed in 3502af43 — aligned the docstring on "Pass None to clear" to match the neighboring `description` parameter. In practice both `None` and `[]` clear (the server normalizes), but only documenting one convention avoids the asymmetry.

logfire/experimental/api_client.py
```diff
@@ -410,6 +413,12 @@ def create_dataset(
             output_type: Type for expected outputs. JSON schema will be generated from this type.
             metadata_type: Type for case metadata. JSON schema will be generated from this type.
             description: Optional description of the dataset.
+            evaluators: Optional dataset-level evaluator specs in
+                `[{"name": "...", "arguments": {...}}]` format. Use
+                `_serialize_evaluators(dataset.evaluators)` to build from a
```
alexmojaki: we can't ask users to call a private function. is the idea that this function is supposed to be low level and closer to the raw API? it still accepts types, not just JSON values. if passing the dataset is somehow bad, what about the evaluators?

docs/evaluate/datasets/sdk.md
```diff
@@ -127,9 +127,24 @@ with LogfireAPIClient(api_key='your-api-key') as client:
 - it uploads all cases through the existing import/upsert API
 - it uses `on_case_conflict='update'` by default, so named cases are updated on repeat pushes
 
-!!! note "Dataset-level evaluators are not uploaded yet"
+!!! note "Round-tripping evaluators"
```
alexmojaki, line 130: this note seems to contain several things not related to its title

logfire/experimental/api_client.py
dmontagu: Good point — addressed in 17c29b01. `create_dataset` and `update_dataset` (sync + async) now accept `Sequence[Evaluator[Any, Any, Any]]` / `Sequence[ReportEvaluator[Any, Any, Any]]` directly and serialize internally via `_serialize_evaluators`, matching the pattern `update_case` already uses. The docstrings no longer reference any private symbols, and `Evaluator` / `ReportEvaluator` are TYPE_CHECKING-only so there's no runtime dependency on pydantic-evals.

docs/evaluate/datasets/sdk.md
dmontagu: Fair — fixed in 0425e753. Dropped the trailing UI-editing sentence so the note stays focused on round-trip behavior.

logfire/experimental/api_client.py
```diff
@@ -848,7 +875,11 @@ def get_dataset(
 
         Dataset, _ = _import_pydantic_evals()
         typed_dataset_cls: type[Dataset[InputsT, OutputT, MetadataT]] = Dataset[input_type, output_type, metadata_type]  # pyright: ignore[reportIndexIssue, reportUnknownVariableType]
-        return typed_dataset_cls.from_dict(data, custom_evaluator_types=list(custom_evaluator_types))  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+        return typed_dataset_cls.from_dict(  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+            data,
+            custom_evaluator_types=list(custom_evaluator_types),
+            custom_report_evaluator_types=list(custom_report_evaluator_types),
+        )
```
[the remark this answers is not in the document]
dmontagu: Good catch — Devin is right. I traced `custom_report_evaluator_types` back to pydantic-ai commit 8a3cd8783 (PR #4243 "Report-level evaluators & experiment-wide analyses"), first released as part of the v1.58.0 tag. The same PR added `Dataset.evaluators`, `Dataset.report_evaluators`, and the `ReportEvaluator` class, so every piece of this feature depends on 1.58.0+.

Bumped both the `datasets` extra and the dev group to `pydantic-evals>=1.58.0` in 3e825cce.

logfire/experimental/api_client.py
```diff
@@ -103,6 +106,51 @@ def _import_pydantic_evals() -> tuple[type, type]:
...
+
+    if data.get('report_evaluators'):
+        warnings.warn(
+            'Hosted dataset has report_evaluators but the installed pydantic-evals does not '
+            'support them. Upgrade to pydantic-evals>=1.58.0 to deserialize report-level evaluators. '
+            'Dropping the field for now.',
+            UserWarning,
+            stacklevel=4,
```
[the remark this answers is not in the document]
dmontagu: Good catch — fixed in 41aedad6. The stack is `_from_dict_compat -> get_dataset -> user`, so `stacklevel=3` lands on the user's call site.

logfire/experimental/api_client.py
```diff
@@ -410,6 +463,10 @@ def create_dataset(
             output_type: Type for expected outputs. JSON schema will be generated from this type.
             metadata_type: Type for case metadata. JSON schema will be generated from this type.
             description: Optional description of the dataset.
+            evaluators: Optional dataset-level evaluators. Instances are serialized
+                to the hosted `{"name": ..., "arguments": ...}` format.
+            report_evaluators: Optional report-level evaluators, serialized the
```
alexmojaki, line 468: docstrings don't need to mention this serialization

tests/test_datasets_client.py
```diff
@@ -312,6 +314,93 @@ def test_import_error(self):
...
+            def from_dict(
+                cls,
+                data: dict[str, Any],
+                custom_evaluator_types: Any = (),
+                custom_report_evaluator_types: Any = (),
+                default_name: str | None = None,
+            ) -> Any:
+                return ('called', data, list(custom_evaluator_types), list(custom_report_evaluator_types))
+
+        assert _from_dict_supports_report_evaluators(Modern) is True
```
alexmojaki, line 332: should test against the real pydantic-evals

PR #1886, opened 2026-04-28
perf: avoid full response model_dump in usage cost path

imp-joshi:
Closes #1844

`get_usage_attributes()` calls `response.model_dump()` on the full response before passing it to `genai_prices.extract_usage()`. For embedding responses this serializes all the embedding vectors even though `extract_usage` only reads two top-level fields (`model` and `usage` for OpenAI/Anthropic, `modelVersion` and `usageMetadata` for Google GenAI).

## What changed

The cost path now calls `model_dump(include=...)` with just those four fields. Falls back to plain `model_dump()` if `include=` is not supported (TypeError), keeping compatibility with custom response wrappers.

## Tests

- `test_model_dump_prefers_include_for_cost` — confirms `include=` is passed on the primary path
- `test_model_dump_include_fallback_to_plain_dump` — confirms the TypeError fallback still computes cost
- Cost path assertions are conditional on `GENAI_PRICES_AVAILABLE` so environments where the pricing backend cannot import (e.g. pydantic 2.4 matrix) do not produce false failures


alexmojaki:
The complication not mentioned in the issue is that this should only apply to embeddings responses so that it doesn't get in the way of https://github.com/pydantic/genai-prices/pull/290 in the future.

imp-joshi:
The latest CI failure looks unrelated to this PR. The changed usage tests passed, and the failure is a timeout in the docs/loguru runnable example. (see [line:258](https://github.com/pydantic/logfire/actions/runs/25061569571/job/73416895847)) 
The exact same ci test on python 3.12 ran successully in its ealier commit [d66bae2](https://github.com/pydantic/logfire/actions/runs/25060338375/job/73412211833)

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -39,7 +39,13 @@ def get_usage_attributes(
     try:
         from genai_prices import calc_price, extract_usage
 
-        response_data = response.model_dump()
+        if provider_id == 'openai' and api_flavor == 'embeddings':
+            try:
+                response_data = response.model_dump(include={'model', 'usage'})
+            except TypeError:
```
alexmojaki: does this actually happen with real objects?

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -39,7 +39,13 @@ def get_usage_attributes(
     try:
         from genai_prices import calc_price, extract_usage
 
-        response_data = response.model_dump()
+        if provider_id == 'openai' and api_flavor == 'embeddings':
```
alexmojaki: do we need to check the provider_id?

logfire/_internal/integrations/llm_providers/usage.py
imp-joshi: No, I don’t think this happens for the real OpenAI/Pydantic response objects. I added it defensively, but it isn’t justified for this path. I removed the fallback and the related test in [d752941](https://github.com/pydantic/logfire/pull/1886/commits/d752941b1cf2fd373d9335388826b201f38c4a67)

logfire/_internal/integrations/llm_providers/usage.py
imp-joshi: I added the `provider_id` check to make the optimization explicitly scoped to OpenAI embeddings, since the reduced dump assumes the `model`/`usage` response shape (and may not be safe for another provider if `api_flavor='embeddings'` is reused later)

In the current code, though, `api_flavor='embeddings'` only comes from the OpenAI path, so the check is redundant today. I can simplify it to `api_flavor == 'embeddings'`.

alexmojaki: i'd prefer that

alexmojaki:
thanks!

PR #1887, opened 2026-04-28
chore: Update dependencies (2026-04-28)

alexmojaki:
Automated dependency update.

PR #1888, opened 2026-04-28
Support new TaskSpanData and TurnSpanData from openai-agents 0.14

alexmojaki:
Stacked on top of #1887.

Adds message templates for the new \`TaskSpanData\` and \`TurnSpanData\` span types introduced in openai-agents 0.14, and flattens the nested \`data\` attribute they expose so the templates can use \`{name}\`, \`{turn}\`, and \`{agent_name}\` directly.


tests/conftest.py
alexmojaki: couldn't the error just be caught here

tests/otel_integrations/test_openai_agents.py
```diff
@@ -13,7 +13,11 @@
 import logfire
 from logfire._internal.exporters.test import TestExporter
 
-pytest.importorskip('agents', reason='openai-agents requires python 3.10', exc_type=ImportError)
+# Importing `agents.sandbox` rather than just `agents` so that DynamicCompactionPolicy gets
+# defined inside importorskip's `warnings.catch_warnings()` block. On pydantic <2.10 its
+# `model_info` field would otherwise emit a protected-namespace UserWarning that gets elevated
+# to an error by `filterwarnings=error`.
+pytest.importorskip('agents.sandbox', reason='openai-agents requires python 3.10', exc_type=ImportError)
```
alexmojaki: make these changes about warnings another separate PR from main

tests/conftest.py
alexmojaki: (Claude) Good call — moved to #1889 and used the simpler approach (just `except (ImportError, UserWarning):` in conftest, no `warnings.catch_warnings()` needed). `pytest.importorskip` in the test modules already silences warnings inside its own catch_warnings, so once conftest doesn't crash the rest takes care of itself.

tests/otel_integrations/test_openai_agents.py
alexmojaki: (Claude) Reverted from this PR in 73814f41 and 67f34b2a. New PR off main: #1889.

logfire/_internal/integrations/openai_agents.py
```diff
@@ -27,7 +27,7 @@
     TranscriptionSpanData,
 )
 from agents.models.openai_responses import OpenAIResponsesModel
-from agents.tracing import ResponseSpanData, response_span
+from agents.tracing import ResponseSpanData, TaskSpanData, TurnSpanData, response_span
```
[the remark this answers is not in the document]
alexmojaki: (Claude) Addressed in 39fc63d4 — bumped the dev constraint to `openai-agents[voice]>=0.14.0`. As you noted, the integration was already using `ResponseSpanData` and other late-added types, so the previous `>=0.9.0` floor was already misleading; this just makes it match what we actually require. The lock file pinned 0.14.3 and CI tested against that, so no behavior change.

PR #1894, opened 2026-05-01
adding .agents for library-skills

adtyavrdhn:


Kludex:
@alexmojaki is the expert

adtyavrdhn:
Skills remain the same as the original repo, only few linting changes to also test these in CI instead of ignoring them.

https://github.com/pydantic/skills/pull/11 -> Same diff is here for easy viewing

alexmojaki:
Thanks, I'm just assuming this works, I'm not checking the instructions now.

PR #1897, opened 2026-05-04
Add support for new `/v2/query` endoint

Viicos:
This switches the existing client to use the `/v2/query` endpoint, for all existing methods. A follow up PR can add support for NDJSON streaming.

I went over a couple solutions to make the switch to the new endpoint. Here are some requirements to keep in mind:
- Behavior should not silently change. For instance, we can't silently have a `min_timestamp` set when existing usage did not set any.
- Having existing usage of the client deprecated is fine, but disruption should ideally happen only once.

---

The first solution would be to introduce new methods, leaving the existing ones unchanged (still targeting `/v1/query`):
- The existing methods would still have to be marked as deprecated, and it is also unclear how to name the new ones (`query_json_v2()` isn't ideal, because if/when we remove the old ones, we end up with this `v2` naming in the method).
- The benefit is we can freely change the data structure to directly match the `/v2/query` result structure (e.g. no `_transform_fields_for_backwards_compatibility()`, etc).

The second solution is to adapt the existing methods, while still preserving backwards compatibility:
- Omitting `min_timestamp` is deprecated, and will be required in the future (internally, we use `2020-01-01` as a `min_timestamp`).
- The data structure is mapped to preserve backwards compatibility (`_map_v2_result()`). This remaps the `columns` field, and only renames the top level `data` -> `rows` (so no concerns of memory/CPU usage client side, we're not iterating/recreating a list of all the rows).
- Only concern: `/v2/query`'s default `limit` when from 100 to 1000, and we rely on the remote default here. I feel it is ok to potentially return more results; could this break existing usage? Anyone could be relying on the limit of 100 in some way?

Imo 2 is the best option, so this is the one I implemented.

---

For the query stream method, I'm wondering if a utility to filter (in a type safe way?) message types would be useful. Maybe having users filtering with `if message['type'] = ...` is enough.


adriangb:
Why not both?

1. Adapt the current methods. We can leave them around longer that way.
2. We add new methods / new clients that don't do (somewhat expensive) transformations.

Viicos:
> Adapt the current methods. We can leave them around longer that way.

Apart from `query_json()` (which we can also not deprecate and keep),  the existing ones I modified aren't meant to go away.

> We add new methods / new clients that don't do (somewhat expensive) transformations.

How would the new ones be named? As per the PR body, I'm not a fan of `v2` being in method names. Also the data transformation added in this PR will not be expensive, mostly because the data isn't deeply modified, just the top level key name is changed. In a major version, we could introduce a breaking change that renames the key names to match the `/v2/query` endpoint verbatim. It will be a one off migration for users, and I don't think it will cause any issues as the SDK is most likely _not_ being really used by other libraries which would need to maintain compatibility with both.

logfire/experimental/query_client.py
```diff
@@ -265,6 +374,34 @@ def _query(
...
+            sql=sql,
+            min_timestamp=min_timestamp,
+            max_timestamp=max_timestamp,
+            limit=limit,
+            params=params,
+            timezone=timezone,
+            deployment_environment=deployment_environment,
+            explain=explain,
+        )
+        response = self.client.post('/v2/query', headers={'accept': accept}, json=body)
```
Viicos, line 510: Self note: wrong base path, need to check where FF exposes it

docs/reference/api/query_client.md
alexmojaki, line 1: https://22a8df9b-pydantic-docs-preview.pydantic.workers.dev/logfire/api/query_client/#logfire.query_client.AsyncLogfireQueryClient.query_json_rows

These don't look good

Image

1. ReadTokenInfo and RowQueryResultsV2 return types don't link anywhere
2. `@async` decorator looking thing
3. Signature shows twice because of deprecated overload

docs/how-to-guides/query-api.md
```diff
@@ -6,7 +6,7 @@ description: "Leverage the Logfire web API to query data via SQL. Export logs &
 This API can be used to retrieve data for export, analysis, or integration with other tools, allowing you to leverage
 your data in a variety of ways.
 
-The API is available at `https://logfire-api.pydantic.dev/v1/query` and requires a **read token** for authentication.
+The API POST endpoint is available at `https://logfire-api.pydantic.dev/v2/query` and requires a **read token** for authentication.
```
alexmojaki:
```suggestion
The API is available at `https://logfire-api.pydantic.dev/v2/query` and requires a **read token** for authentication.
```

I'm not sure if we should still be documenting https://logfire-api.pydantic.dev/ instead of region-specific URLs.

docs/how-to-guides/query-api.md
```diff
@@ -82,26 +83,29 @@ Here's an example of how to use these clients:
...
 
         async with AsyncLogfireQueryClient(read_token='<your_read_token>') as client:
-            # Load data as JSON, in column-oriented format
-            json_cols = await client.query_json(sql=query)
-            print(json_cols)
-
-            # Load data as JSON, in row-oriented format
-            json_rows = await client.query_json_rows(sql=query)
+            # Load data as JSON
+            json_rows = await client.query_json_rows(sql=query,min_timestamp=min_timestamp)
```
alexmojaki:
```suggestion
            json_rows = await client.query_json_rows(sql=query, min_timestamp=min_timestamp)
```

pytest-examples should have failed on this not being formatted, something is wrong. Maybe the whole block being indented gets in the way?

docs/how-to-guides/query-api.md
```diff
@@ -329,11 +337,11 @@ FROM records
 LIMIT 1
 """
 
-# Prepare the query parameters for the GET request
-params = {'sql': query}
+# Prepare the body for the POST request
+body = {'sql': query}
```
alexmojaki: needs a min timestamp

docs/how-to-guides/query-api.md
```diff
@@ -346,18 +354,22 @@ else:
...
     - `application/vnd.apache.arrow.stream`: Returns the data in Apache Arrow format, suitable for high-performance data processing.
     - `text/csv`: Returns the data in CSV format, which is easy to use with many data tools.
     - If no `Accept` header is provided, the default response format is JSON.
-- **Query Parameters**:
+- **Body Parameters**:
     - **`sql`**: The SQL query to execute. This is the only required query parameter.
-    - **`min_timestamp`**: An optional ISO-format timestamp to filter records with `start_timestamp` greater than this value for the `records` table or `recorded_timestamp` greater than this value for the `metrics` table. The same filtering can also be done manually within the query itself.
+    - **`min_timestamp`**: An optional ISO-format timestamp to filter records with `start_timestamp` greater than this value for the `records` table
+      or `recorded_timestamp` greater than this value for the `metrics` table. The same filtering can also be done manually within the query itself.
+      If not provided, defaults to a day from now.
```
alexmojaki:
```suggestion
      If not provided, defaults to a day before the current time.
```

Why have we gone with a default instead of requiring it, and why is it defaulting to a day and not less? I think we want to bias towards users querying only as much time as they need.

docs/how-to-guides/query-api.md
```diff
@@ -346,18 +354,22 @@ else:
...
     - **`max_timestamp`**: Similar to `min_timestamp`, but serves as an upper bound for filtering `start_timestamp` in the `records` table or `recorded_timestamp` in the `metrics` table. The same filtering can also be done manually within the query itself.
-    - **`limit`**: An optional parameter to limit the number of rows returned by the query. If not specified, **the default limit is 500**. The maximum allowed value is 10,000.
-    - **`row_oriented`**: Only affects JSON responses. If set to `true`, the JSON response will be row-oriented; otherwise, it will be column-oriented.
+    - **`limit`**: An optional parameter to limit the number of rows returned by the query. If not specified, **the default limit is 1000**. The maximum allowed value is 10,000.
```
alexmojaki: isn't the default 100?

docs/how-to-guides/query-api.md
```diff
@@ -346,18 +354,22 @@ else:
...
+    - **`params`**: An optional object mapping prepared-statement placeholder names to their substitution values. For example, with the query `SELECT * FROM records WHERE service_name = $svc`, pass `{"svc": "'my_service'"}` as `params` (note the inner quotes — values are inlined verbatim into the SQL).
```
alexmojaki: > values are inlined verbatim into the SQL

There's no proper interpolation? So this doesn't help with things like SQL injection? What makes the feature useful? It seems like getting the quotes wrong is an easy footgun.

docs/reference/api/query_client.md
Viicos: I can looking into 2, for the rest this will be like this until I finish rewriting the proper API docs Starlight plugin (no ETA). I'll note that for 3, I think it was already like this with MkDocs.

logfire/experimental/query_client.py
```diff
@@ -100,30 +150,57 @@ def __init__(self, base_url: str, read_token: str, timeout: Timeout, client: typ
...
+            body['limit'] = limit
+
+        # /v2/query requires aware datetimes, assume UTC:
         if min_timestamp is not None:
-            params['min_timestamp'] = min_timestamp.isoformat()
+            if min_timestamp.tzinfo is None:
+                min_timestamp = min_timestamp.replace(tzinfo=UTC)
+            body['min_timestamp'] = min_timestamp.isoformat()
+        else:
+            body['min_timestamp'] = _MIN_DATETIME
```
alexmojaki, line 178: Is this the deprecated path only? i think a comment in code could help

logfire/experimental/query_client.py
```diff
@@ -176,92 +254,302 @@ def query_json(
...
     ) -> QueryResults:
         """Query Logfire data and return the results as a column-oriented dictionary."""
-        row_results = self.query_json_rows(
-            sql=sql,
-            min_timestamp=min_timestamp,
-            max_timestamp=max_timestamp,
-            limit=limit,
-        )
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', message='Using query_json_rows.*', category=DeprecationWarning)
```
alexmojaki: I'm OK with letting this warning bubble up

logfire/experimental/query_client.py
```diff
@@ -176,92 +254,302 @@ def query_json(
...
+    ) -> RowQueryResultsV2:
+        """Query Logfire data and return the results as a row-oriented dictionary.
+
+        Args:
+            sql: The SQL `SELECT` query to execute.
+            min_timestamp: The minimum timestamp to use when querying data. If the provided
+                [`datetime`][datetime.datetime] doesn't have a timezone set, it is assumed to
+                be UTC.
+
+                /// version-deprecated | v3.35.0
```
alexmojaki:
```suggestion
                /// version-deprecated | v4.35.0
```

Same for the others

docs/how-to-guides/query-api.md
Viicos: Yes, pytest-examples has a poor approach when it comes to detect code blocks. I would recommend migrating to https://docs.astral.sh/ruff/formatter/#markdown-code-formatting, but it is still in preview.

(Also this will run formatting as part of linting, and not tests, which isn't expected to me).

Viicos: It might also be the `skip-run="true"` that prevents linting it.

logfire/experimental/query_client.py
```diff
@@ -176,92 +254,302 @@ def query_json(
...
+                Not providing a `min_timestamp` is deprecated.
+                ///
+            max_timestamp: The maximum timestamp to use when querying data. If the provided
+                [`datetime`][datetime.datetime] doesn't have a timezone set, it is assumed to
+                be UTC.
+            limit: The maximum number of rows to query. This value takes priority over the
+                `LIMIT` clause in the `sql` query.
+            params: Parameters to be used for substitution of prepared statements. For instance,
```
alexmojaki: i'm iffy about this "prepared statements" wording, since there's no caching/reuse.

logfire/experimental/query_client.py
```diff
@@ -176,92 +254,302 @@ def query_json(
...
+                with the query `SELECT * FROM records WHERE service_name = $svc`, it is necessary
+                to provide `{'svc': "'my_service'"}` as `params`.
```
alexmojaki: if we stick with this, with the requirements about quotes, that should be repeated in the docs here.

logfire/experimental/query_client.py
```diff
@@ -176,92 +254,302 @@ def query_json(
...
             limit=limit,
+            params=params,
+            timezone=timezone,
+            environment=environment,
+            explain=False,  # Note: we can expose this in the future
         )
-        return response.json()
+        return _map_v2_result(response.json())
+
+    # Note: on the next major version, move the keyword-only marker after `sql`:
```
alexmojaki, line 339: if we want to actually change things in a major version then i think we'll need a stronger reminder than this

logfire/experimental/query_client.py
```diff
@@ -329,77 +648,252 @@ async def query_json_rows(
...
+        timezone: str | None = None,
+        environment: str | list[str] | None = None,
     ) -> Table:
         """Query Logfire data and return the results as a pyarrow Table.
 
         Note that pyarrow must be installed for this method to succeed.
 
         You can use `polars.from_arrow(result)` to convert the returned table to a polars DataFrame.
+
+        Args:
```
alexmojaki, line 688: If we want args documented on every method, and they're always the same, then i think they should be inserted into the docstring programmatically, not copied

logfire/experimental/query_client.py
```diff
@@ -329,77 +648,252 @@ async def query_json_rows(
...
             raise ImportError('pyarrow is required to use the query_arrow method') from e
 
-        response = await self._query(
+        if min_timestamp is None:  # pragma: no branch
+            warnings.warn(
+                'Using query_arrow() without a min_timestamp is deprecated',
+                DeprecationWarning,
+                stacklevel=2,
+            )
+        response = await self._query_v2(
```
alexmojaki: maybe emit the warning from _query_v2 to dedupe.

probably use warn_at_user_stacklevel.

logfire/db_api.py
```diff
@@ -228,7 +228,7 @@ def execute(self, operation: str, parameters: dict[str, Any] | Sequence[Any] | N
         if parameters is not None:
             sql = _substitute_params(operation, parameters)
 
-        result = self._connection.client.query_json_rows(
+        result = self._connection.client.query_json_rows(  # type: ignore[reportDeprecated]
```
alexmojaki: is omitting a min_timestamp here allowed despite the existing default?

tests/cassettes/test_query_client/test_query_body_params_sync.yaml
alexmojaki, line 1: please confirm that these cassettes were actually generated and not written by AI. the new ones are suspiciously shorter than the old ones and seem to have fewer response headers.

docs/reference/api/query_client.md
alexmojaki: for point 1, wouldn't adding those types to this file so that they get included in docs work?

docs/how-to-guides/query-api.md
alexmojaki: skip-run shouldn't be the problem.

formatting during linting sounds nice. but i think it's good to fail if a python code block is invalid python, looks like ruff wouldn't do that.

Viicos: Opened https://github.com/astral-sh/ruff/issues/25523.

logfire/experimental/query_client.py
Viicos: Do you want me to open an issue? A specific one or general/meta issue about the next major release?

alexmojaki: i think a meta issue would be good

logfire/experimental/query_client.py
Viicos: Inserting them programmatically will require doing this in unified-docs (annoying because decoupled from this repo; and having the ability to do so within this repo will only be possible once we fully migrate to Starlight). But more importantly, we will loose all IDE support if we do so, which I feel is pretty important.

alexmojaki: i just meant setting `__doc__`. ok if you want to keep IDE support. ideally a test would check that they match.

logfire/db_api.py
Viicos: I think similarly to deprecating `None` min_timestamp in the query method, we need to deprecate using the `connect()` function without a min_timestamp

tests/cassettes/test_query_client/test_query_body_params_sync.yaml
Viicos: They were generated, yes. I don't see any cassettes shorter than existing ones, and some have fewer headers possibly because we changed our nginx setup since the last recording.

logfire/experimental/query_client.py
Viicos: Ah I see. I can write a test using Griffe to extract the args documentation, but alternatively now that we have AI reviewing agents, I can add a comment stating they should be kept in sync, usually agents do a good job of catching these.

logfire/experimental/query_client.py
Viicos: https://github.com/pydantic/logfire/issues/1986

logfire/experimental/query_client.py
Viicos: This will require passing the method name to raise the appropriate warning message, is it ok?

alexmojaki: "Querying without a min_timestamp is deprecated" is fine

logfire/experimental/query_client.py
alexmojaki: what about just `assert "Args: ..." in method.__doc__`?

logfire/experimental/query_client.py
alexmojaki: thanks

tests/cassettes/test_query_client/test_info_invalid_schema_sync.yaml
alexmojaki, line 1: i don't understand how only these 2 lines could have changed

Viicos: This is due to some work Jiri did to expose everything through the frontend port iirc

logfire/experimental/query_client.py
Viicos: Yeah as long as we have the proper stacklevel it should be fine

logfire/db_api.py
alexmojaki: yes, and deprecating explicit `None`.

tests/cassettes/test_query_client/test_info_invalid_schema_sync.yaml
alexmojaki: but surely more of the file should have changed if it was generated? there's a date header

Viicos: Claude got really confused here, these tests rely on the backend somehow not returning the necessary JSON keys in the payload. I went with a pragmatic approach here and deleted the test for this, because it requires tweaking the `/info` endpoint on the local Logfire instance to not return the key just to test a simple `try..except`, so I added a `pragma: no cover` instead. Cassettes should now be in a good state.

logfire/experimental/query_client.py
Viicos: Done

logfire/db_api.py
```diff
@@ -141,12 +143,30 @@ def __init__(
...
         self.min_timestamp = min_timestamp
         self.max_timestamp = max_timestamp
         self.limit = limit
 
+    # -- min_timestamp -----------------------------------------------------
+
+    @property
+    def min_timestamp(self) -> datetime | None:
+        """Lower bound for `start_timestamp` filtering."""
+        return self._min_timestamp
```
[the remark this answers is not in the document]
Viicos: At some point we can make the sentinel public. I'm keeping things as is, we can do so when we switch to a PEP 661 sentinel once type checkers support it.

logfire/db_api.py
```diff
@@ -195,10 +215,27 @@ def __init__(self, connection: Connection) -> None:
...
 
+    # -- min_timestamp -----------------------------------------------------
+
+    @property
+    def min_timestamp(self) -> datetime | None:
+        """Per-cursor override for the lower `start_timestamp` bound (inherits from the connection when unset)."""
+        return self._min_timestamp
+
+    @min_timestamp.setter
+    def min_timestamp(self, value: datetime | None) -> None:
```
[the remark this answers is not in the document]
Viicos: It never did. We can do so later if we want to.

logfire/experimental/query_client.py
```diff
@@ -83,11 +91,47 @@ def _rows_to_columns(result: RowQueryResults) -> QueryResults:
...
+        return new_obj
+
+    elif isinstance(obj, list):
+        return [_transform_fields_for_backwards_compatibility(item) for item in obj]  # type: ignore
+
+    else:
+        return obj
+
+
+def _map_v2_result(obj: dict[str, Any]) -> RowQueryResults:
```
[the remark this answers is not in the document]
Viicos: Dumb bot.

docs/reference/api/query_client.md
Viicos: I documented the return type, but still [not working](https://d692608a-pydantic-docs-preview.pydantic.workers.dev/logfire/api/query_client/#logfire.query_client.LogfireQueryClient.query_json_rows)..

alexmojaki: weird, it works elsewhere. i wonder if it's because they live in the experimental package. maybe adding `__all__` to `logfire/query_client.py` might help?

anyway, the fields in those classes would also need docstrings to get something more useful than this

Image

docs/how-to-guides/query-api.md
```diff
@@ -291,10 +300,14 @@ cursor.min_timestamp = datetime.now(timezone.utc) - timedelta(days=14)
 cursor.execute('SELECT start_timestamp, message FROM records LIMIT 10')
 ```
 
+/// version-deprecated | v4.35.0
+Setting `min_timestamp` to `None` to `connect()` or on the cursor is deprecated.
```
alexmojaki:
```suggestion
Setting `min_timestamp` to `None` in `connect()` or on the cursor is deprecated.
```

alexmojaki:
LGTM, final AI pass.

@cubic-dev-ai review.

docs/reference/api/query_client.md
Viicos: I added docstrings for the types. 

> weird, it works elsewhere. i wonder if it's because they live in the experimental package. maybe adding `__all__` to `logfire/query_client.py` might help?

This may be it, although doing so may break existing imports, because it currently re-exports `*` so every non `_` prefix member of the experimental module is importable.

It should work as it is now, and must be fixed when rewriting the API docs extension.

logfire/db_api.py
```diff
@@ -195,10 +197,27 @@ def __init__(self, connection: Connection) -> None:
...
+        self._min_timestamp: datetime | None = _UNSET  # type: ignore[assignment]
         self.max_timestamp: datetime | None = _UNSET  # type: ignore[assignment]
         self.limit: int = _UNSET  # type: ignore[assignment]
 
+    # -- min_timestamp -----------------------------------------------------
+
+    @property
+    def min_timestamp(self) -> datetime | None:
+        """Per-cursor override for the lower `start_timestamp` bound (inherits from the connection when unset)."""
+        return self._min_timestamp
```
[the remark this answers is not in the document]
Viicos: Can't be done for backwards compat, I'll remove the docstring mention for now to avoid confusion.

docs/reference/api/query_client.md
alexmojaki: i don't think we need to worry about preserving `from ... import *`

adriangb:
🥳 🎆

PR #1898, opened 2026-05-04
chore: Update dependencies (2026-05-04)

alexmojaki:
## Summary
- Run `uv sync --upgrade` to bump all dependencies
- Fix typecheck errors from new pyright (Iterator→Generator on contextmanager) and new anthropic (`AnthropicBedrock` no longer publicly exported)
- Update test snapshots and use `IsStr()` for the new non-deterministic `gen_ai.conversation.id`

## Test plan
- [ ] CI passes


logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -6,6 +6,7 @@
 
 import anthropic
 import httpx
+from anthropic.lib.bedrock import AnthropicBedrock, AsyncAnthropicBedrock
```
alexmojaki, line 9: make this type checking only in case it differs between versions

alexmojaki: (Claude) Addressed in c08be822

alexmojaki: (Claude) Reverted in 24056610 — verified `from anthropic.lib.bedrock import AnthropicBedrock, AsyncAnthropicBedrock` works on every supported anthropic version (0.27.0, 0.40.0, 0.50.0, 0.96.0, 0.97.0), so a TYPE_CHECKING-only wrapper isn't needed and would just complicate the runtime path.

Issue #1476, opened 2025-10-08
Why "responses" endpoint are not visualized as "completions" endpoint with litellm wrapper


anotine10:
### Question

i have tried instantiating a litellm router or directly calls litellm "responses" endpoint.
logfire does not show the information as nice as it if were the "completion" endpoint.

Example with "responses" endpoint:
- no input/output token are extracted, no UI for message transcript.
<img width="1198" height="199" alt="Image" src="https://github.com/user-attachments/assets/6dd6ce98-04f3-4bbb-b871-2ae6330c7148" />

Good Example with completion endpoint:

<img width="1452" height="370" alt="Image" src="https://github.com/user-attachments/assets/053b914c-2655-4437-b997-a3ba027ddeb5" />

As a note, it works well if i use the "response" endpoint from Openai or AzureOPenai object (without using litellm)

alexmojaki:
Your screenshot shows a custom logfire log. Did you mean to show the `responses` span above?

Please share example code.

Chances are this just isn't implemented.

If you're using a litellm proxy, try talking to it with an instrumented openai client instead of the litellm sdk.

anotine10:
hi @alexmojaki , 

i paste here a simple snippet code to reproduce:
```
import logfire
from pydantic import BaseModel, Field
from typing import Any, Type
import re
from openai import pydantic_function_tool
from litellm import Router

logfire.instrument_openai()
logfire.instrument_litellm()

#### > utility
def get_openai_tool_config_from_pydantic(
    model: Type[BaseModel],
    name: str | None = None,
    description: str | None = None,
    use_responses_api: bool = False,
) -> dict[str, Any]:
    """Get the tool config for the OpenAI from a Pydantic model

    Args:
        model: The Pydantic model to create the config from
        name: The name of the tool
        description: The description of the tool
        use_responses_api: If True, use the Responses API instead of the Completions API

    OpenAI Responses API uses different JSON schema for the tool config than the Completions API.

    Returns:
        The tool config for the OpenAI
    """
    if not name:
        name = re.sub(r"(?<!^)(?=[A-Z])", "_", model.__name__).lower()

    if not description:
        description = (model.__doc__ or "").strip()

    tool_config = pydantic_function_tool(
        model=model, name=name, description=description
    )

    if use_responses_api:
        return get_upgraded_tool_definition(tool_config)
    else:
        return dict(tool_config)


def get_upgraded_tool_definition(legacy_tool_definition):
[...]

PR #1902, opened 2026-05-04
fix(openai): recognize Responses API input_text/input_image/input_file content parts

### LINKED ISSUE #1901: OpenAI Responses API content types `input_text`/`input_image`/`input_file` not recognized by `instrument_openai`
### Description

When using `logfire.instrument_openai()` with `client.responses.create()` (the OpenAI Responses API), every content part whose `type` is `input_text`, `input_image`, or `input_file` falls through to:

- `gen_ai.unknown` events on the legacy semconv path (`version=1`, the default)
- a generic `{**part, 'type': part_type}` dict on the latest semconv path (`version="latest"`)

Only the Chat Completions content types (`text` / `output_text` / `image_url` / `input_audio`) are currently recognized. This makes Logfire traces for any non-trivial Responses API call very noisy: image, file, and text parts all show as unknown even when they are perfectly valid content per the [official OpenAI docs](https://developers.openai.com/api/docs/guides/images-vision?format=base64-encoded).

### Reproduction

```python
import logfire
from openai import OpenAI

logfire.configure()
logfire.instrument_openai()

client = OpenAI()
client.responses.create(
    model="gpt-4.1",
    input=[
        {
            "role": "user",
            "content": [
                {"type": "input_text", "text": "What is in this image?"},
                {
                    "type": "input_image",
                    "image_url": "data:image/jpeg;base64,<...>",
                },
            ],
        }
    ],
)
```

Console output (simplified):

```
events=[
  {'event.name': 'gen_ai.unknown', 'role': 'unknown',
   'content': 'input_text\n\nSee JSON for details',
   'data': {'type': 'input_text', 'text': '...'}},
  {'event.name': 'gen_ai.unknown', 'role': 'unknown',
   'content': 'input_image\n\nSee JSON for details',
   'data': {'type': 'input_image', 'image_url': '...'}},
]
```

The same content parts are accepted by the API and produce the expected response — this is purely an instrumentation gap.

### Root cause

In `logfire/_internal/integrations/llm_providers/openai.py`:

**Legacy path** — `input_to_events`:

```python
for content_item in content:
    with contextlib.suppress(KeyError):
        if content_item['type'] == 'output_text':
            events.append({'event.name': event_name, 'content': content_item['text'], 'role': role})
            continue
    events.append(unknown_event(content_item))
```

Only `output_text` is recognized; everything else goes to `unknown_event`.

**Latest path** — `_convert_content_part`:

```python
if part_type in ('text', 'output_text'):
    return TextPart(type='text', content=part.get('text', ''))
elif part_type == 'image_url':
    url = part.get('image_url', {}).get('url', '')
    return UriPart(type='uri', uri=url, modality='image')
elif part_type == 'input_audio':
    return BlobPart(...)
else:
    return {**part, 'type': part_type}
```

Note that the recognized image type (`image_url`) follows the **Chat Completions** schema (`{"type": "image_url", "image_url": {"url": "..."}}`), which differs from the **Responses API** schema (`{"type": "input_image", "image_url": "<string>"}` — `image_url` is a flat string, not a nested object).

### Proposed fix

I'd propose adding three cases to `_convert_content_part`:

```python
elif part_type == 'input_text':
    return TextPart(type='text', content=part.get('text', ''))
elif part_type == 'input_image':
    return UriPart(type='uri', uri=part.get('image_url', ''), modality='image')
elif part_type == 'input_file':
    return BlobPart(type='blob', content=part.get('file_data', ''), modality='file')
```

…and the corresponding branches in `input_to_events` for the legacy path.

I'm happy to send the PR myself — just let me know if this is the right shape, or if you'd prefer a different design (e.g. waiting on the upstream `opentelemetry-python-contrib` work tracked in #1586).

### Related

- #1476 — visualization gap when using Responses API via LiteLLM (same root symptom, broader description)
- #1586 — broader migration to OTel semconv
- #1769 — recent refactor that introduced `_convert_content_part_or_parts` but did not add the Responses API types

Mai0313:
## Summary

Fixes #1901.

`logfire.instrument_openai()` currently produces `gen_ai.unknown` events for the standard Responses API content types — `input_text`, `input_image`, `input_file` — because `_convert_content_part` (latest semconv path) and `input_to_events` (legacy semconv path) only know about Chat Completions types (`text`, `output_text`, `image_url`, `input_audio`). Any non-trivial Responses API call (multi-part user message, image input, file input) ends up with most of its content as `gen_ai.unknown`, even though the payload is exactly what the [OpenAI docs](https://developers.openai.com/api/docs/guides/images-vision?format=base64-encoded) prescribe.

This PR teaches both code paths the three missing types.

### Mapping

| Responses API part | semconv mapping | Notes |
|---|---|---|
| `input_text` | `TextPart` (latest) / `gen_ai.{role}.message` (legacy) | Same shape as `output_text`, just on the input side. |
| `input_image` | `UriPart(modality='image')` | `image_url` here is a **flat string** (URL or data URI), distinct from Chat Completions where it's a nested `{url: ...}` dict. |
| `input_file` | `UriPart(modality='document')` when `file_url` or `file_data` is present | `file_id`-only inputs fall through to a generic dict, since there's no URI to point at and no obvious semconv mapping for an opaque ID. |

### Test plan

- [x] Added 4 unit tests covering the three new types in `_convert_content_part_or_parts` and the `input_text` branch in `input_to_events`.
- [x] Full `tests/otel_integrations/test_openai.py` passes (65 tests).
- [x] `tests/otel_integrations/test_openai_agents.py` and `tests/otel_integrations/test_litellm.py` (which share `input_to_events`) pass.
- [x] `ruff check` + `ruff format --check` clean.

I deliberately kept the legacy `input_to_events` change to just `input_text` (mirroring how `output_text` is the only recognized non-string content there) — adding image/file support to the legacy semconv path would require designing a new event shape for media, and that path is on its way out per #1586. Happy to extend if you'd prefer.

Related: #1476, #1586, #1769.

### DIFF
--- a/logfire/_internal/integrations/llm_providers/openai.py
+++ b/logfire/_internal/integrations/llm_providers/openai.py
@@ -282,17 +282,28 @@ def _convert_content_part(part: object) -> MessagePart:
 
     part = cast('dict[str, Any]', part)
     part_type = part.get('type', 'unknown')
-    if part_type in ('text', 'output_text'):
+    if part_type in ('text', 'output_text', 'input_text'):
         return TextPart(type='text', content=part.get('text', ''))
     elif part_type == 'image_url':  # pragma: no cover
         url = part.get('image_url', {}).get('url', '')
         return UriPart(type='uri', uri=url, modality='image')
+    elif part_type == 'input_image':
+        # Responses API: image_url is a flat string (URL or data URI),
+        # not the nested {url: ...} dict used by Chat Completions.
+        return UriPart(type='uri', uri=part.get('image_url', ''), modality='image')
     elif part_type == 'input_audio':  # pragma: no cover
         return BlobPart(
             type='blob',
             content=part.get('input_audio', {}).get('data', ''),
             modality='audio',
         )
+    elif part_type == 'input_file':
+        # Responses API input_file may carry file_url, file_data (data URI),
+        # or just file_id. Prefer URI-bearing fields; fall through for file_id.
+        uri = part.get('file_url') or part.get('file_data')
+        if uri:
+            return UriPart(type='uri', uri=uri, modality='document')
+        return {**part, 'type': part_type}
     else:  # pragma: no cover
         # Return as generic dict for unknown types
         return {**part, 'type': part_type}
@@ -728,10 +739,10 @@ def input_to_events(inp: dict[str, Any], tool_call_id_to_name: dict[str, str]):
             else:
                 for content_item in content:
                     with contextlib.suppress(KeyError):
-                        if content_item['type'] == 'output_text':
+                        if content_item['type'] in ('output_text', 'input_text'):
                             events.append({'event.name': event_name, 'content': content_item['text'], 'role': role})
                             continue
-                    events.append(unknown_event(content_item))  # pragma: no cover
+                    events.append(unknown_event(content_item))
         elif typ == 'function_call':
             tool_call_id_to_name[inp['call_id']] = inp['name']
             events.append(
--- a/tests/otel_integrations/test_openai.py
+++ b/tests/otel_integrations/test_openai.py
@@ -4319,6 +4319,95 @@ def test_convert_responses_inputs_no_inputs() -> None:
     assert (input_messages, system_instructions) == snapshot(([], [{'type': 'text', 'content': 'Be helpful'}]))
 
 
+def test_convert_responses_inputs_input_text() -> None:
+    """Responses API `input_text` content parts should map to TextPart."""
+    from logfire._internal.integrations.llm_providers.openai import convert_responses_inputs_to_semconv
+
+    inputs: list[dict[str, Any]] = [{'role': 'user', 'content': [{'type': 'input_text', 'text': 'Hello there'}]}]
+    input_messages, system_instructions = convert_responses_inputs_to_semconv(inputs, None)
+    assert (input_messages, system_instructions) == snapshot(
+        ([{'role': 'user', 'parts': [{'type': 'text', 'content': 'Hello there'}]}], [])
+    )
+
+
+def test_convert_responses_inputs_input_image() -> None:
+    """Responses API `input_image` content parts should map to UriPart with modality=image."""
+    from logfire._internal.integrations.llm_providers.openai import convert_responses_inputs_to_semconv
+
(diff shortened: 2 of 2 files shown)


logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -282,17 +282,28 @@ def _convert_content_part(part: object) -> MessagePart:
...
     part = cast('dict[str, Any]', part)
     part_type = part.get('type', 'unknown')
-    if part_type in ('text', 'output_text'):
+    if part_type in ('text', 'output_text', 'input_text'):
         return TextPart(type='text', content=part.get('text', ''))
     elif part_type == 'image_url':  # pragma: no cover
         url = part.get('image_url', {}).get('url', '')
         return UriPart(type='uri', uri=url, modality='image')
+    elif part_type == 'input_image':
+        # Responses API: image_url is a flat string (URL or data URI),
```
alexmojaki, line 291: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-input-messages.json specifies that blob parts with the actual bytes should be used instead of data URIs

logfire/_internal/integrations/llm_providers/openai.py
```diff
@@ -282,17 +282,28 @@ def _convert_content_part(part: object) -> MessagePart:
...
             content=part.get('input_audio', {}).get('data', ''),
             modality='audio',
         )
+    elif part_type == 'input_file':
+        # Responses API input_file may carry file_url, file_data (data URI),
+        # or just file_id. Prefer URI-bearing fields; fall through for file_id.
+        uri = part.get('file_url') or part.get('file_data')
+        if uri:
+            return UriPart(type='uri', uri=uri, modality='document')
+        return {**part, 'type': part_type}
```
alexmojaki, line 306: there's a FilePart in https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-input-messages.json, it should be added and used here

dmontagu:
Closing as already fixed: main now handles `input_text`, `input_image` and `input_file` in both the latest-semconv path and the legacy path (`logfire/_internal/integrations/llm_providers/openai.py`), and #1901 is closed. Thanks for the detailed write-up — it matched what landed.

PR #1904, opened 2026-05-04
Optionally emit configuration log after `logfire.configure()`

Kludex:
Adds a `Logfire configured` log emitted at the end of `logfire.configure()`. Off by default for now while we experiment, enabled by setting `advanced.emit_configuration_span`.

Also fixed `advanced.base_url` to always fall back to the environment variable when not set, particularly when `advanced` is otherwise set.


adriangb:
1. Why does this have to be a span instead of sending it out of band? Is it useful to query this data? Is this compatible with any other backend?
2. How do we correlate this with other telemetry from the same SDK instance?
3. How does this play with sampling? Do we force this one span to never be sampled? Presumably that's impossible if there's e.g. an otel collector doing head sampling.

dmontagu:
> 1. Why does this have to be a span instead of sending it out of band? Is it useful to query this data? Is this compatible with any other backend?

I believe it _is_ useful to query this data for a variety of reasons — debugging logfire configuration, knowing when services were started, etc. Obviously better if you could be confident it was never sampled out but I'd prefer to have these spans and have it "misbehave" (i.e., just sometimes be missing these spans) if you have sampling enabled.

Not sure why it would be incompatible with other backends if it uses otel; obviously sending data out-of-band using another API is incompatible with other backends. I feel like making it opt-out of this behavior (rather than opt-in) is reasonable, but making other out-of-band telemetry opt-out rather than opt-in seems harder to justify.

> 2. How do we correlate this with other telemetry from the same SDK instance?

We have the `service.instance.id` resource attribute that I would assume works for this?

> 3. How does this play with sampling? Do we force this one span to never be sampled? Presumably that's impossible if there's e.g. an otel collector doing head sampling.

Presumably plays poorly but I feel like it's better to have than not. We could make it possible to opt out but I'd suggest we have it opt-in by default.

adriangb:
If we go forward with this please lets cherry pick information from `_LogfireConfigData` explicitly, not make a copy of the dict and rely on redaction.

logfire/_internal/config.py
```diff
@@ -202,6 +202,15 @@ class AdvancedOptions:
     serialized configuration sent to child processes. See the [distributed tracing guide](https://logfire.pydantic.dev/docs/how-to-guides/distributed-tracing/#thread-and-pool-executors) for more details.
     """
 
+    emit_configuration_span: bool = False
```
alexmojaki: i think it needs a config param, then samuel can use an env var to enable it while keeping demo code looking clean

alexmojaki (changes requested):

logfire/_internal/config.py
```diff
@@ -395,6 +396,7 @@ def configure(
     code_source: CodeSource | None = None,
     variables: VariablesOptions | LocalVariablesOptions | None = None,
     distributed_tracing: bool | None = None,
+    emit_configuration_span: bool | None = None,
```
alexmojaki: needs to go under advanced options

logfire/_internal/config.py
```diff
@@ -1352,6 +1368,42 @@ def fix_pid():  # pragma: no cover
...
+        objects like span/log processors.
+        """
+        if not self.emit_configuration_span:
+            return
+
+        from logfire._internal.main import Logfire
+
+        with handle_internal_errors:
+            sampling = self.sampling
+            config_dict: dict[str, Any] = {
```
alexmojaki: `local` is worth including. also token count similar to adrian's telemetry header PR. and whether api_key/service_name/service_version/environment/etc is set at all.

logfire/_internal/config.py
```diff
@@ -1352,6 +1368,42 @@ def fix_pid():  # pragma: no cover
...
+                'inspect_arguments': self.inspect_arguments,
+                'min_level': self.min_level,
+                'add_baggage_to_attributes': self.add_baggage_to_attributes,
+                'distributed_tracing': self.distributed_tracing,
+                'head_sample_rate': sampling.head if isinstance(sampling.head, (int, float)) else None,
+                'tail_sampling_enabled': sampling.tail is not None,
+                'code_source_set': self.code_source is not None,
+                'variables_set': self.variables is not None,
+            }
+            Logfire(config=self).info(
```
alexmojaki: use the instance constructed in configure() instead of making one here

logfire/_internal/config.py
Kludex: I thought your previous comment meant the opposite.

alexmojaki: no i just meant it should have a ConfigParam, but it should still be under advanced. although i know that might be more complicated. having it under advanced is the main priority, the ConfigParam is a nice to have.

logfire/_internal/config.py
Kludex: I don't know what you mean, Claude added this stuff based on your comment.

Kludex: I would appreciate if you take this over.

alexmojaki (approved):

Kludex:
Thanks! 🙏

PR #1905, opened 2026-05-05
feat: add X-Logfire-Telemetry header

adriangb:
## Summary

Adds an `X-Logfire-Telemetry: {...}` header (compact JSON) to every Logfire SDK request (OTLP exports, token validation, CRUD client, variable provider, CLI, query client). The value is deliberately small; every field has a concrete product question it answers.

Linear: [PYD-3579](https://linear.app/pydantic/issue/PYD-3579/add-logfire-sdk-version-telemetry-to-all-clients)

> The response-side counterpart (`X-Logfire-Warning` / `X-Logfire-Error` handling) was originally part of this PR; per review it has been split out into #1906.

### What we send and why

**Always sent (`_base_telemetry_pairs`):**
| Field | Why |
|---|---|
| `sdk_version` | Primary signal for deprecation planning — which versions are still in active use, so we know when it is safe to drop one. |
| `sdk_language` | Lets backend ingestion distinguish python from future SDKs (JS, Rust) without parsing User-Agent. |
| `python_version` | Tells us when we can drop support for an older Python. |
| `implementation` | Spots non-CPython users (pypy, graalpy) before we change anything that depends on CPython-specific behaviour. Mirrors `sys.implementation.name`. |
| `os` | Confirms Windows / Linux / macOS coverage before touching platform-sensitive code. |

**Sent when a `LogfireConfig` is available (`_config_telemetry_pairs`):**
| Field | Why |
|---|---|
| `code_source_set` | Adoption signal for the `code_source=` option (newer feature) — informs whether the source-link UI is worth investing in. |
| `variables_set` | Adoption signal for the variables / feature-flag feature (newer feature) — informs whether to keep building on it. |
| `token_count` | How many users configure multiple write tokens in one SDK instance — drives auth/routing roadmap decisions. |
| `service_instance_id` | Mirrors the [`service.instance.id`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-instance-id) OTLP resource attribute (same UUID), letting the backend correlate header-only requests (token validation, variables fetch, CRUD) with the spans this SDK instance exports. |

**Explicitly excluded:** `token`, `api_key`, `service_name`, `environment`, `service_version`, `data_dir`, `code_source.repository|revision|root_path`, and any other free-form string the user supplies. There is also no `min_level`, `console_enabled`, etc.: nothing was added that didn't have a clear product use, and adding new fields is gated by the inline comment in `_config_telemetry_pairs`.

The value is encoded as compact JSON (`json.dumps(..., separators=(',', ':'))`) so future fields with arbitrary string values can't silently break parsing.

### Query client

#1875 added a custom User-Agent on the experimental query client to surface SDK version / Python version / OS to the backend. With this header in place that became duplicate plumbing, so it has been reverted as its own commit and the same `X-Logfire-Telemetry` header is now attached on query client construction instead.

### Test plan
- [x] New `tests/test_telemetry_header.py` covers header construction (with/without config), secret exclusion, OTLP export carrying the same `service.instance.id` as the resource attribute, `LogfireCredentials.from_token`, and `LogfireClient`.
- [x] `make format lint typecheck` clean.
- [x] `uv run pytest tests/test_telemetry_header.py tests/test_query_client.py tests/test_configure.py tests/test_client.py tests/test_variables.py tests/test_cli.py` all pass.

### DIFF
--- a/logfire/_internal/telemetry_header.py
+++ b/logfire/_internal/telemetry_header.py
@@ -1,47 +1,22 @@
 """Build the `X-Logfire-Telemetry` request header.
 
-The header carries non-sensitive information about the SDK and how it is
-configured, encoded as a compact JSON object. The backend uses it to answer
-questions like which SDK versions are still in active use, which Python
-versions we can drop, and which configuration options users actually enable.
+The header carries non-sensitive, config-derived signals about how this SDK
+instance is configured, encoded as a compact JSON object. SDK/runtime identity
+(version, language, Python version, OS, etc.) lives on the standard
+`User-Agent` header instead — see `UA_HEADER` in `_internal/client.py`.
 Secrets (`token`, `api_key`, `service_name`, etc.) are never included.
 """
 
 from __future__ import annotations
 
-import functools
 import json
-import sys
 from typing import TYPE_CHECKING, Any
-
-from logfire.version import VERSION
 
 if TYPE_CHECKING:
     from .config import LogfireConfig
 
 
 TELEMETRY_HEADER_NAME = 'X-Logfire-Telemetry'
-
-
-@functools.cache
-def _base_telemetry_pairs() -> dict[str, Any]:
-    # Each field below has an explicit rationale; do not add a field unless you have one.
-    return {
-        # SDK version: the primary signal for deprecation planning — which versions
-        # are still in active use so we know when it is safe to drop one.
-        'sdk_version': VERSION,
-        # SDK language: lets the same backend ingestion logic distinguish python
-        # from future SDKs (JS, Rust) without having to parse User-Agent.
-        'sdk_language': 'python',
-        # Python version: tells us when we can drop support for an older Python.
-        'python_version': f'{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}',
-        # Implementation: spotting non-CPython users (pypy, graalpy) before
-        # changing anything that depends on CPython-specific behaviour.
-        'implementation': sys.implementation.name,
-        # OS: same idea — confirm Windows / Linux / macOS coverage before
-        # touching platform-sensitive code paths.
-        'os': sys.platform,
-    }
 
 
 def _config_telemetry_pairs(config: LogfireConfig) -> dict[str, Any]:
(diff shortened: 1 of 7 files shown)


alexmojaki:
service_instance_id belongs in the second table under "Sent when a LogfireConfig is available (_config_telemetry_pairs)"

adriangb:
Done

logfire/_internal/telemetry_header.py
```diff
@@ -0,0 +1,134 @@
...
+        # are still in active use so we know when it is safe to drop one.
+        'sdk_version': VERSION,
+        # SDK language: lets the same backend ingestion logic distinguish python
+        # from future SDKs (JS, Rust) without having to parse User-Agent.
+        'sdk_language': 'python',
+        # Python version: tells us when we can drop support for an older Python.
+        'python_version': platform.python_version(),
+        # Runtime: spotting non-CPython users (pypy, graalpy) before changing
+        # anything that depends on CPython-specific behaviour.
+        'runtime': sys.implementation.name,
```
Viicos: I feel like `implementation` better matches the Python documentation wording:

```suggestion
        'implementation': sys.implementation.name,
```

logfire/_internal/telemetry_header.py
```diff
@@ -0,0 +1,134 @@
...
+ERROR_HEADER_NAME = 'X-Logfire-Error'
+
+
+def _format_value(value: object) -> str:
+    if isinstance(value, bool):
+        return 'true' if value else 'false'
+    return str(value)
+
+
+def _base_telemetry_pairs() -> dict[str, str]:
```
Viicos: I think it doesn't hurt to:


```suggestion

@functools.cache
def _base_telemetry_pairs() -> dict[str, str]:
```

Viicos (commented):
I think it would make sense to revert https://github.com/pydantic/logfire/pull/1875 for the query client (let it use the httpx UA), and add the same telemetry header?

Regarding the format, it might be a footgun for the future to use comma-separated; what about using a JSON object encoded as a string?

adriangb:
> Regarding the format, it might be a footgun for the future to use comma-separated; what about using a JSON object encoded as a string?

No strong feelings on my end but why is json better than comma seperated / query param format?

alexmojaki:
why not have everything in _base_telemetry_pairs in the user-agent? we already put the SDK version in there in at least some places

logfire/_internal/telemetry_header.py
```diff
@@ -0,0 +1,129 @@
...
+    warning_message = response.headers.get(WARNING_HEADER_NAME)
+    if warning_message:
+        warnings.warn(warning_message, LogfireServerWarning, stacklevel=2)
+    error_message = response.headers.get(ERROR_HEADER_NAME)
+    if error_message:
+        raise LogfireServerError(error_message)
+    return response
+
+
+def install_logfire_response_hook(session: requests.Session) -> None:
```
alexmojaki: this seems like it should be a separate PR

Viicos:
> No strong feelings on my end but why is json better than comma seperated / query param format?

Your PR body mentioned any value with a comma in it would break parsing

adriangb:
yes but you can still have commas inside of a string inside of json...

adriangb:
@Viicos @alexmojaki I moved the response handling into #1906. I think #1906 stands on it's own and we should move forward with that (in some way).

This PR I leave up to @alexmojaki @Kludex @dmontagu to decide what to do with. I agree with @Viicos that we should consider the overlap with https://github.com/pydantic/logfire/pull/1875.

Maybe we should do this and https://github.com/pydantic/logfire/pull/1904, they're slightly orthogonal. Certainly for our Terraform client and whatnot we cannot send a span so it will have to be headers/user agent.

Viicos:
> yes but you can still have commas inside of a string inside of json...

I might be missing something, but this would not be an issue? The initial proposed format was:

`Logfire-Telemetry: key=val,...` so if `val` had a comma inside it, it would be ambiguous. Now with JSON:

`Logfire-Telemetry: {"key": "val with a , inside", "other_key": "..."}` it removes the ambiguity.

logfire/experimental/query_client.py
```diff
@@ -87,7 +86,6 @@ def _rows_to_columns(result: RowQueryResults) -> QueryResults:
 
 
 _ACCEPT = Literal['application/json', 'application/vnd.apache.arrow.stream', 'text/csv']
-_USER_AGENT = f'logfire-sdk-python/{VERSION} (Python {platform.python_version()}, os {platform.platform()}, arch {platform.machine()})'
```
alexmojaki, line 90: Why remove the user agent? Why not do the opposite and put everything in there?

adriangb: I'm happy to keep the user agent. But does token count (a proxy for project count) which @samuelcolvin requested really belong in the user agent header? I also fear that things (proxies, etc.) expect it to be ~ low cardinality and putting something like the service instance id (which I think is important to be able to correlate what features are used together and segment users) in there might break something that makes that assumption.

alexmojaki: how about using the user agent everywhere for _base_telemetry_pairs?

adriangb: I think we should use the user agent for things that have some semblence of belonging there. But high cardinality things (service instance id) or feature flags / feature usage IMO do not belong there. I spoke with @Viicos today about this and we're on the same page in that regard.

I am pushing a commit to restore User-Agent

PR #1906, opened 2026-05-05
feat: surface X-Logfire-Warning header

adriangb:
Gives the Logfire backend an out-of-band channel — independent of response bodies — to deprecate endpoints or signal hard failures to any client running this SDK.

A new `logfire/_internal/server_response.py` module installs a `requests` response hook on every Logfire-bound HTTP session (OTLP exporters, token validation, CRUD client, variables provider, CLI). Each Logfire response is checked for the custom headers `X-Logfire-Warning: <message>` which calls `warnings.warn(message, LogfireServerWarning)`. Python's default filter dedupes identical messages within a process, so a chatty server only warns once.

### DIFF
--- a/logfire/_internal/server_response.py
+++ b/logfire/_internal/server_response.py
@@ -12,52 +12,37 @@ The server attaches custom headers to API responses:
 
 `install_logfire_response_hook(session)` wires this into a `requests.Session` as
 a response hook so every Logfire-bound HTTP response is inspected. Callers can
-pass a custom `hook` to replace the default behaviour (see
+pass a custom `hook` to replace the default behavior (see
 `AdvancedOptions.transport_response_hook`).
 """
 
 from __future__ import annotations
 
-import warnings
-from typing import Any, Callable
+from typing import Any
 
 import requests
 
-from logfire.exceptions import LogfireServerError, LogfireServerWarning
-
-WARNING_HEADER_NAME = 'X-Logfire-Warning'
-ERROR_HEADER_NAME = 'X-Logfire-Error'
-
-TransportResponseHook = Callable[[requests.Response], object]
-"""Callable invoked for every Logfire API response received by the SDK.
-
-The return value is ignored; raise to abort the call.
-"""
-
-
-def process_logfire_response_headers(response: requests.Response) -> None:
-    """Default transport response hook: surface `X-Logfire-Warning` / `X-Logfire-Error` headers."""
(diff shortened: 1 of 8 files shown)


logfire/_internal/server_response.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Designed to be installed as a `requests` response hook
+    (`session.hooks['response'].append(...)`).
+    """
+    warning_message = response.headers.get(WARNING_HEADER_NAME)
+    if warning_message:
+        warnings.warn(warning_message, LogfireServerWarning, stacklevel=2)
+    error_message = response.headers.get(ERROR_HEADER_NAME)
+    if error_message:
+        raise LogfireServerError(error_message)
```
alexmojaki: I worry that users might have a reason to opt out of this and won't be able to. Maybe that's paranoid.

We could theoretically make this hook configurable, defaulting to this one, but allowing users to take any action for all API responses. Probably overkill.

I'd like to know what concrete use cases we have in mind for both warnings and errors.

adriangb: I added a way to opt out in [278a76c](https://github.com/pydantic/logfire/pull/1906/commits/278a76c4765c8f6be9084ee493424f32b6918b6c)

adriangb: > I'd like to know what concrete use cases we have in mind for both warnings and errors.

From the PR description:

> to deprecate endpoints or signal hard failures to any client running this SDK

alexmojaki: we'd deprecate endpoints by changing code here though, right? or is the point to push users to upgrade their sdk?

hard failures like what? where would the server respond with the error header instead of a 4xx/5xx which would have a similar effect here?

adriangb: Yes the point is that we can propagate deprecation to old versions of the SDK.

> hard failures like what? 

The error part was a bit more just "seems like it could be useful". But I imagine it might be more informative to raise a `LogfireError: Some explanation` Than hope the `4xx` bubbles up correctly in a way that is usable by users.

I also wonder if we could emit an error / warn logfire log so it shows up in users projects 🤔

alexmojaki: you decided you don't want the error option?

adriangb: Yeah. Especially because I think exporters win just swallow it and spit it out to stderr.

PR #1916, opened 2026-05-11
docs: describe new region-picker login flow

summerscope:
## Summary

Update the docs to match the planned apex cutover where `logfire.pydantic.dev` stops being a real landing page and becomes a 302 to the US region. The user-facing change is that a region picker now lives on the regional sign-in screen rather than at the apex, so any prose that described the apex as a "place" — a landing page, signup flow, or region-selection step — was out of date.

This PR is the narrative half. URL canonicalisation (`logfire.pydantic.dev/docs/...` → `pydantic.dev/docs/logfire/...`) lives in a separate PR (#1914) so the diffs stay focused.

## Changes

- **`docs/reference/data-regions.md`** — rewrite the "Choosing a Region" section and the closing "global domain" note. The old text said you'd be redirected away from the global domain *after* picking a region; the new text describes the actual flow (apex 302s to US, picker on the regional sign-in screen, required for first-time visitors, last-used pre-selection for returning users) and adds a list highlighting both states.
- **`docs/index.md`** — expand the first "Set up Logfire" step to explain that users land on the US region's sign-in screen and need to pick a region before continuing, with a pointer to `data-regions.md` for users unsure which to choose.
- **`docs/images/logfire-screenshot-region-picker.png`** — new screenshot showing the picker open with both region options visible, referenced from both pages above.
- **`docs/how-to-guides/sso-setup.md`** — add an HTML-comment TODO marker on Step 7 noting that we still need to confirm whether the regional region picker is bypassed for org-slug SSO links, or whether EU-region SSO users have to pick EU at the picker first. Surfacing this in the diff so it's not forgotten.

## Intentionally untouched

Per a deliberate scoping decision: app URL references (root `logfire.pydantic.dev`, `/login`, `/login/{org-name}`, `/settings/...`, `/-/redirect/...`) stay on the apex because the redirect handles them, and `/login` is specifically meant to land users on the region picker. CHANGELOG history is also out of scope.

## Test plan

- [ ] Render the docs locally and confirm the screenshot displays correctly on both pages
- [ ] Confirm the narrative reads cleanly to someone seeing the new sign-in flow for the first time
- [ ] Resolve the SSO TODO with the platform team (then either delete the marker or add the caveat)

https://claude.ai/code/session_016QT2g2gvpVHYxLcLfifDXq

---
_Generated by [Claude Code](https://claude.ai/code/session_016QT2g2gvpVHYxLcLfifDXq)_


docs/reference/data-regions.md
```diff
@@ -24,7 +24,12 @@ Regions are strictly separated with no data sharing between them:
 
 ## Choosing a Region
 
-When you sign up for Logfire, you'll be prompted to choose a region. After selecting your region, you'll be automatically redirected from the global domain (logfire.pydantic.dev) to your region-specific domain (either logfire-us.pydantic.dev or logfire-eu.pydantic.dev).
+When you sign up for Logfire at [logfire.pydantic.dev](https://logfire.pydantic.dev) you'll be prompted to choose a region, which sends you to the correct subdomain (either [logfire-us.pydantic.dev](https://logfire-us.pydantic.dev) or [logfire-eu.pydantic.dev](https://logfire-eu.pydantic.dev)). Subsequent log ins happen at `logfire-us.pydantic.dev` or `logfire-eu.pydantic.dev`.
```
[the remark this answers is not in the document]
summerscope: IMO this is unnecessary detail for the user. They can see they are redirected if they look at the URL. They will be able to see the URL subdomain change if they change regions.

docs/reference/data-regions.md
Viicos, line 1: I feel like recommending `logfire.pydantic.dev` will be confusing especially for EU users. The admonition says:

> if you open [logfire.pydantic.dev](https://logfire.pydantic.dev) on a new device or in incognito mode, you'll be prompted to pick a region again

But this will be the case _anytime_ you open logfire.pydantic.dev and you are in the EU region, and you will have to select the region each time. Should we just mention that there are two URLs, and you should pick the one you want here?

summerscope: Good call — agreed. Reworked the admonition to flip the framing: instead of "if you visit the apex you'll see the picker again," it now tells users to bookmark and use their regional URL directly, with the data-separation note as the rationale for *why* sessions aren't shared. Apex is no longer mentioned in the admonition body.

Pushed in 80e8d94 (now 1dddc40 after rebase).

---
_Generated by [Claude Code](https://claude.ai/code)_

Viicos: I think my comment wasn't really addressed. The docs here still recommend going to `logfire.pydantic.dev`

summerscope: Yep, removed the reference entirely, LMK if this reads ok to you?

Viicos (approved):

docs/reference/data-regions.md
```diff
@@ -24,7 +24,12 @@ Regions are strictly separated with no data sharing between them:
 
 ## Choosing a Region
 
-When you sign up for Logfire, you'll be prompted to choose a region. After selecting your region, you'll be automatically redirected from the global domain (logfire.pydantic.dev) to your region-specific domain (either logfire-us.pydantic.dev or logfire-eu.pydantic.dev).
+Logfire is hosted independently in each region at [logfire-us.pydantic.dev](https://logfire-us.pydantic.dev) and [logfire-eu.pydantic.dev](https://logfire-eu.pydantic.dev). When you sign up you'll be asked to pick a region; on return visits, log in directly at your regional URL.
+
+![Region picker on the sign-in screen](../images/logfire-screenshot-region-picker.png)
+
+!!! info ""
```
Viicos:
```suggestion
!!! info
```

docs/reference/data-regions.md
Viicos: I think this section is redundant and can be removed?

PR #1919, opened 2026-05-11
Drop Python 3.9 support

alexmojaki:
## Summary
- Raise the main Logfire SDK Python floor to 3.10 and remove the 3.9 CI matrix entries.
- Remove now-redundant Python >=3.10 dependency markers and refresh uv.lock, dropping the <3.10 resolution fork.
- Remove 3.9-only compatibility paths, test skips, docs notes, and apply the ruff pyupgrade cleanup enabled by the new floor.

PR #1920, opened 2026-05-11
Drop support for Python < 3.9.2

alexmojaki:
Moving towards dropping 3.9 entirely in #1919

### DIFF
--- a/tests/otel_integrations/test_claude_sdk.py
+++ b/tests/otel_integrations/test_claude_sdk.py
@@ -10,6 +10,29 @@ from inline_snapshot import snapshot
 from logfire._internal.exporters.processor_wrapper import _transform_langsmith_span_attributes  # type: ignore
 
 
+def test_transform_langsmith_span_uses_completion_model_for_request_model():
+    assert _transform_langsmith_span_attributes(
+        {'logfire.span_type': 'span'},
+        {
+            'gen_ai.completion': {
+                'llm_output': {
+                    'model_name': 'claude-sonnet-4-5-20250929',
+                },
+            },
+        },
+    ) == (
+        {
+            'logfire.span_type': 'span',
+        },
+        {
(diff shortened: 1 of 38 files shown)


tests/otel_integrations/test_claude_sdk.py
```diff
@@ -10,6 +10,29 @@
 from logfire._internal.exporters.processor_wrapper import _transform_langsmith_span_attributes  # type: ignore
 
 
+def test_transform_langsmith_span_uses_completion_model_for_request_model():
+    assert _transform_langsmith_span_attributes(
```
alexmojaki, line 14: this was a 3.9 difference?

alexmojaki: (AI) No, it is not a Python 3.9 behavioral difference. It is portable coverage that came out of the Python 3.9-drop PR while looking at what the branch changed.

alexmojaki: right but how did dropping 3.9 affect coverage that this affects

alexmojaki: (AI) It did not affect this behavior semantically. In #1919, removing Python 3.9 changed the CI/coverage shape and exposed an uncovered branch in `_transform_langsmith_span_attributes`: the path where `gen_ai.completion.llm_output.model_name` supplies both `gen_ai.response.model` and, via `setdefault`, `gen_ai.request.model`. I pulled the test forward because it is portable coverage from that PR, not because the code behaves differently on 3.9.

alexmojaki:
(AI) Intentional. Python 3.9.0/3.9.1 have a broken `collections.abc.Callable` generic implementation that can fail in runtime typing contexts and also reports misleading line events for this branch. This PR keeps Python 3.9 support but makes the declared floor `>=3.9.2`, where that behavior is fixed.

PR #1923, opened 2026-05-12
Support Python 3.14 template strings in formatter

alexmojaki:
## Summary
- run the lint job with Python 3.14 so pyright can resolve string.templatelib
- add guarded runtime support for Python 3.14 template strings in the formatter
- cover template string formatting and conversion errors

## Tests
- uv run --python 3.14 pyright
- uv run --python 3.14 pytest tests/test_formatter.py -q
- uv run --python 3.14 pre-commit run --files .github/workflows/main.yml logfire/_internal/formatter.py tests/test_formatter.py

### DIFF
--- a/logfire/_internal/formatter.py
+++ b/logfire/_internal/formatter.py
@@ -7,10 +7,10 @@ from collections.abc import Callable, Iterator
 from functools import lru_cache
 from string import Formatter
 from types import CodeType
-from typing import TYPE_CHECKING, Any, Literal, TypeGuard, cast
+from typing import TYPE_CHECKING, Any, Literal, cast
 
 import executing
-from typing_extensions import NotRequired, TypedDict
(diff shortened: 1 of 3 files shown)


logfire/_internal/formatter.py
```diff
@@ -30,15 +43,33 @@ class ArgChunk(TypedDict):
     spec: NotRequired[str]
 
 
+def is_template_string(format_string: object) -> TypeGuard[Template]:
```
Viicos, line 46: You should then be able to drop the `assert isinstance(format_string, str)` call below:
```suggestion
def is_template_string(format_string: object) -> TypeIs[Template]:
```

alexmojaki: this is ai slop, ignore for now

PR #1927, opened 2026-05-14
Propagate variable version in baggage alongside label

dmontagu:
## Summary

`ResolvedVariable.__enter__` now sets two baggage entries instead of one:

- `logfire.variables.<name>` — the resolved label (unchanged).
- `logfire.variables.<name>.version` — the resolved version, string-formatted (new).

The version entry is only set when a version actually resolved; code-default resolutions (where `version is None`) emit just the label entry, same as before.

## Why

Today, downstream spans inside a variable's resolution context carry the label via baggage but not the version. The version *is* available on the SDK-emitted `Resolve variable <name>` span as an attribute, but that span is opened and closed entirely inside `_get_result_and_record_span` — before the context manager body runs — so it does not share a `trace_id` with any user-created spans. There is no SQL-only path to attach the version to a downstream span.

That gap is hit hardest by A/B-test analytics. The marquee SQL query for `src/walkthroughs/managed-variables/ab-testing-rollout/` (in the platform repo) aggregates request metrics by `logfire.variables.ranking_config` (the label). After a label gets moved to a new version mid-experiment — the canonical "promote canary v2 → v3" flow — aggregating by label alone mixes data from the old and new versions, which is precisely the failure mode an A/B harness is supposed to surface. The walkthrough today works around it by setting `span.set_attribute('config_version', resolved.version)` in application code. With this change, that workaround becomes unnecessary: `GROUP BY label, attributes->>'logfire.variables.ranking_config.version'` works directly.

## Backwards compatibility

Strictly additive — existing SQL queries that filter or group on `logfire.variables.<name>` (the label entry) continue to work unchanged. New queries can pull the version off the new key when they need it.

The new baggage value is a stringified integer (e.g. `'3'`); OTel baggage values are strings.

## Test plan

- [x] Added an assertion in the existing `test_context_manager_sets_label_in_baggage` test that the version entry is present and matches the resolved version (`'1'` in that fixture).
- [x] Added a new test `test_context_manager_omits_version_for_code_default` confirming that resolutions with no remote config (so `version is None`) emit *only* the label entry, not a version entry.
- [x] All 7 existing context-manager tests still pass.

## Follow-up in platform repo

Once this lands and a release goes out (and the platform's pinned `logfire` revision bumps), the walkthrough's `02_send_queries.py` workaround can be removed and the SQL query simplifies. That cleanup is referenced from the platform PR that bundles the rename work and the walkthrough's interim workaround.


logfire/variables/abstract.py
```diff
@@ -76,17 +76,19 @@ class ResolvedVariable(Generic[T_co]):
...
+    resolution that was active at the time.
 
     Example:
         ```python skip="true"
         my_var = logfire.var(name='my_var', type=str, default='default')
         with my_var.get() as details:
             # Inside this context, baggage is set with:
-            # logfire.variables.my_var = <label> (or '<code_default>' if no label)
+            #   logfire.variables.my_var          = <label> (or '<code_default>' if no label)
+            #   logfire.variables.my_var.version  = <version> (only when a versioned value was resolved)
```
alexmojaki, line 89: please confirm that this can't cause some kind of conflict, i.e. that `foo.version` is rejected as a variable name

tests/test_variables.py
```diff
@@ -721,6 +721,24 @@ def test_context_manager_sets_label_in_baggage(self, config_kwargs: dict[str, An
...
             assert baggage['logfire.variables.cm_var'] == 'my_label'
+            # Version is propagated alongside the label so downstream spans can
+            # be filtered/grouped by `(label, version)`.
+            assert baggage['logfire.variables.cm_var.version'] == '1'
+
+    def test_context_manager_omits_version_for_code_default(self, config_kwargs: dict[str, Any]):
+        """Code-default resolutions have version=None and should not emit a version baggage entry."""
+        lf = logfire.configure(**config_kwargs)
+
+        var = lf.var(name='no_version_var', default='default', type=str)
```
alexmojaki:
```suggestion
    def test_context_manager_omits_version_for_code_default(self):
        """Code-default resolutions have version=None and should not emit a version baggage entry."""
        var = logfire.var(name='no_version_var', default='default', type=str)
```

tests/test_variables.py
```diff
@@ -721,6 +721,24 @@ def test_context_manager_sets_label_in_baggage(self, config_kwargs: dict[str, An
...
+        details = var.get()
+
+        assert details.label is None
+        assert details.version is None
+
+        with details:
+            baggage = logfire.get_baggage()
+            assert baggage['logfire.variables.no_version_var'] == '<code_default>'
+            assert 'logfire.variables.no_version_var.version' not in baggage
```
alexmojaki:
```suggestion
            assert baggage == snapshot(...)
```

logfire/variables/abstract.py
dmontagu: Confirmed — no conflict possible. Variable names must match `^[a-zA-Z_][a-zA-Z0-9_]*$` (Python identifiers, no dots), enforced in two places:

- `logfire/_internal/main.py:2594` — runtime check in `Logfire.var(...)` raises `ValueError` for invalid names.
- `logfire/variables/config.py:194` — `VariableName = Annotated[str, Field(pattern=r'^[a-zA-Z_][a-zA-Z0-9_]*$'), ...]`, applied to `VariableConfig.name` (and `aliases`) on the remote-config payloads.

So `foo.version` cannot exist as a variable name, and the new `logfire.variables.<name>.version` baggage key cannot collide with a `logfire.variables.<other_name>` entry.

tests/test_variables.py
dmontagu: Applied in 4324e660 — dropped the `config_kwargs` parameter and the redundant `logfire.configure(...)` call. The autouse `config` fixture in `conftest.py` already configures the default logfire instance and clears variables between tests, so the extra call wasn't needed here.

tests/test_variables.py
dmontagu: Applied in 4324e660 — switched to `assert baggage == snapshot({'logfire.variables.no_version_var': '<code_default>'})`.

PR #1931, opened 2026-05-18
feat: Artifacts — attach binary blobs to spans

adriangb:
## What

Adds `logfire.Artifact` — attach a binary blob (image, audio, PDF, large JSON) to a span. The blob is uploaded to object storage **out of band**; the span carries only a small **content-addressed** (sha256) reference, so it is not subject to attribute size limits and does not bloat telemetry.

This is the SDK side. The backend side ships separately in the platform repo (PR linked below once open).

## Usage

```python
import logfire

logfire.configure()

logfire.info('chart generated', chart=logfire.Artifact.from_file('chart.png'))
logfire.info('thumbnail', image=logfire.Artifact(png_bytes, content_type='image/png'))
```

- Construct from **bytes**, a **file path** (`Artifact.from_file`), or a **binary handle** (`Artifact.from_file_handle` — handles temp/spooled files).
- `upload` is chosen **per artifact**:
  - `background` (default) — never blocks the caller. If the upload queue is over its byte budget, the artifact is dropped with a warning rather than applying backpressure.
  - `sync` — uploaded inline; the call returns once the blob is stored, so the source can be freed immediately.

## How it works

An artifact serializes to a reference object (`{"type": "logfire.artifact", "sha256": …}`) via the existing `json_schema` / `json_encoder` hooks. A background uploader runs the register → PUT → finalize handshake against the backend; signed object-store URLs are PUT to without the bearer token.

## Changes

- `logfire/_internal/artifacts.py` — `Artifact` + the `ArtifactSource` abstraction (bytes / path; designed so streaming sources can be added later).
- `logfire/_internal/exporters/artifact_uploader.py` — background uploader (bounded queue, drop-on-full).
- `json_schema` / `json_encoder` / `main` / `config` hooks; `Artifact` + `UploadMode` exports.
- `logfire-api` stubs regenerated; docs page at `reference/advanced/artifacts.md`.

## Verification

`tests/test_artifacts.py` — 14 tests pass (construction from each source, schema/encoder hooks, span integration, uploader sync/background/dedup/drop-on-full/error-swallowing). `ruff` + `pyright` clean.

Note: `test_logfire_api.py::test_runtime[with_logfire]` currently fails on an unrelated `dspy` `DeprecationWarning` raised by `instrument_dspy` — pre-existing, not introduced here.

### DIFF
--- a/docs/reference/advanced/artifacts.md
+++ b/docs/reference/advanced/artifacts.md
@@ -1,0 +1,88 @@
+# Artifacts
+
+Artifacts let you attach a **binary blob** — an image, audio clip, PDF, or a large JSON
+payload — to a span. The blob is stored separately from your telemetry, so it is not
+subject to span attribute size limits and does not bloat your traces. The span itself
+carries only a small reference; Logfire uploads the blob out of band.
+
+## Logging an artifact
+
+Wrap your data in `logfire.Artifact` and pass it as a span or log attribute:
+
+```python skip="true"
+import logfire
+
+logfire.configure()
+
+with open('chart.png', 'rb') as f:
+    image_bytes = f.read()
+
+logfire.info('chart generated', chart=logfire.Artifact(image_bytes, content_type='image/png'))
+```
+
+The `chart` argument renders as an image preview on the trace in the Logfire UI, with a
+download link — not as a wall of base64.
+
+### From a file or a file handle
+
+`Artifact.from_file` reads a path lazily (no need to load the bytes yourself), and
+`Artifact.from_file_handle` accepts any open binary handle, including temporary files:
+
+```python skip="true"
+import logfire
+
+logfire.configure()
+
+# From a path — the content type is guessed from the extension.
+logfire.info('report ready', report=logfire.Artifact.from_file('report.pdf'))
+
+# From an open binary handle.
+with open('clip.mp3', 'rb') as handle:
+    logfire.info('audio processed', clip=logfire.Artifact.from_file_handle(handle))
+```
+
+## When the upload happens
+
+Each artifact chooses when its bytes are uploaded, via the `upload` argument:
+
+- **`background`** (the default) — the upload is handed to a background thread and the
+  logging call never blocks. If uploads cannot keep up, queued artifacts are dropped
+  with a warning rather than stalling your program.
+- **`sync`** — the upload runs inline; the logging call returns only once the blob is
+  stored. Use this when you need delivery guaranteed, or when you want to free the
+  source bytes/file immediately afterwards.
+
+```python skip="true"
+import logfire
+
+logfire.configure()
+
+# Block until this artifact is uploaded.
+logfire.info('critical input', data=logfire.Artifact(payload, upload='sync'))
+```
+
+## How it works
+
+Artifacts are **content-addressed**: an artifact's identity is the sha256 of its bytes.
+The same blob logged repeatedly — within a project — is uploaded and stored only once.
+The reference embedded in the span looks like:
+
+```json
+{
+  "type": "logfire.artifact",
+  "sha256": "9f86d0818...",
+  "filename": "chart.png",
+  "content_type": "image/png",
+  "size_bytes": 28421
+}
+```
+
+The blob never travels through the telemetry pipeline. On SaaS the SDK uploads it
+directly to object storage via a signed URL; self-hosted deployments route it through
+the Logfire backend.
+
+## Viewing artifacts
+
+Open a span in the Logfire UI: any artifact-valued argument renders inline — images,
+audio, video, and PDFs as previews, everything else as a download link — alongside its
+filename, content type, and size.
--- a/logfire-api/logfire_api/_internal/exporters/artifact_uploader.pyi
+++ b/logfire-api/logfire_api/_internal/exporters/artifact_uploader.pyi
@@ -1,0 +1,26 @@
+from _typeshed import Incomplete
+
+from ..artifacts import Artifact as Artifact
+from ..utils import log_internal_error as log_internal_error
+
+DEFAULT_MAX_QUEUE_BYTES: Incomplete
+
+class ArtifactUploader:
+    """Uploads artifact blobs to the Logfire backend.
+
+    Owns a daemon worker thread for `background` artifacts; `sync` artifacts are uploaded
+    on the calling thread.
+    """
+    def __init__(self, *, base_url: str, token: str, max_queue_bytes: int = ...) -> None: ...
+    def submit(self, artifact: Artifact) -> None:
+        """Upload an artifact's blob — inline for `sync`, queued for `background`.
+
+        A `background` submit **never blocks the caller**: if the queue is already over
+        its byte budget, the artifact is dropped with a warning rather than applying
+        backpressure. A `sync` submit uploads inline (and so blocks) by the caller's
+        explicit choice. Never raises — upload failures are handled internally.
+        """
+    def flush(self, timeout: float | None = None) -> bool:
+        """Block until queued background uploads have drained. Returns whether they did."""
+    def shutdown(self, timeout: float = 5.0) -> None:
+        """Drain in-flight uploads and stop the worker thread."""
--- a/logfire/__init__.py
+++ b/logfire/__init__.py
@@ -8,6 +8,7 @@ from logfire.propagate import attach_context, get_context
 from logfire.sampling import SamplingOptions
 
 from . import variables as variables
+from ._internal.artifacts import Artifact, UploadMode
 from ._internal.auto_trace import AutoTraceModule
 from ._internal.auto_trace.rewrite_ast import no_auto_trace
 from ._internal.baggage import get_baggage, set_baggage
@@ -213,4 +214,6 @@ __all__ = (
     'get_context',
     'attach_context',
     'url_from_eval',
+    'Artifact',
+    'UploadMode',
 )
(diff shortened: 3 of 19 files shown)


adriangb:
Backend side: pydantic/platform#21850

logfire-api/logfire_api/_internal/exporters/artifact_uploader.pyi
alexmojaki, line 1: Please don't generate pyi files in PRs, they're just clutter. the CLAUDE.md should probably make this more explicit.

docs/reference/advanced/artifacts.md
```diff
@@ -0,0 +1,88 @@
...
+
+## When the upload happens
+
+Each artifact chooses when its bytes are uploaded, via the `upload` argument:
+
+- **`background`** (the default) — the upload is handed to a background thread and the
+  logging call never blocks. If uploads cannot keep up, queued artifacts are dropped
+  with a warning rather than stalling your program.
+- **`sync`** — the upload runs inline; the logging call returns only once the blob is
+  stored. Use this when you need delivery guaranteed, or when you want to free the
```
alexmojaki, line 52: Delivery is far from guaranteed with sync, it still just silently swallows request exceptions without retrying.

logfire/_internal/exporters/artifact_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+import time
+import warnings
+from urllib.parse import urljoin
+
+import requests
+
+from ..artifacts import Artifact
+from ..utils import log_internal_error
+
+# Default ceiling on bytes queued for background upload. When exceeded, `submit` blocks.
```
alexmojaki, line 24: it doesn't block, it drops

logfire/_internal/exporters/artifact_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class ArtifactUploader:
+    """Uploads artifact blobs to the Logfire backend.
+
+    Owns a daemon worker thread for `background` artifacts; `sync` artifacts are uploaded
+    on the calling thread.
+    """
+
+    def __init__(self, *, base_url: str, token: str, max_queue_bytes: int = DEFAULT_MAX_QUEUE_BYTES) -> None:
```
alexmojaki, line 38: max_queue_bytes isn't actually user configurable and probably should be

logfire/_internal/exporters/artifact_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+        if body['status'] == 'exists':
+            # The blob is already stored (content-addressed dedup) — nothing to upload.
+            return
+
+        target = body['upload']
+        # Signed object-store URLs are self-authenticating; sending the bearer token
+        # there can break the signature, so only the backend `/blob` endpoint gets auth.
+        put_headers = self._auth if target['requires_auth'] else {}
+        put = requests.request(
+            target['method'], target['url'], data=artifact.read(), headers=put_headers, timeout=_REQUEST_TIMEOUT
```
alexmojaki, line 143: when the artifact is read from a file, it doesn't live in memory in the queue. but max_queue_bytes seems to be intended to save memory, so should it apply to file artifacts? in fact, what if we stored all artifacts to files to allow a bigger queue?

logfire/_internal/exporters/artifact_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        if artifact.upload == 'sync':
+            self._run_upload(artifact)
+            return
+
+        size = artifact.size_bytes
+        with self._capacity:
+            # Drop rather than block: artifact uploads must never stall the program.
+            # An empty queue always admits one artifact, so a lone large blob still
+            # uploads even if it alone exceeds the budget.
```
alexmojaki, line 65: i don't know if this is wrong, but it doesn't feel intuitive.

logfire/_internal/exporters/artifact_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+        # Signed object-store URLs are self-authenticating; sending the bearer token
+        # there can break the signature, so only the backend `/blob` endpoint gets auth.
+        put_headers = self._auth if target['requires_auth'] else {}
+        put = requests.request(
+            target['method'], target['url'], data=artifact.read(), headers=put_headers, timeout=_REQUEST_TIMEOUT
+        )
+        put.raise_for_status()
+
+        finalized = requests.post(
+            urljoin(self._base_url, f'/v1/artifacts/{sha256}/finalize'),
```
alexmojaki, line 148: why is this needed? the attribute encoding seems to assume the artifact is uploaded regardless. can't the backend finalize?

logfire/__init__.py
```diff
@@ -213,4 +214,6 @@ def loguru_handler() -> Any:
     'get_context',
     'attach_context',
     'url_from_eval',
+    'Artifact',
+    'UploadMode',
```
alexmojaki, line 218: Do users need to use `logfire.UploadMode`? this seems like clutter in the root package.

logfire/_internal/exporters/artifact_uploader.py
```diff
@@ -0,0 +1,152 @@
...
+                    self._capacity.notify_all()
+
+    def _run_upload(self, artifact: Artifact) -> None:
+        try:
+            self._upload(artifact)
+        except requests.RequestException:
+            # Network/HTTP failures are operational, not bugs: the artifact reference is
+            # still recorded on the span, the blob just isn't stored. Best-effort upload —
+            # never crash the caller's logging call over it.
+            pass
```
alexmojaki, line 112: will need retries

logfire/_internal/artifacts.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # From an open binary handle (including temporary / spooled files).
+        with open('report.pdf', 'rb') as handle:
+            logfire.info('report', report=logfire.Artifact.from_file_handle(handle))
+        ```
+    """
+
+    def __init__(
+        self,
+        data: bytes,
```
alexmojaki, line 120: IMO `data` should accept:

- bytes
- str, which is encoded to bytes
- Path, which is like from_file
- a file handle, and then from_file_handle isn't needed at all
- any other object, which goes through the logfire JSON encoding

Then the constructor can handle basically anything, and `from_file` is only a convenience to treat `str` as a path instead of actual data, and to maybe save some memory depending on implementation details.

logfire/_internal/artifacts.py
```diff
@@ -0,0 +1,258 @@
...
+        self.content_type = content_type or _guess_content_type(guess_from)
+        self._digest: tuple[str, int] | None = None
+
+    @property
+    def sha256(self) -> str:
+        """The hex sha256 of the blob — its content-addressed identity."""
+        return self._compute()[0]
+
+    @property
+    def size_bytes(self) -> int:
```
alexmojaki, line 225: does the backend check that this is reported honestly?

logfire/_internal/json_encoder.py
```diff
@@ -214,6 +215,8 @@ def encoder_by_type() -> dict[type[Any], EncoderFunction]:
         Pattern: lambda o, seen: to_json_value(o.pattern, seen),
         UUID: _to_str,
         Exception: _to_str,
+        # An artifact serialises to its reference object; the blob is uploaded separately.
+        Artifact: lambda o, _: o.reference(),
```
alexmojaki, line 219: we need a way to mark the contents of these as always exempt from scrubbing. this shape doesn't allow us to without adding some new machinery. if it was `{'logfire.artifact': {...}}` then adding `logfire.artifact` to the scrubber safe keys would work.

that wouldn't help if someone writes `secret=Artifact(...)`, which will get scrubbed by default either way. not sure if we want to do something about that.

docs/reference/advanced/artifacts.md
adriangb: Do you think we should make the guarantee stronger (error) or make the docstrings match current impl?

logfire/_internal/exporters/artifact_uploader.py
adriangb: should we make it drop, block, spill to disk...?

logfire/_internal/exporters/artifact_uploader.py
adriangb: yeah I think it would make sense to force all artifacts to buffer on disk, they are almost by definition large

logfire/_internal/exporters/artifact_uploader.py
adriangb: i think it's an important property otherwise it's an auto deadlock footgun

logfire/_internal/json_encoder.py
adriangb: maybe we just say scrubbing doesn't apply to artifcats for now and figure it out in a followup?

alexmojaki: i'm saying we need to figure out how to make scrubbing not apply to artifacts, right now it can. not the contents, the reference.

adriangb: that seems like something we can sort out. i'm just asking if we can ship whatever falls out naturally now (even if it's kinda broken) and come up with the right apis, etc. later or if you think it needs to be bundled into this change?

alexmojaki: i thought the goal right now was to settle on a good API? what kind of review are you looking for?

logfire/_internal/exporters/artifact_uploader.py
alexmojaki: for background uploading, i think not blocking is part of the contract. i think it should spill to disk up to a higher limit, then drop.

docs/reference/advanced/artifacts.md
alexmojaki: i think we eventually want a stronger guarantee, but it doesn't have to be in the first pass. until then, the docs should be accurate.

PR #1949, opened 2026-05-21
Address variable structure review findings

alexmojaki:
Stacked on #1947.

Addresses the review comments about:
- preserving provider exceptions when rendered code defaults are used
- avoiding serialized-default work when no render hook is present
- treating Pydantic validator TypeError as a deserialization failure while preserving label/version metadata

Tests: `uv run pytest tests/test_variables.py -q`

### DIFF
--- a/logfire/variables/variable.py
+++ b/logfire/variables/variable.py
@@ -171,11 +171,11 @@ class _BaseVariable(Generic[T_co]):
         """
         return None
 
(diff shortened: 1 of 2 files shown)


dmontagu:
Superseded by #1731 — cherry-picked onto the variable-composition branch as 8b04bfd9 (resolving the conflict with #1731's later revert of the early-return that broke composition for plain `Variable` defaults containing `@{...}@` references). Closing in favor of #1731.

Issue #1950, opened 2026-05-21
Managed variables: template render-time warnings and missing-input handling


dmontagu:
> **Status (2026-05-22)** — Both upstream blockers in `pydantic-handlebars` are resolved and released in [v0.2.0](https://github.com/pydantic/pydantic-handlebars/releases/tag/v0.2.0). The SDK-side work below is now unblocked.

## Context

`logfire.template_var()` resolves a value, expands `@{ref}@` composition references, and renders Handlebars `{{placeholder}}` templates against typed `inputs`. Right now the render step silently substitutes empty strings when a template references a field the inputs don't have:

```py
prompt = logfire.template_var(
    'system_prompt',
    type=str,
    default='Hello {{user_name}}! Welcome to our service.',
    inputs_type=PromptInputs,
)
# inputs has user_name, but the template references user_name2
prompt.get(PromptInputs(user_name='Alice'))
# → "Hello ! Welcome to our service."  (no warning)
```

Meanwhile the composition path *does* warn when a referenced variable is missing:

```
RuntimeWarning: Variable 'agent_prompt' composition failed; falling back to code default:
Referenced variable 'safetyd_rules' could not be resolved.
```

The asymmetry is bad. Surfaced by Alex on #1731 (`templates-and-composition.md:36`).

## Upstream changes shipped in `pydantic-handlebars` 0.2.0

- **`strict=True`** on `HandlebarsEnvironment`, `render()`, and `compile()` — pydantic/pydantic-handlebars#10. Raises `HandlebarsRuntimeError` on missing context references; explicit `None` values still pass.
- **`extract_dependencies(source, ...)`** — pydantic/pydantic-handlebars#10. Statically returns the set of top-level context field names a template depends on. Scope-aware (handles `each`/`with`, block params, `../`, `@root.x`).
- **Configurable mustache delimiters** — pydantic/pydantic-handlebars#11. `open_delim` / `close_delim` on `HandlebarsEnvironment`, `render()`, `compile()`, and `extract_dependencies()`. Lets the SDK run a Handlebars pass that consumes `@{...}@` without disturbing `{{...}}` runtime placeholders. Full Handlebars syntax (block helpers, dotted paths, subexpressions, …) now works under `@{...}@` because we're running the real parser instead of a regex pretending to.

### A. Replace `logfire/variables/reference_syntax.py` with native `@{...}@` rendering

[...]

PR #1951, opened 2026-05-21
Add variable composition and Handlebars template rendering

dmontagu:
Re-opens the work from #1731 with a cleaner two-commit history (one for code, one for docs/demo) so this PR is easier to browse and review.

The diff has been polished since #1731:

1. `TemplateVariable` now subclasses `Variable` directly (rather than both subclassing an internal `_BaseVariable`), so the public API surface no longer exposes a private base class. The `.get(inputs, …)` override on `TemplateVariable` is a knowing Liskov violation marked with `# pyright: ignore[reportIncompatibleMethodOverride]`.
2. `tests/test_docs.py` now skips both `logfire.template_var` *and* `@{…}@` examples on Python 3.9 (composition requires `pydantic-handlebars`, which the `[variables]` extra installs only on 3.10+).

Everything else is functionally identical to `feature/variable-composition`.

## Summary

- **Variable composition**: Managed variable values can reference other managed variables with `@{variable_name}@` syntax. Composition happens during resolution, before deserialization, and supports nested references, dotted access like `@{brand.tagline}@`, escaping, cycle/depth protection, and `composed_from` metadata on resolved variables and spans.
- **Template rendering**: Variable values can contain Handlebars `{{placeholder}}` templates rendered with runtime inputs. Rendering uses `pydantic-handlebars` and preserves `{{...}}` placeholders while `@{...}@` composition is expanded first.
- **`TemplateVariable[T, InputsT]` API**: `logfire.template_var()` adds a single-step flow where `get(inputs)` resolves the variable, expands `@{...}@` references, renders `{{...}}` placeholders, and deserializes to the declared type. `TemplateVariable` is a subclass of `Variable`.
- **Shared resolution path**: `Variable._lookup_serialized` encodes the override → provider → registered code default priority once, and routes both `_resolve` (for `self.name`) and the composition expander's `resolve_ref` (for child `@{ref}@` lookups) through it.
- **Validation and sync support**: Template input schemas are included in variable config/sync, and validation checks composed values for undeclared template fields plus missing references and reference cycles.
- **Docs and demo**: Adds managed-variable docs for templates/composition and a runnable demo covering composition, structured variables, template inputs, and composition-time conditionals.

## Follow-ups (not blocking this PR)

- #1950 — Template render-time warnings, strict-mode for missing inputs, push-/pull-time schema validation, and `@{}@` syntax parity with `{{}}` (block helpers, dotted paths, etc.).

## Supersedes

This PR replaces #1731. All review threads on #1731 have been addressed and resolved. #1731 will be closed once this PR is reviewed and merged.


logfire/variables/_handlebars.py
```diff
@@ -0,0 +1,58 @@
...
+    """Raise a helpful error if pydantic-handlebars is unavailable."""
+    get_handlebars_renderer()
+
+
+@cache
+def get_handlebars_renderer() -> tuple[type[str], Callable[..., str]]:
+    """Return pydantic-handlebars SafeString and render, or raise a helpful error."""
+    try:
+        from pydantic_handlebars import SafeString, render
+    except ImportError as exc:  # pragma: no cover
```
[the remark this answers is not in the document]
dmontagu: Done — `_handlebars.py` now catches `ModuleNotFoundError` with an `exc.name == 'pydantic_handlebars'` guard, so real import failures inside `pydantic_handlebars` (e.g. a broken transitive dep) propagate instead of being reported as the optional extra being missing. Test updated to simulate the missing-module case with `ModuleNotFoundError(name=...)` to match what the runtime actually raises.

logfire/variables/variable.py
```diff
@@ -187,166 +206,260 @@ def refresh_sync(self, force: bool = False):
...
+    ) -> ResolvedVariable[T_co]:
+        """Fall back to the code default and warn after a composition/render failure."""
+        warnings.warn(
+            f"Variable '{self.name}' composition failed; falling back to code default: {exception}",
+            category=RuntimeWarning,
+            stacklevel=2,
+        )
+        return ResolvedVariable(
+            name=self.name,
+            value=self._get_default(targeting_key, attributes),
```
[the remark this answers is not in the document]
dmontagu: Good catch. Threaded `default_value` through `_expand_and_deserialize` (and into `_composition_failure` via that path) using an `_UNSET` sentinel. The composition/render failure paths now reuse the already-computed default from `_resolve_code_default` instead of re-invoking `self._get_default`. `_resolve` (provider-value path, no precomputed default) lazily computes once via a closure when needed. Net effect: callable defaults run at most once per `get()`.

examples/python/variable_composition_demo.py
```diff
@@ -0,0 +1,603 @@
...
+"""
+
+from __future__ import annotations
+
+import json
+
+from pydantic import BaseModel
+
+import logfire
+from logfire._internal.config import LocalVariablesOptions
```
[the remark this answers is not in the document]
dmontagu: Done — switched to `from logfire import LocalVariablesOptions`.

logfire/variables/variable.py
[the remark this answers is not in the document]
dmontagu: Update: pushed a follow-up that simplified this further. After the `_lookup_serialized` extraction, the `_resolve_code_default` path was only entered when the registered default could not be serialized to JSON; the rest of its body (composition/render-through-the-default plus the `_UNSET` plumbing I added to address this comment) was effectively dead code, so I dropped it. The remaining double-call surface is `_composition_failure` falling back to `_get_default` after `_lookup_serialized` already invoked it once via `_get_serialized_default` in the code-default tier — preserving that as-is since closing it cleanly needs `_lookup_serialized` to thread the unserialized value through, which felt out of proportion for the rare callable-default-plus-composition-failure case. Happy to revisit in a follow-up if you'd rather see it strictly single-call.

logfire/variables/variable.py
```diff
@@ -187,166 +201,247 @@ def refresh_sync(self, force: bool = False):
...
+
+        # 1. Context override (only for variables whose type we know)
+        if context_overrides is not None and name in context_overrides and variable is not None:
+            override_value = context_overrides[name]
+            if is_resolve_function(override_value):
+                override_value = override_value(targeting_key, attributes)
+            try:
+                serialized = variable.type_adapter.dump_json(override_value).decode('utf-8')
+            except (ValueError, TypeError, RuntimeError):
+                pass  # Fall through to provider/code default
```
alexmojaki, line 291: This silent fallthrough seems user-unfriendly for an explicit `override(...)`. If an override value cannot be serialized for composition/template processing, we should probably emit a warning at minimum before ignoring it and using provider/code-default resolution. Otherwise tests can pass while callers think their override took effect when it did not.

logfire/_internal/main.py
```diff
@@ -2603,27 +2612,114 @@ def var(
...
+        )
+        self._variables[name] = variable
+
+        return variable
+
+    def template_var(
+        self,
+        name: str,
+        *,
+        type: type[T],
```
alexmojaki, line 2629: Question: `var()` lets `type` be omitted and infers it from a non-callable default, while `template_var()` requires `type` even though it also has a non-callable default in the common case. Is that difference intentional? If not, it seems like `template_var("prompt", default="Hello {{name}}", inputs_type=Inputs)` could mirror `var()` and only require `type` when the default is a resolve function.

logfire/variables/abstract.py
```diff
@@ -422,6 +530,90 @@ def _check_type_label_compatibility(
...
+    warnings_list: list[str] = []
+
+    # Collect all known variable names (local + server)
+    all_names: set[str] = {v.name for v in variables} | set(server_config.variables.keys())
+
+    # Build a reference graph: variable_name -> set of referenced names
+    ref_graph: dict[str, set[str]] = {}
+
+    # Scan local variable defaults for references
+    for variable in variables:
```
alexmojaki, line 555: P2: This only builds the reference graph from variables passed in by local code, so refs inside server-only variables reached through composition are not validated.

Concrete remote repro from `scratch_2934.py`:

```python
logfire.configure()
foo2 = logfire.var('foo2', type=str, default='foo2 references @{foo1}@')
print(logfire.variables_validate())
foo1 = logfire.var('foo1', type=str, default='foo1')
print(logfire.variables_validate())
```

With remote `foo1.latest_version = "foo1 references @{foo3}@"`, the first validation reports `reference_errors=[]` because local `foo2` references `foo1` and `foo1` exists remotely. Only after registering local `foo1` does validation scan `foo1`'s remote values and report `Variable 'foo1' references '@{foo3}@' which does not exist.`

So strict validation can miss missing refs/cycles inside server-only composed fragments that are reachable from local variables. It probably needs to traverse the composed reference graph, not just scan server values for names that are already local variables.

alexmojaki: The backend just let me set that invalid template `foo1 references @{foo3}@`, i'm not sure if that's an issue on its own.

Also I think validating foo2 should complain about the missing local var even when the server remote value is valid. Fallback to the code default should always be an option.

logfire/variables/variable.py
```diff
@@ -187,166 +201,247 @@ def refresh_sync(self, force: bool = False):
...
+        Handles composition between the provider fetch and Pydantic deserialization.
+        When render_fn is provided, it is applied after composition and before deserialization.
+        """
+        assert serialized_result.value is not None
+
+        serialized_value = serialized_result.value
+        composed: list[ComposedReference] = []
+
+        # Expand @{references}@ if any are present
+        if has_references(serialized_value):
```
alexmojaki, line 343: This guard makes escaped-only composition syntax behave differently from the same escaped syntax when a real reference is also present. Repro on this stack:

```python
import logfire

logfire.configure()

baz = logfire.var('baz', type=str, default='baz')

assert logfire.var('bar1', type=str, default=r'@{baz}@').get().value == 'baz'
assert logfire.var('bar2', type=str, default=r'\@{baz}@').get().value == r'\@{baz}@'
assert logfire.var('bar3', type=str, default=r'@{baz}@ and \@{baz}@').get().value == 'baz and @{baz}@'
```

The escaped `@{baz}@` keeps the backslash when it is the only composition-looking syntax, because `has_references(serialized_value)` is false and `expand_references()` never gets a chance to unescape it. When there is also a real reference, expansion runs and the escaped reference becomes a literal `@{baz}@`. So the observable escaping behavior depends on whether another reference is present in the same value.

logfire/variables/composition.py
```diff
@@ -0,0 +1,369 @@
...
+    serialized_value: str,
+    variable_name: str,
+    resolve_fn: ResolveFn,
+    *,
+    _visited: tuple[str, ...] = (),
+    _depth: int = 0,
+) -> tuple[str, list[ComposedReference]]:
+    """Expand `@{var}@` references in a serialized variable value.
+
+    Uses the Handlebars engine so that `@{}@` supports simple references,
```
alexmojaki, line 103: This docstring also understates the syntax after #1952. The runtime is now using native pydantic-handlebars dependency extraction/rendering, so block helper conditions are not limited to top-level names; for example `@{#if user.active}@...@{/if}@` works. Keeping this wording makes the code comments disagree with the behavior introduced later in the stack.

docs/reference/advanced/managed-variables/templates-and-composition.md
```diff
@@ -0,0 +1,216 @@
...
+The `@{}@` syntax supports a small Handlebars-compatible subset for composing variables. It supports simple references, dotted field reads, and block helpers that branch or iterate over a top-level referenced variable:
+
+| Syntax | Description |
+|--------|-------------|
+| `@{variable_name}@` | Insert a variable's value |
+| `@{variable.field}@` | Access a nested field |
+| `@{#if variable}@...@{else}@...@{/if}@` | Conditional on whether a variable is set |
+| `@{#each items}@...@{/each}@` | Iterate over a list variable |
+
+Block helper conditions and iterables must be top-level variable names. Use `@{#if user}@...@{user.active}@...@{/if}@` rather than `@{#if user.active}@`.
```
alexmojaki, line 162: This restriction becomes stale once #1952 is applied. The native Handlebars path accepts dotted expressions in block helper headers, e.g. `@{#if user.active}@premium@{/if}@` renders correctly (I verified it returns `premium`). So the docs should either remove the top-level-only restriction here or explicitly document the broader syntax now supported by #1952.

tests/test_variable_templates.py
```diff
@@ -0,0 +1,471 @@
...
+            default=Config(code='OK'),
+            inputs_type=Inputs,
+        )
+
+        # `model_construct` bypasses the constructor's validation so the override can hold
+        # the unrendered template; rendering then produces 'abc123', which fails the
+        # pattern constraint and exercises the outer error handler's code-default fallback.
+        bad_override = Config.model_construct(code='{{code}}')
+        with var.override(bad_override):
+            resolved = var.get(Inputs(code='abc123'))
```
alexmojaki, line 467: This looks worth tightening. The templated value itself is not really "bad" here: `Config.model_construct(code="{{code}}")` is valid as an unrendered template, and the rendered result only becomes invalid because `Inputs(code="abc123")` violates the `^[A-Z]+$` constraint. Could this use names like `templated_config` and `invalid_inputs` to make that clearer? Also, if this behavior is intentional, should the fallback path emit a warning when render inputs produce a value that fails validation before returning the code default with `reason="other_error"`?

logfire/variables/variable.py
alexmojaki: also, if unserializable overrides aren't allowed, why are unserializable code defaults allowed? this is weird:

```python
import logfire

x1 = object()
x2 = object()

fragment = logfire.var(name='fragment', default=x1, type=object)

with fragment.override(x2):
    assert fragment.get().value is x1
```

dmontagu:
Closing in favor of #1954

PR #1952, opened 2026-05-22
Run @{...}@ composition through native pydantic-handlebars

dmontagu:
First in a series implementing the SDK follow-ups from #1950, stacked on #1951.

Replaces the regex-and-translate workaround in `reference_syntax.py` and `composition.py` with calls into `pydantic-handlebars >= 0.2.0`'s native configurable-delimiter API.

## What this enables

`@{...}@` accepts the full Handlebars syntax now:

- Block helpers with dotted-path headers — `@{#if user.active}@premium@{else}@free@{/if}@`. The old regex translator silently dropped the condition because `user.active` isn't an identifier.
- `{{#each}}` parent-context references with `../`.
- Helper subexpressions (`@{lookup obj key}@`).

This closes the syntax-parity item Alex flagged on `templates-and-composition.md:186` of #1731.

## What changed

- `pyproject.toml`: bumps `pydantic-handlebars` to `>=0.2.0` in the `[variables]` extra and dev group; exempts it from the `exclude-newer` filter alongside the other Pydantic packages so `uv lock` picks up the just-released version.
- `_handlebars.py`: adds `get_environment()` (cached `HandlebarsEnvironment(open_delim='@{', close_delim='}@')`) and `extract_composition_dependencies(template)` wrapping `pydantic_handlebars.extract_dependencies` with the same delimiters.
- `reference_syntax.py`: `render_once()` shrinks to a one-line call into that environment. The sentinel-protect-then-regex-translate code path is gone. `{{...}}` runtime placeholders survive untouched because they're plain content under the configured delimiters.
- `composition.py`: `find_references()` / `_collect_ref_names()` go through `extract_composition_dependencies` for AST-correct detection (block helpers, dotted paths, subexpressions, helper-name exclusion). A textual word-boundary scan supplies first-occurrence ordering so the public `find_references` signature stays a `list[str]`.

The unresolved-dotted-reference protection in `_render_value` is kept for behaviour compat (literal `@{name.field}@` is retained when `name` can't be resolved). That's the misrender-retention parity question — a separate PR in this series.

## Tests

- All 501 existing variable / template / composition / push / template-validation tests pass.
- New `TestExpandReferencesNativeHandlebarsSyntax` covers `@{#if user.active}@`, `@{#each list}@@{this}@@{/each}@`, `../` parent refs, and `@{lookup obj key}@`.
- New `TestFindReferencesNativeHandlebarsSyntax` checks the AST-aware dependency extraction (helper names excluded, scope-shifting respected).

## Follow-ups in #1950

Next in the sequence: push-time validation (D + E), then render-time strict mode (C), then misrender-retention parity (F), then internal cleanups (G + H).


logfire/variables/composition.py
```diff
@@ -132,9 +130,10 @@ def expand_references(
...
 
     # JSON-decode the serialized value so we can work with actual strings.
-    try:
-        decoded = json.loads(serialized_value)
-    except (json.JSONDecodeError, TypeError):
+    decoded = _safe_json_load(serialized_value)
+    if decoded is None and not serialized_value.lower().startswith('null'):
+        # The string isn't valid JSON at all — bail out rather than try to
+        # render an invalid value.
         return serialized_value, composed
```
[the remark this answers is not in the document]
dmontagu: Fixed in 9a44b699 — tightened the check to `serialized_value.strip() != 'null'`. JSON `null` is always lowercase, and exact equality avoids the `'nullify'` false positive. I dropped the `isinstance(..., str)` guard from the suggestion because pyright (correctly) flags it as redundant: the parameter is already typed `str`, so the type contract is the right place to enforce that.

logfire/variables/_handlebars.py
alexmojaki: do we need this dep check over and over

logfire/variables/composition.py
```diff
@@ -32,19 +32,17 @@
     'has_references',
 )
 
-# Matches unescaped @{ (not preceded by \).
-# In JSON-serialized strings, a real backslash is \\, so \\@{ is an escaped ref.
+# Matches unescaped @{ (not preceded by \). Used as a cheap gate so we only
+# parse strings that actually contain composition syntax. Real reference
+# extraction goes through `pydantic_handlebars.extract_dependencies` so block
+# helpers, dotted paths, and subexpressions are all handled AST-correctly.
 _HAS_REFERENCE = re.compile(r'(?<!\\)@\{')
```
alexmojaki: This gate is coupled to the current pydantic-handlebars escaping behavior, where any backslash immediately before `@{` escapes the tag. If pydantic-handlebars is changed to match Handlebars.js double-escape behavior, this regex becomes incorrect: `\\@{x}@` should contain a real reference with a literal leading backslash, but `(?<!\\)@\{` would still return false because the `@` is immediately preceded by `\`. So either this should be explicitly treated as a temporary approximation of the current renderer behavior, or the gate should be parity-aware / delegated to the parser so it keeps matching the renderer if escaping is fixed upstream.

alexmojaki: here's the problematic behaviour in pydantic-handlebars:

```python
from pydantic_handlebars import render

context = {'x': 'X'}

assert render(r'{{x}}', context) == 'X'
assert render(r'\{{x}}', context) == r'{{x}}'
assert render(r'\\{{x}}', context) == r'\{{x}}'
assert render(r'\\\{{x}}', context) == r'\\{{x}}'
```

compare with https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22%7B%7Bx%7D%7D%20%5C%5C%7B%7Bx%7D%7D%20%5C%5C%5C%5C%7B%7Bx%7D%7D%5Cn%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%5Cn%20%20x%3A%20%5C%22X%5C%22%5Cn%7D%5Cn%22%2C%22output%22%3A%22X%20%7B%7Bx%7D%7D%20%5C%5CX%5Cn%22%2C%22preparationScript%22%3A%22%22%2C%22handlebarsVersion%22%3A%224.7.9%22%7D

alexmojaki: in other words, you can't currently 'double escape', i.e. have a literal backslash right before a template field, regardless of delimiter.

logfire/variables/reference_syntax.py
```diff
@@ -1,66 +1,29 @@
...
-        .replace(right_runtime_placeholder, '}}')
-        .replace(escaped_reference_start, '@{')
-    )
+    r"""Single-pass `@{...}@` Handlebars render.
+
+    `{{...}}` runtime placeholders in *template* are not touched — they
+    are plain content under the configured delimiters. The escape
+    sequence `\@{` produces a literal `@{` in the output.
+    """
+    return get_environment().render(template, context)
```
alexmojaki: Since this path is now native pydantic-handlebars, it may be worth caching compiled templates here rather than only caching the environment. `HandlebarsEnvironment.render()` calls `compile(source)` on every call, and `compile()` reparses the source each time. For stable managed-variable values, a small bounded cache around `get_environment().compile(template)` would avoid repeated parsing when the same composed value is resolved multiple times, e.g. `@lru_cache(maxsize=1024)` returning a `CompiledTemplate` and then calling `.render(context)`. The same idea may also apply separately to dependency extraction / `{{...}}` template rendering, but this wrapper is the easiest place to avoid repeated parse work on the composition render path.

dmontagu:
Closing in favor of #1954

PR #1953, opened 2026-05-22
Validate template inputs schemas at push/sync time

dmontagu:
Second in the #1950 series, stacked on #1952. Implements items D and E.

Wires the previously-orphaned `logfire.variables.template_validation.validate_template_composition` into the push/validate paths so a template variable's `{{field}}` references are checked against its declared `inputs_type` JSON schema before sync.

## What gets checked

For each `TemplateVariable`, the existing validator walks the composition graph and runs every template string through `pydantic_handlebars.check_template_compatibility` against the variable's `inputs_type` JSON schema. Coverage:

- **Local code defaults** authored against an incompatible schema.
- **Server-stored label values + latest version** authored against an older or different schema (item E — server template references a field the current local `inputs_type` doesn't declare).
- **Composed-in fragments**: an incompatible `{{field}}` inside a variable reached via `@{ref}@` is reported with the composition path so the reader can see *why* the template is implicated.

## How it surfaces

- `VariableDiff.template_field_issues` and `ValidationReport.template_field_issues` are new `list[TemplateFieldIssue]` fields populated by a new internal helper `_collect_template_field_issues(variables, server_config)`.
- `_format_diff` and `ValidationReport.format()` render the issues in a dedicated section showing the field name, the variable + label where it was found, and (when relevant) the composition path.
- `push_variables(strict=True)` fails when any issues are present, with guidance to fix the template or update the `inputs_type`. Without `strict`, issues are displayed but the push proceeds.
- `ValidationReport.is_valid` returns False when issues are present.

## Tests

Three new cases in `tests/test_push_variables.py`:

- `test_compute_diff_template_field_issues_local_default` — local code default references undeclared field.
- `test_compute_diff_template_field_issues_server_label` — server-stored label authored against an older schema.
- `test_compute_diff_template_field_issues_follow_composition` — incompatible field inside a fragment reached via `@{...}@`, with `reference_path` reported.

The new helper only iterates over `TemplateVariable` instances. `TemplateVariable` can only be constructed on Python 3.10+ (it requires `pydantic-handlebars`), so on 3.9 the helper is a no-op and the test cases are skipped.

## Series

Next: PR 3 (render-time strict mode + config plumbing — item C), then PR 4 (misrender retention parity — item F), then PR 5 (internal cleanups — items G and H).


dmontagu:
Addressed in 893bbe0c by adding an `is_handlebars_available()` guard and making `_collect_template_field_issues` a no-op when pydantic-handlebars is missing. Same change also fixes the Python 3.9 CI failure (`test_compute_diff_template_inputs_schema_change` was triggering the import via `_compute_diff`).

logfire/variables/abstract.py
```diff
@@ -530,6 +567,73 @@ def _check_type_label_compatibility(
...
+        result: dict[str | None, str] = {}
+        server_var = server_config.variables.get(name)
+        if server_var is not None:
+            for label, labeled in server_var.labels.items():
+                if isinstance(labeled, LabeledValue):
+                    result[label] = labeled.serialized_value
+            if server_var.latest_version is not None and None not in result:
+                result[None] = server_var.latest_version.serialized_value
+        local_var = locals_by_name.get(name)
+        if local_var is not None and None not in result and not is_resolve_function(local_var.default):
```
alexmojaki: P2: This skips validating the local code default whenever the server has a `latest_version`. `result[None]` is populated from `latest_version` above, then this branch is skipped because of `None not in result`.

That means the exact value that can be served at runtime is not checked: rollout can still serve code default even when `latest_version` exists (for example, empty rollout / UI showing 100% code default). I reproduced this with a remote variable whose latest version is valid (`{{user_name}}`) but whose local default is `Local default is invalid: {{local_missing}}`: `variables_validate()` reports `template_field_issues=[]`, `push_variables(strict=True, dry_run=True)` proceeds, and `get()` returns `Local default is invalid: `.

We probably need to validate the local default as its own source instead of sharing the `None` key with `latest_version`, or otherwise always include it when code default can be served.

logfire/variables/abstract.py
```diff
@@ -530,6 +567,73 @@ def _check_type_label_compatibility(
...
+        (under the ``None`` key) when one is available. ``None`` represents
+        "the latest version" in `validate_template_composition`'s contract;
+        we reuse it to mean "the value we'd render in the absence of a
+        provider hit" — i.e. the code default.
+        """
+        result: dict[str | None, str] = {}
+        server_var = server_config.variables.get(name)
+        if server_var is not None:
+            for label, labeled in server_var.labels.items():
+                if isinstance(labeled, LabeledValue):
```
alexmojaki: Minor reporting point: `LabelRef`s are skipped here, so when labels point at `latest`, template field issues are reported against `found_in_label=None` rather than against the labels that can serve the value.

Example remote state:

```python
labels = {
    'v1': LabeledValue(version=1, serialized_value='"{{user_name}} 1"'),
    'v2': LabelRef(version=2, ref='latest'),
    'latest': LabelRef(version=2, ref='latest'),
}
latest_version = LatestVersion(version=2, serialized_value='"{{user_name}} 2"')
```

With local `inputs_type` containing `user_name2` but not `user_name`, validation reports `v1` plus `found_in_label=None`. The second issue is technically the version-2/latest value, but the user-facing labels `v2` and `latest` are what can serve it. Not a correctness miss, just potentially confusing output; it might be clearer to follow `LabelRef`s for reporting or include the labels that point at the invalid version.

tests/test_push_variables.py
```diff
@@ -499,6 +610,216 @@ def test_push_variables_strict_fails_with_reference_errors(mock_logfire_instance
...
+    prompt = TemplateVariable[str, Inputs](
+        name='prompt',
+        default='@{fragment}@ {{user_name}}',
+        type=str,
+        inputs_type=Inputs,
+        logfire_instance=mock_logfire_instance,  # type: ignore
+    )
+    # `fragment.default = object()` is not JSON-serializable; the walker should swallow
+    # the dump_json error and continue.
+    fragment = Variable[object](
```
alexmojaki: Why isn’t this `logfire.var(...)`?

alexmojaki: (AI) Addressed in 634a33c7

dmontagu:
Consolidated into #1954. This PR's push/sync template validation now lives in #1954 (structured `TemplateFieldIssue` / `template_field_issues` with composition `reference_path`, adopted as the convergence over the interim string-based version and kept alongside #1954's `reference_cycles` distinction). Integrated in 744dec4e; review fixes (RecursionError guard on the graph-build walk, cross-root dedup, multi-hop path display, docs) in b11c187d. GitHub auto-marked this merged since its commits are now contained in #1954's branch.

PR #1954, opened 2026-05-22
Composition and templating with native handlebars

alexmojaki:
Adds two composable capabilities to managed variables — **composition** (`@{other_variable}@`) and **templating** (`{{placeholder}}`) — built on the [`pydantic-handlebars`](https://github.com/pydantic/pydantic-handlebars) engine. Both are especially useful for AI applications, where a prompt is assembled from reusable fragments and personalized with per-request inputs, and edited from the UI without a redeploy.

## New public API

- **`logfire.template_var(name, *, type=..., default=..., inputs_type=..., description=None, template_mismatch_policy=None)`** — defines a variable whose value contains Handlebars `{{placeholder}}` expressions. `get(inputs)` resolves, composes, renders, and deserializes in one call. `inputs_type` (a Pydantic model or any `TypeAdapter`-supported type) drives type-safe `get(inputs)` calls and generates a `template_inputs_schema` that is synced to the backend.
- **`@{variable_name}@` composition syntax** — usable in any variable's value (plain `var()` or `template_var()`); references are expanded at resolution time.
- New exports: `TemplateVariable`, `ComposedReference`, `TemplateInputsMismatchError`, `VariableCompositionError`, `VariableCompositionCycleError`, `TemplateMismatchPolicy`.

## Resolution pipeline

Every `get()` now runs a four-stage pipeline:

1. **Resolve** — fetch the serialized value from the provider (or use the code default).
2. **Compose** — expand `@{ref}@` references.
3. **Render** — render `{{placeholder}}` templates against the typed inputs (`TemplateVariable` only).
4. **Deserialize** — validate to the variable's declared type.

## Composition (`@{variable}@`)

- Runs through the full Handlebars engine with `@{`/`}@` as the delimiter pair, so simple references, dotted field access (`@{brand.product}@`), block helpers (`@{#if user.active}@…@{/if}@`, `@{#each items}@…@{/each}@`), and sub-expressions all work. `{{…}}` runtime placeholders are left untouched for the render pass.
- **Recursive resolution** — a referenced variable is fully resolved (including its own `@{…}@` references) before being spliced in, walking the reference graph leaf-first. This differs from plain Handlebars, which does one-shot substitution.
- **Tracking** — every expansion is recorded on `ResolvedVariable.composed_from` (a tree of `ComposedReference` with name/value/label/version/reason and nested entries) and mirrored into span attributes, so the full composition chain is visible in traces.
- **Failure modes:**
  - A **provider/stored value (and an `override()`) is composed strictly**: an unresolved `@{ref}@`/`@{ref.field}@`, an undecodable reference, a cycle (`A → @{B}@`, `B → @{A}@`), or a depth overflow (`MAX_COMPOSITION_DEPTH = 20`) discards the value and falls back to the variable's **code default**, with a `RuntimeWarning`, `reason='other_error'`, and the triggering exception on `ResolvedVariable.exception`. So a value that can't fully compose is replaced by a known-complete one rather than served half-assembled.
  - The **code default is the lenient last resort**: when it is composed (as the active value, or as a fallback target) a missing `@{ref}@` renders as an **empty string** and `@{#if missing}@` is falsy — standard Handlebars, since there is nowhere further to fall back to. A `RuntimeWarning` still names it; only a structural failure (cycle/parse) in the code default itself falls back one more step, to the raw uncomposed default.
- The `\@{…}@` escape sequence yields a literal `@{…}@`.

## Templating (`{{field}}`)

- Standard Handlebars: interpolation, dot-notation, `{{#if}}`/`{{#unless}}`/`{{#each}}`/`{{#with}}`, comments. HTML auto-escaping is disabled (these are prompts/config, not HTML).
- **Structured types** render in place: only string fields containing `{{…}}` are rendered; other fields pass through unchanged.
- **Render-time mismatch policy** (`template_mismatch_policy`) controls what happens when the resolved (post-composition) template references a `{{field}}` not declared in `inputs_type`:
  - `'warn'` (default) — emit a `RuntimeWarning` and render anyway (undeclared field → empty string, matching Handlebars).
  - `'error'` — raise `TemplateInputsMismatchError` instead of rendering.
  - `'ignore'` — render silently.

  Configurable per-variable on `template_var()` or per-instance on `VariablesOptions` / `LocalVariablesOptions`. The variable-level value wins when set (even when it relaxes the instance setting); otherwise the instance setting applies, falling back to `'warn'`.

## Push / sync and validation

- `template_inputs_schema` is synced alongside the variable's JSON schema on `variables_push()` / `variables_push_config()` and persisted on the backend.
- `variables_push()` and `variables_validate()` now walk the **full reachable composition graph** (local code defaults *and* server-stored label values, including the `latest` version) and report:
  - **reference errors** — `@{ref}@` pointing at a variable that doesn't exist anywhere in the push or on the server.
  - **reference cycles** — including cycles whose midpoints are server-only variables.
  - **template field issues** — `{{field}}` references (in a value or anywhere reachable through composition) that aren't declared in the owning `TemplateVariable`'s `inputs_type`, reported with the composition path that led to them.
- **Strict semantics:** cycles always block a push (they can never resolve). Missing references and template-field issues block under `strict=True` and warn otherwise (they may legitimately resolve in another environment, and an undeclared field renders to an empty string). `ValidationReport` / `VariableDiff` gain `reference_errors`, `reference_cycles`, and `template_field_issues`; their `format()` output and the push diff render these sections.

## Overrides participate in composition

`Variable.override(value)` now runs a JSON-serializable override through the same compose → render → deserialize pipeline as a stored value, so an override is a faithful stand-in for a candidate stored value (e.g. during iterative optimization): `@{ref}@` references are expanded against the live config, and `{{…}}` rendering still applies for `TemplateVariable`. A non-serializable override (an arbitrary Python object) is returned verbatim under `reason='context_override'`, with no round-trip.

## Default-latest fallback

When a variable's rollout routes to `latest` but no version exists yet, resolution falls back to the **code default** with `reason='code_default'`, `label=None`, `version=None`, and `<code_default>` baggage — so a variable configured to serve `latest` returns its code default until the first version is created, then serves it automatically with no targeting change. `'latest'` and `'code_default'` are documented as reserved label names.

## Other changes

- **Validation hardening:** resolution warnings are filter-independent (a `filterwarnings=error` config can't turn an informational warning into a swallowed `other_error`); validation failures warn before falling back to the code default; callable code defaults are memoised per `get()` so a failing/expensive default isn't invoked twice.
- **`var()` / `template_var()` now reject `default=None` without an explicit `type`** (the value type can't be inferred from `None`).
- **Remote provider robustness:** SSE reconnect backoff resets only after data actually arrives (no tight reconnect loop on a 200-then-close), clean stream end backs off before reconnecting, fork re-init resets shutdown state, a failed initial fetch no longer blocks every resolve, `refresh(force=True)` guarantees read-after-write, and variable-type list/upsert surface network errors as `VariableWriteError`.
- **Local provider** returns a deep-copied config snapshot so concurrent writes can't raise "dictionary changed size during iteration" during a push diff.
- **Scrubbing:** adds the `pylf_v\d+_` default pattern.

## Dependencies

The `logfire[variables]` extra now requires `pydantic-handlebars>=0.2.1`. `logfire.template_var()` raises immediately if the package is unavailable, so an application never silently uses an unrendered template; plain `import logfire` still works without it.

## Docs & examples

- New **Templates and Composition** reference page; updates to the managed-variables overview, configuration reference, remote, and scrubbing docs.
- New runnable `examples/python/variable_composition_demo.py` covering composition, nested/subfield references, template rendering, conditionals/iteration, rollout overrides, and baggage propagation against a purely local config.


logfire/variables/variable.py
```diff
@@ -187,166 +201,247 @@ def refresh_sync(self, force: bool = False):
...
+        if context_overrides is not None and name in context_overrides and variable is not None:
+            override_value = context_overrides[name]
+            if is_resolve_function(override_value):
+                override_value = override_value(targeting_key, attributes)
+            try:
+                serialized = variable.type_adapter.dump_json(override_value).decode('utf-8')
+            except (ValueError, TypeError, RuntimeError):
+                pass  # Fall through to provider/code default
+            else:
+                return ResolvedVariable(name=name, value=serialized, reason='context_override')
```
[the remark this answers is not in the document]
alexmojaki: already raised in https://github.com/pydantic/logfire/pull/1951

logfire/variables/composition.py
```diff
@@ -0,0 +1,395 @@
...
+    syntax through `pydantic_handlebars.extract_dependencies`, so block
+    helpers (`@{#if var}@`), dotted paths (`@{var.field}@`), and
+    subexpressions (`@{lookup obj key}@`) are all picked up correctly.
+
+    Args:
+        serialized_value: The raw JSON-serialized variable value to scan.
+
+    Returns:
+        List of unique top-level variable names referenced, in order of
+        first occurrence.
```
alexmojaki: the 'in order' part is not quite true:

```python
from logfire.variables.composition import find_references

print(find_references('"@{foo}@ @{bar}@"'))
print(find_references('"bar: @{foo}@ @{bar}@"'))

# ['foo', 'bar']
# ['bar', 'foo']
```

I say drop this line/requirement, simplify the code, let the result be out of order, sort alphabetically in tests.

logfire/variables/variable.py
```diff
@@ -187,166 +201,247 @@ def refresh_sync(self, force: bool = False):
...
+            provider_result = provider.get_serialized_value(name, targeting_key, attributes)
 
-            if serialized_result.value is None:
-                # Provider had no value; surface that the code default was used.
+        if provider_result.value is not None:
+            return provider_result
+
+        # 3. Registered code default
+        if variable is not None:
+            serialized_default = variable._get_serialized_default(targeting_key, attributes)
```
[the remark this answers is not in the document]
dmontagu: Fixed in #1964 (stacked on this PR). Introduced `_DEFAULT_CACHE: ContextVar[dict[int, T] | None]` and `Variable._get_default_cached(...)`; `_resolve` sets an empty cache at the top of each `get()` and resets on exit, and every fallback path (`_fallback_to_default`, the deserialize-error branch, `_resolve_code_default`) goes through the cached helper. First call populates, subsequent calls in the same `get()` return the cached value. Outside `_resolve` the cache is `None` and the helper is a passthrough.

New regression test `test_callable_default_invoked_once_on_composition_failure` forces provider-empty + callable-default + composition-failure (via monkeypatched `expand_references`) and asserts the callable ran exactly once.

logfire/variables/abstract.py
```diff
@@ -303,10 +405,16 @@ def format(self, *, colors: bool = True) -> str:
...
+        if self.reference_errors:
+            lines.append(f'\n{red}=== Reference errors ==={reset}')
+            for error in self.reference_errors:
+                lines.append(f'  {red}✗ {error}{reset}')
+
         # Summary line
         if not self.is_valid:
-            error_count = variables_with_errors + len(self.variables_not_on_server)
-            lines.append(f'\n{red}Validation failed: {error_count} error(s) found.{reset}')
+            issue_count = variables_with_errors + len(self.variables_not_on_server) + len(self.reference_errors)
```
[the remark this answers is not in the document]
dmontagu: Fixed in #1964 (stacked on this PR). Added `and self.is_valid` to the "Valid" section's gate so it only renders for fully-clean reports. Reference-error / template-field-issue reports no longer show the contradictory pair.

logfire/variables/composition.py
dmontagu: Done in #1964 (stacked on this PR). Dropped the ordering claim entirely — `find_references` now returns `set[str]`. The internal `_order_by_first_position` helper was a regex word-boundary scan that matched names in literal content (your `'bar: @{foo}@ @{bar}@'` repro reordered them); we'd need to walk the AST for true source-order, which the parser is the only thing that can do correctly. Tests that asserted list-order now use set comparison or `{ref.name for ref in composed}`.

In-SDK callers (`abstract.py`, `template_validation.py`) all consume the result via `.update()` or iteration, so the API switch from `list` to `set` is observable but not breaking.

logfire/variables/variable.py
```diff
@@ -356,6 +557,62 @@ def _get_default(
...
+            # Defensive: every production call site is inside `_resolve`,
+            # which sets the cache. Falling back to a direct compute keeps
+            # the helper safe if someone reaches in from an unexpected
+            # entry point in the future.
+            return self._get_default(targeting_key, merged_attributes)
+        key = id(self)
+        if key not in cache:
+            cache[key] = self._get_default(targeting_key, merged_attributes)
+        return cache[key]
+
```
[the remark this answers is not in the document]
dmontagu: Fixed in #1965. Cache now records `(ok: bool, value_or_exception: Any)` — failing callables cache the exception and re-raise on subsequent lookups without re-invoking. New regression test `test_failing_callable_default_invoked_once_per_get` asserts a raising default fires exactly once per `get()` (was 3 without the fix).

tests/test_variable_templates.py
```diff
@@ -0,0 +1,553 @@
...
+
+        # `model_construct` bypasses the constructor's validation so the override can hold
+        # the unrendered template — `templated_config` is itself a *valid* template; it's
+        # only the rendered result (`abc123`) that violates the pattern constraint.
+        # Exercises the outer error handler's code-default fallback for inputs that
+        # produce a constraint-violating render.
+        templated_config = Config.model_construct(code='{{code}}')
+        invalid_inputs = Inputs(code='abc123')
+        with var.override(templated_config):
+            resolved = var.get(invalid_inputs)
```
alexmojaki: The point about the warning in https://github.com/pydantic/logfire/pull/1951#discussion_r3289312130 remains

dmontagu: Addressed in the latest push: composition and render failures already warned before falling back to the code default, but a successfully-rendered value (or a context override that renders to an invalid value) that then failed validation fell back silently. `Variable._expand_and_deserialize` and `_resolve_context_override` now emit a `RuntimeWarning` ("value failed validation; falling back to code default") in those paths too, so the substitution is observable. Tests in `test_variables.py`/`test_variable_templates.py` updated to assert the warning. Resolving.

logfire/variables/composition.py
```diff
@@ -0,0 +1,389 @@
...
+    Fatal failures (a cycle or depth overflow) are structural — the value can't be
+    meaningfully composed at all — so a consumer should discard the value and fall back.
+    Non-fatal failures (an unresolved/missing reference, or a referenced variable with a
+    malformed value) are *soft*: the literal `@{ref}@` text is left in place and the rest
+    of the value still renders, so the partially-composed value is usable. Always `False`
+    when *error* is `None`.
+    """
+    composed_from: list[ComposedReference] = field(default_factory=list)  # pyright: ignore[reportUnknownVariableType]
+    """Nested references that were expanded within this reference."""
+
```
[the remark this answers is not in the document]
dmontagu: Good catch — `composed_from` was the last field at file creation, and inserting `fatal` ahead of it did shift its positional index. All internal call sites use keyword args (so nothing here breaks), and field order doesn't affect serialization or the `fatal`/`composed_from` consumers, but since this module is shared verbatim with the backend OFREP endpoint, keeping a stable field order is worth it. Moved `fatal` to after `composed_from` as suggested.

docs/reference/advanced/managed-variables/configuration-reference.md
```diff
@@ -131,6 +132,31 @@ def test_premium_config_handling():
     # Back to normal after context exits
 ```
 
+#### Overrides and composition
```
alexmojaki, line 135: This section is mostly stating what the user would naturally expect and is clearly AI feeling the need to document fixing this when it didn't behave like this initially. only the non-serializable bit is interesting, and could actually do with a bit of expanding.

docs/reference/advanced/managed-variables/configuration-reference.md
```diff
@@ -14,7 +14,8 @@
 | `json_schema` | JSON Schema for validation (optional) |
 | `description` | Human-readable description (optional) |
 | `aliases` | Alternative names that resolve to this variable (optional, for migrations) |
-| `example` | JSON-serialized example value, used as template in UI (optional) |
+| `example` | JSON-serialized example value, used as starting point when creating versions in the UI (optional) |
+| `template_inputs_schema` | JSON Schema for template `{{placeholder}}` inputs (optional, set automatically when template inputs are declared via `logfire.template_var()`) |
```
alexmojaki, line 18: I see 0 support for template_inputs_schema in the backend. Because of this, `logfire.variables_push()` always thinks that there are changes to apply when there's a template var, whereas with normal vars it can recognise when there's nothing new to push.

logfire/_internal/main.py
```diff
@@ -142,6 +144,7 @@
     ExcInfo = SysExcInfo | BaseException | bool | None
 
 T = TypeVar('T')
+InputsT = TypeVar('InputsT')
```
alexmojaki, line 147: i think this should have bounds on it restricting it to mappings and basemodels, and/or runtime warnings when declaring a template_var with an invalid inputs_type

logfire/variables/abstract.py
```diff
@@ -142,6 +158,96 @@ def __exit__(self, exc_type: type[BaseException] | None, exc_val: BaseException
...
+            a `dict`, or any `Mapping`. If `None`, returns an empty dict.
+
+    Returns:
+        A dict suitable for use as a Handlebars template context.
+
+    Raises:
+        TypeError: If inputs is not a supported type.
+    """
+    if inputs is None:
+        return {}
```
alexmojaki, line 179: Do we actually want this None fallback?

logfire/_internal/main.py
alexmojaki: actually, should we be allowing inputs to be other objects like dataclasses? in particular we could either use anything that a pydantic.TypeAdapter[Any] can dump, which has well defined semantics, or we could use logfire's json encoding to handle values more broadly. either way i guess we probably can't check types in that case.

dmontagu: I was happy to allow arbitrary pydantic-serializable/deserializable types. Not sure if there's a reason not to other than the fact that you can't check that someone used such a type as the parameter, but I think that's fine, in practice I can't imagine someone wanting to use a managed variable for something non-serializable for the obvious reason that — how would they plan to make use of the "managed" aspect?

To  be more concrete: yes, I am happy allowing other objects (like dataclasses, or more importantly typeddicts) as the inputs, and I don't think we need to change anything here — Python's type system just doesn't provide the expressiveness necessary to constrain the typevar in a way that I think is worthwhile here. (In particular, I'm pretty sure there's no way to say "this typevar should allow any TypedDict, but not arbitrary types".)

Either way, debating this point is definitely not worth delaying a merge, imo, we can change typing semantics later. I'm okay allowing arbitrary types as inputs and relying on pydantic `to_json`/`to_jsonable_python` or similar.

docs/reference/advanced/managed-variables/configuration-reference.md
dmontagu: 🤷‍♂️ I think it's worth including these notes, I read this and was happy with it (despite having told it not to do the thing where it documents changes to unmerged behaviors in various other places). Imo it's fairly easy to imagine the override behavior being different than this, maybe I'm biased as someone so involved with the implementation, but I feel like it's worth a note on these points. Maybe a more concise note?

logfire/variables/abstract.py
dmontagu: I can see some marginal benefits (being able to use optional items and then using {{#if ...}} for the rendering) and it doesn't seem to have much downside, so I'm inclined to retain the fallback. (Maybe the optional type thing is broken now, not sure, but I don't see downside to having things in place to better support that regardless of whether it's fully supported today?)

So I would suggest we retain it for now.

logfire/variables/variable.py
```diff
@@ -187,166 +273,392 @@ def refresh_sync(self, force: bool = False):
...
+                attributes=attributes,
+                serialized_result=serialized_result,
+                composed=composed,
+            )
 
+        # Soft, per-reference failures (a missing/unresolvable `@{ref}@`, or a referenced
+        # variable with a malformed value) are NOT fatal: composition leaves the literal
+        # `@{ref}@` text in place and the rest of the value still renders. Keep the
+        # partially-composed value rather than discarding the whole thing for the code
+        # default, but warn so the unexpanded reference isn't silent (the per-reference
```
alexmojaki: this is surprising and differs from what i thought we agreed on, why is it not falling back to code default?

dmontagu: 1. I thought we discussed and agreed that showing the @{ref}@ was more useful for debugging, and it was a bit more likely to go undetected if we just immediately fall back to the code default, maybe I misremembered.
2. Now that we support composition in the code default, you end up kicking the can down the road, you can't just fall back to code default for missed references in that case. I feel like the minor improvement consistency and debugability is worth the degradation in behavior when the variables are misconfigured.

alexmojaki: Codex's reading of the huddle:

- David argued several times that preserving the unresolved syntax is more debuggable.
- Alex did not strongly object to that early on, but also pushed to avoid expanding PR scope.
- Later, Alex’s explicit preference was: favor successfully rendering some template, use strict rendering for the server template, and fall back to the code-default template when that fails.
- David pushed back on predictability/debuggability, but the conversation ends with “I’ll note that opinion,” not with agreement.

So I’d summarize it as:

> The transcript records a disagreement/tradeoff, not a settled agreement. The PR behavior matches David’s debugging preference, but it does not match Alex’s final stated preference of falling back to a successfully rendered code default when the server-side value/template is incompatible.

logfire/variables/variable.py
alexmojaki, line 1: This code:

```python
import logfire

bar = logfire.template_var('bar111', default="Hello {{name}}", inputs_type=dict)
foo = logfire.var('foo111', default="Foo: @{bar111}@")

print(foo.get().value)
```

happily prints `Foo: Hello {{name}}`. A non-template `var` referencing a `template_var` should probably emit a warning at declaration time since there's no good reason to do that which I can think of since there will be no template inputs, and probably represents a user mistake. not a blocker.

logfire/variables/variable.py
alexmojaki: so i was also misremembering agreement.

1. why does the missing var case differ from other errors, where it falls back to the code default?
2. we can do various things to increase visibility and debuggability, like showing warnings in logfire. if the user doesn't notice those warnings, they're likely to have a hard time figuring out the fix based purely on the presence of `@{ref}@`. exposing `@{ref}@` pretty much just tells the user that they should go looking for warnings to explain the problem unless it's something like an obvious typo.
3. in response to 2: we can warn (possibly in a future PR) at declaration time if your code defaults don't compose, which should help surface that kind of problem at development time.

logfire/variables/template_validation.py
```diff
@@ -0,0 +1,219 @@
...
+)
+
+# Matches {{identifier}} and {{path.to.identifier}} Handlebars variable references.
+# Excludes block helpers ({{#if}}), closing tags ({{/if}}), partials ({{> name}}),
+# and comments ({{! text}}).
+TEMPLATE_FIELD_PATTERN = re.compile(r'\{\{\s*([a-zA-Z_]\w*(?:\.[a-zA-Z_]\w*)*)\s*\}\}')
+
+
+@dataclass
+class TemplateFieldIssue:
```
alexmojaki, line 34: I think this issue object is losing the most important attribution when a template variable composes another template variable.

Repro from `scratch_3001.py`:

```py
class Bar(pydantic.BaseModel):
    name: str

class Bar2(pydantic.BaseModel):
    name2: str

bar = logfire.template_var('bar111', default='Hello {{name2}}', inputs_type=Bar2)
foo = logfire.template_var('foo111', default='Foo: @{bar111}@', inputs_type=Bar)

print(logfire.variables_validate().format())
```

Current output:

```text
=== Template field issues ===
  ✗ {{name2}} in bar111 via @{bar111}@ is not declared in the inputs_type schema
```

That makes it sound like `bar111` is invalid, but `bar111` is valid against its own `inputs_type=Bar2`. The problem is specifically that `foo111` composes `bar111`, so `bar111`'s `{{name2}}` is being checked against `foo111`'s `inputs_type=Bar`, which only has `name`.

I think `TemplateFieldIssue` needs to carry the root/validated template variable whose schema was applied, e.g. `root_variable='foo111'` (and possibly root label), and the formatted output should name it:

```text
✗ foo111: {{name2}} found in bar111 via @{bar111}@ is not declared in foo111's inputs_type schema
```

Related: `_collect_template_field_issues()` currently dedups across roots on `(field_name, found_in_variable, found_in_label)`. That can collapse multiple affected parents into one issue, again hiding which template variable(s) are actually broken by composing the shared fragment.

docs/reference/advanced/managed-variables/configuration-reference.md
dmontagu: You need to be looking at https://github.com/pydantic/platform/pull/21424, I believe template_inputs_schema etc. is handled correctly on that branch; we need to release both together.

logfire/variables/variable.py
dmontagu: makes sense, eventually we could probably go further and warn for incompatibility on inputs on composed values (across any variants of the relevant variables). I don't think we need to do that yet (unless it's straightforward) but I agree that having a template var referenced from a non-template-var is bad. We should call this out in the docs as something you should not do, and further discuss the potential issues with template input type incompatiblity when using composition like this.

For now let's just make it emit a warning if you compose a variable with a template_inputs_schema into one that doesn't have this.

logfire/variables/variable.py
dmontagu: > why does the missing var case differ from other errors, where it falls back to the code default?

I felt like this made sense, but if this behavior is what needs to change for you to approve this, I don't want to debate, I just want to get this merged, so let's change it to work the way you are describing — missing variable references lead to fallback to code default, and missing variable references in the code defaults are retained.

If you wanted to match handlebars semantics we could also do that and just have missing references be empty strings but that feels worse enough on most fronts that I prefer (at least until convinced otherwise) to handle it differently.

logfire/variables/template_validation.py
dmontagu: Makes sense, I'm okay making this change if it's straightforward; if it's fundamentally difficult, I don't think this should block merging, but what you are describing does seem better than the current behavior.

logfire/variables/composition.py
```diff
@@ -0,0 +1,460 @@
...
+    Returns `True` whenever the string contains the composition open
+    delimiter, regardless of preceding backslashes. The actual
+    escape-or-real decision is made by `pydantic_handlebars` at render time
+    — distinguishing an escaped `\@{x}@` from an unescaped `@{x}@` here
+    would require a variable-width lookbehind (the renderer counts
+    backslash parity to match Handlebars.js semantics) and is unnecessary:
+    `extract_composition_dependencies` returns an empty set for escaped-only
+    strings, and the renderer correctly leaves the literal text in place.
+    """
+    return _HAS_OPEN_DELIM in serialized_value
```
alexmojaki, line 111: This gate only proves the value contains the substring `@{`, not that it contains a real composition reference. If `pydantic-handlebars` is not installed, runtime cannot make the parse/escape decision described above: a plain `var()` provider value containing literal `@{` enters composition, raises `HandlebarsDependencyError`, and falls back to the code default with `reason="other_error"`, silently dropping the provider value. That seems like a strong reason to require `pydantic-handlebars` whenever managed-variable composition can run, or to leave values untouched unless the parser is available and confirms an actual reference.

docs/reference/advanced/managed-variables/configuration-reference.md
dmontagu: Trimmed this down to the part that isn't obvious and expanded the non-serializable case: such an override is returned exactly as passed (`reason='context_override'`, no compose/render/deserialize round-trip), and any `@{ref}@`/`{{...}}` inside it is left untouched since there's no serialized form to run the engine against. 5f4b3a54

logfire/_internal/main.py
dmontagu: Left `inputs_type` unconstrained — agreed the type system can't usefully express "any TypedDict/dataclass but not arbitrary types". I did make arbitrary types actually work at runtime though: `_inputs_to_context` now serializes any pydantic-serializable inputs (dataclasses, TypedDicts, attrs, …) via `to_jsonable_python` instead of raising `TypeError` on e.g. a dataclass instance. Added unit + end-to-end tests for a dataclass `inputs_type`. 5f4b3a54

logfire/variables/abstract.py
dmontagu: Kept the `inputs=None → {}` fallback. It's consistent with the rest of the rendering model — a missing `{{field}}` renders empty just like a missing composition reference now does — and keeps optional inputs usable via `{{#if ...}}`. No change. 5f4b3a54

logfire/variables/variable.py
dmontagu: Reworked per our discussion. Provider/stored values (and overrides) are now composed in **strict** mode: an unresolved `@{ref}@`/`@{ref.field}@`, a cycle, or a depth overflow discards the value and falls back to the code default. The **code default is the lenient last resort** — a missing ref in it renders empty (standard Handlebars), since there's nowhere further to fall back. Templates `{{}}` are unchanged. The literal-preservation machinery is gone; this rides on pydantic-handlebars' `strict` flag with no new config. Docs + tests updated. 5f4b3a54

logfire/variables/variable.py
dmontagu: Added a declaration-time `RuntimeWarning` when a variable without `inputs_type` composes a `template_var`, checked in both directions so it fires regardless of which is declared first. Also called this out in the Templates & Composition docs as something to avoid. 5f4b3a54

logfire/variables/template_validation.py
dmontagu: `TemplateFieldIssue` now carries `root_variable`, and the formatted message names it, e.g. `foo111: {{name2}} found in bar111 via @{bar111}@ is not declared in foo111's inputs_type schema`. Also fixed the cross-root dedup so a shared fragment incompatible with several roots is reported once per affected root instead of being collapsed to a single line. 5f4b3a54

logfire/variables/composition.py
dmontagu: Fixed in 6db846de. Rather than special-case the missing-parser path, using managed variables now requires the `logfire[variables]` extra, checked eagerly: `logfire.var()` / `template_var()` (and the public `logfire.variables` config classes, via `__getattr__`) raise a clear `ImportError` if `pydantic` / `pydantic-handlebars` aren't installed. So the silent drop can't happen — a value can't reach composition without the parser present. Plain `import logfire` and the rest of the SDK keep working without the extra (the check runs only on *use*, not at import). The now-redundant `ensure_handlebars_available()` is gone, and the `@{` substring gate is harmless since the parser is always available to make the real escape-vs-reference decision.

alexmojaki:
@cubic-dev-ai review

PR #1957, opened 2026-05-22
Require handlebars for template field validation

alexmojaki:
## Summary

- remove the Python 3.9-era no-op path from template field validation
- require the handlebars-backed validator to run when collecting template field issues
- remove stale skip decorators from the template field validation push tests

## Tests

- `uv run pytest tests/test_push_variables.py -q`
- `uv run ruff check logfire/variables/_handlebars.py logfire/variables/abstract.py tests/test_push_variables.py`
- `uv run pyright logfire/variables/_handlebars.py logfire/variables/abstract.py tests/test_push_variables.py`

PR #1958, opened 2026-05-22
Add template_mismatch_policy for render-time {{field}} mismatches

dmontagu:
Third in the #1950 series, stacked on #1953. Implements item C: a configurable render-time policy for `TemplateVariable.get(inputs)` when the resolved template references a `{{field}}` the inputs don't declare. Closes the asymmetry Alex flagged on `templates-and-composition.md:36` (composition warned, render path silently substituted empty).

## New knob

`template_mismatch_policy: Literal['warn', 'error', 'ignore']`:

- `'warn'` (default): emit a `RuntimeWarning` and render the missing field as the empty string (Handlebars' default behaviour).
- `'error'`: raise `TemplateInputsMismatchError` instead of rendering.
- `'ignore'`: render silently, no warning.

## Three configuration levels, variable wins

1. **Per-variable** via `template_var(..., template_mismatch_policy='error')`. **Variable-level wins**, even when relaxing — set this when one specific variable should opt out of (or into) strict enforcement without affecting siblings.
2. **Per-Logfire-instance** via `VariablesOptions.template_mismatch_policy` or `LocalVariablesOptions.template_mismatch_policy` (the two carry the field independently — they're peers, not subclasses of each other).
3. Falls back to `'warn'` when nothing is set.

```python
# Tighten one variable while the rest of the app stays lenient
strict_prompt = logfire.template_var(
    'prompt',
    type=str,
    default='Hi {{user_name}}',
    inputs_type=Inputs,
    template_mismatch_policy='error',
)

# Or flip the default at the configure() level
logfire.configure(variables=LocalVariablesOptions(config=..., template_mismatch_policy='error'))
```

## Why a dedicated exception type

For `'error'`, we raise a new `TemplateInputsMismatchError(Exception)` rather than a `HandlebarsError`. The SDK's normal composition-failure fallback (`Variable._resolve`'s outer `except Exception`) catches `HandlebarsError` and friends to degrade gracefully to the code default — but the `'error'` policy is explicitly opt-in *loud* mode, so we route past that fallback. A dedicated `except TemplateInputsMismatchError: raise` in `_resolve` lets the exception escape to the caller.

## Tests

Eight new cases in `TestTemplateMismatchPolicy`:

- Default policy is `'warn'`, fires on mismatch, silent when inputs satisfy.
- Per-variable `'error'` raises `TemplateInputsMismatchError`.
- Per-variable `'ignore'` renders silently.
- Instance-level `'error'` raises.
- Instance-level `'error'` + variable `'ignore'` → renders silently (variable-level relaxes).
- Instance-level `'warn'` + variable `'error'` → raises (variable-level escalates).
- Instance-level `'warn'` + variable `'ignore'` → renders silently.

## Series

Next: PR 4 (misrender retention parity — item F), then PR 5 (internal cleanups — items G and H).


alexmojaki:
coverage will likely be resolved by https://github.com/pydantic/logfire/pull/1957

dmontagu:
Consolidated into #1954. This PR's render-time `template_mismatch_policy` ('warn'/'error'/'ignore') + `TemplateInputsMismatchError` (plumbed via `template_var()` / `VariablesOptions` / `LocalVariablesOptions`) now lives in #1954. Integrated in ff54fdb1 (plus a fix keeping the 'warn' policy correct under `-W error`), with docs/export follow-ups in b11c187d. GitHub auto-marked this merged since its commits are now contained in #1954's branch.

PR #1959, opened 2026-05-22
Add `db.query.text` to scrubber SAFE_KEYS

bendrucker:
Adds `'db.query.text'` to `BaseScrubber.SAFE_KEYS` alongside the deprecated `'db.statement'`.

Upstream OTel semconv replaced `db.statement` with `db.query.text` Until the scrubber's allow list tracks the new name, anyone emitting `db.query.text` loses SQL text to scrubbing whenever a query contains trigger tokens like `privateKey`, `password`, or `secret`.

`'db.statement'` is kept for backwards compatibility .

See the [OTel SQL semconv](https://opentelemetry.io/docs/specs/semconv/db/sql/) for the current canonical attribute names.

### DIFF
--- a/logfire/_internal/scrubbing.py
+++ b/logfire/_internal/scrubbing.py
@@ -140,6 +140,7 @@ class BaseScrubber(ABC):
         'http.target',
         'http.route',
         'db.statement',
+        'db.query.text',
         'db.plan',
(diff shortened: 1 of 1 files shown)


alexmojaki:
Thanks!

PR #1960, opened 2026-05-23
Add update_genai_prices flag to logfire.configure()

JonathanTsen:
## Problem

Pydantic AI uses [`genai-prices`](https://github.com/pydantic/genai-prices) to populate the `operation.cost` attribute on LLM spans. The catalogue mirrors the upstream provider pricing pages — for example [Google Gemini API pricing](https://ai.google.dev/gemini-api/docs/pricing) and [OpenAI API pricing](https://openai.com/api/pricing/) — and is shipped as a snapshot inside the package. So when a brand-new model is released (e.g. `gemini-3.5-flash`) `calc_price` raises `LookupError`, which pydantic-ai then swallows silently in `pydantic_ai/_instrumentation.py`. The result is that newly released models show up in Logfire with correct token counts but **no `operation.cost`**, until a new `genai-prices` release is published, pulled in transitively, and reinstalled.

`genai-prices` itself already exposes an opt-in `UpdatePrices` background updater that refreshes the catalogue from the upstream repo. This PR makes that updater available through a single Logfire configure flag.

## Change

A new opt-in kwarg `update_genai_prices: bool | None = None` on `logfire.configure()`.

- Default: `False` (no behaviour change for existing users).
- When `True`, Logfire instantiates `genai_prices.update_prices.UpdatePrices` and starts the daemon thread that polls upstream every hour.
- Lifecycle is fully managed: stopped automatically on reconfigure and on interpreter exit via `atexit`.
- Skipped under emscripten (no threads).
- If `genai-prices` is not installed (or fails to import — e.g. on pydantic < 2.5 environments missing `pydantic.Tag`), the flag emits a one-shot `UserWarning` and is a no-op.
- Also exposed via the `LOGFIRE_UPDATE_GENAI_PRICES=1` environment variable, following the existing `ConfigParam` pattern.

## Usage

```python
import logfire

logfire.configure(update_genai_prices=True)
logfire.instrument_pydantic_ai()
```

Or via env var:

```bash
LOGFIRE_UPDATE_GENAI_PRICES=1
```

## Pricing source of truth

The model prices used to compute `operation.cost` are maintained in [`pydantic/genai-prices`](https://github.com/pydantic/genai-prices) and mirror the official provider catalogues:

- Google Gemini: <https://ai.google.dev/gemini-api/docs/pricing>
- OpenAI: <https://openai.com/api/pricing/>

This PR doesn't change how the prices are derived — it only lets Logfire users pick up upstream updates without waiting for a `genai-prices` release.

## Files changed

- `logfire/_internal/config_params.py` — new `UPDATE_GENAI_PRICES` `ConfigParam` and registration in `CONFIG_PARAMS`.
- `logfire/_internal/config.py` — kwarg threaded through `configure()`, `_LogfireConfigData`, `_load_configuration`, `LogfireConfig.__init__`/`configure`; lifecycle in `_initialize` via two helpers (`_maybe_start_genai_prices_updater` / `_stop_genai_prices_updater`); new field surfaced in `emit_configuration_span`.
- `logfire-api/logfire_api/_internal/config.pyi` and `config_params.pyi` — regenerated stubs.
- `docs/integrations/llms/pydanticai.md` — new "Keep model pricing up to date" section.
- `tests/test_configure.py` — 6 new tests covering default-off, enabled-starts, env var, graceful `ImportError`, reconfigure-stops-old, and emscripten-skip; existing `test_configuration_span_emitted_when_opted_in` snapshot updated to include the new field.

## Tests

All new tests mock `genai_prices.update_prices.UpdatePrices`, so no network or real threads are needed in CI.

```
$ uv run pytest tests/test_configure.py tests/test_logfire_api.py tests/test_logfire.py
199 passed, 2 skipped
$ make lint
All checks passed
$ make typecheck
0 errors, 0 warnings, 0 informations
```

### DIFF
--- a/docs/integrations/llms/pydanticai.md
+++ b/docs/integrations/llms/pydanticai.md
@@ -68,7 +68,7 @@ To always pick up new prices without waiting for a release, pass `update_genai_p
 to `logfire.configure()`. A daemon thread will refresh the catalogue from upstream every hour
 in the background:
 
-```python
+```python skip-run="true" skip-reason="external-connection"
 import logfire
 
 logfire.configure(update_genai_prices=True)
--- a/docs/integrations/llms/pydanticai.md
+++ b/docs/integrations/llms/pydanticai.md
@@ -61,8 +61,11 @@ reference or the [Pydantic AI docs on instrumenting](https://pydantic.dev/docs/a
 
 Pydantic AI uses [`genai-prices`](https://github.com/pydantic/genai-prices) to populate the
 `operation.cost` span attribute. The model price catalogue is shipped as a snapshot inside the
-package, so newly released models (e.g. `gemini-3.5-flash`, `claude-4`, …) only get a cost
-calculation once a new version of `genai-prices` is released and pulled in.
+package, mirroring the official provider pricing pages (e.g.
+[Google Gemini API pricing](https://ai.google.dev/gemini-api/docs/pricing) and
+[OpenAI API pricing](https://openai.com/api/pricing/)), so newly released models
+(e.g. `gemini-3.5-flash`, `gpt-5.1`, …) only get a cost calculation once a new version of
+`genai-prices` is released and pulled in.
 
 To always pick up new prices without waiting for a release, pass `update_genai_prices=True`
 to `logfire.configure()`. A daemon thread will refresh the catalogue from upstream every hour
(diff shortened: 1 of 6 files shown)


adtyavrdhn:
Hey @JonathanTsen 

Thanks for this, I was going for something similar on [`Pydantic AI`](https://github.com/pydantic/pydantic-ai/pull/4841) and this just made us realize it should just be in `genai-prices`.

The plan would be to then just allow opting in and out from both `Pydantic AI` and `logfire` which would affect their own respective points where they rely on `genai-prices` and `genai-prices` to manage the lifecycle of this thread to update its prices for its consumers. 

I'll have a look at what that might look like on `genai-prices` and follow up on it soon :)

adtyavrdhn:
Thank you @JonathanTsen for identifying this and for the original implementation. I opened #2364 with the updated shared-updater lifecycle, focused tests, and documentation, and credited your work there.

PR #1961, opened 2026-05-23
feat: capture Gemini cache and thinking tokens in instrument_google_genai

JonathanTsen:
## Summary

- Patches upstream `_GenerateContentInstrumentationHelper` to extract `cached_content_token_count`, `thoughts_token_count` and `tool_use_prompt_token_count` from `response.usage_metadata` and emit them as `gen_ai.usage.cache_read.input_tokens`, `gen_ai.usage.details.thoughts_tokens` and `gen_ai.usage.details.tool_use_prompt_tokens`.
- Computes `operation.cost` via `genai-prices` when available (silent failure if the package is missing or the model is unknown).
- Closes the gap where direct `google.genai` users (not going through `pydantic-ai`) were missing cache and thinking metrics in their spans.

## Implementation notes

- Defensive monkey-patch of two upstream methods (`_maybe_update_token_counts`, `create_final_attributes`), following the same pattern already used elsewhere in `logfire/_internal/integrations/google_genai.py`. Wrapped in `try/except` at module load so a future upstream rename keeps the base instrumentor working.
- Methods confirmed present in `opentelemetry-instrumentation-google-genai` 0.7b0; current pinning in `pyproject.toml` is `>= 0.4b0`.
- Streaming uses the upstream "keep last non-zero" semantics — walrus operator naturally ignores `None`/`0`, so partial chunks don't overwrite a final value.
- Unlike Anthropic, the Gemini API's `prompt_token_count` already includes cached tokens, so we expose `cache_read` separately rather than summing.
- `genai-prices` is invoked with `response.model_dump(by_alias=True)` because the extractor expects camelCase JSON keys (`usageMetadata`, `modelVersion`).

## Test plan

- [x] `uv run pytest tests/otel_integrations/test_google_genai.py` — 8 passing (3 new + 5 existing; existing VCR snapshots updated to include the new attributes that real Gemini responses already carry).
- [x] `make lint`
- [x] `make typecheck`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## [unreleased]
+
+Integrations:
+
+* Capture Gemini `cache_read`, `thoughts` and `tool_use_prompt` tokens in `instrument_google_genai`; compute `operation.cost` via `genai-prices` when available.
+
 ## [v4.33.0] (2026-05-13)
 
 CLI:
--- a/docs/integrations/llms/google-genai.md
+++ b/docs/integrations/llms/google-genai.md
@@ -53,10 +53,13 @@ following attributes may appear depending on the response:
 
 - `gen_ai.usage.input_tokens` — total prompt tokens (already includes cached, see below)
 - `gen_ai.usage.output_tokens` — completion tokens
-- `gen_ai.usage.cache_read.input_tokens` — tokens served from context cache (cache hit)
-- `gen_ai.usage.details.thoughts_tokens` — reasoning tokens (Gemini 2.5 / 3.x)
-- `gen_ai.usage.details.tool_use_prompt_tokens` — tokens used for tool definitions
-- `operation.cost` — calculated price in USD (requires [`genai-prices`](https://pypi.org/project/genai-prices/))
+- `gen_ai.usage.cache_read.input_tokens` — tokens served from [context cache](https://ai.google.dev/gemini-api/docs/caching) (cache hit)
+- `gen_ai.usage.details.thoughts_tokens` — [reasoning tokens](https://ai.google.dev/gemini-api/docs/thinking) (Gemini 2.5 / 3.x)
+- `gen_ai.usage.details.tool_use_prompt_tokens` — tokens used for [tool definitions](https://ai.google.dev/gemini-api/docs/function-calling)
+- `operation.cost` — calculated price in USD using the [official Gemini pricing tables](https://ai.google.dev/gemini-api/docs/pricing) via [`genai-prices`](https://pypi.org/project/genai-prices/)
 
 Note that, unlike Anthropic, the Gemini API's `prompt_token_count` already includes
-the cached tokens; Logfire does not sum them again.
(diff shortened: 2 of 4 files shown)


docs/integrations/llms/google-genai.md
```diff
@@ -45,3 +45,21 @@ This creates a span which shows the conversation in the Logfire UI:
...
+- `gen_ai.usage.cache_read.input_tokens` — tokens served from [context cache](https://ai.google.dev/gemini-api/docs/caching) (cache hit)
+- `gen_ai.usage.details.thoughts_tokens` — [reasoning tokens](https://ai.google.dev/gemini-api/docs/thinking) (Gemini 2.5 / 3.x)
+- `gen_ai.usage.details.tool_use_prompt_tokens` — tokens used for [tool definitions](https://ai.google.dev/gemini-api/docs/function-calling)
+- `operation.cost` — calculated price in USD using the [official Gemini pricing tables](https://ai.google.dev/gemini-api/docs/pricing) via [`genai-prices`](https://pypi.org/project/genai-prices/)
+
+Note that, unlike Anthropic, the Gemini API's `prompt_token_count` already includes
+the cached tokens; Logfire does not sum them again. This is documented in the
+[`GenerateContentResponseUsageMetadata.prompt_token_count`](https://googleapis.github.io/python-genai/genai.html#genai.types.GenerateContentResponseUsageMetadata.prompt_token_count)
+field description: *"When `cached_content` is set, this also includes the number
+of tokens in the cached content."*
```
alexmojaki, line 65: I don't think this is needed

logfire/_internal/integrations/google_genai.py
```diff
@@ -59,6 +59,68 @@ def wrapped_to_dict(obj: object) -> object:
...
+                return
+            # "keep last non-zero" — streaming sends partial chunks; cached/thoughts/tool_use
+            # counts typically only appear in the final chunk.
+            if cached := getattr(metadata, 'cached_content_token_count', None):
+                self._lf_cache_read = cached
+            if thoughts := getattr(metadata, 'thoughts_token_count', None):
+                self._lf_thoughts = thoughts
+            if tool_use := getattr(metadata, 'tool_use_prompt_token_count', None):
+                self._lf_tool_use_prompt = tool_use
+            self._lf_response = response
```
alexmojaki, line 83: it looks like _lf_response is the only thing that needs to be stored, and the rest can be retrieved in _wrapped_create_final_attributes

tests/otel_integrations/test_google_genai.py
```diff
@@ -278,6 +283,189 @@ class ResponseData(pydantic.BaseModel):
...
 
 
+def _stub_generate_content(response: Any) -> Any:
+    def _generate(self: Any, **kwargs: Any) -> Any:
+        return response
+
+    return _generate
+
+
+def _build_fake_genai_response(
```
alexmojaki, line 293: is extra mocking needed? can we stick to vcr for the new tests?

CHANGELOG.md
alexmojaki, line 1: please don't edit this file

PR #1962, opened 2026-05-24
Tier 1 fixes for #1954 (variable composition review feedback)

dmontagu:
Stacked on #1954. Four correctness fixes addressing review feedback Alex left on #1951 that's now folded into #1954.

## 1. Transitive reachability in the reference walker (r3288517914)

`_check_reference_errors` only iterated locally-registered variables when seeding the ref graph, so a missing reference reachable only via composition into a server-only fragment was never reported, and a cycle whose midpoints were server-only went silent. Refactored to walk the full graph with a FIFO queue from each local variable, pulling values out of `VariablesConfig` for any name reached. `VariablesConfig` is treated as the self-contained source of truth for substitution — `@{name}@` whose `name` is in neither local registration nor `server_config` is the same kind of error as a name in nothing at all.

Two regression tests in `test_push_variables.py`:
- `test_compute_diff_reference_errors_through_server_only_chain` — Alex's exact repro: local `foo2` → server-only `foo1` → missing `foo3`.
- `test_compute_diff_detects_cycle_through_server_only_chain` — `foo` → server-only `server_a` → server-only `server_b` → `foo`.

## 2. Top-level context-override fast path (r3287492856 / r3289439477 / Devin r3286255835)

#1951 unified override / provider / code-default through `_lookup_serialized`, which added a `dump_json` / `validate_json` round trip on the override path. Unserializable overrides (e.g. `var.override(some_object)` with a non-JSON-friendly value) silently fell through to provider / code default. Inconsistent with the pre-#1951 typed-Python return, and inconsistent within the SDK itself — unserializable *code defaults* still take effect.

Added `Variable._resolve_context_override` and route to it from the top of `_resolve` before `_lookup_serialized` is consulted. Tries serialization (so `render_fn` can run for `TemplateVariable`); on failure, returns the user's value verbatim. The `_lookup_serialized` override branch stays in place for *composition-reference* overrides — composition substitutes into strings, so serialization is unavoidable there, and falling through on failure remains the right call per the design discussion.

New test in `test_variables.py::TestVariableContext::test_override_unserializable_value_returned_typed`.

## 3. Escape-only composition consistency (r3288986490)

`_expand_and_deserialize` short-circuited on `has_references(value)`. The regex requires *unescaped* `@{`, so a value containing only `\@{baz}@` produced no refs, was returned unchanged with its backslash; the same escape combined with a real `@{ref}@` ran through composition (which *does* unescape) and produced literal `@{baz}@`. Behaviour depended on whether another reference was present.

Dropped the short-circuit. `expand_references()` already handles the no-refs case efficiently (early-return after `_unescape_serialized`), so both code paths now route through the same unescape logic.

New test in `test_variable_composition.py::TestCompositionIntegration::test_escape_only_value_is_unescaped_consistently`.

## 4. `template_var()` type inference (r3287772951)

`template_var()` required an explicit `type=` even when `default` was a non-callable. `var()` infers in that case; the inconsistency was easy to trip on when authoring docs. Added a no-`type=` overload mirroring `var()`'s pattern, with the same runtime check that a resolve-function `default` still requires an explicit `type=`.

## Verification

442 variable / composition / template / push tests pass. Pyright + ruff format + ruff check clean.

Recommend merging this into #1954 (or once #1954 lands, retargeting to main) before continuing with Tier 2 (push-time validation improvements — local-default-vs-latest split, `LabelRef` reporting) and Tier 3 (#1958 docstring exception-name fixes).

### DIFF
--- a/logfire/variables/_handlebars.py
+++ b/logfire/variables/_handlebars.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
-from collections.abc import Callable
 from functools import cache, lru_cache
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
+    from types import ModuleType
+
     from pydantic_handlebars import CompiledTemplate, HandlebarsEnvironment
 
 
@@ -21,7 +22,7 @@ class _FallbackHandlebarsError(Exception):
 
 try:
     from pydantic_handlebars import HandlebarsError as _ImportedHandlebarsError
-except ImportError:  # pragma: no cover
+except ImportError:
     _ImportedHandlebarsError = _FallbackHandlebarsError
 
 HandlebarsError: type[Exception] = _ImportedHandlebarsError
@@ -31,16 +32,29 @@ class HandlebarsDependencyError(ImportError):
     """Raised when a Handlebars feature is used without pydantic-handlebars installed."""
 
 
-def _dependency_error() -> HandlebarsDependencyError:
-    return HandlebarsDependencyError(
-        'Handlebars template rendering requires the `pydantic-handlebars` package, '
-        'which is installed by the `logfire[variables]` extra.'
-    )
+@cache
+def _pydantic_handlebars() -> ModuleType:
+    """Return the cached `pydantic_handlebars` module, or raise a helpful error.
+
+    The import is cached so every accessor below — `get_environment`,
+    `compile_composition_template`, `extract_composition_dependencies`,
+    etc. — can grab attributes off the returned module without each one
+    repeating its own try/except dance. Failure surfaces once, here, with
+    the install hint.
+    """
+    try:
+        import pydantic_handlebars
+    except ModuleNotFoundError as exc:
+        raise HandlebarsDependencyError(
+            'Handlebars template rendering requires the `pydantic-handlebars` package, '
+            'which is installed by the `logfire[variables]` extra.'
+        ) from exc
+    return pydantic_handlebars
 
 
 def ensure_handlebars_available() -> None:
     """Raise a helpful error if pydantic-handlebars is unavailable."""
-    get_environment()
+    _pydantic_handlebars()
 
 
 @cache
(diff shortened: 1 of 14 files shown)


logfire/variables/_handlebars.py
```diff
@@ -72,20 +72,41 @@ def get_handlebars_renderer() -> tuple[type[str], Callable[..., str]]:
...
+
+@cache
+def _extract_dependencies_fn() -> Callable[..., set[str]]:
+    """Cached lookup of `pydantic_handlebars.extract_dependencies`."""
     try:
         from pydantic_handlebars import extract_dependencies
     except ModuleNotFoundError as exc:  # pragma: no cover
         if exc.name == 'pydantic_handlebars':
             raise _dependency_error() from exc
         raise
```
alexmojaki: my point before was not about caching. it's that these 5 try/except lines are repeated 4 times verbatim and make my eyes hurt. i don't even think they need to be deduped. `ensure_handlebars_available()` being called should be enough to ensure that you don't need to check the imports in any other places.

also why do they all have pragmas? sounds like there isn't an actual test of a user-level function leading to the dependency error.

alexmojaki: or just have a function like this that just imports the whole module instead of one or two names at a time. the longer you look at these functions, the dumber they look. python imports don't need extra caching.

logfire/variables/_handlebars.py
```diff
@@ -72,20 +72,41 @@ def get_handlebars_renderer() -> tuple[type[str], Callable[..., str]]:
     return SafeString, render
 
 
-def extract_composition_dependencies(template: str) -> set[str]:
-    """Return the top-level `@{name}@` references in *template*.
+@lru_cache(maxsize=1024)
+def compile_composition_template(source: str) -> CompiledTemplate:
```
alexmojaki, line 101: can do similar for regular {{}} templates

logfire/variables/_handlebars.py
```diff
@@ -72,20 +72,41 @@ def get_handlebars_renderer() -> tuple[type[str], Callable[..., str]]:
...
         from pydantic_handlebars import extract_dependencies
     except ModuleNotFoundError as exc:  # pragma: no cover
         if exc.name == 'pydantic_handlebars':
             raise _dependency_error() from exc
         raise
-    return extract_dependencies(template, open_delim=COMPOSITION_OPEN_DELIM, close_delim=COMPOSITION_CLOSE_DELIM)
+    return extract_dependencies
+
+
+def extract_composition_dependencies(template: str) -> set[str]:
```
alexmojaki, line 120: this should also be lru_cached. or better yet, it feels like you should be able to get this from the already cached compiled template.

logfire/variables/_handlebars.py
dmontagu: Collapsed in b89515cc — all four call sites now go through a single `_pydantic_handlebars()` cached module accessor. The dance is gone and `get_environment` / `get_safe_string_cls` / `compile_*` etc. just reach in via attribute access.

For the pragmas: dropped them all. The dep-error path is now one branch in one function, covered in-process by a new `test_missing_handlebars_raises_helpful_error` (cache-clear + `__import__` monkeypatch). Also dropped the defensive `exc.name == 'pydantic_handlebars'` discriminator since the only remaining un-coverable branch would have been that one — pydantic-handlebars is small enough that a transitive ModuleNotFoundError is pathological. 100% coverage on the file now.

logfire/variables/_handlebars.py
dmontagu: Done in b89515cc. Added `compile_runtime_template(source)` paralleling `compile_composition_template`, both `@lru_cache(maxsize=1024)`. `render_serialized_string` in abstract.py routes through it instead of calling the module-level `pydantic_handlebars.render` (which re-parses every call).

logfire/variables/_handlebars.py
dmontagu: Promoted `extract_composition_dependencies` itself to `@lru_cache(maxsize=1024)` in b89515cc. Considered piggybacking on the compiled template — there's no public way to get dep info out of a `CompiledTemplate` in pydantic-handlebars today though, so the separate LRU is the simpler path. Could revisit if that surface grows.

logfire/variables/_handlebars.py
```diff
@@ -31,16 +32,29 @@ class HandlebarsDependencyError(ImportError):
...
+    repeating its own try/except dance. Failure surfaces once, here, with
+    the install hint.
+    """
+    try:
+        import pydantic_handlebars
+    except ModuleNotFoundError as exc:
+        raise HandlebarsDependencyError(
+            'Handlebars template rendering requires the `pydantic-handlebars` package, '
+            'which is installed by the `logfire[variables]` extra.'
+        ) from exc
```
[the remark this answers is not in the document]
dmontagu: Good catch. Restored the discriminator in d6073ce4 — `ModuleNotFoundError` for anything other than `pydantic_handlebars` itself now re-raises unchanged. Added `test_unrelated_module_not_found_propagates` that monkeypatches `__import__` to raise `ModuleNotFoundError(name='some_other_dep')` from inside the `pydantic_handlebars` import and asserts the raw error propagates (not wrapped as `HandlebarsDependencyError`). 100% line + branch coverage on the file with no pragmas.

logfire/variables/_handlebars.py
dmontagu: Quick update — cubic pointed out (r3295580826) that dropping the `exc.name` discriminator would reframe a missing transitive dep as "install pydantic-handlebars". Valid point, so I restored the discriminator in d6073ce4 and added `test_unrelated_module_not_found_propagates` that covers the re-raise branch in-process. Net result: discriminator is back, no pragmas, both branches covered, still 100% on the file.

PR #1970, opened 2026-05-25
docs: restore JavaScript integration pages

bruno-espino:
- restore the JavaScript integration docs that are still referenced from mkdocs nav and public docs pages
- fixes the Cloudflare Pages strict MkDocs build failing with missing integrations/javascript targets


bruno-espino:
Closing this approach: the JavaScript pages were intentionally removed in #1924 and should not be restored here. I'll fix the remaining stale MkDocs references instead.

PR #1978, opened 2026-05-27
Support OpenTelemetry SDK 1.42

alexmojaki:


logfire/_internal/integrations/system_metrics.py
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in cc1613ae by including the OpenTelemetry dependency bump that introduces `process.disk.io` in upstream `_DEFAULT_CONFIG`. With that bump, `FULL_CONFIG` picks it up through `_DEFAULT_CONFIG` and the focused test passes against the updated lockfile.

tests/otel_integrations/test_system_metrics.py
```diff
@@ -37,37 +49,39 @@ def test_default_system_metrics_collection(metrics_reader: InMemoryMetricReader)
...
+        'system.network.dropped_packets',
+        'system.network.errors',
+        'system.network.io',
+        'system.network.packets',
+        'system.swap.usage',
+        'system.swap.utilization',
+        'system.thread_count',
+    ]
+    if process_disk_io_supported():
+        expected_metric_names.insert(7, 'process.disk.io')
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in cc1613ae by including the OpenTelemetry instrumentation 0.63 lockfile update. I verified the installed upstream `SystemMetricsInstrumentor` creates `process.disk.io` and catches unsupported platforms in its callback, which is why the test keeps a platform-support guard.

.github/workflows/main.yml
```diff
@@ -161,7 +161,7 @@ jobs:
 
       - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f  # v6.3.0
         with:
-          node-version: "23"
+          node-version: "24"
```
alexmojaki: was this confirmed to be necessary

logfire/version.py
```diff
@@ -1,5 +1,5 @@
 """Version information for logfire package."""
 
-import importlib_metadata
+from importlib.metadata import version
```
alexmojaki, line 3: what are the implications of this change

pyodide_test/package.json
```diff
@@ -9,6 +9,6 @@
   "license": "MIT",
   "description": "",
   "dependencies": {
-    "pyodide": "^0.29.3"
+    "pyodide": "^0.29.4"
```
alexmojaki: was this confirmed to be necessary

tests/otel_integrations/test_sqlalchemy.py
```diff
@@ -435,49 +431,49 @@ async def test_sqlalchemy_async_instrumentation(parameter: str, exporter: TestEx
                 },
             },
             {
-                'name': IsStr(regex=r'PRAGMA .*test_(engine|engines)\.db$'),
+                'name': IsStr(regex=r'PRAGMA( .*test_(engine|engines)\.db)?$'),
```
alexmojaki: is this inconsistent? is fuzzy matching still needed?

tests/otel_integrations/test_sqlalchemy.py
```diff
@@ -435,49 +431,49 @@ async def test_sqlalchemy_async_instrumentation(parameter: str, exporter: TestEx
...
                 'end_time': 4000000000,
-                'attributes': {
-                    'logfire.span_type': 'span',
-                    'logfire.msg': 'PRAGMA main.table_info("auth_records")',
-                    'db.statement': 'PRAGMA main.table_info("auth_records")',
-                    'db.system': 'sqlite',
-                    'db.name': IsStr(regex=r'.*test_(engine|engines)\.db$'),
-                    'db.operation': IsStr(),
-                },
+                'attributes': IsPartialDict(
```
alexmojaki: why the IsPartialDict?

tests/otel_integrations/test_system_metrics.py
alexmojaki: you can still use a snapshot here by popping before comparing. make this easy to update in the future without magic numbers like 7

.github/workflows/main.yml
alexmojaki: (AI) Addressed in e3e18d2c. Confirmed Node 24 was not necessary: the Pyodide test passes locally under Node 23 after the actual `importlib.metadata` fix, so I reverted this back to Node 23.

logfire/version.py
alexmojaki: (AI) The implication is that `VERSION` now uses Python's stdlib `importlib.metadata`, available for our supported Python >=3.10, instead of the `importlib_metadata` backport. That removes this code path's runtime dependency on the backport and fixes Pyodide, where the backport is not installed. Normal installations should behave the same: it still reads the installed `logfire` distribution metadata.

pyodide_test/package.json
alexmojaki: (AI) Addressed in e3e18d2c. Confirmed the Pyodide bump was not necessary: `pyodide` 0.29.3 passes locally under Node 23 after the actual `importlib.metadata` fix, so I reverted this package change.

logfire/version.py
alexmojaki: what installed it before? are we sure this is safe?

tests/otel_integrations/test_system_metrics.py
alexmojaki: (AI) Addressed in 068060ea. The test now pops `process.disk.io` by name when supported and compares the remaining metric list to an inline snapshot, so there is no hard-coded insertion index.

tests/otel_integrations/test_sqlalchemy.py
alexmojaki: (AI) Addressed in 068060ea. I removed the fuzzy name matching by normalizing the optional DB path suffix before comparing the snapshot, so the snapshot now asserts exact SQL operation names.

tests/otel_integrations/test_sqlalchemy.py
alexmojaki: (AI) Addressed in 068060ea. Removed `IsPartialDict`; after normalizing the optional `db.name` and `db.operation` attributes, the snapshot can assert the exact remaining attributes.

logfire/version.py
alexmojaki: (AI) It was present in the dev/test environment incidentally via `litellm` in `uv.lock`, not via Logfire's runtime dependencies. The package itself does not declare `importlib-metadata`, and Pyodide did not install it, which is why the Pyodide job exposed the issue. This is safe because Logfire requires Python >=3.10, where `importlib.metadata` is stdlib and provides the same distribution metadata lookup we need here.

PR #1982, opened 2026-05-31
docs: add complete MCP tool inventory with scope notes

strawgate:
## Summary
- add a full MCP tool inventory table for the `/mcp` endpoint
- group tools by capability with concise descriptions
- note that visible tools depend on token scopes

## Testing
- docs-only change


strawgate:
I think we want to not include codemode yet

PR #1990, opened 2026-06-02
Add on_change callbacks for managed variables

dmontagu:
## Summary

Adds support for registering callbacks that fire when a managed variable's configuration changes, enabling reactive patterns like cache invalidation and feature-flag updates.

Reworked on top of current `main` now that composition and templating (#1954) have landed, addressing the review feedback about composition and duplicate registration.

## What's added

- `Variable.on_change()` decorator + `_notify_change()` (variable.py)
- **Composition-aware change expansion**: `expand_config_changes()` (variable.py) expands provider-level config changes to every registered variable that (transitively) composes a changed variable via `@{ref}@` references — whether the reference appears in a server-stored value (any label or `latest_version`) or in a code default. Reference and changed names are normalized through the alias map.
- `VariableProvider.add_on_config_change()` / `_notify_config_change()` (abstract.py), with idempotent registration. Providers report changed names *including aliases* (from both old and new configs), so dependents referencing a deleted variable via an alias are still notified.
- Change notifications on local `create`/`update`/`delete` (local.py) and remote refresh diffing (remote.py). The remote provider notifies outside its refresh lock to avoid deadlocks if a callback triggers resolution or another refresh.
- **Listeners live on `LogfireConfig`, not the provider** (config.py): every provider the config creates (including the lazy-init path) is wired to the config's dispatcher, so notifications survive `configure()` re-runs — including for `with_settings()` instances — and there is no setup flag to get out of sync. `Logfire._on_variables_config_change` dispatches to registered variables (main.py).
- Docs section in `managed-variables/remote.md`

## Notification contract

- **Only resolution-relevant changes fire**: labels, rollout, overrides, latest version, aliases. Metadata-only edits (e.g. changing a description in the UI) don't notify. Local and remote providers diff identically, so an identical-config `update_variable` is silent on both.
- **Callbacks must be idempotent** (documented): notifications are deliberately conservative — a change to a label the rollout never serves, or to a value only served for other targeting keys, still fires. Rollout-aware filtering would require per-targeting-key evaluation, so the contract is "re-read and reconcile".
- **No notification on the provider's initial load** (that would race registration); to reconcile once at startup, call your handler directly after registering it — `on_change` returns the callback to make that easy.
- Dispatch snapshots the variables registry and callback lists, so concurrent or re-entrant registration can't disrupt an in-flight notification cycle. Exceptions in callbacks are logged, not raised.

## Usage

```python
feature_enabled = logfire.var('feature_enabled', default=False)


@feature_enabled.on_change
def on_feature_change():
    new_value = feature_enabled.get().value
    invalidate_cache()


on_feature_change()  # optionally, reconcile once at startup too
```

If another variable composes `feature_enabled` (e.g. a prompt containing `@{feature_enabled}@`), its `on_change` callbacks fire too when `feature_enabled` changes.

## Tests

- 30 new tests: basics, composition (direct/transitive/code-default/`latest_version`), aliases (registration, reference, deleted target), duplicate prevention after `variables_clear()`, reconfigure, remote diff detection, and characterization tests pinning the contract (metadata-only edits don't fire; unserved-label changes do; re-entrant registration is safe)
- Full variables suites + configure: 675 passed; variables modules at 100% coverage
- pyright clean, ruff clean


logfire/variables/abstract.py
```diff
@@ -640,6 +640,26 @@ def _update_variable_schema(
...
 
+    _on_config_change: Callable[[set[str]], None] | None = None
+
+    def set_on_config_change(self, callback: Callable[[set[str]], None]) -> None:
+        """Set a callback that will be called when variable configurations change.
+
+        Args:
+            callback: Called with the set of variable names whose configs changed.
+        """
+        self._on_config_change = callback
```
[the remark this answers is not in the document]
dmontagu: Good catch — fixed in 629dbb64. The provider now holds a list of subscribers (`add_on_config_change` appends rather than overwrites), and `_notify_config_change` dispatches to all of them. This way multiple `Logfire` instances sharing a provider via `with_settings()` each get their variables notified. Providers are recreated on every `configure()`, so the callback list resets with the new provider and there's no accumulation across reconfigures. Added a regression test (`test_on_change_fires_for_variables_on_shared_provider_instances`).

logfire/_internal/main.py
```diff
@@ -2574,8 +2575,24 @@ def var(
...
+            self._change_notifications_setup = True
+
         return variable
 
+    def _setup_variable_change_notifications(self) -> None:
+        """Set up the on_config_change callback on the provider to dispatch to variables."""
+        provider = self.config.get_variable_provider()
+
+        def on_config_change(changed_names: set[str]) -> None:
+            for name in changed_names:
```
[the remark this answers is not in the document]
dmontagu: Fixed in 629dbb64 — `on_config_change` now iterates `sorted(changed_names)` for deterministic dispatch order.

logfire/variables/remote.py
```diff
@@ -324,9 +324,22 @@ def refresh(self, force: bool = False):
...
                 self._last_fetched_at = datetime.now(tz=timezone.utc)
+
+                # Detect changed variables and notify
+                if old_config is not None:
+                    changed: set[str] = set()
+                    all_names = set(old_config.variables.keys()) | set(new_config.variables.keys())
+                    for name in all_names:
+                        old_var = old_config.variables.get(name)
+                        new_var = new_config.variables.get(name)
+                        if old_var != new_var:
```
alexmojaki: i think this isn't enough with composition. if a referenced var changes then its 'parent' has also effectively changed even if its own config hasn't

dmontagu: You're right — fixed. The implementation has been reworked on top of main: providers now report config-level diffs (including changed variables' aliases), and the `Logfire` instance expands them via `expand_config_changes()`, which builds the `@{ref}@` reference graph from server-stored values (all labels + `latest_version`) *and* registered code defaults, normalizes names through the alias map, and computes reverse reachability to a fixpoint. So when a referenced variable changes, every registered variable that transitively composes it gets its `on_change` callbacks fired too — including via aliases and via references that only exist in a code default. Covered by the new `TestOnChangeComposition` tests.

logfire/_internal/main.py
```diff
@@ -2574,8 +2575,24 @@ def var(
...
+        """Set up the on_config_change callback on the provider to dispatch to variables."""
+        provider = self.config.get_variable_provider()
+
+        def on_config_change(changed_names: set[str]) -> None:
+            for name in sorted(changed_names):
+                variable = self._variables.get(name)
+                if variable is not None:
+                    variable._notify_change()  # pyright: ignore[reportPrivateUsage]
+
+        provider.add_on_config_change(on_config_change)
```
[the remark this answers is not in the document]
dmontagu: Confirmed and fixed in the rework: the append-only registration + reset flag is gone entirely. Change listeners now live on `LogfireConfig` (surviving reconfiguration; each provider the config creates is wired to the config's dispatcher at creation, including the lazy-init path), and registration is idempotent via bound-method equality — so `variables_clear()`/re-registration cycles can't duplicate notifications. Covered by `test_no_duplicate_notifications_after_clear_and_redefine` and `test_add_on_config_change_is_idempotent`.

docs/reference/advanced/managed-variables/remote.md
```diff
@@ -393,3 +393,49 @@ merged = server_config.merge(local_config)
...
+
+```python skip="true"
+feature_enabled = logfire.var('feature_enabled', default=False)
+
+
+@feature_enabled.on_change
+def on_feature_change():
+    new_value = feature_enabled.get().value
+    logfire.info('feature_enabled changed to {new_value}', new_value=new_value)
+    invalidate_cache()
```
alexmojaki:
```suggestion
```python
import logfire

feature_enabled = logfire.var('feature_enabled', default=False)

@feature_enabled.on_change
def on_feature_change():
    new_value = feature_enabled.get().value
    logfire.info('feature_enabled changed to {new_value}', new_value=new_value)
```

docs/reference/advanced/managed-variables/remote.md
```diff
@@ -393,3 +393,49 @@ merged = server_config.merge(local_config)
...
+
+Other key points:
+
+- Callbacks receive no arguments; call `variable.get()` to see the current value
+- Callbacks may run on the provider's polling thread — keep them fast and non-blocking
+- Don't create, update, or delete variables from inside a callback (that would re-enter
+  change notification)
+- Multiple callbacks can be registered on the same variable
+- Exceptions in callbacks are caught and logged (they don't crash the polling thread)
+- The initial load of remote configuration does not fire callbacks; to reconcile once at
```
alexmojaki: It also doesn't start the polling thread, so if you only register callbacks without getting any variable eagerly, then they don't get called. This probably wouldn't happen in practice, but it happened when i tried it out which made me think it didn't work. So maybe registering a callback should start the thread, but it looks like that's not simple because registering callbacks before configuring is meant to work. Also, if i configure, then get a variable and thus start the thread, then configure again, the previous thread is shutdown and so it's not polling again.

logfire/variables/remote.py
```diff
@@ -350,14 +353,31 @@ def refresh(self, force: bool = False):
...
 
             try:
+                old_config = self._config
                 new_config = VariablesConfig.model_validate(variables_config_data)
                 self._config = new_config
                 self._last_fetched_at = datetime.now(tz=timezone.utc)
+
+                # Detect which variables' configs changed since the previous fetch. The first
+                # fetch (old_config is None) is not a "change" — variables are only beginning
+                # to resolve against the provider at that point.
```
alexmojaki: It's possible to get a value for a variable before the first fetch succeeds with block_before_first_resolve=False, or (according to codex) if the first fetch fails. Then from the user's perspective the value changes from the code default that they saw in the initial reconciliation to the server value.

logfire/variables/variable.py
```diff
@@ -1181,6 +1236,70 @@ def _static_composition_refs(variable: Variable[Any]) -> set[str]:
...
+        for labeled_value in variable_config.labels.values():
+            if isinstance(labeled_value, LabeledValue):
+                add_refs(variable_config.name, labeled_value.serialized_value)
+        if variable_config.latest_version is not None:
+            add_refs(variable_config.name, variable_config.latest_version.serialized_value)
+
+    # The code default participates in resolution (and composition through it) too, so its
+    # references count as edges even for variables with no server-side values.
+    for name, variable in registry.items():
+        for ref in _static_composition_refs(variable):
```
alexmojaki, line 1304: Question/contract gap: this only tracks refs discoverable from `_static_composition_refs()`, so callable defaults are skipped. I agree we shouldn't invoke user code just to discover dependencies, since a resolve function may be slow, side-effectful, or return different refs per targeting context. But resolution *does* compose the callable result, so `default=lambda ...: "hi @{name}@"` can have an effective value that changes when `name` changes while `on_change` stays silent.

I think we should either document this as an explicit limitation of `on_change` (static code defaults only), or provide some explicit dependency declaration path for dynamic defaults. Otherwise users can get stale cache invalidation behavior from two defaults that resolve identically in simple cases: static `"hi @{name}@"` notifies, callable returning `"hi @{name}@"` does not.

dmontagu: I think let's just make it so that if you have a callable code default, we assume it references _every_ other variable from the perspective of whether to call the on_change handler. This will be an edge case anyway, and we've already mentioned the on_change function should be idempotent.

logfire/variables/remote.py
dmontagu: I'm fine with calling on_change for the first values, we've already told people the call should be idempotent so I think we can go with the version where it gets called pretty eagerly/unnecessarily early on.

docs/reference/advanced/managed-variables/remote.md
dmontagu: Applied — added `import logfire`, dropped the undefined `invalidate_cache()` call, and removed `skip="true"` so the snippet is now linted and run by the docs test suite. I moved the `on_change()` startup-reconcile pattern into its own runnable example under the "initial load is best-effort" bullet. (6a7a49da)

logfire/variables/variable.py
dmontagu: Done — `expand_config_changes()` now treats a variable with a callable code default as referencing *every* variable, so any change fires its `on_change`. We don't introspect or invoke the callable (it may be slow, side-effectful, or return different refs per targeting context); we just conservatively assume a dependency, which is fine given the documented idempotency contract. New test `test_on_change_fires_for_callable_default_on_unrelated_change`. (6a7a49da)

logfire/variables/remote.py
dmontagu: Done — the first successful fetch now fires `on_change`. The refresh diff treats the initial load as a diff against an empty config, so every variable is reported as changed (it transitions from the code default to the server value). Deliberately eager, consistent with the idempotent "re-read and reconcile" contract. New test `test_first_fetch_notifies_all_variables`; updated the existing change-detection tests and the docs/docstring note (initial-load delivery is now described as best-effort, since it can fire but may race registration). (6a7a49da)

docs/reference/advanced/managed-variables/remote.md
dmontagu: Good catch — fixed both halves.

1. Registering a callback now starts the poller: `Variable.on_change()` calls `Logfire._ensure_variable_change_polling()`, which triggers the lazy provider init (and thus the background poll) when configuration has already happened — so callbacks fire even if you never resolve a variable eagerly.
2. The request is recorded on `LogfireConfig` (`_variable_change_polling_requested`), so it survives reconfiguration: a `configure()` that rebuilds the provider as a no-op restarts the poller. Registering before `configure()` still works — we remember the request and start on the next configure.

Covered by `TestOnChangePollingLifecycle` (`test_registering_on_change_starts_lazy_poller`, `test_on_change_polling_survives_reconfigure`). (6a7a49da)

docs/reference/advanced/managed-variables/remote.md
```diff
@@ -393,3 +393,62 @@ merged = server_config.merge(local_config)
...
+- Callbacks may run on the provider's polling thread — keep them fast and non-blocking
+- Don't create, update, or delete variables from inside a callback (that would re-enter
+  change notification)
+- Multiple callbacks can be registered on the same variable
+- Exceptions in callbacks are caught and logged (they don't crash the polling thread)
+- Delivery of the *initial* load of remote configuration is best-effort: the provider may
+  fetch it before your callback is registered, so don't rely on it firing. To reliably
+  reconcile once at startup, call your handler yourself right after registering it:
+
+```python
```
alexmojaki, line 442: this is basically the same snippet as the first

logfire/_internal/config.py
```diff
@@ -706,6 +706,13 @@ def configure(
     # are started when first accessed via get_variable_provider().
     if config.variables is not None:
         config.get_variable_provider().start(logfire_instance if config.variables.instrument else None)
+    elif config._variable_change_polling_requested:  # pyright: ignore[reportPrivateUsage]
```
alexmojaki, line 743: Maybe polling should start if any variable is defined? Then it has a chance of being ready by the first `.get()`.

logfire/variables/variable.py
```diff
@@ -1181,6 +1240,86 @@ def _static_composition_refs(variable: Variable[Any]) -> set[str]:
...
+    while True:
+        additions = {node for node, refs in edges.items() if node not in closure and refs & closure}
+        if not additions:
+            break
+        closure |= additions
+
+    result = {name for name in registry if canonical(name) in closure}
+    # Any change at all effectively changes a variable with a callable default (it may compose
+    # the changed variable). `changed_names` is always non-empty here, so include them all.
+    if changed_names:
```
alexmojaki, line 1318: "`changed_names` is always non-empty here"

so why the if condition?

alexmojaki:
Reviewing this led to 2 small PRs from main: https://github.com/pydantic/logfire/pull/2033 and https://github.com/pydantic/logfire/pull/2034. The latter affects how things are done here.

logfire/_internal/main.py
```diff
@@ -2603,6 +2603,7 @@ def var(
             description=description,
         )
         self._variables[name] = variable
+        self._config.add_variables_change_listener(self._on_variables_config_change)
```
alexmojaki, line 2606: I think https://github.com/pydantic/logfire/pull/2034 should be merged first, and then this should change so that Logfire doesn't need to own callbacks. then there'll be 3 layers of callbacks instead of 4.

PR #1994, opened 2026-06-03
Include project URL in configuration span message

alexmojaki:
When `emit_configuration_span` is enabled and Logfire already knows the project URL during configuration, include that URL in the `Logfire configured` span message instead of printing a separate project URL line.

This uses the existing `_project_url` state as the source of truth, so credentials-file URLs are available eagerly even when a token is supplied by another source. If async token validation later confirms the same URL, Logfire now avoids printing a duplicate `Logfire project URL` line.

Validation covered in this PR:

- configuration span URL message with credentials-file token
- configuration span URL message with an explicit token override
- no real OTLP export leakage from the regression test
- pyright-safe project URL access via a typed `project_url` property

https://claude.ai/code/session_01BaJtdKyNpFQhjPmBLofWty

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -860,7 +860,7 @@ class _LogfireConfigData:
 
         self.additional_span_processors = additional_span_processors
         self._project_url: str | None = None
-        self._instant_project_url: str | None = None
+        self._configuration_span_project_url: str | None = None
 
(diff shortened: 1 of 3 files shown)


alexmojaki:
Replacing with https://github.com/pydantic/logfire/pull/2007

PR #2006, opened 2026-06-09
Release v4.36.0

dmontagu:
Bumping version to v4.36.0.


release/prepare.py
```diff
@@ -74,7 +74,8 @@ def update_history(new_version: str, notes: str) -> None:
         sys.exit(1)
 
     new_chunk = f'{title}\n\n{notes}\n\n'
-    updated_content = re.sub(r'(# Release Notes\n\n)', rf'\1{new_chunk}', history_content)
+    marker = '# Release Notes\n\n'
+    updated_content = history_content.replace(marker, marker + new_chunk, 1)
```
dmontagu, line 78: Needed this change because:

```
  new_chunk contains the freshly-generated release notes, and one of this release's PR titles is "Add pylf_v\d+_ pattern to default scrubbing patterns" (#1993) — it has a literal \d in it. When that text gets
  interpolated into the replacement string argument of re.sub, Python parses the replacement as a template and tries to interpret \d as an escape sequence. \1 is a valid backreference, but \d is not a valid
  replacement escape, so it raises re.error and the whole script aborts.
```

PR #2012, opened 2026-06-11
Add `resource_attributes` and `resource_detectors` options to `logfire.configure`

dmontagu:
## Context

The (beta) Hosts view needs the `host.name` resource attribute to associate telemetry with a machine, and more generally there was no programmatic way to shape resource attributes from `logfire.configure()` — the only options were the `OTEL_RESOURCE_ATTRIBUTES` / `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` env vars.

This PR builds on #2013 (now merged), which pre-populates `host.*`/`os.*`/`process.runtime.*` by default so machines show up in the Hosts view out of the box, and adds first-class, programmatic control over the resource.

## What this PR adds

- **`resource_attributes`** (top-level `logfire.configure()` argument): a static mapping of extra resource attributes, e.g. to override the auto-detected `host.name` or add custom attributes — a programmatic alternative to `OTEL_RESOURCE_ATTRIBUTES`.
- **`resource_detectors`** (on `AdvancedOptions`): OpenTelemetry resource detectors to run when building the resource. Each item is a `ResourceDetector` instance or the name of a detector registered under the standard `opentelemetry_resource_detector` entry-point group (so third-party detector packages work too). `opentelemetry-sdk` provides `'os'`, `'host'`, and `'process'`; the special name `'*'` runs every registered detector. Since `os.*`/`host.*` are already populated by default, `'process'` is the main one you'd add.
- Both are also configurable via **environment variables** (`LOGFIRE_RESOURCE_ATTRIBUTES`, `LOGFIRE_RESOURCE_DETECTORS`) and **`pyproject.toml`** under `[tool.logfire]`.

```python
import logfire

logfire.configure(
    resource_attributes={
        'host.name': 'my-meaningful-host-name',  # override an auto-detected attribute
        'datacenter.region': 'eu-west-1',         # and/or add arbitrary custom attributes
    },
    advanced=logfire.AdvancedOptions(resource_detectors=['process']),  # add process.* attributes
)
logfire.instrument_system_metrics()
```

## Precedence

Resource attributes are merged from lowest to highest precedence:

1. **Auto-detected defaults** — `process.runtime.*`, `host.*`, `os.*`, a random `service.instance.id`, and the git-hash `service.version` fallback. These deliberately sit *below* the env vars so they can be overridden.
2. **Env vars** — `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` (applied by `Resource.create`).
3. **The `resource_detectors` argument.**
4. **Explicit `logfire.configure()` attributes** — `resource_attributes`, `service_name`/`service_version`/`environment`, etc.

So explicit `logfire.configure()` settings win over everything, the `resource_detectors` argument beats the env-var detectors, and `OTEL_RESOURCE_ATTRIBUTES` still overrides the auto-detected defaults (e.g. a customer whose `socket.gethostname()` is a useless container ID can override `host.name` cleanly).

## Design notes

- Detectors are run via the OTel SDK's public `get_aggregated_resources()` plus entry-point loading, rather than writing env vars — there's a clean programmatic path, so no global-state mutation, and `ResourceDetector` instances are supported, not just names.
- An unknown detector name (or one that fails to load) **emits a `LogfireConfigWarning` and is skipped** rather than raising, so a typo can never crash the app — matching OTel's tolerant handling of the env var.
- The git-hash `service.version` fallback is now resolved in `_initialize` (where the resource is built) instead of `_load_configuration` (which runs at import). This removes a `git` subprocess call on `import logfire`, keeps the git hash as a low-precedence fallback that `OTEL_RESOURCE_ATTRIBUTES` can override, and — because it's never stored on the config — means child processes (`ProcessPoolExecutor`) re-detect it as a fallback rather than treating it as an explicit value.
- `resource_detectors` lives on `AdvancedOptions` because it's a niche option (the common `os`/`host` attributes are populated by default), while `resource_attributes` stays top-level since overriding `host.name` / adding custom attributes is a common need.

Platform-side note (not addressed here): the default `base='basic'` config doesn't emit `system.cpu.utilization` — its CPU metric is the logfire-specific `system.cpu.simple_utilization`, which the Hosts page queries don't currently include — so basic-config hosts appear (via `system.memory.utilization`) but with an empty CPU column.

## Docs

- System metrics integration page restructured: the Hosts view sits alongside the dashboard recommendation, with a focused "Customising resource attributes" note.
- Cross-links between the Hosts page and the system-metrics integration.
- Resource-attribute configuration documented under the SQL reference `otel_resource_attributes` column.

## Checklist

- Tests: both new options (instances, names, `'*'`, bare-string coercion, unknown-name warning, precedence vs env vars, env var + `pyproject.toml`), default host/os population and env-var override, git-hash fallback + serialization to child processes.
- `logfire-api` shim needs no changes (its `configure` accepts `**kwargs`; `.pyi` stubs are regenerated at release).


logfire/_internal/config.py
```diff
@@ -810,6 +842,8 @@ def _load_configuration(
         self.service_name = param_manager.load_param('service_name', service_name)
         self.service_version = param_manager.load_param('service_version', service_version)
         self.environment = param_manager.load_param('environment', environment)
+        self.resource_attributes = dict(resource_attributes or {})
+        self.resource_detectors = list(resource_detectors or ())
```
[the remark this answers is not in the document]
dmontagu: Good catch — a bare string satisfies `Sequence[str]` statically but would be split into characters. `_load_configuration` now raises a `TypeError` with a "did you mean `resource_detectors=['host']`" hint (this also covers both the `configure()` path and executor deserialization), with a test. Fixed in dd4fac1f.

logfire/_internal/config.py
```diff
@@ -810,6 +842,14 @@ def _load_configuration(
         self.service_name = param_manager.load_param('service_name', service_name)
         self.service_version = param_manager.load_param('service_version', service_version)
         self.environment = param_manager.load_param('environment', environment)
+        self.resource_attributes = dict(resource_attributes or {})
+        if isinstance(resource_detectors, str):
+            # Prevent e.g. `resource_detectors='host'` from being treated as `['h', 'o', 's', 't']`.
+            raise TypeError(
+                '`resource_detectors` must be a sequence of detector names and/or `ResourceDetector` instances, '
+                f'not a string. Did you mean `resource_detectors=[{resource_detectors!r}]`?'
+            )
```
```suggestion
            resource_detectors = [resource_detectors]
```

logfire/_internal/config.py
```diff
@@ -1078,6 +1127,16 @@ def _initialize(self) -> None:
...
             resource = Resource.create(otel_resource_attributes)
 
+            if self.resource_detectors:
+                detectors = [
+                    detector if isinstance(detector, ResourceDetector) else _load_resource_detector(detector)
+                    for detector in self.resource_detectors
+                ]
+                # `resource` is merged on top so that everything above
+                # (explicit attributes and detectors from `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`)
+                # takes precedence over these detectors, matching the semantics of `Resource.create`.
```
alexmojaki: OTEL_EXPERIMENTAL_RESOURCE_DETECTORS should have the lowest precedence, but otherwise i agree

logfire/_internal/config.py
dmontagu: Done in b76ec720 — applied your suggestion: a bare-string `resource_detectors` is now coerced to a single-element list (e.g. `'host'` → `['host']`) rather than raising. Updated `test_resource_detectors_string` to verify the coercion.

logfire/_internal/config.py
dmontagu: Good catch — fixed in b76ec720. `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` now has the lowest precedence. The resource is built in three layers: `Resource.create({})` (OTel defaults + the env-var detectors + `OTEL_RESOURCE_ATTRIBUTES`), then the `resource_detectors` kwarg, then the explicit attributes on top. So explicit `resource_detectors` override the env-var detectors, while explicit attributes still win over both. Updated the docs precedence note and added `test_resource_detectors_take_precedence_over_env_var_detectors`.

logfire/_internal/config.py
```diff
@@ -2138,6 +2203,20 @@ def default_project_name():
     return sanitize_project_name(os.path.basename(os.getcwd()))
 
 
+def _load_resource_detector(name: str) -> ResourceDetector:
```
alexmojaki: this looks like a reduced version of _build_resource_detectors from otel. what are we missing in this reduction?

logfire/_internal/config.py
alexmojaki: let's remove this if possible, it seems to be duplicating otel behaviour, which probably isn't terrible but is making it harder to think through all the layers of priority.

logfire/_internal/config.py
```diff
@@ -1045,13 +1091,15 @@ def _initialize(self) -> None:
...
-                'service.name': self.service_name,
                 'process.pid': os.getpid(),
                 # https://opentelemetry.io/docs/specs/semconv/resource/process/#python-runtimes
                 'process.runtime.name': sys.implementation.name,
                 'process.runtime.version': get_runtime_version(),
                 'process.runtime.description': sys.version,
             }
+            if self.service_name:
+                # When unset, `Resource.create` below applies its own `unknown_service` default.
+                otel_resource_attributes['service.name'] = self.service_name
```
alexmojaki: i’m questioning the behaviour asserted in the existing test test_otel_otel_resource_attributes_env_var. OTEL_RESOURCE_ATTRIBUTES should probably not take priority over logfire.configure(service_name, service_version)

logfire/_internal/config.py
alexmojaki: in fact it's explicitly backwards now, logfire.configure params (incl resource_attributes) should always take priority over env vars

logfire/_internal/config.py
```diff
@@ -810,6 +842,12 @@ def _load_configuration(
         self.service_name = param_manager.load_param('service_name', service_name)
         self.service_version = param_manager.load_param('service_version', service_version)
         self.environment = param_manager.load_param('environment', environment)
+        self.resource_attributes = dict(resource_attributes or {})
```
alexmojaki: why are these not using param_manager.load_param?

docs/integrations/system-metrics.md
```diff
@@ -23,6 +23,46 @@ logfire.instrument_system_metrics()
...
+## Host information
+
+By default, the metrics produced by `instrument_system_metrics` don't say anything about the machine they come from,
+such as the `host.name` resource attribute that the Logfire UI uses to associate metrics with a host.
+To include this information, pass resource detectors to [`logfire.configure()`][logfire.configure]:
+
+```py
+import logfire
+
+logfire.configure(resource_detectors=['os', 'host'])
```
alexmojaki: this will become redundant after https://github.com/pydantic/logfire/pull/2013 is merged.

logfire/_internal/config.py
dmontagu: Done in c5e5bd1c — removed the manual `OTEL_RESOURCE_ATTRIBUTES` re-parse. `logfire.configure()` attributes now take precedence over env vars. The resource is built lowest→highest: OTel defaults + `OTEL_RESOURCE_ATTRIBUTES`/`OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` (both via `Resource.create`) → the `resource_detectors` argument → explicit configure attributes.

One subtlety worth flagging: `service_version` silently falls back to the git commit hash, and that fallback should *not* override an explicit `OTEL_RESOURCE_ATTRIBUTES` service.version. So the git-hash is applied as a low-precedence layer below the env vars (an explicit `service_version` still wins).

logfire/_internal/config.py
dmontagu: Agreed, and addressed in c5e5bd1c. With the manual parse gone, `OTEL_RESOURCE_ATTRIBUTES` no longer overrides `service_name`/`service_version`. The only existing-behaviour change is `test_otel_service_name_has_priority_on_otel_resource_attributes_service_name_env_var`: `OTEL_SERVICE_NAME` now wins over `service.name` in `OTEL_RESOURCE_ATTRIBUTES` (banana→potato) — which actually matches OTel's own `OTELResourceDetector` (and the OTel spec, where `OTEL_SERVICE_NAME` takes precedence over `service.name` in `OTEL_RESOURCE_ATTRIBUTES`). Updated that test.

`test_otel_otel_resource_attributes_env_var` stays green: there the values come only from the env var, and the git-hash `service.version` fallback now sits below it.

logfire/_internal/config.py
dmontagu: Switched both to `param_manager.load_param` in c5e5bd1c. They now support `LOGFIRE_RESOURCE_ATTRIBUTES` (a comma-separated `key=value` list) and `LOGFIRE_RESOURCE_DETECTORS` (a comma-separated list of names) env vars, plus `pyproject.toml` config. `ResourceDetector` instances and the bare-string coercion still pass through the runtime path. Added `dict[str, str]` casting to `ParamManager` for the attributes.

logfire/_internal/config.py
dmontagu: Good question — the differences from OTel's `_build_resource_detectors` are intentional:

- We don't append the implicit `otel` detector, because `Resource.create({})` already runs `OTELResourceDetector` (handling `OTEL_RESOURCE_ATTRIBUTES`/`OTEL_SERVICE_NAME`) at the base layer.
- We raise on an unknown name instead of logging-and-skipping, since an explicit `resource_detectors=['typo']` kwarg should fail loudly (vs. the env var, which OTel tolerates).
- We don't support the `'*'` wildcard expansion — that's the one genuine capability gap. Happy to add it (scan all registered `opentelemetry_resource_detector` entry points) if you think it's worth it; it just wasn't needed for the host/os/process case.

logfire/_internal/config.py
alexmojaki, line 1: include counts of the new params in the config span

logfire/_internal/config.py
dmontagu: Done in 0fe100d8, going further per the discussion with David:

- Added `'*'` support — it expands to every registered detector except `'otel'` (whose `OTEL_RESOURCE_ATTRIBUTES`/`OTEL_SERVICE_NAME` handling is already applied by `Resource.create`), matching the env var.
- Unknown detector names (from the kwarg, `LOGFIRE_RESOURCE_DETECTORS`, or pyproject.toml) now emit a `LogfireConfigWarning` and are skipped rather than raising, so a typo can never crash the app — matching OTel's tolerant handling of the env var. `test_resource_detector_unknown_name` now asserts the warning instead of a `ValueError`.

So the only remaining intentional difference from `_build_resource_detectors` is that we don't append the implicit `otel` detector, since `Resource.create` already runs it.

logfire/_internal/config.py
dmontagu: Done in 896b06a3 — added `resource_attributes` and `resource_detectors` counts (`len(...)`) to the `Logfire configured` span, alongside the existing `additional_span_processors` count.

logfire/_internal/config.py
```diff
@@ -906,9 +948,13 @@ def _load_configuration(
             metrics = MetricsOptions()
         self.metrics = metrics
 
+        # Whether `service_version` was auto-detected from git rather than set explicitly. An explicit
+        # version takes precedence over `OTEL_RESOURCE_ATTRIBUTES`, but the git-hash fallback does not.
+        self._service_version_from_git = False
         if self.service_version is None:
             try:
                 self.service_version = get_git_revision_hash()
+                self._service_version_from_git = True
```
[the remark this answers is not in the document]
dmontagu: Fixed in ec66ca4d.

You're right: `_service_version_from_git` isn't a dataclass field, so `asdict()` in `serialize_config()` dropped it and the child treated the git-hash `service_version` as explicit, giving it precedence over `OTEL_RESOURCE_ATTRIBUTES`.

Rather than promote the git hash in the child, `serialize_config()` now drops the git-detected `service_version` (sets it to `None`) when `_service_version_from_git` is set, so the child re-detects it via its own `_load_configuration` and keeps it as a low-precedence fallback — matching the parent, where `OTEL_RESOURCE_ATTRIBUTES` wins. An explicitly-set `service_version` is left untouched and still round-trips. Added `test_git_service_version_dropped_on_serialization` covering both cases.

docs/integrations/system-metrics.md
```diff
@@ -23,6 +23,57 @@ logfire.instrument_system_metrics()
 
 Then in your project, click on 'Dashboards' in the top bar, click 'New Dashboard', and select 'Basic System Metrics (Logfire)' from the dropdown.
 
+## Resource attributes
```
alexmojaki: This section doesn't make much sense here like this. Resource detectors in particular no longer seem relevant.

I think related docs improvements about metrics that would make sense (doesn't have to be in this PR) would be:

- Linking to this page from https://pydantic.dev/docs/logfire/observe/hosts/ and vice versa
- Briefly mentioning how to customise any resource attributes such as `host.name` to make the Hosts page more useful, and how that relates to system metrics.

The reference-style stuff here would fit under https://pydantic.dev/docs/logfire/reference/sql/#otel_resource_attributes.

logfire/_internal/config.py
```diff
@@ -1046,10 +1100,16 @@ def _initialize(self) -> None:
...
         with suppress_instrumentation():
             otel_resource_attributes: dict[str, Any] = {
-                **common_resource_attributes(),
-                'service.name': self.service_name,
+                RESOURCE_ATTRIBUTES_VERSION: VERSION,
                 'process.pid': os.getpid(),
+                # https://opentelemetry.io/docs/specs/semconv/resource/process/#python-runtimes
+                'process.runtime.name': sys.implementation.name,
+                'process.runtime.version': get_runtime_version(),
+                'process.runtime.description': sys.version,
```
alexmojaki: I liked having these cached by common_resource_attributes. And I don't think they should override env vars any less than the host/os stuff. But the version and pid can stay here.

docs/integrations/system-metrics.md
dmontagu: Agreed — trimmed in 7c5392ce. The section is now a short "Host information" note that:

- leads with the `host.*`/`os.*` attributes populated by default (so the machine shows up in the Hosts view), and links to the [Hosts page](https://pydantic.dev/docs/logfire/guides/web-ui/hosts/);
- shows customising `host.name` (and other attributes) via `resource_attributes` to make the Hosts page more useful;
- drops the reference-style detector/env-var/precedence detail, pointing instead to the `resource_attributes`/`resource_detectors` args of `logfire.configure()` (whose docstring already documents them).

Leaving the rest of your suggestions for a follow-up since they touch other pages: the reverse link from the Hosts page back here, and relocating the reference material under the SQL `otel_resource_attributes` docs.

logfire/_internal/config.py
```diff
@@ -1059,22 +1119,45 @@ def _initialize(self) -> None:
                 )
                 if self.code_source.root_path:
                     otel_resource_attributes[RESOURCE_ATTRIBUTES_CODE_ROOT_PATH] = self.code_source.root_path
-            if self.service_version:
+            if self.service_version and not self._service_version_from_git:
+                # The git-hash fallback is applied lower down so it doesn't override env vars.
```
alexmojaki: this dance feels messy. i think we can avoid _service_version_from_git by only calling get_git_revision_hash here instead of in _load_configuration, which is probably a bad place for it to live anyway because that gets called immediately when logfire is imported, before calling configure().

docs/integrations/system-metrics.md
dmontagu: (Real David here — I'm telling it to not leave your other suggestions for follow-up, but to do them now.)

logfire/_internal/config.py
dmontagu: agreed on all counts

logfire/_internal/config.py
```diff
@@ -924,6 +970,8 @@ def __init__(
         service_name: str | None = None,
         service_version: str | None = None,
         environment: str | None = None,
+        resource_attributes: Mapping[str, Any] | None = None,
+        resource_detectors: Sequence[ResourceDetector | str] | None = None,
```
alexmojaki: resource_detectors feels obscure and possible better suited to `AdvancedOptions`.

logfire/_internal/config.py
```diff
@@ -1059,22 +1119,45 @@ def _initialize(self) -> None:
...
+            #      view works out of the box) and the git-hash `service.version`. These deliberately sit
+            #      below the env vars so an explicit `OTEL_RESOURCE_ATTRIBUTES` (e.g. a meaningful
+            #      `host.name` or `service.version`) still wins.
+            #   2. OTel's default attributes plus the env vars `OTEL_RESOURCE_ATTRIBUTES` and
+            #      `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` (both applied by `Resource.create`).
+            #   3. The `resource_detectors` argument.
+            #   4. Explicit `logfire.configure()` attributes (logfire's own plus `resource_attributes`).
+            resource = Resource.create({})
+            # Always non-empty: `detected_host_os_resource_attributes()` always provides `host.*`/`os.*`.
+            fallback_resource_attributes: dict[str, Any] = dict(detected_host_os_resource_attributes())
```
alexmojaki: service.instance.id should go in here

logfire/_internal/config.py
```diff
@@ -2136,6 +2221,52 @@ def default_project_name():
...
+
+    Names are looked up in the `opentelemetry_resource_detector` entry point group — the same group used by the
+    `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` environment variable. The special name `'*'` expands to every
+    registered detector except `'otel'` (whose `OTEL_RESOURCE_ATTRIBUTES`/`OTEL_SERVICE_NAME` handling is already
+    applied by `Resource.create`), matching the env var.
+
+    Unknown names, or detectors that fail to load, emit a warning and are skipped rather than raising, so that a
+    typo can never crash the application — matching OpenTelemetry's tolerant handling of the env var.
+    """
+    from importlib.metadata import entry_points
```
alexmojaki: move import up

logfire/_internal/config.py
dmontagu: Done in 6cbd5641. Moved `process.runtime.*` back into the cached `common_resource_attributes()` alongside `host.*`/`os.*`, and that function is now applied at the low-precedence fallback layer — so the runtime attrs are cached again and are overridable by env vars just like the host/os ones. `RESOURCE_ATTRIBUTES_VERSION` and `process.pid` stay inline in the high-precedence dict (pid in particular must not be cached — it’s re-set per fork).

logfire/_internal/config.py
dmontagu: Agreed — done in 6cbd5641. Removed `_service_version_from_git` entirely and moved the `get_git_revision_hash()` call into `_initialize`, where the resource is built: when no explicit `service_version` is set, the git hash is added directly to the low-precedence fallback layer. So it never lands on `self.service_version`, which means:

- no more import-time `git` subprocess (it only runs when building the resource), and
- the serialize-to-child workaround from the earlier review is gone too — `service_version` is just `None` on the config when auto-detected, so it serializes as `None` and the child re-detects it as a fallback rather than treating it as explicit.

Updated `test_configure_service_version` to assert via the resource (+ the no-git case), flipped the `service_version` bool in the config-span snapshot, and slimmed the serialization test.

docs/integrations/system-metrics.md
dmontagu: Follow-up: did the in-repo parts of your suggestions now too (cf8e151d) — added the reverse link from the Hosts page to the system-metrics SDK integration, and moved the reference-style resource_attributes/resource_detectors detail (env vars, pyproject, precedence, default host.*/os.* population) under the SQL `otel_resource_attributes` column docs.

tests/test_configure.py
```diff
@@ -943,6 +949,222 @@ def test_otel_service_name_has_priority_on_otel_resource_attributes_service_name
...
+
+    logfire.info('test1')
+
+    [span] = exporter.exported_spans_as_dict(include_resources=True)
+    attributes = span['resource']['attributes']
+    # From the `host`/`os` detectors:
+    assert attributes['host.name'] == socket.gethostname()
+    assert attributes['os.type'] == platform.system().lower()
+    # From the `process` detector:
+    assert 'process.runtime.name' in attributes
```
alexmojaki: all these attributes are always present by default and pointless to check, this test currently says nothing. it could check other attributes from the process detector tho. parametrising this to try `['*']`, `'process'`, and `['process']` should produce similar results and then test_resource_detectors_by_name (it has the same problem of checking os/host) and test_resource_detectors_string can be removed.

tests/test_configure.py
```diff
@@ -943,6 +949,222 @@ def test_otel_service_name_has_priority_on_otel_resource_attributes_service_name
...
+
+    logfire.info('test1')
+
+    [span] = exporter.exported_spans_as_dict(include_resources=True)
+    assert span['resource']['attributes'] == IsPartialDict({'host.name': socket.gethostname()})
+
+
+def test_resource_detector_unknown_name(config_kwargs: dict[str, Any], exporter: TestExporter) -> None:
+    # An unknown detector name warns and is skipped rather than raising, so a typo can't crash the app.
+    config_kwargs['resource_detectors'] = ['nonexistent-detector', 'host']
```
alexmojaki:
```suggestion
    config_kwargs['resource_detectors'] = ['nonexistent-detector', 'process']
```

tests/test_configure.py
```diff
@@ -943,6 +949,222 @@ def test_otel_service_name_has_priority_on_otel_resource_attributes_service_name
...
+        configure(**config_kwargs)
+
+    logfire.info('test1')
+
+    [span] = exporter.exported_spans_as_dict(include_resources=True)
+    attributes = span['resource']['attributes']
+    # `resource_detectors` takes precedence over `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`.
+    assert attributes['host.name'] == 'from-kwarg-detector'
+    # The env var detector still contributes attributes that the kwarg detector doesn't override.
+    assert attributes['host.arch'] == platform.machine()
```
alexmojaki: goddamn i'm so tired of this please AI update these tests using host

docs/reference/sql.md
```diff
@@ -268,6 +268,8 @@ Technically, each call to `logfire.configure()` can create a different set of re
 
 In **Logfire** and other OpenTelemetry SDKs, you can set arbitrary resource attributes by setting the `OTEL_RESOURCE_ATTRIBUTES` environment variable to a comma-separated list of key-value pairs, e.g. `OTEL_RESOURCE_ATTRIBUTES=service.name=my-service,service.version=1.0.0`.
 
+With **Logfire** specifically, you can also set them programmatically with the `resource_attributes` argument of [`logfire.configure()`][logfire.configure] (or the `LOGFIRE_RESOURCE_ATTRIBUTES` environment variable, or `pyproject.toml`), and run OpenTelemetry [resource detectors](https://opentelemetry.io/docs/languages/python/resources/) via the `resource_detectors` argument (or `LOGFIRE_RESOURCE_DETECTORS` / the standard `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`). Attributes set explicitly in `logfire.configure()` take precedence over detected attributes, which take precedence over the `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` environment variables. By default Logfire pre-populates `host.*`, `os.*` and `process.runtime.*`, so machines show up in the [Hosts view](../guides/web-ui/hosts.md) out of the box.
```
alexmojaki: Rather start by talking about how to do it in the *python logfire sdk* and then fallback to other otel sdks.

docs/integrations/system-metrics.md
```diff
@@ -23,6 +23,32 @@ logfire.instrument_system_metrics()
 
 Then in your project, click on 'Dashboards' in the top bar, click 'New Dashboard', and select 'Basic System Metrics (Logfire)' from the dropdown.
 
+## Host information
+
+By default, Logfire sets the `host.name`, `host.arch`, `os.type` and `os.version`
```
alexmojaki: The essential information is "there's a cool UI page that instantly shows these metrics", which belongs with the dashboard recommendation. Resource attributes are secondary. no one cares that host.arch is set.

alexmojaki: more broadly, each page currently only has one continuous section of lines added. look at the page as a whole and think about how best to update it so that the end results contains the desired information organised in the best way. make edits as needed, not just additions. just stay within the PR scope.

logfire/_internal/config.py
dmontagu: Done in cbd4ef06 — moved `resource_detectors` into `AdvancedOptions` (kept `resource_attributes` top-level, since setting `host.name` is a common need). The `LOGFIRE_RESOURCE_DETECTORS` env var and `pyproject.toml` support carry over, loaded alongside `base_url`/`emit_configuration_span`. Tests use `AdvancedOptions(resource_detectors=...)` and the docs point there.

logfire/_internal/config.py
dmontagu: Done in 3622ce00 — `service.instance.id` now goes into the low-precedence fallback dict (next to the `process.runtime.*`/`host.*`/`os.*` defaults and the git-hash `service.version`) instead of a separate trailing merge. Same precedence as before (a detector / env var / explicit value still overrides it), just one merge instead of two. It stays out of the `@functools.cache`d `common_resource_attributes()` since the UUID must be fresh per process.

logfire/_internal/config.py
dmontagu: Done in 3622ce00 — hoisted `from importlib.metadata import entry_points` to the module top.

tests/test_configure.py
dmontagu: Done in 3622ce00 — replaced the three tests with a single parametrized `test_resource_detectors` over `[*]`, `process`, and `[process]`, asserting process-detector-only attributes (`process.owner` / `process.executable.*`) that are not present by default. `test_resource_detectors_by_name` and `test_resource_detectors_string` are removed — the `[process]` and bare-string `process` params cover them (including the string coercion).

tests/test_configure.py
dmontagu: Done in 3622ce00 — switched to `[nonexistent-detector, process]` and assert a process-only attribute, so the test actually proves the valid detector ran (rather than checking `host.name`, which is populated by default anyway).

tests/test_configure.py
dmontagu: Done in 3622ce00 — swept the whole PR to stop using `host`/`os` as example detectors (redundant now that `host.*`/`os.*` are populated by default). The resource-detector tests, the env-var and pyproject tests, the bare-string coercion comment, and the docstring example all use the `process` detector now. `host.name` only remains as a resource-*attribute* example (the override use case) and in the default-population assertions.

docs/reference/sql.md
dmontagu: Done in b2dc1122 — the `otel_resource_attributes` section now leads with the Logfire Python SDK (`resource_attributes` / `AdvancedOptions.resource_detectors`, env vars, pyproject) and then falls back to `OTEL_RESOURCE_ATTRIBUTES` for other OTel SDKs.

docs/integrations/system-metrics.md
dmontagu: Done in b2dc1122 — restructured the page rather than just appending: the Hosts view now sits with the dashboard recommendation under Usage (the instant UI for these metrics), and resource attributes are demoted to a short "Customising resource attributes" note focused on overriding `host.name` (dropped the `host.arch`/`os.*` enumeration). Also updated for the `resource_detectors` → `AdvancedOptions` move.

alexmojaki:
@cubic-dev-ai review the whole PR, not just recent commits

alexmojaki:
@cubic-dev-ai review

logfire/_internal/config.py
```diff
@@ -246,6 +246,26 @@ def hook(helper: ServerResponseCallbackHelper):
     Raise from the hook to abort the calling code path.
     """
 
+    resource_detectors: Sequence[ResourceDetector | str] | None = None
```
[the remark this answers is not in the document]
dmontagu: Good catch on the mismatch, but it is the PR description that was stale, not the code. Unknown detector names emit a `LogfireConfigWarning` and are skipped (they never raise) — a deliberate change earlier in the review (0fe100d8) to match OpenTelemetry tolerant handling of `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`, so a typo cannot crash the app. The docstring is correct; I have updated the PR description to match (it previously described an earlier `ValueError` design that was dropped).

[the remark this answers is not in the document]
dmontagu: Good catch on the mismatch, but it is the PR description that was stale, not the code. Unknown detector names emit a `LogfireConfigWarning` and are skipped (they never raise) — a deliberate change earlier in the review (0fe100d8) to match OpenTelemetry tolerant handling of `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`, so a typo cannot crash the app. The docstring is correct; I have updated the PR description to match (it previously described an earlier `ValueError` design that was dropped).

docs/integrations/system-metrics.md
```diff
@@ -21,7 +21,32 @@ logfire.configure()
...
+    resource_attributes={
+        'host.name': 'my-meaningful-host-name',
+        'datacenter.region': 'eu-west-1',
+    },
+)
+
+logfire.instrument_system_metrics()
+```
+
+To customise further — for example to add OpenTelemetry resource detectors for `process.*` or cloud metadata —
```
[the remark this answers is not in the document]
dmontagu: Intentional. `resource_detectors` was deliberately de-emphasized on this page during review: now that `host.*`/`os.*` are populated by default it is a niche option, and it was moved off the top-level `configure()` API onto `AdvancedOptions`. This section focuses on the common case (`resource_attributes`, e.g. overriding `host.name`); the inline `resource_detectors=[process]` example lives in the `AdvancedOptions.resource_detectors` docstring that the sentence links to. Adding a second prominent example here would work against that simplification.

docs/integrations/system-metrics.md
```diff
@@ -21,7 +21,32 @@ logfire.configure()
 logfire.instrument_system_metrics()
 ```
 
-Then in your project, click on 'Dashboards' in the top bar, click 'New Dashboard', and select 'Basic System Metrics (Logfire)' from the dropdown.
+The machine running your code then shows up automatically in the [Hosts view](../guides/web-ui/hosts.md), which
+charts CPU, memory, load, disk and network per host. For a customisable view, click 'Dashboards' in the top bar,
+click 'New Dashboard', and select 'Basic System Metrics (Logfire)' from the dropdown.
```
alexmojaki: 1. the dashboard isn't customisable
2. the instructions need updating anyway

Image

docs/reference/sql.md
```diff
@@ -266,7 +266,9 @@ You can query it using the `->>` operator, similar to the [`attributes`](#attrib
 
 Technically, each call to `logfire.configure()` can create a different set of resource attributes, so it should only be called once if possible.
 
-In **Logfire** and other OpenTelemetry SDKs, you can set arbitrary resource attributes by setting the `OTEL_RESOURCE_ATTRIBUTES` environment variable to a comma-separated list of key-value pairs, e.g. `OTEL_RESOURCE_ATTRIBUTES=service.name=my-service,service.version=1.0.0`.
+In the **Logfire** Python SDK, set resource attributes with the `resource_attributes` argument of [`logfire.configure()`][logfire.configure] (or the `LOGFIRE_RESOURCE_ATTRIBUTES` environment variable, or `pyproject.toml`), and run OpenTelemetry [resource detectors](https://opentelemetry.io/docs/languages/python/resources/) via `resource_detectors` on [`AdvancedOptions`][logfire.AdvancedOptions] (or `LOGFIRE_RESOURCE_DETECTORS`). Attributes set explicitly in `logfire.configure()` take precedence over detected attributes, which take precedence over the environment variables below. By default Logfire pre-populates `host.*`, `os.*` and `process.runtime.*`, so machines show up in the [Hosts view](../guides/web-ui/hosts.md) out of the box.
```
alexmojaki: > detected attributes, which take precedence over the environment variables below

that sounds wrong

alexmojaki: it's worth being detailed here, this is a reference. show code.

logfire/_internal/config.py
```diff
@@ -246,6 +246,26 @@ def hook(helper: ServerResponseCallbackHelper):
...
+    """OpenTelemetry resource detectors to run when building the resource.
```
alexmojaki: link to otel docs

logfire/_internal/config.py
```diff
@@ -1063,28 +1103,58 @@ def _initialize(self) -> None:
                 otel_resource_attributes['service.version'] = self.service_version
             if self.environment:
                 otel_resource_attributes[RESOURCE_ATTRIBUTES_DEPLOYMENT_ENVIRONMENT_NAME] = self.environment
-            otel_resource_attributes_from_env = os.getenv(OTEL_RESOURCE_ATTRIBUTES)
-            if otel_resource_attributes_from_env:
-                for _field in otel_resource_attributes_from_env.split(','):
-                    key, value = _field.split('=', maxsplit=1)
-                    otel_resource_attributes[key.strip()] = value.strip()
+            otel_resource_attributes.update(self.resource_attributes)
```
alexmojaki: i think this should be higher, the service_name etc args should take priority over resource_attributes

logfire/_internal/config.py
```diff
@@ -2144,24 +2260,28 @@ def get_runtime_version() -> str:
...
+    The `host.*`/`os.*` attributes let the Logfire Hosts page surface a row without the customer opting into
+    the experimental `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` env var or passing `resource_detectors`. All of
+    these are applied below the env vars (and below the `resource_detectors`/`resource_attributes` configure
+    arguments), so any of those still wins — e.g. a customer whose `socket.gethostname()` is a useless
+    container ID can override `host.name` cleanly.
+    """
     return {
-        RESOURCE_ATTRIBUTES_VERSION: VERSION,
         # https://opentelemetry.io/docs/specs/semconv/resource/process/#python-runtimes
-        # TODO use ProcessResourceDetector?
```
alexmojaki, line 2151: leave this todo

tests/test_configure.py
```diff
@@ -1802,8 +1995,10 @@ def test_configuration_span_emitted_when_opted_in(config_kwargs: dict[str, Any],
                         'token_count': 0,
                         'api_key': False,
                         'service_name': False,
-                        'service_version': True,
+                        'service_version': False,
```
alexmojaki: why did this change

docs/integrations/system-metrics.md
dmontagu: Fixed in 45a64755. You are right that the standard dashboard is not customisable — I dropped that wording, and rather than re-spell the (now-changed) click path I point to the maintained UI docs instead. The "Usage" section now lists where the metrics actually surface: the [Hosts view](../guides/web-ui/hosts.md), the [Metrics explorer](../guides/web-ui/metrics-explorer.md), the **Basic System Metrics (Logfire)** standard dashboard on the [Dashboards page](../guides/web-ui/dashboards.md) (noting it is read-only, with a pointer to using it as a template for a custom dashboard), and querying the `metrics` table in [Explore](../guides/web-ui/explore.md). Thanks for the screenshot — confirmed against the current frontend.

docs/reference/sql.md
dmontagu: Good catch — reworded in 45a64755. The ambiguity was "detected attributes": the `resource_detectors` *argument* does beat the env vars, but the auto-detected *defaults* (host/os) sit below them, and the old sentence blurred the two. Replaced it with an explicit highest-to-lowest precedence list: (1) explicit `configure()` attributes, (2) the `resource_detectors` argument, (3) `OTEL_RESOURCE_ATTRIBUTES`/`OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`, (4) the pre-populated defaults.

dmontagu: Done in 45a64755 — expanded the section with `logfire.configure()`, `pyproject.toml`, and `OTEL_RESOURCE_ATTRIBUTES` code examples plus the full precedence list. (This is also where cubic wanted the `resource_detectors` example — the reference page is the right home for it, rather than the system-metrics integration page.)

logfire/_internal/config.py
dmontagu: Done in 45a64755 — the `AdvancedOptions.resource_detectors` docstring now links to the OpenTelemetry resource-detectors docs: https://opentelemetry.io/docs/concepts/resources/#resource-detectors. (The `languages/python/resources/` URL I first used was a 404; corrected in a follow-up commit.)

logfire/_internal/config.py
dmontagu: Agreed — fixed in 45a64755. `resource_attributes` now forms the base of the explicit layer, with the dedicated `service_name`/`service_version`/`environment` args (and `code_source`) applied on top, so a dedicated arg wins over the same key set generically via `resource_attributes`. A key with no dedicated arg still comes from `resource_attributes`. Added `test_dedicated_args_take_precedence_over_resource_attributes`.

docs/reference/sql.md
```diff
@@ -266,9 +266,33 @@ You can query it using the `->>` operator, similar to the [`attributes`](#attrib
...
+resource_detectors = ["process"]
+```
+
+By default Logfire pre-populates `host.*`, `os.*` and `process.runtime.*`, so machines show up in the [Hosts view](../guides/web-ui/hosts.md) out of the box.
+
+The precedence, from highest to lowest:
+
+1. Attributes set explicitly in `logfire.configure()`: `resource_attributes`, and the dedicated `service_name` / `service_version` / `environment` arguments (which override the same keys set via `resource_attributes`).
+2. The `resource_detectors` argument.
+3. The standard `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` environment variables — also the mechanism used by other OpenTelemetry SDKs, e.g. `OTEL_RESOURCE_ATTRIBUTES=service.name=my-service,service.version=1.0.0`.
```
alexmojaki: split 1 and 3 into pairs of items, also mention the LOGFIRE_ env vars

docs/reference/sql.md
dmontagu: Verified against the OTel SDK source and confirmed empirically. Stock OTel `Resource.create` builds its detector list with the `otel` detector (which reads `OTEL_RESOURCE_ATTRIBUTES`) placed *last*, so `OTEL_RESOURCE_ATTRIBUTES` takes precedence over the `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` detectors. Logfire delegates that interaction entirely to `Resource.create({})`, so it matches stock OTel exactly — empirically, with `OTEL_RESOURCE_ATTRIBUTES=host.name=X` + `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS=host`, `host.name` is `X`. The only logfire-specific layer is the `resource_detectors` *argument* (on `AdvancedOptions`), which sits above the env vars — consistent with `resource_attributes` (a configure param) beating `OTEL_RESOURCE_ATTRIBUTES`. I have rewritten the precedence list to make all of this explicit (see r3415168148).

logfire/_internal/config.py
dmontagu: Done — links to https://opentelemetry.io/docs/concepts/resources/#resource-detectors (the first URL I used, `languages/python/resources/`, was a 404; this one is verified real, anchor included).

logfire/_internal/config.py
dmontagu: Restored in 6d6983f4 — kept the `# TODO use ProcessResourceDetector?` comment.

tests/test_configure.py
dmontagu: Side effect of moving the git-hash `service_version` detection out of `_load_configuration` into `_initialize` (the refactor from the "this dance feels messy" thread). The git hash is no longer stored on the config object, so `config.service_version` is `None` when nothing was set explicitly, and this span field — `bool(config.service_version)` — now reflects whether a `service_version` was *explicitly set* rather than "a value exists including the git fallback". That is arguably the more useful signal, but happy to compute it differently (e.g. off the built resource) if you would rather it keep the old meaning.

docs/reference/sql.md
dmontagu: Done in 6d6983f4 — split into one item per layer (dedicated args > `resource_attributes` > `resource_detectors` > `OTEL_RESOURCE_ATTRIBUTES` > `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS` > defaults), noted the `LOGFIRE_*` env var / `pyproject.toml` equivalents for the first three, and added a line that items 4-5 are handled by OTel `Resource.create` (so `OTEL_RESOURCE_ATTRIBUTES` beating the env-var detectors matches stock OTel).

tests/test_configure.py
alexmojaki: can _initialize set self.service_version?

docs/integrations/system-metrics.md
```diff
@@ -21,7 +21,42 @@ logfire.configure()
...
+  break them down by dimension, no SQL required.
+- **[Dashboards](../guides/web-ui/dashboards.md)** — enable the built-in **Basic System Metrics (Logfire)**
+  standard dashboard. Standard dashboards are read-only; to adjust it,
+  [use it as a template](../guides/web-ui/dashboards.md#using-a-standard-dashboard-as-a-template) for a custom
+  dashboard.
+
+You can also query the metrics directly in the [Explore](../guides/web-ui/explore.md) view via the `metrics`
+table (see the [SQL reference](../reference/sql.md)).
+
+## Customising resource attributes
```
alexmojaki, line 38: now this section can just very briefly mention `host.name` and then link to the SQL reference (with an anchor) which has plenty of detail.

docs/guides/web-ui/hosts.md
```diff
@@ -103,6 +103,10 @@ The Hosts inventory keys hosts on `host.id` and `host.name`. Without them, a sin
 
 Hosts inside a Kubernetes cluster are recognized as nodes and appear on the [Kubernetes view](kubernetes.md) as well as in the Hosts inventory. The dedup hinges on `host.name` matching `k8s.node.name`. If you run a node-scoped DaemonSet collector, set both attributes from the downward API (`spec.nodeName`) so the same string lands in both places — otherwise the Hosts list will show one row per pod replica instead of one row per node.
 
+### From a Python application
+
+If your workload is a Python app already using Logfire, you can emit system metrics directly from the SDK with [`logfire.instrument_system_metrics()`](../../integrations/system-metrics.md) instead of (or alongside) running a collector. Logfire pre-populates the `host.name` resource attribute (overridable via `resource_attributes`), so the app's host shows up here automatically.
```
alexmojaki: link to sql reference

logfire/_internal/config.py
```diff
@@ -246,6 +246,27 @@ def hook(helper: ServerResponseCallbackHelper):
     Raise from the hook to abort the calling code path.
     """
 
+    resource_detectors: Sequence[ResourceDetector | str] | None = None
```
alexmojaki, line 249: link to sql reference

logfire/_internal/config.py
```diff
@@ -515,6 +537,13 @@ def configure(
 
             Defaults to the `LOGFIRE_ENVIRONMENT` environment variable.
 
+        resource_attributes: Additional
```
alexmojaki, line 542: link to sql reference

logfire/_internal/config.py
```diff
@@ -1045,11 +1084,18 @@ def _initialize(self) -> None:
...
             otel_resource_attributes: dict[str, Any] = {
-                **common_resource_attributes(),
-                'service.name': self.service_name,
+                **self.resource_attributes,
+                RESOURCE_ATTRIBUTES_VERSION: VERSION,
                 'process.pid': os.getpid(),
             }
+            if self.service_name:
+                # When unset, `Resource.create` below applies its own `unknown_service` default.
+                otel_resource_attributes['service.name'] = self.service_name
```
alexmojaki: emit a warning when any of the specific arguments like service_name overrides self.resource_attributes. do it in a DRY way.

logfire/_internal/config.py
```diff
@@ -1063,28 +1109,57 @@ def _initialize(self) -> None:
...
+                detectors = _load_resource_detectors(self.advanced.resource_detectors)
+                if detectors:
+                    # Detectors passed explicitly via `resource_detectors` take precedence over those from
+                    # `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`, but not over the explicit attributes merged below.
+                    resource = resource.merge(get_aggregated_resources(detectors, Resource.get_empty()))
+
+            # Explicit `logfire.configure()` attributes (logfire's own plus the `resource_attributes`
+            # argument) take precedence over everything detected above, including the env vars
+            # `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`.
+            resource = resource.merge(Resource(otel_resource_attributes))
```
alexmojaki: move the whole resource construction to a method

logfire/_internal/config.py
```diff
@@ -2136,6 +2213,50 @@ def default_project_name():
...
+            )
+            continue
+        _load_resource_detector_entry_point(matching[0], result)
+    return result
+
+
+def _load_resource_detector_entry_point(entry_point: Any, result: list[ResourceDetector]) -> None:
+    try:
+        result.append(entry_point.load()())
+    except Exception as exc:  # pragma: no cover
```
alexmojaki: is the pragma needed?

logfire/_internal/config.py
```diff
@@ -2136,6 +2213,50 @@ def default_project_name():
...
+                category=LogfireConfigWarning,
+            )
+            continue
+        _load_resource_detector_entry_point(matching[0], result)
+    return result
+
+
+def _load_resource_detector_entry_point(entry_point: Any, result: list[ResourceDetector]) -> None:
+    try:
+        result.append(entry_point.load()())
```
```suggestion
def _load_resource_detector_entry_point(entry_point: Any):
    try:
        return entry_point.load()()
```

alexmojaki: is the type really unknown?

tests/test_configure.py
dmontagu: Done in faf8dd01 — `_build_resource()` now stores the git-hash fallback back on `self.service_version` (so the configuration span field and serialization reflect it again), while still applying it via the low-precedence `fallback_resource_attributes` so `OTEL_RESOURCE_ATTRIBUTES` keeps winning *within* the process. Updated `test_configure_service_version` and `test_auto_detected_service_version_serialized` to match, and the config-span snapshot back to `service_version: True`.

docs/integrations/system-metrics.md
dmontagu: Done in faf8dd01 — trimmed this section to a one-line pointer (set a clearer `host.name`, or add other attributes such as `process.*`/cloud metadata) linking to the SQL reference [`#resource-attributes`](../reference/sql.md#resource-attributes) anchor, and dropped the redundant code sample.

docs/guides/web-ui/hosts.md
dmontagu: Done in faf8dd01 — added a link to the [SQL reference](../../reference/sql.md#resource-attributes) next to the `resource_attributes` mention.

logfire/_internal/config.py
dmontagu: Done in faf8dd01 — the `resource_detectors` docstring now links to the SQL reference `#resource-attributes` anchor for the full precedence list and how to query the attributes.

logfire/_internal/config.py
dmontagu: Done in faf8dd01 — the `resource_attributes` docstring now links to the SQL reference `#resource-attributes` anchor.

logfire/_internal/config.py
dmontagu: Done in faf8dd01 — `_build_resource()` collects the dedicated args into a `dedicated_attributes` dict and warns once per key (single loop) when it overrides a *different* value the user also set via `resource_attributes`. Covered by the updated `test_dedicated_args_take_precedence_over_resource_attributes`.

logfire/_internal/config.py
dmontagu: Done in faf8dd01 — extracted the whole resource construction into `LogfireConfig._build_resource()`; `_initialize` now just does `resource = self._build_resource()`.

logfire/_internal/config.py
dmontagu: Good catch — the pragma was hiding an untested branch. Removed it and added `test_resource_detector_load_failure`, which patches `entry_points` with a detector that raises on load and asserts the warning fires and configuration still succeeds.

logfire/_internal/config.py
dmontagu: Done in faf8dd01 — `_load_resource_detector_entry_point` now returns `ResourceDetector | None` (callers append when non-None) instead of mutating a `result` list. And no, the type wasn't really unknown — annotated it as `importlib.metadata.EntryPoint`.

PR #2013, opened 2026-06-11
Pre-populate `host.*` and `os.*` resource attributes

strawgate:
Pre-populate four resource attributes the OTel SDK ships detectors for but leaves opt-in behind `OTEL_EXPERIMENTAL_RESOURCE_DETECTORS`:

| Attribute | Source |
|---|---|
| `host.name` | `socket.gethostname()` |
| `host.arch` | `platform.machine()` |
| `os.type` | `platform.system().lower()` |
| `os.version` | `platform.release()` (Windows/Solaris: `platform.version()`) |

Values match what `_HostResourceDetector` + `OsResourceDetector` would emit. Follows the same pattern as the existing `process.runtime.*` pre-pop in this file. `OTEL_RESOURCE_ATTRIBUTES` still wins (processed after this dict).


strawgate:
Thanks — addressed in 16bbfa56: the default-host test now clears `OTEL_RESOURCE_ATTRIBUTES` via `os.environ.pop` inside a `patch.dict` so an inherited env var can't override the SDK defaults.

strawgate:
Re: `os.description`-dropped flag — intentional. An earlier draft of this PR added `os.description` so the Hosts page OS column would have a longer label; on review we decided that was the SDK conforming to a downstream UX gap rather than the SDK doing the right thing. The final shape strictly mirrors what OTel's `OsResourceDetector` would emit (`os.type` + `os.version`). The Hosts page can render `os.type + os.version` together if it wants a richer cell. Customers who need `os.description` populated can set `OTEL_RESOURCE_ATTRIBUTES=os.description=…`.

alexmojaki (approved):
@strawgate maybe check my changes

PR #2014, opened 2026-06-12
Add `system.cpu.load_average.1m` and `system.process.count`

strawgate:
Upstream `SystemMetricsInstrumentor` emits neither metric, so the Logfire wrapper adds both via the same callback pattern as `system.cpu.simple_utilization`.

Closes the gap where the Hosts page's **1-minute load** and **Running process count** columns were permanently empty for SDK-instrumented hosts.

## Naming

Review asked for both metrics to be renamed toward OTel semantic conventions. Only one of them is actually semconv:

- **`system.process.count` is semconv** — an UpDownCounter in `{process}` with a Recommended `process.state` attribute. Renamed from `system.processes.count` and switched from a gauge to `metric_up_down_counter_callback` accordingly. `process.state` is left out: `psutil.pids()` is one cheap syscall, whereas splitting by state needs a `process_iter` over every process and multiplies the series count — which matters now that this is in the basic base.
- **`system.cpu.load_average.1m` stays.** Semconv defines no load-average metric. The [system metrics spec](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/) mentions `system.linux.cpu.load_1m` only as a hypothetical illustration of OS-prefixed naming ("could be named"), not as a definition. The name, gauge instrument, and `{thread}` unit here match the OTel Collector's `hostmetrics` load scraper, which is where most host load averages reaching Logfire come from — so SDK-reported and Collector-reported hosts land on one series rather than two.

## Basic base

Both metrics moved from `base='full'` to the basic base. Each is a single number per collection with no attributes, and both fill columns the Hosts page shows by default.

## Follow-up needed on the platform side

The Hosts inventory query matches on `system.processes.count` only. It needs `system.process.count` added alongside it, or the **Running process count** column stays empty for SDK-instrumented hosts. The Collector keeps emitting the older `system.processes.count`, so both names need to be accepted.

## Notes

- On Windows, `psutil.getloadavg()` is an emulated polling shim: the first call returns 0 and values update every ~5s. Documented in the helper's docstring.
- Our recommended Collector config for the Hosts page sets `load: cpu_average: true`, which normalises load by CPU count; `psutil.getloadavg()` does not. Same metric name, slightly different scale depending on source. Left as raw load average here since that is the conventional meaning, but worth a look.

### DIFF
--- a/logfire/_internal/integrations/system_metrics.py
+++ b/logfire/_internal/integrations/system_metrics.py
@@ -31,6 +31,7 @@ MetricName: type[
         'system.cpu.simple_utilization',
         'system.cpu.time',
         'system.cpu.utilization',
+        'system.cpu.load_average.1m',
         'system.memory.usage',
         'system.memory.utilization',
         'system.swap.usage',
@@ -43,6 +44,7 @@ MetricName: type[
         'system.network.errors',
         'system.network.io',
         'system.network.connections',
+        'system.processes.count',
         'system.thread_count',
         'process.open_file_descriptor.count',
         'process.context_switches',
@@ -62,6 +64,7 @@ MetricName: type[
     'system.cpu.simple_utilization',
     'system.cpu.time',
     'system.cpu.utilization',
+    'system.cpu.load_average.1m',
     'system.memory.usage',
     'system.memory.utilization',
     'system.swap.usage',
@@ -74,6 +77,7 @@ MetricName: type[
     'system.network.errors',
     'system.network.io',
     'system.network.connections',
+    'system.processes.count',
     'system.thread_count',
     'process.open_file_descriptor.count',
     'process.context_switches',
@@ -109,6 +113,11 @@ FULL_CONFIG: Config = {
     'process.cpu.core_utilization': None,
     'system.cpu.time': CPU_FIELDS,
     'system.cpu.utilization': CPU_FIELDS,
+    # The Hosts page reads these — upstream `_DEFAULT_CONFIG` doesn't include
+    # them and they aren't expensive (one number per scrape each), so add them
+    # so `base='full'` covers the page without a follow-up `config={…}` knob.
+    'system.cpu.load_average.1m': None,
+    'system.processes.count': None,
     # For usage, knowing the total amount of bytes available might be handy.
     'system.memory.usage': MEMORY_FIELDS + ['total'],
     # For utilization, the total is always just 1 (100%), so it's not included.
@@ -172,6 +181,14 @@ def instrument_system_metrics(logfire_instance: Logfire, config: Config | None =
         # Override OTEL here to avoid emitting 0 in the first measurement.
         measure_process_cpu_utilization(logfire_instance)
         del config['process.cpu.utilization']
+
+    if 'system.cpu.load_average.1m' in config:
+        measure_system_cpu_load_average_1m(logfire_instance)
+        del config['system.cpu.load_average.1m']
+
+    if 'system.processes.count' in config:
+        measure_system_processes_count(logfire_instance)
+        del config['system.processes.count']
 
     instrumentor = SystemMetricsInstrumentor(config=config)  # pyright: ignore[reportArgumentType]
     instrumentor.instrument(meter_provider=logfire_instance.config.get_meter_provider())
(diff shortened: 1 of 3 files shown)


logfire/_internal/integrations/system_metrics.py
```diff
@@ -248,3 +265,41 @@ def callback(_options: CallbackOptions) -> Iterable[Observation]:
...
+
+    Upstream `SystemMetricsInstrumentor` doesn't include this metric; we add
+    it so `base='full'` covers the Hosts page's `Procs` column.
+    """
+
+    def callback(_options: CallbackOptions) -> Iterable[Observation]:
+        yield Observation(len(psutil.pids()))
+
+    logfire_instance.metric_gauge_callback(
+        'system.processes.count',
```
alexmojaki: I think this should probably be `system.process.count` as in https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemprocesscount, which is an updowncounter. If including the `process.state` attribute isn't easy it's probably fine to leave out.

logfire/_internal/integrations/system_metrics.py
```diff
@@ -248,3 +265,41 @@ def callback(_options: CallbackOptions) -> Iterable[Observation]:
...
+    so `base='full'` covers the Hosts page's `Load 1m` column. On Windows,
+    `psutil.getloadavg()` is an emulated polling shim — first call returns 0
+    and subsequent values update every ~5s.
+    """
+
+    def callback(_options: CallbackOptions) -> Iterable[Observation]:
+        yield Observation(psutil.getloadavg()[0])
+
+    logfire_instance.metric_gauge_callback(
+        'system.cpu.load_average.1m',
```
alexmojaki, line 292: https://opentelemetry.io/docs/specs/semconv/system/system-metrics/ mentions a hypothetical `system.linux.cpu.load_1m`. Not a real semconv, but worth mimicking. If we don't want to include the OS prefix, then `system.cpu.load_1m`.

logfire/_internal/integrations/system_metrics.py
```diff
@@ -109,6 +113,11 @@
     'process.cpu.core_utilization': None,
     'system.cpu.time': CPU_FIELDS,
     'system.cpu.utilization': CPU_FIELDS,
+    # The Hosts page reads these — upstream `_DEFAULT_CONFIG` doesn't include
+    # them and they aren't expensive (one number per scrape each), so add them
+    # so `base='full'` covers the page without a follow-up `config={…}` knob.
```
alexmojaki: If these are so cheap, and they're used in a default UI view, then they should be in the basic config.

tests/otel_integrations/test_system_metrics.py
```diff
@@ -186,6 +200,8 @@ def test_full_base():
         'cpython.gc.collected_objects': None,
         'cpython.gc.collections': None,
         'cpython.gc.uncollectable_objects': None,
+        'system.cpu.load_average.1m': None,
+        'system.processes.count': None,
     }, 'Docs and the MetricName type need to be updated if this test fails'
```
alexmojaki, line 209: docs weren't updated

alexmojaki (changes requested):

logfire/_internal/integrations/system_metrics.py
dmontagu: So the point is — rename the attribute to system.process.count, and make it an updowncounter not a gauge. If convenient, we can add the process.state attribute, but note that could increase the number of series significantly, which we probably wouldn't want to do in default config if it costs a decent amount. @strawgate do you have thoughts about whether we should add that attribute?

Alternatively we could change the name more significantly, I don't think it should be close to the semantic convention one though if we do that (like it is now). But I'm okay using this name and not having the attribute if we think this is good.

logfire/_internal/integrations/system_metrics.py
dmontagu: @strawgate Can we just change the name to this? Is there a reason you used the existing name?

tests/otel_integrations/test_system_metrics.py
dmontagu: (the test explains that if it fails the docs need to be updated, we're guessing the AI fixed the test pre-emptively but didn't realize it also needed to update the docs.)

strawgate:
Rebased on main and worked through the review. Splitting the two rename requests, because only one of them is actually semconv:

**`system.processes.count` → `system.process.count` — done.** [semconv defines it](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemprocesscount) as an UpDownCounter in `{process}`, so it's renamed and switched from a gauge to `metric_up_down_counter_callback`. Left `process.state` out: `psutil.pids()` is one cheap syscall, while splitting by state needs a `process_iter` over every process and multiplies the series count — which matters more now that it's in the basic base. Easy to add later if we want it.

**`system.cpu.load_average.1m` — I'd like to keep this one.** Semconv defines no load-average metric at all. The [system metrics spec](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/) mentions `system.linux.cpu.load_1m` only as a hypothetical illustration of OS-prefixed naming ("An instrument for load average over 1 minute on Linux *could be named*..."), which matches @alexmojaki's own read that it's "not a real semconv". So there's no convention to conform to, and the alternative is inventing a name nobody emits.

What does exist is the OTel Collector's `hostmetrics` load scraper, which emits `system.cpu.load_average.1m` as a gauge in `{thread}`. That's where most host load averages reaching Logfire come from, and it's what the Hosts inventory query matches on. Matching it puts SDK-reported and Collector-reported hosts on one series instead of two. I've also switched the unit from `{run_queue_item}` to the Collector's `{thread}` for the same reason.

**Basic config — done.** Both moved out of `base='full'`. Each is one number per collection with no attributes, and both fill columns the Hosts page shows by default, which was your point.

**Docs — done.** `docs/integrations/system-metrics.md` now lists both in the basic and full base examples, with annotations for what each measures.

One thing this surfaces: the Hosts inventory query matches `system.processes.count` only, so it needs `system.process.count` added alongside it or the column stays empty for SDK-instrumented hosts. Noted in the PR body; happy to open the platform PR.

@dmontagu re "is there a reason you used the existing name?" — for the process count, no good one, hence the rename. For load average, the reason is the Collector receiver above.

strawgate:
Opened the companion platform PR: pydantic/platform#34048. It teaches the Hosts inventory and detail queries to match both `system.process.count` and the Collector's `system.processes.count`, so the Running process count column keeps working for Collector-reported hosts and starts working for SDK-reported ones.

Worth landing that one first, or together with this: merging this SDK rename alone would blank the column for SDK-instrumented hosts until it ships.

alexmojaki (approved):

PR #2015, opened 2026-06-12
Add `logfire.variables_used` array attribute listing managed variables in scope

dmontagu:
## Motivation

Managed variables stamp spans with one attribute key per variable (`logfire.variables.<name> = <label>`, propagated via baggage). Answering "which spans used variable X?" therefore requires JSON **key-existence** predicates, which the Logfire backend cannot index or prune — every such query is a full scan of the time window, and the caller must enumerate every possible key from the project's variable definitions up front. On a large project this made the agent-detail page's managed-variables query time out (it was ~30 unprunable full scans; mitigated server-side in pydantic/platform#23417, but a full scan remains the floor).

Emitting the variable names as an attribute **value** turns this into a term query: no key enumeration, and the backend can shred + index the array so file pruning applies (fusionfire already supports `array_has` pruning predicates for indexed array columns — follow-up on the platform side).

## Change

Spans created while managed variables are in scope get one additional array-valued attribute alongside the existing per-variable keys:

```
logfire.variables.agent_prompt = 'production'          # existing
logfire.variables.agent_prompt.version = '3'           # existing
logfire.variables_used = ['agent_prompt', 'ai_chat']   # new
```

- Derived in `DirectBaggageAttributesSpanProcessor.on_start`, which already iterates baggage entries on every span start — no extra pass; spans with no variables in scope don't get the attribute.
- `.version` companion entries are excluded; unambiguous because variable names match `^[a-zA-Z_][a-zA-Z0-9_]*$` (no dots).
- New `ATTRIBUTES_VARIABLES_PREFIX` / `ATTRIBUTES_VARIABLES_USED_KEY` constants; `variables/abstract.py` now builds its baggage keys from the same prefix constant so the two sides can't drift.
- Since variables propagate via baggage, downstream-service spans get the array too — same coverage as the per-variable keys.
- Docs: `managed-variables/index.md` now lists the concrete attributes downstream spans receive.

### DIFF
--- a/docs/reference/advanced/managed-variables/index.md
+++ b/docs/reference/advanced/managed-variables/index.md
@@ -226,6 +226,12 @@ The recommended pattern is to use the variable's `.get()` method as a context ma
 
 When using the Logfire SDK, baggage values are automatically added as attributes to all downstream spans. This means any spans created inside the context manager will be tagged with which label and version was used, making it easy to filter and compare behavior in the Logfire UI.
 
+Concretely, spans created inside the context manager get:
+
+- `logfire.variables.<name>`: the selected label (or `<code_default>` when no label was selected)
+- `logfire.variables.<name>.version`: the resolved version, when a versioned value was used
+- `logfire.variables_used`: an array of the names of all variables currently in scope, so you can efficiently query for spans that used a given variable without knowing every possible attribute key
+
 ```python skip="true"
 from pydantic_ai import Agent
 
--- a/logfire/_internal/baggage.py
+++ b/logfire/_internal/baggage.py
@@ -7,6 +7,7 @@ from contextlib import contextmanager
 from opentelemetry import baggage, context
 from opentelemetry.sdk.trace import Span, SpanProcessor
 
+from .constants import ATTRIBUTES_VARIABLES_PREFIX, ATTRIBUTES_VARIABLES_USED_KEY
 from .json_encoder import logfire_json_dumps
 
 __all__ = (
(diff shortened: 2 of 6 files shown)


docs/reference/advanced/managed-variables/index.md
```diff
@@ -226,6 +226,12 @@ The recommended pattern is to use the variable's `.get()` method as a context ma
 
 When using the Logfire SDK, baggage values are automatically added as attributes to all downstream spans. This means any spans created inside the context manager will be tagged with which label and version was used, making it easy to filter and compare behavior in the Logfire UI.
 
+Concretely, spans created inside the context manager get:
+
+- `logfire.variables.<name>`: the selected label (or `<code_default>` when no label was selected)
+- `logfire.variables.<name>.version`: the resolved version, when a versioned value was used
+- `logfire.variables_used`: an array of the names of all variables currently in scope, so you can efficiently query for spans that used a given variable without knowing every possible attribute key
```
alexmojaki, line 233: FF doesn't actually currently support a good way to check if a JSON array (which is different from a real array like `tags`) contains an item. Will ingest convert this attribute to an array?

logfire/_internal/baggage.py
```diff
@@ -106,4 +114,11 @@ def on_start(self, span: Span, parent_context: context.Context | None = None) ->
                     continue
                 k = 'baggage_conflict.' + k
             attrs[k] = v
+        if variable_names:
+            # Also emit the variable names as a single array-valued attribute: querying
+            # "spans that used variable X" by *value* doesn't require enumerating every
+            # possible `logfire.variables.<name>` key, and array element values can be
+            # indexed for file pruning where JSON key-existence checks cannot.
```
alexmojaki, line 121: It sure feels like FF *should* be able to index `?`

tests/test_baggage.py
alexmojaki, line 1: these tests are redundant AFAICT, the variables test covers everything

dmontagu:
I'm going to close this for now, I am not confident enough in FF and this should probably be discussed with Adrian in more depth, but this isn't a problem we need to fix today.

PR #2017, opened 2026-06-12
fix: guard None text in Anthropic TextBlock during streaming

### LINKED ISSUE #1998: instrument_anthropic crashes streaming responses with TypeError when a text block has text=None (web_search/web_fetch citations)
### Description

**What I'm seeing**

`logfire.instrument_anthropic()` aborts an in-progress streaming response with:

```
File ".../logfire/_internal/integrations/llm_providers/anthropic.py", line 288, in get_response_data
    return {'combined_chunk_content': ''.join(texts), 'chunk_count': self._chunk_count}
TypeError: sequence item 0: expected str instance, NoneType found
```

**Root cause**

`AnthropicMessageStreamState.get_response_data` joins the text of every text block in the accumulated message:

```python
texts = [block.text for block in self._message.content if isinstance(block, (TextBlock, BetaTextBlock))]
return {'combined_chunk_content': ''.join(texts), ...}
```

`TextBlock.text` / `BetaTextBlock.text` are typed `str`, but the Anthropic SDK accumulates the streaming snapshot via `construct_type(...)` (`anthropic/lib/streaming/_beta_messages.py`), which **bypasses validation**. A text block that is opened (`content_block_start`) and only ever receives a `citations_delta` — no `text_delta` — keeps `text=None`. This happens routinely with the **`web_search` / `web_fetch` builtin tools**, whose responses carry citation-bearing text blocks. The result is `''.join([None, ...])` → `TypeError`.

**Why this is more than a dropped span attribute**

The join runs in `record_streaming`'s `__exit__`, i.e. **inside the instrumented stream iterator** (`llm_providers/llm_provider.py`, `LogfireInstrumentedAsyncStream.__stream__`). So the exception propagates out of the async generator into application code and **kills a response the model already produced successfully** — it is not contained to telemetry. We saw 50+ aborted end-user chat responses in production from this (model `claude-opus-4-6`, all using `web_search`).

**Expected**

Instrumentation must never raise into the stream. Skipping `None` text (as the OpenAI path already does in `OpenaiCompletionStreamState.record_chunk`, which only appends truthy content) resolves it:

```python
texts = [
    block.text
    for block in self._message.content
    if isinstance(block, (TextBlock, BetaTextBlock)) and block.text is not None
]
```

Still reproduces on `main` (4.35.0). I'd be happy to open a PR with a fix + test if that's helpful.

**Note:** `convert_response_to_semconv` (the `'latest'` semconv path) also reads `block.text` and would store `content=None` for such a block — doesn't crash, but may warrant the same coalescing.

### Python, Logfire & OS Versions, related packages

```toml
logfire="4.34.0"  # also reproduces on main (4.35.0)
anthropic="0.96.0"
platform="Linux (prod); also repro'd on macOS-26.0-arm64"
python="3.11"
[related_packages]
pydantic="2.13.4"
opentelemetry-api="1.39.1"
opentelemetry-sdk="1.39.1"
opentelemetry-semantic-conventions="0.60b1"
```

axelray-dev:
Fixes #1998

## Problem

`instrument_anthropic` crashes with `TypeError` during streaming when Anthropic's API returns `TextBlock` objects with `text=None`. This happens with `web_search` and `web_fetch` tool results where `construct_type` produces blocks with `text=None`.

The crash occurs in two locations:

1. `get_response_data()`: `''.join(texts)` fails because `texts` contains `None` values from `[block.text for block in ...]`
2. `convert_response_to_semconv()`: passes `None` content to `TextPart`, which can cause downstream issues

## Fix

Guard both locations against `None` text using `getattr(block, 'text', None)` (pyright-compatible with Anthropic SDK's `str` annotation, since `model_construct` bypasses validation):

- `convert_response_to_semconv()`: skip TextBlock/BetaTextBlock when `text` is `None` (falls through to generic dict handler)
- `get_response_data()`: filter out blocks with `None` text from the joined content

## Tests

Added `test_messages_stream_text_block_none` that directly tests both crash sites:
- Creates a `Message` with a `text=None` TextBlock and a normal TextBlock
- Verifies `get_response_data()` returns only the non-None text
- Verifies `convert_response_to_semconv()` handles the None-text block gracefully

```
uv run pytest tests/otel_integrations/test_anthropic.py -k test_messages_stream_text_block_none
uv run pytest tests/otel_integrations/test_anthropic.py  # all 23 tests pass
uv run pyright logfire/_internal/integrations/llm_providers/anthropic.py  # 0 errors
uv run ruff format --check logfire/_internal/integrations/llm_providers/anthropic.py  # clean
```

### DIFF
--- a/logfire/_internal/integrations/llm_providers/anthropic.py
+++ b/logfire/_internal/integrations/llm_providers/anthropic.py
@@ -236,8 +236,9 @@ def convert_response_to_semconv(message: Message | BetaMessage) -> OutputMessage
     parts: list[MessagePart] = []
 
     for block in message.content:
-        if isinstance(block, (TextBlock, BetaTextBlock)):
-            parts.append(TextPart(type='text', content=block.text))
+        text = getattr(block, 'text', None)
+        if isinstance(block, (TextBlock, BetaTextBlock)) and text is not None:
+            parts.append(TextPart(type='text', content=text))
         elif isinstance(block, (ToolUseBlock, BetaToolUseBlock)):
             parts.append(
                 make_tool_call_part(
@@ -284,7 +285,11 @@ class AnthropicMessageStreamState(StreamState):
     def get_response_data(self) -> Any:
         if self._message is None:
             return {'combined_chunk_content': '', 'chunk_count': 0}
-        texts = [block.text for block in self._message.content if isinstance(block, (TextBlock, BetaTextBlock))]
+        texts = [
+            text
+            for block in self._message.content
+            if isinstance(block, (TextBlock, BetaTextBlock)) and (text := getattr(block, 'text', None)) is not None
+        ]
         return {'combined_chunk_content': ''.join(texts), 'chunk_count': self._chunk_count}
 
     def get_attributes(self, span_data: dict[str, Any]) -> dict[str, Any]:
--- a/tests/otel_integrations/test_anthropic.py
+++ b/tests/otel_integrations/test_anthropic.py
@@ -633,8 +633,8 @@ def test_messages_stream_text_block_none() -> None:
         usage=Usage(input_tokens=25, output_tokens=55),
     )
     state = AnthropicMessageStreamState()
-    state._message = message
-    state._chunk_count = 1
+    setattr(state, '_message', message)
+    setattr(state, '_chunk_count', 1)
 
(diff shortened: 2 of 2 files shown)


alexmojaki:
Thanks!

PR #2042, opened 2026-06-30
Raise opentelemetry-sdk ceiling to <1.44.0 (allow 1.43)

### LINKED ISSUE #2041: Raise opentelemetry-sdk ceiling to allow >=1.43 (blocks otel-instrumentation-fastapi 0.64b0 FastAPI 0.137 fix)
### Summary

logfire 4.37.0 pins `opentelemetry-sdk<1.43.0,>=1.39.0`. This ceiling transitively prevents upgrading `opentelemetry-instrumentation-fastapi` to `0.64b0`, which contains the fix for a 500-error regression on FastAPI >= 0.137. Could the `opentelemetry-sdk` upper bound be raised to allow `>=1.43`?

### Background / dependency chain

- **FastAPI 0.137** introduced `_IncludedRouter` — routes added via `include_router()` are no longer flattened into `app.routes`, so some route objects no longer expose a `.path` attribute.
- This crashes `opentelemetry-instrumentation-fastapi`'s `_get_route_details()` with `AttributeError: '_IncludedRouter' object has no attribute 'path'`. Because the OTelASGI middleware runs on every request, this **500s every CORS preflight (`OPTIONS`) request**, which browsers then surface as CORS errors on the real request.
  - Upstream bug: open-telemetry/opentelemetry-python-contrib#4699
  - Fix (handles full + partial route matches): open-telemetry/opentelemetry-python-contrib#4700
  - Duplicate preflight report: open-telemetry/opentelemetry-python-contrib#4732
- The fix shipped in **`opentelemetry-instrumentation-fastapi 0.64b0`**
  (released 2026-06-24, tagged "1.43.0/0.64b0").
- `0.64b0` requires **`opentelemetry-sdk >= 1.43`**.
- **logfire 4.37.0 pins `opentelemetry-sdk<1.43.0,>=1.39.0`**, so we cannot pull `0.64b0`.

### Impact

Anyone using `logfire.instrument_fastapi()` is effectively pinned to FastAPI `<0.137`. Upgrading FastAPI to 0.137+ while on logfire 4.37.0 breaks all browser CORS preflights with 500s. We've had to pin `fastapi>=0.136.3,<0.137` as a workaround.

### Ask

Bump the `opentelemetry-sdk` upper bound (and any related `opentelemetry-*` pins) to allow `>=1.43`, so `opentelemetry-instrumentation-fastapi 0.64b0` can be installed and FastAPI 0.137+ is usable again.

### Environment

- logfire: 4.37.0
- opentelemetry-sdk: 1.39.1 (capped by `opentelemetry-sdk<1.43.0,>=1.39.0`)
- opentelemetry-instrumentation-fastapi: 0.60b1
- fastapi: pinned to <0.137 as a workaround

hramezani:
Closes #2041

## Summary

logfire pins `opentelemetry-sdk<1.43.0` (and the matching `opentelemetry-exporter-otlp-proto-http`). That ceiling transitively blocks `opentelemetry-instrumentation-fastapi 0.64b0`, which requires `opentelemetry-sdk>=1.43` and contains the fix for the FastAPI ≥0.137 `_IncludedRouter` regression — `_get_route_details()` raises `AttributeError`, 500-ing every CORS preflight (`OPTIONS`) request (open-telemetry/opentelemetry-python-contrib#4700).

Anyone on `logfire.instrument_fastapi()` is therefore effectively pinned to FastAPI `<0.137`.

This bumps the upper bound to `<1.44.0` so the 1.43.0 line (and fastapi-instrumentation `0.64b0`) can be installed.

```diff
-    "opentelemetry-sdk >= 1.39.0, < 1.43.0",
-    "opentelemetry-exporter-otlp-proto-http >= 1.39.0, < 1.43.0",
+    "opentelemetry-sdk >= 1.39.0, < 1.44.0",
+    "opentelemetry-exporter-otlp-proto-http >= 1.39.0, < 1.44.0",
```

## Verification

Installed the exact combination from the issue (fastapi 0.138.x + opentelemetry-sdk 1.43.0 + opentelemetry-instrumentation-fastapi 0.64b0) and ran the relevant suites:

- `tests/otel_integrations/test_fastapi.py`, `test_starlette.py`, `test_asgi.py` → **all pass** (22).
- Core config/exporter tests compared on 1.42.1 vs 1.43.0 on the same interpreter: **no functional regression**, only test-snapshot churn from two upstream OTel changes (both handled below).

## Test robustness for 1.43 (included in this PR)

OTel 1.43 makes two output changes that only surface in tests, addressed so CI stays green whether the lock resolves 1.42 or 1.43:

1. **`service.instance.id` is now a dashed UUID4** on the default resource (was the 32-char `uuid4().hex` form in ≤1.42). The `TestExporter` normaliser only recognised the hex form, so a random UUID leaked into every resource snapshot. Fixed centrally in `logfire/_internal/exporters/test.py` by normalising the dashed form to the same fixed value — no snapshot edits needed, and it works on both versions.
2. **Serialized protobuf body size** changed (897045 → 897108 bytes), so `test_max_body_size_bytes` now asserts the message shape via a regex instead of a hardcoded byte count.

Both fixes were verified against `opentelemetry-sdk` **1.42.1 and 1.43.0** (affected tests pass on each).

> Note: as of 2026-07-01 the 1.43.0 line has aged past uv's `exclude-newer = "1 week"` window, so a fresh `uv lock --upgrade` now resolves 1.43.0 / 0.64b0. The committed lock is intentionally left at 1.42.1 (this PR only raises the ceiling); the test-robustness changes above ensure CI won't break when the lock is next regenerated.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2042?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


alexmojaki:
> Today's lockfile still resolves 1.42.1 (1.43.0 is within uv's `exclude-newer = "1 week"` window), so CI is unaffected until it ages in.

This is the key problem

logfire/_internal/exporters/test.py
```diff
@@ -150,7 +150,12 @@ def process_attribute(
         assert value == os.getpid()
         return 1234
     if name == 'service.instance.id':
-        if re.match(r'^[0-9a-f]{32}$', value):
+        # OpenTelemetry <=1.42 set this to `uuid4().hex` (32 hex chars); 1.43+ uses the
```
alexmojaki, line 153: No, otel didn't set this before at all, logfire did. This points to the need for a change in config.py, not tests.

The otel change is here: https://github.com/open-telemetry/opentelemetry-python/pull/5259/changes#diff-d2c46fd8da8afdf67c6ad3dd42b0a243c5049f3c720d9bb3351f1d5522338b06R472

The spec is here: https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-instance-id

Also see the docs: https://pydantic.dev/docs/logfire/reference/sql/#service_instance_id

The otel change is causing a change in the documented logfire behaviour. logfire generates a new random UUID for each call to logfire.configure(), whereas it looks like otel holds onto a stable global value. A decision needs to be made about this.

cc @dmontagu who was recently involved in code changes in this area.

hramezani: We have two options:
* Defer to OTel — drop our fallback, update docs
* Preserve current behaviour — re-assert our own id above OTel's detector while still letting OTEL_RESOURCE_ATTRIBUTES/user detectors win.

anthony2261: Would love to see this merged (or any solution that bumps up the otel version). Is there someone who should take a look at the regex change who could help out?

alexmojaki: I've decided this isn't worth worrying about, pushing ahead

PR #2054, opened 2026-07-06
`logfire.instrument_pydantic()`: Stop instrumenting third-party Pydantic models by default

### LINKED ISSUE #2053: Pydantic plugin: docs say third-party modules are not instrumented by default, but empty include instruments everything
### Description

The `logfire.instrument_pydantic` docstring for the `include` parameter states:

> By default, third party modules are not instrumented. This option allows you to include specific modules.

But in the installed implementation (v4.34.0), `_include_model` in `logfire/integrations/pydantic.py` contains no third-party detection at all. The decision is:

1. a hardcoded ignore list of exactly three modules (`fastapi`, `logfire_backend`, `fastui`),
2. user `exclude` patterns,
3. user `include` patterns — **and if `include` is empty, it returns `True` for every model**.

So `logfire.instrument_pydantic(record="failure")` with no `include` instruments every pydantic model in the process. In our Django app that was **2,221 model classes at startup** (verified by adding a print inside `_include_model` at runtime), the overwhelming majority third-party (openai SDK, opik, mem0, ...).

### Why it matters

With `record="failure"`, this records third-party libraries' *internal, by-design* validation fallbacks as warnings:

- Opik's OpenAI usage parser deliberately tries a Completions-shaped schema first, catches the `ValidationError`, and falls back to a Responses-shaped one — one recorded "Validation on OpenAICompletionsUsage failed" per LLM call.
- The OpenAI SDK's `construct_type` tries strict validation of its stream-event union before falling back to lenient construction — each mismatch records "Validation on union failed" with all ~101 enumerated union-member errors as attributes, per event.

None of these are actionable failures — they're libraries' internal try/except control flow — but they generate warning-level noise and span/attribute volume (quota) on every LLM call.

### Expected

Either the documented default (don't instrument third-party modules unless included) is implemented, or the docstring/docs are corrected to say the default is instrument-everything, so users know they must pass `include` themselves. We worked around it with `include={r"^apps\..*", ...}` — note also that since `_include_model` matches with `re.search(f'{pattern}$', ...)` (end-anchored only), users must remember the leading `^` or patterns like `apps\..*` unexpectedly match third-party modules such as `django.apps.config` mid-path — worth a docs example if the current behavior is kept.

### Python, Logfire & OS Versions

```
logfire 4.34.0
pydantic 2.13.4
python 3.14.0
platform: linux (Docker, Django app)
```

chocoHacks33:
## Summary
- make the Pydantic plugin skip non-user modules by default when `include` is empty
- keep `include=` as an explicit override so third-party modules can still be instrumented intentionally
- add regression coverage for the default skip path and the include override

## Why
Fixes #2053. The current implementation instruments every model outside a tiny hardcoded ignore list when `include` is empty, which contradicts the documented default and can create high-volume noise from third-party validation fallbacks.

## Validation
- `git diff --check`
- direct regression script against the editable checkout covering:
  - default exclusion of a third-party module path
  - default inclusion of a user module path
  - `include=` override for an explicit third-party module
- full `pytest tests/test_pydantic_plugin.py -q` is blocked in this environment before test collection by an existing OpenTelemetry/importlib deprecation warning promoted to error by the repo's pytest warning policy

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2054?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


logfire/integrations/pydantic.py
```diff
@@ -368,6 +372,56 @@ def new_schema_validator(
 _pydantic_plugin_config_value: PydanticPlugin | None = None
 
 
+@lru_cache
+def _non_user_code_prefixes() -> tuple[Path, ...]:
```
alexmojaki: it looks like this logic should be folded into `logfire._internal.stack_info.NON_USER_CODE_PREFIXES`

tests/test_pydantic_plugin.py
```diff
@@ -236,7 +327,7 @@ class MyModel(BaseModel, plugin_settings={'logfire': {'record': 'failure'}}):
                             'exemplars': [],
                         },
                     ],
-                    'aggregation_temporality': AggregationTemporality.DELTA,
```
alexmojaki, line 239: please revert these unrelated changes

tests/test_pydantic_plugin.py
```diff
@@ -1292,9 +1383,7 @@ class Hero(sqlmodel.SQLModel, table=True):
                     'logfire.level_num': 9,
                     'logfire.span_type': 'span',
                     'success': True,
-                    'result': '{"id":1}'
-                    if get_version(pydantic.__version__) >= get_version('2.7.0')
-                    else '"Hero(id=1)"',
+                    'result': '{"id":1}',
```
alexmojaki: why has this changed?

tests/test_pydantic_plugin.py
```diff
@@ -157,6 +159,95 @@ def test_logfire_plugin_include_exclude_models(
...
+        assert _new_pydantic_plugin_result('namespace_pkg.models', 'NamespaceModel') != (None, None, None)
+
+
+def test_pydantic_plugin_excludes_builtin_modules_by_default() -> None:
+    def fake_find_spec(module: str) -> SimpleNamespace | None:
+        if module == 'builtins':
+            return SimpleNamespace(origin='built-in', submodule_search_locations=None)
+        return None
+
+    with patch('logfire.integrations.pydantic.importlib.util.find_spec', side_effect=fake_find_spec):
```
alexmojaki: why is this patching needed?

tests/test_pydantic_plugin.py
```diff
@@ -157,6 +159,95 @@ def test_logfire_plugin_include_exclude_models(
...
+
+    def fake_find_spec(module: str) -> SimpleNamespace | None:
+        if module == 'third_party_pkg.models':
+            return SimpleNamespace(
+                origin=os.path.join(purelib, 'third_party_pkg', 'models.py'),
+                submodule_search_locations=[],
+            )
+        return None
+
+    with patch('logfire.integrations.pydantic.importlib.util.find_spec', side_effect=fake_find_spec):
```
alexmojaki: why is this patching needed?

tests/test_pydantic_plugin.py
```diff
@@ -157,6 +159,95 @@ def test_logfire_plugin_include_exclude_models(
...
+        if module == 'third_party_pkg.models':
+            return SimpleNamespace(
+                origin=os.path.join(purelib, 'third_party_pkg', 'models.py'),
+                submodule_search_locations=[],
+            )
+        if module == 'tests.test_pydantic_plugin':
+            return SimpleNamespace(origin=__file__, submodule_search_locations=[])
+        return None
+
+    with patch('logfire.integrations.pydantic.importlib.util.find_spec', side_effect=fake_find_spec):
```
alexmojaki: why is this patching needed?

tests/test_pydantic_plugin.py
```diff
@@ -157,6 +159,95 @@ def test_logfire_plugin_include_exclude_models(
...
+def _new_pydantic_plugin_result(
+    module: str,
+    name: str,
+    *,
+    include: set[str] | None = None,
+    exclude: set[str] | None = None,
+) -> tuple[object | None, object | None, object | None]:
+    getattr(logfire_pydantic, '_module_is_non_user_code').cache_clear()
+    getattr(logfire_pydantic, '_non_user_code_prefixes').cache_clear()
+    logfire.configure(
```
alexmojaki: why configure logfire?

tests/test_pydantic_plugin.py
```diff
@@ -157,6 +159,95 @@ def test_logfire_plugin_include_exclude_models(
...
+        send_to_logfire=False,
+        metrics=logfire.MetricsOptions(additional_readers=[InMemoryMetricReader()]),
+    )
+    logfire.instrument_pydantic(record='all', include=include or set(), exclude=exclude or set())
+    plugin = LogfirePydanticPlugin()
+    return cast(
+        tuple[object | None, object | None, object | None],
+        plugin.new_schema_validator(
+            core_schema.int_schema(), None, SchemaTypePath(module=module, name=name), 'BaseModel', None, {}
```
alexmojaki: why this complex internal call? this looks very hard to understand, isn't there a better way to test?

tests/test_pydantic_plugin.py
```diff
@@ -2,7 +2,9 @@
```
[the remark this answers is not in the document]
dmontagu: Not applicable: this test suite normalizes `code.lineno` to a fixed `123` in every exported-span snapshot (`tests/utils.py` passes `fixed_line_number=123` into `exported_spans_as_dict`), so the inline snapshots never contain physical line numbers and are unaffected by tests shifting within the file. The full suite passes on the current head.

dmontagu:
Pushed 9c9757a8 to tighten the diff: reverted unrelated inline-snapshot churn in existing tests (the `AggregationTemporality.DELTA` -> `1` rewrites and the dropped pydantic-version conditional in `test_sqlmodel_pydantic_plugin`, both presumably from an `--inline-snapshot=fix` run). The test diff is now purely the new regression tests. Full file passes locally (52 passed, 2 xfailed).

alexmojaki (commented):
just found that i reviewed this ages ago and forgot to submit the review, sorry

logfire/_internal/stack_info.py
```diff
@@ -108,23 +108,30 @@ def get_user_frame_and_stacklevel() -> tuple[FrameType | None, int]:
     return None, 0
 
 
+def is_non_user_path(path: str | Path) -> bool:
+    """Check if the file path is in a location that doesn't contain user code.
+
+    Specifically the standard library, site-packages (wherever opentelemetry is installed),
+    the logfire package, and anything registered with `add_non_user_code_prefix`.
```
alexmojaki, line 119: add_non_user_code_prefix needs to document this

PR #2071, opened 2026-07-13
chore: Update dependencies

alexmojaki:
Routine dependency update.

Three deps are held back rather than bundling their breaking changes into a routine bump:

- **pydantic-ai-slim / pydantic-evals** pinned `<2.0`. The 2.x major bump removes
  instrumentation format version=1, deprecates versions 2-4 (warning becomes an error
  in tests), restructures the MCP API (`MCPServer`→`MCPToolset`, needs `fastmcp`), and
  makes `Dataset.name` required. The public `logfire[datasets]` extra is capped too so
  users don't pull an unvalidated 2.x. Tracked for a dedicated migration.
- **openai-agents** pinned `<0.17.5` (main's version). 0.17.5+ changes the scheduling
  of concurrent input guardrails vs the first model turn, making `test_input_guardrails`
  flaky: a tripped guardrail's cancelled model call intermittently emits an extra span,
  which shifts the test exporter's sequential span ids and can't be normalised away.
  Unpin once the test is made deterministic.

Also re-added `types-requests` explicitly: `requests` 2.34 now ships inline `py.typed`
annotations that regress some types (`Session`/`Response.headers`), and `openai-agents`
no longer pulls `types-requests` in transitively.


pyproject.toml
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 53e70a29 — capped the public `datasets` extra at `pydantic-evals>=1.0.0,<2` to match the dev pin, so `logfire[datasets]` users won't pull the unvalidated 2.x release either. Both caps will be lifted together in the 2.x migration.

pyproject.toml
```diff
@@ -117,6 +119,9 @@ dev = [
     "pytest-pretty >= 1.2.0",
     "pydantic >= 2.11.0",
     "requests >= 2.31.0",
+    # requests >= 2.34 ships inline py.typed stubs that regress some annotations
+    # (e.g. Session/Response.headers); keep using the more accurate types-requests stubs.
```
alexmojaki: explain the 'more accurate' claim

alexmojaki: (AI) Fair — 'more accurate' was too vague and not even uniformly true (the two stub sets disagree in both directions: types-requests types `Response.headers` as `CaseInsensitiveDict[str]` while requests' inline types say `CaseInsensitiveDict[str | bytes]`).

The concrete reason we want types-requests is `Session.headers`: requests' inline types type it as `CaseInsensitiveDict[str]`, which rejects the `str | bytes` header values requests actually accepts at runtime and that `DiskRetryer` passes to a `Session` (`exporters/otlp.py`). types-requests types it as `MutableMapping[str, str | bytes]`, matching how the code uses it. Rewrote the comment to say exactly that in 7d537375.

PR #2074, opened 2026-07-13
Upgrade openai-agents to 0.17.7

alexmojaki:
## Summary

- make the input guardrail test independent of the Agents SDK's cancellation scheduling
- retain the exact snapshot for the successful guardrail path
- remove the temporary `openai-agents` upper bound and upgrade to 0.17.7

## Testing

- `uv run pytest tests/otel_integrations/test_openai_agents.py -q`
- `uv run ruff format tests/otel_integrations/test_openai_agents.py --check`
- `uv run ruff check tests/otel_integrations/test_openai_agents.py`
- `uv run pyright tests/otel_integrations/test_openai_agents.py`


tests/otel_integrations/test_openai_agents.py
```diff
@@ -1069,8 +1080,8 @@ async def zero_guardrail(_context: Any, _agent: Agent[Any], inp: Any) -> Guardra
                     'logfire.span_type': 'span',
                     'type': 'custom',
                     'name': 'Agent workflow',
-                    'sdk_span_type': 'task',
                     'gen_ai.system': 'openai',
+                    'sdk_span_type': 'task',
```
alexmojaki, line 1083: clean up the diff

alexmojaki: (AI) Addressed in 3aa3406b and 81cfe3da: the test keeps the full snapshot, filters only the scheduling-dependent incomplete response span before simplification, and uses IsInt() only for the four affected end times.

tests/otel_integrations/test_openai_agents.py
```diff
@@ -997,27 +994,41 @@ async def zero_guardrail(_context: Any, _agent: Agent[Any], inp: Any) -> Guardra
...
+    )
+
+    exporter.exported_spans.clear()
+    with pytest.raises(InputGuardrailTripwireTriggered):
+        await Runner.run(agent, '0?')
+
+    spans = exporter.exported_spans_as_dict(parse_json_attributes=True)
+    # The model call runs concurrently with input guardrails. Depending on when it is
+    # cancelled, the Agents SDK may or may not emit an incomplete Responses API span.
+    # Assert the guardrail telemetry independently of that span and its effect on IDs.
```
alexmojaki: are IDs affected or not?

tests/otel_integrations/test_openai_agents.py
alexmojaki: (AI) Addressed in 96280b41: restored the original attribute-key order so the PR no longer shows those lines as moves.

tests/otel_integrations/test_openai_agents.py
alexmojaki: (AI) They are not affected in the reproduced race: the optional response span starts after the enclosing spans have been created, so only their generated end times shift. Corrected the comment in bc26a97c.

PR #2077, opened 2026-07-14
Route managed variables by API key region

### LINKED ISSUE #2056: Managed-variables provider ignores API key region and defaults base URL to US prod
### Description

When `logfire.configure()` is given a **region-scoped API key** (e.g. `pylf_v2_stagingeu_…`) for the managed-variables API, the `LogfireRemoteVariableProvider` does **not** derive its base URL from the key's region — it defaults to US production (`https://logfire-us.pydantic.dev`). The span-export path, by contrast, *does* resolve the backend from the token's region. This inconsistency means a staging-EU (or any non-US-prod) key silently queries the wrong backend, where the key isn't valid, so every variable resolves to its code default (`reason='missing_config'`) with no error surfaced.

You must pass `advanced=AdvancedOptions(base_url=...)` explicitly to point the variables provider at the right backend, even though the key already encodes the region.

### Reproduction

```python
import logfire
from logfire.variables import Variable

# api_key is a stagingeu key: pylf_v2_stagingeu_...
logfire.configure(api_key="pylf_v2_stagingeu_...", variables=logfire.VariablesOptions(), send_to_logfire=False)

prov = logfire.DEFAULT_LOGFIRE_INSTANCE.config.get_variable_provider()
print(prov._base_url)  # -> https://logfire-us.pydantic.dev  (WRONG: key is stagingeu)

v = Variable("prompt__x", type=str, default="d")
print(v.get()._reason)  # -> 'missing_config'  (fetch against wrong backend)
```

Passing `advanced=logfire.AdvancedOptions(base_url="https://logfire-eu.pydantic.info")` fixes it:
`prov._base_url` becomes the EU URL and `v.get()` resolves correctly (`'unrecognized_variable'` for a missing var, real value for an existing one).

### Expected

The variables provider should derive its base URL from the API key's region (matching how the span/token path resolves the region), so a region-scoped key "just works" without an explicit `base_url`.

### Environment

Observed while wiring pydantic-ai-harness managed capabilities against a staging-EU preview.

alexmojaki:
## Summary

- infer the Logfire backend from the stable API-key prefix instead of requiring the entire key to match the current token format
- route eager and lazy managed-variable providers through the same `AdvancedOptions.generate_base_url()` path used by telemetry export
- cover v2 staging-EU API keys at both the resolver and remote-provider configuration layers

The reported harness reproduction was locked to Logfire 4.33.0. Its token parser did not yet accept the organization UUID in v2 API keys, so backend inference silently fell back to US. Although the current parser recognizes that key shape, routing should not depend on validation of the evolving key suffix; this keeps region selection stable across token-format changes.

Closes #2056

## Tests

- `uv run pytest tests/test_configure.py -q` (103 passed)
- `uv run pytest tests/test_variables.py -q` (299 passed)
- pre-commit: Ruff, Ruff Format, Pyright


logfire/_internal/config.py
```diff
@@ -137,6 +137,7 @@
 COMMON_REQUEST_HEADERS = {'User-Agent': f'logfire/{VERSION}'}
 """Common request headers for requests to the Logfire API."""
 PROJECT_NAME_PATTERN = r'^[a-z0-9]+(?:-[a-z0-9]+)*$'
+LOGFIRE_TOKEN_REGION_PATTERN = re.compile(r'^pylf_v[0-9]+_(?P<region>[a-z]+)_')
```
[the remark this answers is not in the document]
alexmojaki: (AI) Thanks, no action needed. This is a pre-existing display concern outside the routing fix; production v2 keys currently use lowercase organization UUIDs and alphanumeric secret suffixes, while this regression intentionally uses a suffix shape the strict display parser rejects to ensure backend routing depends only on the stable prefix.

logfire/_internal/config.py
```diff
@@ -1665,9 +1664,8 @@ def _lazy_init_variable_provider(self) -> VariableProvider:
             from logfire.variables.remote import LogfireRemoteVariableProvider
 
             options = VariablesOptions()
-            base_url = self.advanced.base_url or get_base_url_from_token(api_key)
             provider = LogfireRemoteVariableProvider(
-                base_url=base_url,
+                base_url=self.advanced.generate_base_url(api_key),
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 3972b1b6

tests/test_variables.py
```diff
@@ -1596,6 +1597,37 @@ def test_refresh_skips_when_recently_fetched(self) -> None:
 
 @pytest.mark.filterwarnings('ignore::pytest.PytestUnhandledThreadExceptionWarning')
 class TestApiKeySupport:
+    def test_api_key_region_configures_provider_base_url(
+        self, config_kwargs: dict[str, Any], monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        monkeypatch.delenv('LOGFIRE_BASE_URL', raising=False)
```
alexmojaki: do this in conftest like with LOGFIRE_TOKEN etc

tests/test_variables.py
```diff
@@ -1596,6 +1597,37 @@ def test_refresh_skips_when_recently_fetched(self) -> None:
...
+                    'test_var': {
+                        'name': 'test_var',
+                        'labels': {'default': {'version': 1, 'serialized_value': '"region_value"'}},
+                        'rollout': {'labels': {'default': 1.0}},
+                        'overrides': [],
+                    }
+                }
+            },
+        )
+        request_mocker.get('https://logfire-eu.pydantic.info/v1/variable-updates/', status_code=404)
```
alexmojaki: this looks like excessive effort to avoid getting the protected var `_base_url`

tests/test_variables.py
alexmojaki: (AI) Addressed in ad546035

tests/test_variables.py
alexmojaki: (AI) Addressed in 7fc5e3d8

PR #2084, opened 2026-07-16
fix(db_statement_summary): bound INSERT summary length like the SELECT path

chuenchen309:
## Summary

`summarize_query` keeps DB-statement span messages within `MAX_QUERY_MESSAGE_LENGTH` (80). The SELECT path enforces this by ending in `truncate(summary, MAX_QUERY_MESSAGE_LENGTH)`, and the fallback path is bounded too — but the `INSERT` branch returned its f-string directly:

```python
return f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}'
```

`columns`/`values` are truncated, but `table` is not, so a long (e.g. schema-qualified) table name pushes the summary past the limit.

## Reproduction

```python
>>> from logfire._internal.db_statement_summary import summarize_query
>>> q = 'INSERT INTO "analytics"."very_long_events_table_name_for_testing" (a, b, c, d) VALUES (1, 2, 3, 4)'
>>> len(summarize_query(q))
112   # > MAX_QUERY_MESSAGE_LENGTH (80)
```

The equivalent `SELECT` with the same long table stays bounded (~74 chars).

## Fix

Wrap the INSERT result in the same final `truncate(..., MAX_QUERY_MESSAGE_LENGTH)` that the SELECT path already applies, so every branch stays bounded.

## Tests

Added `test_insert_long_table`; the existing `test_insert` (short table) is unchanged. Full `tests/test_db_statement_summary.py` passes (24), and `ruff format/check` + `pyright` are clean on the touched files.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2084?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/logfire/_internal/db_statement_summary.py
+++ b/logfire/_internal/db_statement_summary.py
@@ -79,7 +79,10 @@ def summarize_query(db_statement: str) -> str | None:
         return select(expr, table, match_end=select_match.end(), db_statement=db_statement, ctes=ctes)
     elif insert_match := INSERT_RE.match(db_statement):
         table, columns, values = insert_match.groups()
-        return f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}'
+        return truncate(
+            f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}',
+            MAX_QUERY_MESSAGE_LENGTH,
+        )
     else:
         return fallback(db_statement)
 
--- a/logfire/_internal/db_statement_summary.py
+++ b/logfire/_internal/db_statement_summary.py
@@ -79,10 +79,10 @@ def summarize_query(db_statement: str) -> str | None:
         return select(expr, table, match_end=select_match.end(), db_statement=db_statement, ctes=ctes)
     elif insert_match := INSERT_RE.match(db_statement):
         table, columns, values = insert_match.groups()
-        return truncate(
-            f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}',
-            MAX_QUERY_MESSAGE_LENGTH,
-        )
+        suffix = f' {truncate(columns, 25)} VALUES {truncate(values, 25)}'
+        # the table gets whatever length is left over because this is the best identifier of the query
+        table = truncate(table, MAX_QUERY_MESSAGE_LENGTH - len('INSERT INTO ') - len(suffix))
+        return f'INSERT INTO {table}{suffix}'
     else:
         return fallback(db_statement)
 
--- a/tests/test_db_statement_summary.py
+++ b/tests/test_db_statement_summary.py
@@ -153,3 +153,16 @@ def test_insert():
         message_from_db_statement(attrs(q), None, 'INSERT')
         == 'INSERT INTO table (apple, bana…n, egg, fig) VALUES (1, 2, 3, 4, 5, 6)'
     )
+
+
+def test_insert_long_table():
+    # A long (e.g. schema-qualified) table name is not truncated on its own, so without a
+    # final length bound the INSERT summary blew past MAX_QUERY_MESSAGE_LENGTH (was 112 chars
+    # here) — unlike the SELECT path, which always ends with truncate(summary, MAX). The
+    # summary must stay bounded like every other branch.
+    q = 'INSERT INTO "analytics"."very_long_events_table_name_for_testing" (apple, banana, carrot, durian, egg, fig) VALUES (1, 2, 3, 4, 5, 6)'
+    # insert_assert(message_from_db_statement(attrs(q), None, 'INSERT'))
+    assert (
+        message_from_db_statement(attrs(q), None, 'INSERT')
+        == 'INSERT INTO "analytics"."very_long_event…a…n, egg, fig) VALUES (1, 2, 3, 4, 5, 6)'
+    )
--- a/tests/test_db_statement_summary.py
+++ b/tests/test_db_statement_summary.py
@@ -1,4 +1,4 @@
-from logfire._internal.db_statement_summary import message_from_db_statement
+from logfire._internal.db_statement_summary import MAX_QUERY_MESSAGE_LENGTH, message_from_db_statement
 
 
 def test_no_db_statement():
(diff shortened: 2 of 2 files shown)


chuenchen309:
Good catch that `truncate(summary, MAX_QUERY_MESSAGE_LENGTH)` can return `MAX + 1` chars — but that's a property of the shared `truncate()` helper, not something this PR introduces: for an even `length`, `truncate` returns `2 * (length // 2) + 1` (so `80 → 81`). The main/SELECT path already bounds its summary the exact same way (`return truncate(summary, MAX_QUERY_MESSAGE_LENGTH)`), so wrapping the INSERT branch in the same helper keeps the two paths consistent — which is the point of this fix (the INSERT summary was previously unbounded at ~112 chars).

Making the INSERT result strictly `<= MAX` while the SELECT path stays at `MAX + 1` would introduce an inconsistency. The clean way to get a hard `<= MAX` guarantee is to fix `truncate` itself (e.g. `half_length = (length - 1) // 2`), but that changes every even-length caller (`10`, `80`) and their existing snapshots, which is out of scope for "bound the INSERT summary." Happy to do that as a follow-up if you'd prefer the strict bound.

logfire/_internal/db_statement_summary.py
```diff
@@ -79,7 +79,10 @@ def summarize_query(db_statement: str) -> str | None:
         return select(expr, table, match_end=select_match.end(), db_statement=db_statement, ctes=ctes)
     elif insert_match := INSERT_RE.match(db_statement):
         table, columns, values = insert_match.groups()
-        return f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}'
+        return truncate(
+            f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}',
+            MAX_QUERY_MESSAGE_LENGTH,
+        )
```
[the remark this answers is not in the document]
chuenchen309: Added an explicit `len(result) <= MAX_QUERY_MESSAGE_LENGTH + 1` assertion to the test to make the bound self-documenting.

On the `+ 1`: the 81-vs-80 is a pre-existing property of `truncate()` itself — it returns `2 * (length // 2) + 1` characters for an even `length`, so `truncate(s, 80)` yields up to 81. That behavior is **shared by the SELECT path**, which already ends in `truncate(summary, MAX_QUERY_MESSAGE_LENGTH)`. The goal of this PR is only to make the INSERT branch honor the same bound the SELECT branch already does; changing `truncate()` to reserve a slot for the ellipsis would alter the output of every existing SELECT summary and break the current exact-match tests (`test_long_select`, `test_one_join_long`, etc.), so it is out of scope here. Happy to tighten `truncate()` in a separate PR if that is wanted.

(The Ruff S608 "SQL injection" hint is a false positive — this f-string builds a truncated display summary for a span message, it is never executed as SQL.)

chuenchen309:
Putting a number on the "out of scope" claim I made above, since it was an assertion and this is checkable:

Applying the suggested `truncate` change (reserving one character for `…`, i.e. `half_length = (length - 1) // 2`) **fails 9 of the 24 tests** in `tests/test_db_statement_summary.py` — `test_long_select`, `test_one_join_long` and friends, all of which pin exact 81-character SELECT output that predates this PR.

That's the reason this PR leaves `truncate` alone: the 81-vs-80 is a property of `truncate` itself (`f'{s[:length // 2]}…{s[-(length // 2):]}'` → `2 * (length // 2) + 1`), shared by the SELECT path that already calls it. This PR only makes the INSERT branch honour the same bound the SELECT branch already honours; tightening `truncate` would change every existing SELECT summary and is a separate, deliberate decision.

Happy to send that as its own PR if the team wants the bound to be exact — it'd need the expected literals in those 9 tests updated too.

tests/test_db_statement_summary.py
```diff
@@ -153,3 +153,13 @@ def test_insert():
         message_from_db_statement(attrs(q), None, 'INSERT')
         == 'INSERT INTO table (apple, bana…n, egg, fig) VALUES (1, 2, 3, 4, 5, 6)'
     )
+
+
+def test_insert_long_table():
+    # The table name isn't truncated on its own, so this was 112 chars before the bound.
+    q = 'INSERT INTO "analytics"."very_long_events_table_name_for_testing" (apple, banana, carrot, durian, egg, fig) VALUES (1, 2, 3, 4, 5, 6)'
+    # insert_assert(message_from_db_statement(attrs(q), None, 'INSERT'))
```
alexmojaki: rather use inline_snapshot

logfire/_internal/db_statement_summary.py
```diff
@@ -79,7 +79,10 @@ def summarize_query(db_statement: str) -> str | None:
         return select(expr, table, match_end=select_match.end(), db_statement=db_statement, ctes=ctes)
     elif insert_match := INSERT_RE.match(db_statement):
         table, columns, values = insert_match.groups()
-        return f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}'
+        return truncate(
+            f'INSERT INTO {table} {truncate(columns, 25)} VALUES {truncate(values, 25)}',
```
alexmojaki: rather truncate `table`

dmontagu:
Hi @chuenchen309 — thanks for catching this and for the clear write-up and repro. I hope you don't mind: I pushed a commit directly to your branch applying @alexmojaki's two review suggestions, so this can move forward without more round-trips.

What changed:

- **Truncate `table` rather than the whole message.** Wrapping the composed f-string meant the summary got a second ellipsis dropped into the middle of the column list (`INSERT INTO "analytics"."very_long_event…a…n, egg, fig) VALUES …`). Now `table` is truncated as its own component, the way `columns` and `values` already are. It gets whatever length is left over from `MAX_QUERY_MESSAGE_LENGTH` after the columns and values, since the table is the best identifier of the query — so it is only shortened when the message would actually overflow, and short table names are never touched:

  ```
  INSERT INTO "analyti…testing" (apple, bana…n, egg, fig) VALUES (1, 2, 3, 4, 5, 6)
  INSERT INTO "analytics"."ve…me_for_testing" (apple) VALUES (12345678901234567890)
  ```

- **`inline_snapshot` for the new tests**, replacing the hand-written expected strings.

Your reasoning on the CodeRabbit `truncate()` off-by-one was right, and I left the shared helper alone — the INSERT path now has exactly the same bound characteristics as the SELECT path, which was the goal of the PR. The explicit length assertion went away with the snapshot rewrite, since any regression in the bound would change the snapshotted string anyway.

`test_insert` and every other existing expectation are unchanged; 25 passed locally, plus the sqlalchemy/mysql/sqlite3 integration tests, and ruff + pyright are clean.

PR #2106, opened 2026-07-22
Send spec-compliant User-Agent: `logfire-python/<version>`, retaining the OTLP exporter identifier on exports

dmontagu:
## Summary

Aligns the SDK's `User-Agent` handling with the OpenTelemetry exporter specification and with the other Logfire SDKs, and unifies it across every HTTP surface:

| Surface | Before | After |
|---|---|---|
| OTLP exports (traces/metrics/logs) | `logfire/<v>` (upstream exporter identifier dropped) | `logfire-python/<v> OTel-OTLP-Exporter-Python/<exporter-v>` |
| API client, token validation, variables | `logfire/<v>` | `logfire-python/<v>` |
| Experimental query client | `logfire-sdk-python/<v> (Python ..., os ..., arch ...)` | `logfire-python/<v>` |
| CLI session | requests default (`python-requests/<v>`) | `logfire-python/<v>` |

## Why

The OTel exporter spec's [User Agent section](https://opentelemetry.io/docs/specs/otel/protocol/exporter/#user-agent) recommends that when a distribution adds its identifier to the `User-Agent`, the exporter's default identifier is retained, typically by prepending:

> "The resulting User-Agent SHOULD include the exporter's default User-Agent string. The intention is to support an identifier for an OpenTelemetry SDK/Agent distribution. Typically an exporter would *prepend* the given identifier to its own. For example: `MyDistribution/x.y.z OTel-OTLP-Exporter-Python/1.2.3`"

That guidance was added in [open-telemetry/opentelemetry-specification#4560](https://github.com/open-telemetry/opentelemetry-specification/pull/4560) (Sept 2025, motivated by [#4479](https://github.com/open-telemetry/opentelemetry-specification/issues/4479)). Our current behavior predates it: the `logfire/<version>` header dates to #250 (Sept 2023) and, once reused for the OTLP exporters, silently replaced the upstream `OTel-OTLP-Exporter-Python/<v>` identifier because the Python exporter applies custom headers over its defaults. Since the Python exporter does not (yet) implement the spec's `user-agent` enrichment option, we rebuild its default identifier from the installed package version and prepend ours; a test compares against a freshly constructed exporter's actual default so we notice if the upstream format ever changes.

Retaining the exporter identifier also has direct debugging value: it tells the backend which upstream exporter version produced a request without having to ask the user.

### Choices and their grounding

- **Product token `logfire-python/<v>`**: matches the sibling SDKs (`logfire-js/<v>`, `logfire-rust/<v>`) and the spec's `MyDistribution/x.y.z` product-token form (RFC 9110).
- **No parenthetical platform details**: the Python version, OS, and architecture added to the query client's UA in #1875 are removed, as proposed by their author in review of #1905. On OTLP exports this information is already carried properly by resource attributes (`process.runtime.*`, `os.*`, `host.*`, see #2013), and a single low-cardinality product token keeps the UA usable for filtering and aggregating by SDK version.
- **One constant for every surface**: previously three different formats existed (API/OTLP, query client, CLI). All requests now share `UA_HEADER`, with OTLP exports additionally retaining the upstream identifier via `OTLP_EXPORTER_UA_HEADER`.

This complements (and does not replace) the request-telemetry-header discussion in #1905; richer runtime/config signals belong there, not in the UA.

## Compatibility notes

- The UA is informational; no behavior depends on it. Anything matching the exact string `logfire/<version>` (e.g. proxy rules) should match the `logfire-python/` prefix instead. Older SDK versions keep sending the old formats, so consumers should expect both during the transition.
- User-supplied `user-agent` headers on the query client are still respected (`setdefault`).

## Test plan

- New `tests/test_user_agent.py`: product-token format; OTLP UA exactly equals `<our token> <upstream exporter default>` (drift guard against the installed exporter's real default); end-to-end `logfire.configure()` asserting the UA per endpoint (`/v1/info` vs `/v1/traces`) via `requests_mock`; API client, query client, and CLI request headers.
- `make format lint typecheck` clean; `tests/test_configure.py`, `tests/test_cli.py`, `tests/test_query_client.py`, `tests/exporters/test_otlp_session.py`, `tests/test_variables.py` all pass (625 tests).

https://claude.ai/code/session_0159PYpdaBhM35pcPFiVcotC

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


logfire/experimental/query_client.py
```diff
@@ -149,7 +147,7 @@ def __init__(self, base_url: str, read_token: str, timeout: Timeout, client: typ
         self.timeout = timeout
         headers = client_kwargs.pop('headers', {})
         headers['authorization'] = read_token
-        headers.setdefault('user-agent', _USER_AGENT)
+        headers.setdefault('user-agent', UA_HEADER)
```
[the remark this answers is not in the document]
dmontagu: Addressed in 5d4236f. The headers are now normalized through `httpx.Headers` (case-insensitive) before `setdefault`, so a caller-supplied `User-Agent` in any casing is respected and exactly one entry is sent; this also makes the `authorization` assignment replace rather than duplicate a caller-cased `Authorization`. Added `test_query_client_user_agent_caller_override` asserting `get_list("user-agent") == ["my-agent/1.0"]` for a caller-supplied `User-Agent`, and tightened the existing test to assert a single-entry list.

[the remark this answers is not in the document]
dmontagu: Addressed in 5d4236f: headers are normalized through case-insensitive `httpx.Headers` before the `setdefault`, so a differently-cased caller `User-Agent` is detected and no duplicate entry is sent. Regression test added (`test_query_client_user_agent_caller_override`).

logfire/_internal/config.py
```diff
@@ -1309,7 +1310,7 @@ def check_tokens():
                     for token in token_list:
                         base_url = self.advanced.generate_base_url(token)
                         otlp_forwarding_destinations.append((base_url, token))
-                        headers = {'User-Agent': f'logfire/{VERSION}', 'Authorization': token}
+                        headers = {'User-Agent': OTLP_EXPORTER_UA_HEADER, 'Authorization': token}
```
alexmojaki, line 1313: consider also _forwarding_user_agent in logfire/_internal/forwarding.py

logfire/experimental/query_client.py
```diff
@@ -138,7 +137,6 @@ def _map_v2_result(obj: dict[str, Any]) -> RowQueryResults:
 
 
 _ACCEPT = Literal['application/json', 'application/vnd.apache.arrow.stream', 'text/csv']
-_USER_AGENT = f'logfire-sdk-python/{VERSION} (Python {platform.python_version()}, os {platform.platform()}, arch {platform.machine()})'
```
alexmojaki, line 141: It's not super clear what the harm of keeping the other info in the UA is. I guess it makes parsing and querying harder? But if we're going to remove info then it seems important to merge some version of https://github.com/pydantic/logfire/pull/1905 first so that we don't lose it entirely.

logfire/experimental/query_client.py
```diff
@@ -147,9 +145,11 @@ def __init__(self, base_url: str, read_token: str, timeout: Timeout, client: typ
         self.base_url = base_url
         self.read_token = read_token
         self.timeout = timeout
-        headers = client_kwargs.pop('headers', {})
+        # `Headers` is case-insensitive, so a caller-supplied `User-Agent` (any casing) is
+        # respected rather than a second `user-agent` entry being added alongside it.
+        headers = Headers(client_kwargs.pop('headers', None))
         headers['authorization'] = read_token
-        headers.setdefault('user-agent', _USER_AGENT)
+        headers.setdefault('user-agent', UA_HEADER)
```
alexmojaki, line 152: why do we even support custom headers? and should our UA be appended to a custom UA?

logfire/_internal/exporters/otlp.py
```diff
@@ -24,9 +25,20 @@
...
 
+# The upstream OTLP exporters emit `OTel-OTLP-Exporter-Python/<version>` by default, but drop it
+# entirely when custom headers include a `User-Agent`. The OTLP exporter spec recommends that a
+# distribution's identifier be *prepended* while the exporter's default identifier is retained
+# (https://opentelemetry.io/docs/specs/otel/protocol/exporter/#user-agent), so rebuild that
+# default here and prepend ours. `tests/test_user_agent.py` guards against the upstream format
+# drifting.
+OTLP_EXPORTER_UA_HEADER = (
+    f'{UA_HEADER} OTel-OTLP-Exporter-Python/{package_version("opentelemetry-exporter-otlp-proto-http")}'
+)
```
Viicos: `importlib.metadata` is known to be slow to import/use, and on AWS lambda deployments it is sometimes not usable:

```suggestion
from opentelemetry.exporter.otlp.proto.http.version import __version__

OTLP_EXPORTER_UA_HEADER = (
    f'{UA_HEADER} OTel-OTLP-Exporter-Python/{__version__}'
)
```

logfire/experimental/query_client.py
Viicos: I'd say it's fine to leave the telemetry in this header, and we rename `logfire-sdk-python/...` to `logfire-python/...`. I initially had concerns with parsing/querying because I have a dashboard where I keep track of SDK queries by UA, but if we properly follow the RFC 9110 grammar and have the correct UDF to do to the parsing then it should be fine.

cc @jirikuncar

logfire/experimental/query_client.py
Viicos: Unfortunately removing the option to set it would be breaking in some way, but I agree it's questionable. I'd also add the httpx one as a component after ours

logfire/_internal/exporters/otlp.py
alexmojaki: well that's a problem for `logfire/version.py`

Viicos: Might be fine to leave like this and circle back to it if we get issues

logfire/experimental/query_client.py
dmontagu: I just want to follow the otel spec here and not rely on this too much for extra info. I feel like we don't depend on the python/os/arch stuff _so much_ that we can't afford a gap in awareness of this.

logfire/_internal/exporters/otlp.py
dmontagu: Done in 66bbefe6, with a tweak: the version now comes from `opentelemetry.exporter.otlp.proto.http.version` (aliased import), which is the exact module the exporter uses to build its own default UA, so there's no drift risk. Beyond import speed, the old code called `importlib.metadata.version(...)` at import time, so a bundle with stripped dist-info would raise `PackageNotFoundError` on `import logfire`. That failure mode still exists in `logfire/version.py` as Alex notes, but no reason to add a second instance.

logfire/_internal/config.py
dmontagu: I looked at it and left it alone deliberately: `_forwarding_user_agent` already follows the same convention this PR adopts, prepending `logfire-proxy/<version>` while retaining the incoming client's UA verbatim (the spec's prepend-and-retain pattern). The one thing it doesn't convey is which SDK is doing the proxying, since the token is `logfire-proxy` rather than e.g. `logfire-python-proxy`, but the retained inner UA already identifies the originating SDK. Happy to rename the token in a follow-up if you think language attribution on the proxy is worth it.

logfire/experimental/query_client.py
dmontagu: Agreed with Victorien that removing custom-header support would be breaking, so I've kept it, including the existing semantics where a caller-supplied UA fully replaces ours (it doubles as an opt-out). On appending `python-httpx/<v>` after ours: I'd rather not. The spec guidance we're following only asks that the *exporter's* identifier be retained; httpx is just the transport, and consistency would then also suggest `python-requests/<v>` on the API client and CLI, re-growing the UA right after this PR trimmed it to one low-cardinality token per surface. If we want richer client-runtime detail later, I think #1905's telemetry header is the right home for it.

alexmojaki: The spec is irrelevant here, this isn't the exporter.

httpx/requests does naturally seem like it fits in the UA. I think it fits there better than in the telemetry header. And it's not currently there anyway.

logfire/experimental/query_client.py
alexmojaki: Are we currently using the info in this UA? Is there pressure to change this UA soon?

logfire/experimental/query_client.py
Viicos: And this is a common practice also described as an example in the RFC: https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.5-6.

alexmojaki (changes requested):

PR #2109, opened 2026-07-24
Add capfire.reconfigure()

### LINKED ISSUE #1689: Add ability to reconfigure a Logfire object
### Description

It would be useful to be able to reconfigure a `Logfire` object.

Right now, any call to `.configure()` needs to be aware of the full configuration for the logging and can't just update/reconfigure certain aspects. This can be problematic when needing to just update one aspect of the logger, like when testing.

For example, if I want to add a test exporter for tests

```python
# app.py
import logfire

logfire.configure(
    scrubbing=logfire.ScrubbingOptions()
)

def main():
    logfire.info(...)
```

```python
# test.py
import logfire
from logfire.testing import SimpleSpanProcessor, TestExporter

from app import main

def test_scrubbing():
    exporter = TestExporter()

    logfire.configure(
        scrubbing=logfire.ScrubbingOptions(),
        additional_span_processors=[SimpleSpanProcessor(exporter)],
    )

    main()
```

Here, Logire was already configured in `app.py`. I want to test my scrubber but I need to reconfigure it in the test. `logfire.reconfigure` would allow me to just pass `additional_span_processors`.

---

There is option of using [`capfire`](https://logfire.pydantic.dev/docs/reference/advanced/testing/#capfire-fixture), but that has the inverse problem.

```python
# test.py
import logfire
from logfire.testing import CaptureLogfire

from app import main

def test_scrubbing(capfire: CaptureLogfire):
    exporter = TestExporter()

    logfire.configure(
        scrubbing=logfire.ScrubbingOptions(),
        additional_span_processors=[SimpleSpanProcessor(exporter)],
    )

    main()
```

This time, the `capfire` fixture has made a call to `logfire.configure()`, overwriting the configuration from `app.py`. In order to add the scrubbing back in, I also need to configure my own test exporter. `logfire.reconfigure` would allow me to just pass `scrubbing`.

frangelbarrera:
Adds CaptureLogfire.reconfigure(**kwargs) so tests can call logfire.configure mid-test without losing capfire's exporter wiring. additional_span_processors, advanced.log_record_processors, and metrics.additional_readers extend capfire's defaults; other kwargs replace them. exporter and log_exporter are reused; id_generator and ns_timestamp_generator are reused as the base for the merge, and when overridden via advanced= the corresponding attribute on capfire is updated to match. metrics_reader is always replaced with a fresh InMemoryMetricReader (OTel forbids reusing one across MeterProvider instances). metrics=False disables metrics.

Closes #1689

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -122,6 +122,7 @@ from .utils import (
     handle_internal_errors,
     platform_is_emscripten,
     suppress_instrumentation,
+    track_explicit_fields,
 )
 
 if TYPE_CHECKING:
@@ -178,6 +179,9 @@ class ConsoleOptions:
     """The output stream to write console output to (default: stdout)."""
 
 
+# `track_explicit_fields` lets `logfire.testing.CaptureLogfire.reconfigure` merge a user-supplied
+# instance with its own defaults without silently dropping fields explicitly set to a default value.
+@track_explicit_fields
 @dataclass
 class AdvancedOptions:
     """Options primarily used for testing by Logfire developers."""
@@ -300,6 +304,8 @@ class PydanticPlugin:
     """Exclude specific modules from instrumentation."""
 
 
+# See the comment on `AdvancedOptions`.
+@track_explicit_fields
 @dataclass
 class MetricsOptions:
     """Configuration of metrics."""
--- a/logfire/_internal/utils.py
+++ b/logfire/_internal/utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import dataclasses
 import functools
 import hashlib
 import inspect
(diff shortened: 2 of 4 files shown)


dmontagu:
Thanks for this @frangelbarrera — the design matches what was agreed in #1689 and the test coverage is good. I pushed one commit to your branch (c1f3a1d8) addressing the cubic review finding; happy to revert if you'd rather take it a different way.

**The bug.** `_merge_dataclass` decided whether a field of `advanced=` / `metrics=` overrode the capfire default by comparing its value against the field's dataclass default. That conflates "the caller didn't set this field" with "the caller passed a value that happens to equal the default", so the latter was silently dropped. Confirmed with a test before fixing:

```python
capfire.reconfigure(advanced=logfire.AdvancedOptions(ns_timestamp_generator=time.time_ns))
logfire.info('hi')
[span] = capfire.exporter.exported_spans
assert span.start_time > 1_700_000_000_000_000_000  # AssertionError: 1000000000
```

The same applied to `id_generator=SeededRandomIdGenerator(None)` — capfire's `IncrementalIdGenerator` was kept and the trace ID was still `1`.

**The fix.** The information simply isn't recoverable from a constructed dataclass, so `AdvancedOptions` and `MetricsOptions` now record which fields were actually passed to `__init__` (`track_explicit_fields` in `_internal/utils.py`), and the merge overrides exactly those. That keeps the useful behaviour — `advanced=AdvancedOptions(base_url=...)` still keeps capfire's deterministic ID/timestamp generators, which matters because otherwise a plain `advanced=AdvancedOptions(log_record_processors=[...])` would silently lose determinism — while making an explicit override always win. `_merge_dataclass` gets quite a bit shorter as a result. The tracking attribute is invisible to `dataclasses.asdict`/`fields`/`__eq__`/`replace` and to stubgen, so `logfire-api` and the executor config serialisation are unaffected.

@alexmojaki this is the one part worth a second opinion: it touches two public config dataclasses for the benefit of a test helper. The alternative is to let `advanced=` replace wholesale (except `log_record_processors`), but then passing `advanced=` at all quietly gives up capfire's deterministic IDs and timestamps, which seemed worse.

**On the CodeRabbit comments:**

- Snapshots for the span-producing tests — applied where span content is the subject of the test (`test_reconfigure_preserves_span_exporter`, `test_reconfigure_preserves_id_and_timestamp_generators`, `test_reconfigure_replaces_scrubbing`). Left the plumbing assertions alone in `test_reconfigure_extends_additional_span_processors` and `test_reconfigure_accepts_console_and_send_to_logfire_kwargs`, where the claim is "the extra processor also saw the span" rather than anything about span content, and a snapshot would only add noise.
- `# noqa: S105/S106` for the `'hunter2'` literals — skipped: this repo doesn't enable flake8-bandit (`extend-select` in `pyproject.toml` has no `S`), and it does enable `RUF100`, so adding those directives makes `make lint` fail with "Remove unused `noqa` directive". Verified locally.

`make lint`, `make typecheck` and `uv run pytest tests/test_testing.py tests/test_configure.py` are all clean.

frangelbarrera:
Thanks @dmontagu — I appreciate the review and the commit. This direction makes sense to me, and I’m happy to adjust if @alexmojaki prefers a different approach.

PR #2111, opened 2026-07-24
Log transient variables-refresh failures as warnings, not errors

dmontagu:
## Summary

The remote variables provider polls `GET /v1/variables/` every 60 seconds (as a fallback to SSE). Any failed poll — including a one-off 503 from a load balancer during a rolling deploy, a connection reset, or a read timeout — was logged as a full error-level exception record, even though a previously fetched config keeps serving values and the next poll retries within a minute. In practice this turns routine infrastructure blips into a steady stream of error-level records in the receiving project.

Separately, none of the provider's requests to the Logfire API were suppressed, so anyone with an instrumented HTTP client got a span for every poll, every write, and every SSE reconnect.

## Changes

Refresh failures:

- Transient failures (connection errors, timeouts, 5xx/408/429 responses) now log a **warning without the exception attached** while a previously fetched config is still cached. The exception type and message are recorded as attributes.
- Sustained outages still surface: after 5 consecutive failed polls (~5 minutes at the default 60s interval), failures escalate to error level.
- Non-transient HTTP errors (notably 401/403 auth misconfigurations) still log an error immediately.
- Failures before any config has been fetched still log an error immediately (the app is running entirely on code defaults).
- A successful fetch resets the consecutive-failure counter.

Instrumentation:

- Every request the provider makes to the Logfire API now runs under `suppress_instrumentation()` — the poll, the five read/write methods, and the SSE stream. Users who call `logfire.instrument_requests()` (or equivalent) no longer see spans for Logfire's own bookkeeping traffic.

No public API changes.


logfire/variables/remote.py
```diff
@@ -188,6 +196,35 @@ def _log_error(self, message: str, exc: Exception) -> None:
...
+        """Whether a refresh failure is likely to resolve on its own by the next poll.
+
+        Connection errors, timeouts, and 5xx/408/429 responses are transient (server restarts,
+        rolling deploys, load-balancer blips). Other HTTP statuses, notably 401/403, indicate a
+        misconfiguration that will not fix itself.
+        """
+        if isinstance(exc, UnexpectedResponse):
+            status = exc.response.status_code
+            return status >= 500 or status in (408, 429)
+        return isinstance(exc, RequestException)
```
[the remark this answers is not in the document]
dmontagu: Not changing this, for three reasons:

1. The suggested narrowing to `ConnectionError`/`Timeout` would not change the `SSLError` behavior: `requests.exceptions.SSLError` is a subclass of `ConnectionError`, so it would still be classified as transient. That classification is also intentional — TLS handshake failures during load-balancer churn are among the transient blips this PR is quieting, and they are indistinguishable at this level from a persistent certificate problem.
2. `InvalidURL`/`InvalidSchema`/`MissingSchema` are deterministic properties of the `base_url` and token, both fixed at construction. They fail on the very first fetch, before any config has been cached, so they always take the immediate-error path; the cached-config warning branch is unreachable for them.
3. Any persistent failure that does slip through as transient (e.g. `TooManyRedirects` appearing mid-lifetime, or a server cert expiry) is bounded by the consecutive-failure counter: it escalates to error level after 5 failed polls (~5 minutes at the default interval). The asymmetric risk runs the other way — classifying a genuinely transient failure as persistent reintroduces the unbounded error-level noise this PR removes.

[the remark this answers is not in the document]
dmontagu: Keeping the broad classification, with the rationale now documented in the docstring (d6a41eb5):

1. With `base_url` and the token fixed at construction, the only reachable `RequestException` subclasses this branch decides differently from a connection/timeout allowlist are `JSONDecodeError` on a 2xx and `TooManyRedirects`. URL/schema errors are deterministic and fail the very first fetch, before any config is cached, so they always take the immediate-error path.
2. A malformed 2xx body is as likely a misbehaving intermediary (LB error page, captive portal) as a server defect, and a server defect that persists escalates to error level via the consecutive-failure counter within 5 polls (~5 minutes at the default interval), so it is not hidden.
3. The asymmetry favors transient: a misclassified persistent failure costs at most 4 warning-level polls before errors flow, while a misclassified transient blip produces an error-level exception record on every occurrence, which is the exact noise this PR removes. Also note `SSLError` subclasses `ConnectionError`, so a connection/timeout allowlist would not change certificate handling anyway.

logfire/variables/remote.py
```diff
@@ -188,6 +196,41 @@ def _log_error(self, message: str, exc: Exception) -> None:
         else:
             warnings.warn(f'{message}: {exc}', category=RuntimeWarning)
 
+    def _log_warning(self, message: str, exc: Exception) -> None:
+        """Log a warning using logfire if available, otherwise warnings.
+
+        Unlike `_log_error`, no traceback is attached: this is for transient failures that are
+        expected to self-heal, and recording the exception would turn every blip into an
+        exception record (and a tracked issue) in the receiving project.
```
alexmojaki: this would not produce an issue by default since the level is warning

logfire/variables/remote.py
```diff
@@ -188,6 +196,41 @@ def _log_error(self, message: str, exc: Exception) -> None:
...
+
+        Args:
+            message: The warning message.
+            exc: The exception that occurred.
+        """
+        if self._logfire is not None:
+            self._logfire.warn('{message}: {error}', message=message, error=str(exc))
```
alexmojaki: i think this loses the exception type

logfire/variables/remote.py
alexmojaki, line 427: we also need logfire.suppress_instrumentation here, users who have requests instrumented are getting logs of this request (errors and otherwise)

logfire/variables/remote.py
dmontagu: You're right — `create_issue` defaults to `_record_exception and (level_is_unset or level >= 'error') and parent_span is None`, so nothing is fingerprinted at warning level and the default issues pipeline (`is_exception and attributes ? 'logfire.exception.fingerprint'`) never sees it.

Fixed the docstring in e3faeef9. There's still a reason not to attach the exception, just not that one: `is_exception` is set from the exception event regardless of level, and the standard error predicate is `otel_status_code = 'ERROR' OR level >= 17 OR is_exception`, so a warning with an exception event still counts towards service error rates. The docstring now says that instead.

logfire/variables/remote.py
dmontagu: Fixed in e3faeef9 — the type is now a separate attribute rather than relying on `str(exc)`:

```python
self._logfire.warn(
    '{message}: {error_type}: {error}',
    message=message,
    error_type=type(exc).__name__,
    error=str(exc),
)
```

The `warnings.warn` fallback (when no logfire instance is bound) includes it too. `test_transient_failure_with_cached_config_logs_warning` now asserts `error_type == 'UnexpectedResponse'`.

logfire/variables/remote.py
dmontagu: Done in e3faeef9. Applied to every request the provider makes, not just the poll:

- `refresh()` — the poll. The `except` handler runs outside the suppressed block, so failures are still logged.
- `_sse_listener` — this one mattered most: an instrumented client would otherwise open a span per reconnect, each lasting as long as a stream meant to stay open indefinitely.
- `create_variable` / `update_variable` / `delete_variable` / `list_variable_types` / `upsert_variable_type`.

Two tests cover it: one asserting no spans are exported for a poll + all four write calls under `logfire.instrument_requests()`, and a control asserting a user's own request in the same block still gets a span (suppression isn't leaking).

logfire/variables/remote.py
```diff
@@ -188,6 +196,48 @@ def _log_error(self, message: str, exc: Exception) -> None:
...
+        """
+        if self._logfire is not None:
+            self._logfire.warn(
+                '{message}: {error_type}: {error}',
+                message=message,
+                error_type=type(exc).__name__,
+                error=str(exc),
+            )
+        else:
+            warnings.warn(f'{message}: {type(exc).__name__}: {exc}', category=RuntimeWarning)
```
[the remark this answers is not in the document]
dmontagu: Not changing this.

B028 isn't enabled in this repo — `[tool.ruff.lint].extend-select` is `Q, RUF100, C90, UP, I, D, DTZ005, FA`, with no `B`. `make lint` and the `lint` CI job both pass on this line.

On the merit rather than the lint rule: this warning is emitted from `refresh()` running on the provider's background daemon thread, so `stacklevel=2` would point at `refresh()` in this same file rather than at any user code — no more useful than the default. `_log_error` directly above uses the same `warnings.warn(...)` shape without `stacklevel`, so adding it to only one of the pair would just make them inconsistent.

logfire/variables/remote.py
alexmojaki: wdym by "standard error predicate"? is such a thing documented? if anything it should be `level >= 'error'`.

dmontagu: "Standard" was the wrong word and the claim doesn't hold — I've dropped it in c8f6c235.

What I was referring to: `otel_status_code = 'ERROR' OR level >= 17 OR is_exception` appears in three places in the platform repo, deliberately kept in sync with each other — `ERROR_PREDICATE` in `services/logfire-frontend/src/app/services/red-metrics.ts`, the `errors_hll` branch of the `service_red` default MV, and `_RECORDS_ERROR_PREDICATE` in `logfire-db/crud/recommended_slos.py`. So it's an internal convention shared by the Services page and the SLO presets, not anything documented or contractual.

But it doesn't apply here regardless: `service_red` is `WHERE kind = 'span'`, and `logfire.warn()` sets `logfire.span_type: 'log'` → `kind = 'log'`, so the Services error rate can't see this record either way. The SLO presets are provider-scoped. I checked after your comment rather than before writing it — I shouldn't have leaned on a backend detail from an SDK docstring, and the one I picked was wrong.

The docstring now gives the reason that actually stands on its own:

> Unlike `_log_error`, the exception is not attached to the record; its type and message are recorded as attributes instead. These are transient failures expected to self-heal, and the traceback runs through `requests` internals rather than any user code, so attaching it would add a stacktrace to every blip without saying anything the type and message don't.

If you'd rather not keep the distinction at all now that the issues concern is gone, happy to just pass `_exc_info=exc` here too — that collapses `_log_warning` into `_log_error` with a different level. Your call.

alexmojaki (approved):

PR #2158, opened 2026-07-28
docs: fix silent OTLP data loss for non-Python clients + wrong Get Help description

strawgate:
Two fixes surfaced by a persona-journey docs audit.

## Alternative clients (any-language OpenTelemetry path)
The generic env-var recipe omitted `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`. Logfire receives OpenTelemetry data over HTTP only, but many SDKs (Java, .NET, anything using a gRPC exporter) default to gRPC — so they send **nothing, usually with no error**, and a user following this page had no way to know why. This is the catch-all page for languages without a dedicated guide, so it was the most likely place to hit the problem and the least equipped to explain it.

- Add `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` with a warning explaining the gRPC-default failure mode.
- Add `OTEL_SERVICE_NAME` (without it, data lands under `(unknown)` in the Services view).
- Add **Verify** and **Troubleshooting** sections to match the dedicated language pages.

## Get Help page
Its meta description was copy-pasted from a Slack-alerts page and described completely different content. Rewritten to match the page.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2158?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -1,6 +1,6 @@
 ---
 title: "Get Help with Pydantic Logfire: Docs, Support & GitHub"
-description: "Need to send alerts to Slack from Logfire? This guide shows how to create a Slack webhook and define SQL-based alert criteria for your Slack alert system."
+description: "Where to get help with Pydantic Logfire: the community Slack, GitHub issues, documentation, and email support."
 hide:
 ---
 # Getting help with Pydantic Logfire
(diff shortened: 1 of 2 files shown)


strawgate:
Addressed the review feedback:
- Added `OTEL_SERVICE_NAME` to the Rust example block (it only exported endpoint + headers, so Rust users would still land under `(unknown)`). Node already sets the service name in code via `Resource`. Thanks @coderabbitai / cubic.
- Replaced the em-dash in the warning with a comma, per the style guide (§5, no em-dashes).

strawgate:
@coderabbitai the Rust `OTEL_SERVICE_NAME` and the em-dash are both fixed — please re-review.

Issue #520, opened 2024-10-19
It should be possible to install `logfire` without automatically enabling `logfire-api` use


samuelcolvin:
### Description

The ideal would be I guess

```py
logfire.enable_logfire_api('module.path')
```

But that would be a breaking change. Perhaps we do this via an env variable, so 

https://github.com/pydantic/logfire/blob/cc79cf448a71f8c7a0f57deacaf3676e603168c4/logfire-api/logfire_api/__init__.py#L9-L14

becomes some thing like

```py
_real_logfire_enabled = False
if not os.getenv('LOGFIRE_API_DISABLED'):
    try:
        logfire_module = importlib.import_module('logfire')
        sys.modules[__name__] = logfire_module
        _real_logfire_enabled = True
    except ImportError:
        pass

if not _real_logfire_enabled:
    if not TYPE_CHECKING:  # pragma: no branch
        ...
```

Either way, I need a way to have `logfire` installed without activating `logfire-api` calls.

samuelcolvin:
I still think this would come in useful, but it's no longer urgent.

Kludex:
I don't understand the reason for it. The idea was exactly to use logfire when installed.

alexmojaki:
I am also confused.

A library using logfire-api typically won't call `configure` because that's up to the application. So the application isn't forced to actually use logfire just because it's installed, they just need to set `LOGFIRE_IGNORE_NO_CONFIG`.

samuelcolvin:
For the same reason you can opt to enable logging for just some of the modules that have logging integrated - you might want use logfire but not collect data from all code which has `logfire-api` integrated.

I also think it's unfortunate if installing logfire causes code which was previously be behaving in one way to suddenly "break" - e.g. raise an error because you haven't called `logfire.configure()`.

Datapoints that support the idea that you should be able to switch `logfire-api` off for certain libraries/modules:
* rust's `tracing` doesn't suddenly start sending data if you install `tracing`, you need to enable it for each package
* (as above) you can control which modules you collect logs from
* otel library instrumentation doesn't automatically start collecting data just because you have the relevant package installed

I think it's pretty uncontraverial that you should be able to decide which modules you want to receive traces/spans from.

alexmojaki:
> I also think it's unfortunate if installing logfire causes code which was previously be behaving in one way to suddenly "break" - e.g. raise an error because you haven't called `logfire.configure()`.

[...]

PR #2162, opened 2026-07-29
fix(anthropic): price Bedrock calls under the aws provider

arthi-arumugam-git:
`AnthropicBedrock` calls come out with token counts but no `operation.cost`, every time.

`get_anthropic_usage_attributes` hardcodes `provider_id='anthropic'`. Through Bedrock the model ID is a Bedrock one, so the lookup is for `anthropic.claude-3-haiku-20240307-v1:0` inside the `anthropic` provider, which has no such model. `calc_price` raises `LookupError`, the `except Exception: pass` in `get_usage_attributes` swallows it, and the cost is dropped. Tokens are set before that block, which is why they still show up and the failure looks like "prices missing for Bedrock" rather than an error.

The prices are not missing. genai-prices resolves all of these under `aws` today:

```
anthropic.claude-3-haiku-20240307-v1:0                        -> $0.00000425
us.anthropic.claude-3-7-sonnet-20250219-v1:0                  -> $0.000051
eu.anthropic.claude-3-7-sonnet-20250219-v1:0                  -> $0.000051
arn:aws:bedrock:eu-west-1:...:inference-profile/eu.anthropic. -> $0.000051
```

### Why just changing the provider ID doesn't work

That was my first patch, and it fails differently:

```
ValueError: Missing value at `usage.inputTokens`
```

genai-prices' `aws` extractor expects the Bedrock Converse response shape, `usage.inputTokens`. `AnthropicBedrock` is not the Converse API; it returns the Anthropic Messages shape, `usage.input_tokens`. So `extract_usage` can read the body or the `aws` prices, but not both.

The fix is to stop going through `extract_usage` for this case. We already compute `input_tokens` and `output_tokens` from the response two lines earlier, so `get_usage_attributes` takes an optional `model_ref` (plus the cache token split) and prices those directly, bypassing the body-shape step. Everything else keeps the existing path untouched.

### Why the model ID and not the client type

`instrument_anthropic` knows perfectly well whether it was handed a Bedrock client, but that information doesn't reach here. Streaming arrives via `AnthropicMessageStreamState.get_attributes`, which calls `get_anthropic_usage_attributes` with an accumulated message and no client, so threading the client through would mean changing the shared `instrument_llm_provider` signature and carrying it into `StreamState` as well.

Matching the model ID covers both paths with one branch, and it fails safe: a native Anthropic model name cannot take the `[region.]anthropic.` or `arn:aws:bedrock:` shape, so anything that doesn't match falls through to the `anthropic` provider exactly as it does today. Happy to do it the other way if you'd rather.

### Scope

This is the `anthropic` SDK path. The reporter on #1023 is using pydantic-ai's `BedrockConverseModel`, where the cost is recorded by pydantic-ai rather than here, so that half is not addressed by this and I did not want to claim otherwise. What I can say is that this repo has the same bug on the client it does own, and `tests/otel_integrations/test_anthropic_bedrock.py` had zero `operation.cost` assertions against 27 in `test_anthropic.py`, so it wasn't going to be caught.

### Tests

The existing `test_sync_messages` snapshot gains the `operation.cost` it should have had. New parametrized test covers the plain model ID, the `us.`/`eu.` cross-region inference profiles and the full inference-profile ARN.

```
tests/otel_integrations/test_anthropic_bedrock.py .......  6 passed
tests/otel_integrations/test_openai.py + test_llm_usage_attributes.py + test_anthropic_bedrock.py  86 passed
```

(`test_anthropic.py::test_async_beta_messages` fails for me on a clean checkout too, before any of this, so I left it alone.)

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -304,6 +305,16 @@ def get_attributes(self, span_data: dict[str, Any]) -> dict[str, Any]:
...
 
 
+# Bedrock model IDs are the Anthropic model name behind a `anthropic.` vendor prefix,
+# optionally behind a geography prefix for cross-region inference profiles, and optionally
+# behind a full inference profile ARN:
+#   anthropic.claude-3-haiku-20240307-v1:0
+#   eu.anthropic.claude-3-7-sonnet-20250219-v1:0
+#   arn:aws:bedrock:eu-west-1:123456789012:inference-profile/eu.anthropic.claude-3-7-sonnet-20250219-v1:0
+# Native Anthropic model names never take that shape, so this only matches Bedrock.
+_BEDROCK_MODEL_RE = re.compile(r'^(arn:aws:bedrock:|([a-z]{2,4}\.)?anthropic\.)')
```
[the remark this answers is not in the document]
arthi-arumugam-git: Good catch, and it was the exact failure this PR exists to remove: an ID we don't recognise loses its cost silently.

`global` and `us-gov` are both real and both missed by the 2-4 letter rule. genai-prices prices both, and `global.anthropic.claude-sonnet-4-5-20250929-v1:0` resolves to its own model entry rather than the regional one, so at 1M in / 1M out it is $0.018 against $0.0198 regional. Dropping it would have been a real number, not a rounding artifact.

Widened to any single label before `anthropic.` in b2dc784, since `anthropic.` is the part that actually identifies Bedrock and a native Anthropic model name never contains it. Added `apac`, `us-gov` and `global` to the parametrized test.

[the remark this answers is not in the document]
arthi-arumugam-git: Fixed in b2dc784, thanks. Same point cubic raised, and it was worth catching.

On the other note about using the full span-snapshot pattern: I left the new test as targeted assertions deliberately. It is parametrized over seven model IDs and the only thing that varies between them is the cost, so seven near-identical 90-line snapshots would bury the one line that matters. `test_sync_messages` right above it still does the full snapshot assertion for the span shape, and it now carries the `operation.cost` it was missing, so that coverage is not lost. Happy to switch if you would rather have it uniform.

arthi-arumugam-git:
Gentle ping on this one. It routes Bedrock pricing through the aws provider, and it is green and mergeable. Is anything else needed before it can go in, or is it just waiting on a maintainer pass?

arthi-arumugam-git:
Gentle ping on this one. CI is green and it still merges cleanly.

Context that may help it get a look: the sibling fix in pydantic/genai-prices (#520, Bedrock prices for Writer Palmyra X4 and X5) was merged on 5 August. This is the same class of problem one layer up, where Anthropic calls made through Bedrock get priced under the anthropic provider instead of aws, so the cost attributed to a Bedrock run comes out wrong with nothing raising.

Happy to rebase or split it if that makes review easier.

logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -313,8 +329,28 @@ def get_anthropic_usage_attributes(response: Any) -> dict[str, Any]:
...
-    input_tokens = usage.input_tokens + (usage.cache_read_input_tokens or 0) + (usage.cache_creation_input_tokens or 0)
+    cache_read_tokens = usage.cache_read_input_tokens or 0
+    cache_write_tokens = usage.cache_creation_input_tokens or 0
+    input_tokens = usage.input_tokens + cache_read_tokens + cache_write_tokens
     output_tokens = usage.output_tokens
+
+    model = getattr(response, 'model', None)
+    if isinstance(model, str) and _BEDROCK_MODEL_RE.match(model):
+        # Bedrock-hosted models are priced under the `aws` provider, not `anthropic`.
+        # We can't hand the response to genai_prices' `aws` extractor though: that expects the
```
alexmojaki: provider_id='aws', api_flavor='anthropic' should work

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -37,20 +45,36 @@ def get_usage_attributes(
...
-        else:
-            response_data = response.model_dump()
-        extract_kwargs: dict[str, Any] = {'provider_id': provider_id}
-        if api_flavor is not None:
-            extract_kwargs['api_flavor'] = api_flavor
-        usage_data = extract_usage(response_data, **extract_kwargs)
-        if usage_data.model is not None:
-            result['operation.cost'] = float(
-                calc_price(usage_data.usage, model_ref=usage_data.model.id, provider_id=provider_id).total_price
+        if model_ref is not None:
```
alexmojaki: conditioning behaviour on the presence of a model is confusing. rather replicate this:

https://github.com/pydantic/pydantic-ai/blob/28a58d53b4d75b4da6d3b372edb9651f9cbe2411/pydantic_ai_slim/pydantic_ai/usage.py#L327-L329

that should fix the issue by inferring the correct provider from the url, rather than the model, and help more generally

arthi-arumugam-git: Done in 16b1250. get_usage_attributes now resolves the provider the way usage.py does in pydantic-ai: the client's base URL first, the integration's provider id as the fallback, each candidate tried in order. The URL is read from the instrumented client (the bound instance when classes are instrumented) and threaded through to on_response and the stream states, and the model-id regex is gone.

You were right that it helps more generally. Two costs that were silently missing start appearing, both covered by updated snapshots: beta Anthropic messages, whose response shape the extractor rejected, and OpenRouter models called through the OpenAI client, which never resolved under the openai provider so calc_price raised and the cost was dropped.

logfire/_internal/integrations/llm_providers/anthropic.py
arthi-arumugam-git: I tried this first: extract_usage(data, provider_id='aws', api_flavor='anthropic') does work. What stopped me from using it for both candidates in the URL-first chain is that the flavor is provider-specific: on genai-prices 0.1.1 the anthropic provider raises "Unknown api_flavor 'anthropic', allowed values: default, chat", so the aws candidate and the anthropic fallback would need different flavors. Since genai-prices isn't a pinned dependency here, I went with pricing off the token counts the integration already extracts, which works identically for every candidate and any installed version. Happy to switch to per-candidate flavors instead if you prefer that shape.

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -36,21 +51,62 @@ def get_usage_attributes(
...
+            for candidate_id, candidate_url in candidates:
+                try:
+                    result['operation.cost'] = float(
+                        calc_price(
+                            price_usage, model_ref=model_ref, provider_id=candidate_id, provider_api_url=candidate_url
+                        ).total_price
+                    )
+                    break
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
arthi-arumugam-git: S110 and BLE001 are not enabled in this repo's ruff config (extend-select is Q, RUF100, C90, UP, I, D, DTZ005, FA), so make lint passes as written; the report above ran ruff with --select S110 explicitly. The blanket except around pricing is deliberate and predates this PR in this same function: cost calculation must never be able to break a user's request or telemetry, and genai_prices is not a pinned dependency here, so its exception surface is not stable across installed versions (0.1.1 raises LookupError and ValueError in the cases I probed, but that is not a contract). It is also the same per-candidate 'except Exception: pass' used by pydantic-ai's RequestUsage.extract, which this PR was reviewed toward replicating. 'continue' instead of 'pass' changes nothing here since the try is the last statement in the loop body. Leaving as is unless the maintainers want the S rules adopted repo-wide.

arthi-arumugam-git:
Catching up on the cubic review, which I had not answered in writing. Two of the five are now addressed on the branch and I disagree with one.

**Streaming path untested.** Fair, and it was the useful one: coverage counted the line but nothing asserted the number it produced. A faithful Bedrock streaming mock turned out to be disproportionate, since Bedrock streams over AWS binary event-stream framing rather than SSE and mocking that means hand-building checksummed frames. So I asserted the wiring where it lives instead: the stream state carries the client base URL and the accumulated message prices under `aws` because of it, with the same call scoring nothing without it.

**Loop falls through and the cost is dropped.** Also covered now, by a test where no candidate can price the model: tokens are still recorded and `operation.cost` is simply absent, which is the behaviour I want rather than a zero.

**Catch only the expected genai_prices errors.** Leaving this. Cost calculation must never be able to break a user's request or their telemetry, genai_prices is not pinned here so its exception surface is not a contract, and this is the same per-candidate pattern that pydantic-ai's `RequestUsage.extract` uses, which is what this PR was reviewed toward. Happy to narrow it if the maintainers would rather.

The snapshot and the `provider_api_url` observations look like descriptions of the change rather than problems, so I have not acted on those.

arthi-arumugam-git:
@coderabbitai review

The broad except was answered above rather than changed, and the two findings that were real (the untested streaming path, and the loop dropping the cost) are fixed on the branch since your review. Re-reviewing so the stale changes-requested state does not sit on a green PR.

arthi-arumugam-git:
Small housekeeping note, since it is not visible from the outside: this PR still shows "changes requested", and that is the CodeRabbit review from yesterday rather than a human one. I answered it in the thread above (the broad except is deliberate, and the two findings that were real are fixed on the branch), but I cannot ask the bot to re-review because it only takes commands from org members.

So the state is: green on all checks, mergeable, and nothing outstanding that I know of. Happy to change anything else if you would rather the except were narrowed.

alexmojaki:
Please respond to review comments in the threads

logfire/_internal/integrations/llm_providers/llm_provider.py
```diff
@@ -101,13 +113,14 @@ def _instrumentation_setup(*args: Any, **kwargs: Any) -> Any:
                 stream_cls = kwargs['stream_cls']
                 assert stream_cls is not None, 'Expected `stream_cls` when streaming'
                 original_context = get_context()
+                stream_base_url = _client_base_url(args)
```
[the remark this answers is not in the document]
arthi-arumugam-git: Sorry, I answered this in a comment on the PR instead of in here.

This one is fixed. There is a test now for the streaming path, though not the way you suggested. Bedrock streams over AWS binary event framing rather than SSE, so mocking a real stream means hand building checksummed frames, which felt like a lot of machinery for one assertion. Instead the test goes at the wiring: the stream state carries the client base URL, and the same message prices under aws because of it, and scores nothing without it.

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -36,21 +51,62 @@ def get_usage_attributes(
...
+        if model_ref is not None:
+            from genai_prices import Usage as PriceUsage
+
+            price_usage = PriceUsage(
+                input_tokens=input_tokens,
+                output_tokens=output_tokens,
+                cache_read_tokens=cache_read_tokens,
+                cache_write_tokens=cache_write_tokens,
             )
+            for candidate_id, candidate_url in candidates:
```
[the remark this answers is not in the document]
arthi-arumugam-git: Also fixed, and sorry for answering this on the PR rather than here.

There is a test for it now with a model none of the candidates can price. Tokens still get recorded and operation.cost is simply missing, which is what I wanted rather than a zero sitting there looking like a real number.

tests/otel_integrations/test_anthropic.py
```diff
@@ -2017,8 +2017,7 @@ async def test_async_beta_messages(exporter: TestExporter) -> None:
                         'usage': {
                             'cache_creation': {'ephemeral_1h_input_tokens': 0, 'ephemeral_5m_input_tokens': 0},
                             'cache_creation_input_tokens': 0,
-                            'cache_read_input_tokens': 0,
-                            'inference_geo': 'not_available',
+                            'cache_read_input_tokens': 0, 'fallback_credit': None, 'inference_geo': 'not_available',
```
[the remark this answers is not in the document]
arthi-arumugam-git: This one I think is fine as is.

The snapshot is the repo's own format and that field comes from their fixture, so it is not something my change introduced. Happy to look again if you meant something more specific.

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -36,21 +51,62 @@ def get_usage_attributes(
...
+                try:
+                    result['operation.cost'] = float(
+                        calc_price(
+                            price_usage, model_ref=model_ref, provider_id=candidate_id, provider_api_url=candidate_url
+                        ).total_price
+                    )
+                    break
+                except Exception:
```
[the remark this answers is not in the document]
arthi-arumugam-git: This one I am leaving, and sorry for replying on the PR rather than in the thread.

Working out a cost should never be able to break someone's request or their tracing. genai-prices is not pinned here either, so what it raises is not really a contract I can rely on. It is also the same shape pydantic-ai uses in RequestUsage.extract, which is what this PR was pointed at. If the maintainers would rather it were narrower I will happily change it.

logfire/_internal/integrations/llm_providers/usage.py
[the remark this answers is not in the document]
arthi-arumugam-git: Answering here rather than on the PR, sorry.

Same as above: the broad except is deliberate so a pricing lookup can never take down a user's request or their spans. The two findings in this review that were real, the untested streaming path and the fallthrough dropping the cost, are both fixed on the branch now.

arthi-arumugam-git:
Understood, sorry about that. I have gone back and replied in each of the review threads instead. Your two are answered in place as well.

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -21,7 +25,18 @@ def get_usage_attributes(
...
     Returns only attributes that have values.
 
-    Token/raw-usage and cost fail independently — a cost error does not prevent tokens
+    provider_url is the client's base URL. When present it takes precedence over provider_id
+    for identifying the provider, since the URL identifies who actually served (and bills)
+    the request, e.g. Anthropic models served by Bedrock are priced under `aws`. This mirrors
+    pydantic-ai's resolution order:
+    https://github.com/pydantic/pydantic-ai/blob/28a58d53b4d75b4da6d3b372edb9651f9cbe2411/pydantic_ai_slim/pydantic_ai/usage.py#L327-L329
+
+    Pass model_ref when the response body isn't shaped the way genai_prices' extractors
```
alexmojaki: this isn't what's been done

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -36,21 +51,64 @@ def get_usage_attributes(
...
+    # (provider_id, provider_api_url) candidates, URL first: genai_prices raises on a
+    # candidate that doesn't resolve, and the next one is tried. `calc_price` and
+    # `extract_usage` overload the two provider arguments as mutually exclusive, so a
+    # candidate is either a URL or an id and each is passed on its own.
+    candidates: list[str] = [provider_id]
+    if provider_url is not None:
+        candidates.insert(0, provider_url)
+
+    def is_url(candidate: str) -> bool:
+        return candidate.startswith(('http://', 'https://'))
```
alexmojaki: this is very strange. you already have the provider_id and provider_url separately, you know which is which. this feels like a weird AI choice that wasn't reviewed by a human. otherwise it needs a comment explaining the is_url approach.

logfire/_internal/integrations/llm_providers/usage.py
```diff
@@ -36,21 +51,64 @@ def get_usage_attributes(
...
-        if api_flavor is not None:
-            extract_kwargs['api_flavor'] = api_flavor
-        usage_data = extract_usage(response_data, **extract_kwargs)
-        if usage_data.model is not None:
-            result['operation.cost'] = float(
-                calc_price(usage_data.usage, model_ref=usage_data.model.id, provider_id=provider_id).total_price
+        if model_ref is not None:
+            from genai_prices import Usage as PriceUsage
+
+            price_usage = PriceUsage(
```
alexmojaki: constructing usage should be a last resort, try to use genai-prices extractors as much as possible. https://github.com/pydantic/pydantic-ai/blob/28a58d53b4d75b4da6d3b372edb9651f9cbe2411/pydantic_ai_slim/pydantic_ai/usage.py#L327-L329 shows how to get the provider before extracting. if the provider is known to be aws, set the api_flavor based on that.

logfire/_internal/integrations/llm_providers/usage.py
arthi-arumugam-git: You're right, and it was a bad answer to a real constraint rather than a considered design.

What happened: `calc_price` and `extract_usage` declare `provider_id` and `provider_api_url` in mutually exclusive overloads, so passing both with one set to None matches neither and pyright rejects it. I flattened the pair to get around that, and then had to recover which was which by looking at the string, which is exactly as backwards as it reads.

`find_provider` takes model_ref, provider_id and provider_api_url together, so the constraint isn't there at all. That's what the pydantic-ai code you linked uses, and I should have followed it rather than working around the wrong function. The candidates are back to `[(None, provider_url), (provider_id, None)]` and nothing inspects a string.

logfire/_internal/integrations/llm_providers/usage.py
arthi-arumugam-git: Fixed. `provider.extract_usage(...)` is the default path now, and the hand-built `Usage` is only used when the caller passes `model_ref`, which is the case where the response body isn't shaped the way the extractors expect.

On the flavor: `anthropic` is a flavor on the aws provider only, so it's chosen from the resolved provider rather than assumed, which also means Bedrock responses go through the extractor instead of being constructed by hand.

logfire/_internal/integrations/llm_providers/usage.py
arthi-arumugam-git: You're right, the docstring was describing the intent rather than the code. It's accurate now: the resolution really does go through `find_provider` in the same order pydantic-ai uses, URL first and the provider id second.

PR #2163, opened 2026-07-29
docs: describe a trace accurately (not only as a request), across getting-started and AI pages

strawgate:
A trace isn't always an HTTP request — it's any top-level operation: a background job, a scheduled task, or an **agent run**. Several pages glossed a trace as "one request" / "the journey of one request," which is inaccurate (an agent run isn't a request), and a few were even self-contradictory (e.g. the AI observability overview said "turns an agent run into one trace (the full journey of one request…)").

## Changes
- **Getting-started definitions**: `first-trace.md`, `for-ai-engineers.md`, `for-backend-and-sre.md` — reworded to name jobs / tasks / agent runs, not just requests.
- **AI/agent pages** (11): the AI observability overview, the evaluate-an-agent cookbook, and the LLM integration pages (`openai`, `anthropic`, `google-genai`, `langchain`, `litellm`, `llamaindex`, `magentic`, `mcp`, `mirascope`) — the shared gloss now reads "one request or agent run."

## Scope note
The same "full journey of one request" gloss also appears on the **web-framework / database / HTTP** integration pages, but those pages are specifically about request-handling code, so "one request" is contextually accurate there and is intentionally left unchanged.

### DIFF
--- a/docs/ai-observability.md
+++ b/docs/ai-observability.md
@@ -13,7 +13,7 @@ description: "See every agent decision, tool call, and downstream service in one
 
 <div class="agent-observability-intro__detail-copy">
 
-<p>Logfire turns an agent run into one trace (the full journey of one request, made of nested spans): the model calls, tool calls, database queries, API requests, and response your user receives.</p>
+<p>Logfire turns an agent run into one trace (the full journey of one request or agent run, made of nested spans): the model calls, tool calls, database queries, API requests, and response your user receives.</p>
 
 </div>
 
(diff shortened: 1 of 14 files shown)


strawgate:
Took the substance of this — first-trace now reads "one request, job, task, or agent run", and I broadened the AI/agent pages too. I kept "the full record … from start to finish" rather than reverting to "journey", since a human reviewer specifically flagged "journey of one request" as too abstract on the sibling TypeScript page.

PR #2166, opened 2026-07-29
Raise an error instead of asserting on unexpected query API responses

adriangb:
`_BaseLogfireQueryClient.handle_response_errors` classified every query API response by status code:

```python
if response.status_code == 400:
    ...
    raise QueryExecutionError(data)
if response.status_code == 422:
    ...
    raise QueryRequestError(data)
assert response.status_code == 200, response.content
```

That trailing `assert` was the only guard against a non-200, non-400, non-422 response, which is a problem in two ways:

1. **`python -O` (and `PYTHONOPTIMIZE`) strips `assert` statements.** A 500 or 503 then falls straight through, and the caller parses the error body as if the query had succeeded. If the failing response happens to be shaped like a result (an empty `{"schema": ..., "data": []}` from a proxy or an error handler, for instance), the caller gets a silently wrong answer back from a query client, which is about the worst failure mode available to an observability tool.
2. **Even with assertions enabled, a server-side failure surfaces as a bare `AssertionError`**, not one of the module's own exception types, so callers that already catch `QueryExecutionError` / `QueryRequestError` have nothing meaningful to catch it with.

## The change

A new `UnexpectedResponseError` is raised for any status code that isn't 200, 400 or 422:

```python
if response.status_code != 200:
    raise UnexpectedResponseError(f'Unexpected response status code {response.status_code}: {response.text!r}')
```

It is a `RuntimeError` subclass like its three siblings, and is exported from `logfire.query_client` alongside them (`logfire/query_client.py` is only a re-export shim, so `logfire/experimental/query_client.py` is the single implementation site). The message keeps the status code and the body, so it is at least as useful for debugging as the old `response.content` assertion message.

The 400 and 422 branches, including their content-type handling, are untouched: this is strictly additive, not a reclassification of statuses that already worked. Their `# pragma: no cover` markers are dropped because the new tests now cover them.

Notes on two alternatives that were considered:

* **Reusing `QueryExecutionError`.** Its docstring is "Raised when the query execution fails on the server", which fits a 500 but not a 404, a 503 from a load balancer, or a redirect. It would also silently widen the meaning of an exception people already catch, and `handle_response_errors` is shared with `/v1/read-token-info`, where "query execution" means nothing.
* **`response.raise_for_status()`.** It raises `httpx.HTTPStatusError`, which is catchable and standard, but it drags an httpx exception type into a public API that otherwise raises its own `RuntimeError` subclasses, and it does nothing for 2xx-but-not-200 statuses such as 204.

The body of an unexpected response is deliberately not decoded as JSON the way the 400/422 bodies are: an error page from a proxy or a truncated body would raise a `JSONDecodeError` and mask the actual failure.

## Tests

`tests/test_query_client_errors.py` drives both the sync and async clients through an `httpx.MockTransport`, rather than the cassettes used by `test_query_client.py`, since a healthy server does not produce these responses. It covers:

* 204, 301, 404, 500, 502 and 503 raising `UnexpectedResponseError` from `query_json_rows()` and from `info()`
* an unexpected response that claims `content-type: application/json` but has a truncated body
* the unchanged 400 / 422 behaviour, for JSON bodies, plain-text bodies, and a content type carrying parameters
* a subprocess run under `python -O` that fails if no error is raised, with an `if __debug__: sys.exit(...)` guard so it cannot pass vacuously when assertions are enabled

Against the previous code, 16 of the 24 new tests fail (the 8 that pass are the unchanged 400 / 422 cases).

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2166?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tests/test_query_client_errors.py
```diff
@@ -0,0 +1,178 @@
...
+    async with AsyncLogfireQueryClient(
+        read_token=READ_TOKEN, base_url=BASE_URL, transport=mock_transport(status_code, json={'detail': 'nope'})
+    ) as client:
+        with pytest.raises(expected_error) as exc_info:
+            await client.query_json_rows(SQL, min_timestamp=MIN_TIMESTAMP)
+
+    assert exc_info.value.args == ({'detail': 'nope'},)
+
+
+def test_unexpected_status_code_raises_with_assertions_disabled():
```
adriangb: This seems over-engineered to me. I would remove this test.

alexmojaki (approved):

tests/test_query_client_errors.py
dmontagu: Agreed - removed this test in 80008b95.

tests/test_query_client_errors.py
```diff
@@ -0,0 +1,178 @@
...
+    lambda request: httpx.Response(500, json={{'schema': {{'fields': []}}, 'data': []}})
+)
+with LogfireQueryClient(read_token={READ_TOKEN!r}, base_url={BASE_URL!r}, transport=transport) as client:
+    try:
+        result = client.query_json_rows({SQL!r}, min_timestamp=datetime(2020, 1, 1, tzinfo=timezone.utc))
+    except UnexpectedResponseError as exc:
+        print(exc)
+    else:
+        sys.exit(f'no error was raised for a 500 response, got {{result}}')
+"""
```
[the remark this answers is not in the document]
dmontagu: This test was removed in 80008b95 (see the thread above).

PR #2168, opened 2026-07-30
chore: Update dependencies

alexmojaki:


pyproject.toml
```diff
@@ -255,6 +256,9 @@ line-length = 120
 extend-exclude = ["logfire-api/logfire_api/*", "tests/auto_trace_samples/param_spec.py"]
 
 [tool.ruff.lint]
+# Ruff 0.16 greatly expanded the default rule set (59 -> 413 rules).
+# Restore the pre-0.16 defaults until we deliberately migrate to the new defaults.
+select = ["E4", "E7", "E9", "F"]
```
alexmojaki, line 261: explain how this line works

docs/reference/advanced/baggage.md
```diff
@@ -38,7 +38,7 @@ logfire.configure()
 
 with logfire.set_baggage(user_id='scolvin'):
     print(logfire.get_baggage())  # (just for demonstration, not usually needed)
-    #> {'user_id': 'scolvin'}
+    # > {'user_id': 'scolvin'}
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 03b878ec, which reverts the Markdown formatting commit entirely, and dc6a6fe9, which adds `exclude = ["*.md"]` to `[tool.ruff.format]` so ruff 0.16's new Markdown code-block formatting can't mangle the `#>` markers again. pytest-examples remains the authority on docs code formatting; `tests/test_docs.py` passes again as of f5dd20fc.

pyproject.toml
alexmojaki: (AI) `select` sets the *base* rule selection, replacing ruff's built-in default set (whereas `extend-select` below adds rules on top of that base, and `ignore` subtracts from the result). Each entry is a rule-code prefix that selects every rule under it: `F` is all of pyflakes, and `E4`/`E7`/`E9` are the pycodestyle error subgroups for imports (e.g. E401), statements (e.g. E711/E722), and syntax/IO errors (E9xx).

`["E4", "E7", "E9", "F"]` was ruff's documented default selection before 0.16. Ruff 0.16.0 changed the default from those 59 rules to 413 rules spanning many more categories (BLE, S, SIM, PIE, TRY, ...), which immediately flagged dozens of violations in this repo — many of them intentional patterns like broad `except Exception` in code that must never crash the host app. This line pins the base selection back to the pre-0.16 defaults so the deps bump doesn't bundle a lint-policy migration; adopting (some of) the new defaults can be a deliberate follow-up.

PR #2177, opened 2026-08-01
ci: run pytest with xdist (-n logical) in the test matrix

dmontagu:
The pytest step is 96% of each paid test-matrix job's wall time (~312s of ~324s), running single-process on a 4-core runner. Run it with xdist instead, mirroring the Makefile's `testcov` target exactly (`coverage run -m pytest -n logical --dist=loadgroup`), which the suite already uses for every local run:

- `-n logical` uses all vCPUs (Ubicloud vCPUs are hyperthreaded, so `-n auto` would only use half; `psutil` is already in the dependency tree).
- `--dist=loadgroup` preserves the `xdist_group` co-scheduling the local runs rely on.
- Coverage already collects from xdist workers via `patch = ["subprocess"]` in `pyproject.toml`, and the combine job merges the per-leg parallel data files unchanged.

This should substantially cut both merge-queue latency and paid runner minutes per CI round.

Previously parked as draft behind #2179: the suite had cross-test interference under xdist that is now fixed on main (#2194, #2195, #2196). Eight consecutive full-suite `-n logical --dist=loadgroup` runs on this branch pass locally (2291 passed, 0 failed).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -125,7 +125,7 @@ jobs:
           uv pip install 'opentelemetry-exporter-otlp-proto-http==${{ matrix.otel-version }}.*'
 
       - run: mkdir coverage
-      - run: uv run --no-sync coverage run -m pytest --junitxml=coverage/test-results.xml
+      - run: uv run --no-sync coverage run -m pytest -n logical --dist=loadgroup --junitxml=coverage/test-results.xml
         env:
           COVERAGE_FILE: coverage/.coverage.${{ runner.os }}-py${{ matrix.python-version }}-pydantic-${{ matrix.pydantic-version }}-otel-${{ matrix.otel-version }}
       - name: store coverage files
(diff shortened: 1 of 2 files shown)


dmontagu:
Parking this as draft: the runner change itself works (test legs drop from ~5-6 min to ~1.5-2.5 min, pytest step 312s to 82s), but running the suite in parallel surfaces a population of load/ordering-sensitive tests beyond the two already fixed (celery in #2176, pytest-plugin env in #2178): the claude-agent-sdk cassette test races its tool-use hooks, a test leaks global requests instrumentation that breaks a variables test under certain worker co-locations, and a local full-suite run also caught test_anthropic and test_pydantic_ai_mcp failures. Queueing this before those are fixed would reintroduce exactly the merge-queue flake tax we just eliminated. Tracking the remaining work in the follow-up issue.

PR #1745, opened 2026-03-01
feat: add instrument_asyncio() method for asyncio instrumentation

Br1an67:
## Summary

Add `logfire.instrument_asyncio()` method that combines OpenTelemetry asyncio instrumentation with Logfire's slow callback detection into a single API call.

### Changes
- New integration module `logfire/_internal/integrations/asyncio_.py`
- `instrument_asyncio(slow_duration=0.1)` method on `Logfire` class
- Uses `AsyncioInstrumentor` for coroutine/future/to_thread tracing
- Uses `log_slow_async_callbacks()` for event loop blocking detection
- Optional dependency: `pip install 'logfire[asyncio]'`
- Updated docs page and integration index table
- Added `logfire-api` stubs

Closes #939

Issue #939, opened 2025-03-20
Adding support for opentelemetry-instrumentation-asyncio


tyong920:
### Description

Currently, the project lacks support [Correct me if i'm wrong] for opentelemetry-instrumentation-asyncio, which would allow better tracing and monitoring of asyncio-based operations. Integrating this instrumentation would enhance observability, making it easier to diagnose performance bottlenecks and errors in async workflows.

Official package: [opentelemetry-instrumentation-asyncio](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-asyncio)

alexmojaki:
Pretty much any `opentelemetry-instrumentation-*` will just work as described with logfire activated. Using the first example from their docs:

```python
import asyncio
import os

import logfire
from opentelemetry.instrumentation.asyncio import AsyncioInstrumentor

logfire.configure()
os.environ['OTEL_PYTHON_ASYNCIO_COROUTINE_NAMES_TO_TRACE'] = 'sleep'

AsyncioInstrumentor().instrument()


async def main():
    await asyncio.create_task(asyncio.sleep(0.1))


asyncio.run(main())
```

![Image](https://github.com/user-attachments/assets/b298dcda-2c7c-4fd9-883f-f19ab157c7c2)

Also check out https://logfire.pydantic.dev/docs/reference/api/logfire/#logfire.Logfire.log_slow_async_callbacks . I see we need to make this more discoverable in docs.

tyong920:
Appreciate the info—I’ll check it out!

Br1an67:
I'd like to work on this. Added a docs page covering OpenTelemetry asyncio instrumentation setup and `logfire.log_slow_async_callbacks()`.

PR: #1745

PR #2186, opened 2026-08-04
Add `log_event_loop_pauses`: detect event loop pauses on any loop implementation, including uvloop

dmontagu:
`logfire.log_slow_async_callbacks()` works by patching `asyncio.events.Handle._run`, which uvloop never calls (its callback handles are implemented in Cython), so under uvloop it silently detects nothing.

This adds `logfire.log_event_loop_pauses(slow_duration=0.1, check_interval=None)`, which detects event loop pauses without depending on the loop implementation:

- A repeating heartbeat callback on the loop records when it last ran.
- A daemon thread watches the heartbeat. When it falls more than `slow_duration` behind schedule, the loop is blocked right now, and the thread samples the loop thread's stack (via `sys._current_frames()`) to capture what is blocking it.
- Once the loop recovers, a warning is logged with the measured pause duration and the sampled stack, in the same `logfire.asyncio` scope as `log_slow_async_callbacks`:

  > Event loop blocked for 0.532 seconds

Because it measures loop responsiveness rather than timing each callback, it also catches pauses that aren't caused by a single slow callback: garbage collection, other threads holding the GIL, CPU starvation. The trade-off is that attribution is a best-effort stack sample rather than exact callback timings, so on a stock asyncio loop `log_slow_async_callbacks` still gives better attribution; the two compose fine and use the same default threshold.

Notes:

- It must be called while the loop to monitor is running (it needs the loop to schedule the heartbeat on) and monitors only that loop. Calling it with no running loop raises `RuntimeError`.
- A pause is logged once the loop recovers from it, so a permanently blocked loop logs nothing. If that case matters we could also log "blocked for at least X seconds" while still blocked; left out for now to keep this simple.
- Overhead is one timer callback plus one watchdog thread wake per `check_interval` (default `slow_duration / 4`, i.e. 40/s at the default threshold).
- uvloop is added as a dev dependency (non-Windows only) so the test suite covers the uvloop case with a real uvloop event loop.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2186?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/logfire/_internal/async_.py
+++ b/logfire/_internal/async_.py
@@ -4,6 +4,9 @@ import asyncio
 import asyncio.events
 import asyncio.tasks
 import inspect
+import sys
+import threading
+import time
 from contextlib import AbstractContextManager, contextmanager
 from pathlib import Path
 from types import CoroutineType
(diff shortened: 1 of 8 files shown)


logfire/_internal/async_.py
```diff
@@ -120,3 +123,119 @@ def _callback_attributes(callback: Any) -> _CallbackAttributes:
...
+def log_event_loop_pauses(
+    logfire: Logfire, slow_duration: float, check_interval: float | None
+) -> AbstractContextManager[None]:
+    """Log a warning whenever the current event loop is unresponsive for too long.
+
+    See Logfire.log_event_loop_pauses.
+    """
+    loop = asyncio.get_running_loop()
+    if check_interval is None:
+        check_interval = slow_duration / 4
```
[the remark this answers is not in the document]
dmontagu: Fixed in 05f457d61c74e810fb53d82713968bfee3bb4b7e: both arguments are now validated as positive before the watchdog starts, with a test covering both `ValueError`s.

alexmojaki (changes requested):
I like the general idea, but

1. I don't like the API. It seems modeled after log_slow_async_callbacks. That's from very early in SDK history, before the `instrument_*` convention, and it probably wouldn't be named that nowadays. https://github.com/pydantic/logfire/pull/1745 proposes folding it into a new `instrument_asyncio`.
2. I worry about the overhead or other possible negatives that are hard to spot in a code review.

I think we should try this out ourselves first before exporting it at the top level, so either:

1. Implement this in our backend first, since I don't see anything here at a glance that *needs* SDK changes.
2. Or put this under the experimental package.

Once it's been dogfooded then we can worry more about details and API design.

logfire/_internal/async_.py
```diff
@@ -120,3 +123,123 @@ def _callback_attributes(callback: Any) -> _CallbackAttributes:
...
+    See Logfire.log_event_loop_pauses.
+    """
+    if slow_duration <= 0:
+        raise ValueError('slow_duration must be positive')
+    if check_interval is not None and check_interval <= 0:
+        raise ValueError('check_interval must be positive')
+    loop = asyncio.get_running_loop()
+    if check_interval is None:
+        check_interval = slow_duration / 4
+    logfire = logfire.with_settings(custom_scope_suffix='asyncio')
```
alexmojaki, line 142: probably not this scope suffix since this goes beyond asyncio

dmontagu:
Agreed on all points. I have moved this into our backend to dogfood it first (pydantic/platform#30951): same mechanism, started from `configure_observability` with an env-var kill switch, and using a `logfire.event_loop` scope instead of `logfire.asyncio` per the inline comment, since it catches GC/GIL pauses too.

Closing this PR. Once we have prod data on overhead and usefulness, we can revisit upstreaming it, presumably folded into `instrument_asyncio` (#939 / #1745) rather than as another `log_*` method.

logfire/_internal/async_.py
dmontagu: Agreed; the internal implementation uses a `logfire.event_loop` scope instead.

PR #2187, opened 2026-08-04
Scrub spans and logs lazily in the exporter thread

Kludex:
Scrubbing ran in `MainSpanProcessorWrapper.on_end` and `MainLogProcessorWrapper.on_emit`, i.e. in whichever thread ended the span or emitted the log - often an application's event loop.

Spans and log records are now wrapped in `LazilyScrubbedReadableSpan` / `LazilyScrubbedLogRecord`, which scrub on first read of their contents. That read happens inside `SpanExporter.export()`, which for the Logfire exporter runs on the batch processor's background thread. The result is cached under a lock, so it's computed once no matter how many exporters read the span. Every exporter still receives scrubbed data, including user-provided `additional_span_processors` - only the timing moved.

## Measurements

Time spent inside `logfire.info()` - the part that runs on the event loop - with a 200-item nested payload, 50 calls:

| | mean | p95 | max |
|---|---|---|---|
| before | 7.2ms | 7.8ms | 13.3ms |
| after | 2.4ms | 2.7ms | 3.8ms |

Tagging each scrubbing callback invocation with `threading.current_thread().name` confirms it moved from `MainThread` to `OtelBatchSpanRecordProcessor`.

**Caveat:** measuring event-loop heartbeat lag instead gives 12.1ms after vs 12.9ms before - essentially unchanged. Scrubbing is pure Python, so on the worker thread it still holds the GIL. What changed is that the work is now preemptible at `sys.getswitchinterval()` granularity instead of one monolithic block, and it's off the critical path of the coroutine that logged. For an app genuinely CPU-bound on scrubbing this redistributes the cost rather than eliminating it.

## Notes

- With `SimpleSpanProcessor` (the console exporter, `TestExporter`) the read still happens in the calling thread - unavoidable, since those export synchronously.
- With `sampling=SamplingOptions(tail=...)`, `TailSamplingProcessor` reads `span.attributes` to get the level, which triggers the scrub inline again. Not addressed here.
- An exporter that never reads `span.attributes` now skips scrubbing entirely. The real OTLP path serializes every attribute, so it's unaffected.
- `force_flush()` scrubs on whatever thread calls it, since the pending batch is drained there.
- The scrubbing callback now runs on a background thread, so it can't rely on the state of the thread that created the span. Documented in `docs/how-to-guides/scrubbing.md`.
- One assertion in `test_otel_logs.py` compared the log record by identity; it now compares body and attributes.

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/how-to-guides/scrubbing.md
+++ b/docs/how-to-guides/scrubbing.md
@@ -82,6 +82,9 @@ def scrubbing_callback(match: logfire.ScrubMatch):
 logfire.configure(scrubbing=logfire.ScrubbingOptions(callback=scrubbing_callback))
 ```
 
+Scrubbing runs when a span or log is about to be sent, which usually happens on a background thread, so
+the callback shouldn't rely on the state of the thread that created the span, e.g. context variables.
+
 ## Security tips
 
 ### Use message templates
--- a/logfire/_internal/exporters/logs.py
+++ b/logfire/_internal/exporters/logs.py
@@ -4,7 +4,7 @@ from opentelemetry.sdk._logs import ReadWriteLogRecord
 
 import logfire
 from logfire._internal.exporters.wrapper import WrapperLogProcessor
-from logfire._internal.scrubbing import BaseScrubber
+from logfire._internal.scrubbing import BaseScrubber, LazilyScrubbedLogRecord
 from logfire._internal.utils import is_instrumentation_suppressed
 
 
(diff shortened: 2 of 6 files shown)


Kludex:
It doesn't seem this solves the problem I was trying to solve.

PR #2189, opened 2026-08-04
Fix metric resource PID after fork

alexmojaki:
## Summary

- update metric, span, and log resources with the child PID after `os.fork()`
- perform the updates without taking OpenTelemetry locks or rerunning its process-dependent resource detectors
- snapshot OpenTelemetry's private resource updater implementations so dependency drift fails loudly
- cover parent and child resource PIDs end to end, including a fork while provider locks are held

## Root cause

`MeterProvider` stores the resource used during collection in `_sdk_config.resource`. Assigning `meter_provider._resource` created an unused attribute, so child metrics continued to export the parent process PID. Cached SDK tracer and logger instances also retain their original resource and need to be updated after the fork.

Logfire's child callback runs before OpenTelemetry resets inherited provider locks. It therefore updates the same backing state as OpenTelemetry's private `_update_resource()` methods without calling those lock-taking methods or `_handle_fork()`, which also reruns resource detection.

## Validation

- `uv run pytest tests/test_configure.py` — 106 passed
- focused regression tests with OpenTelemetry SDK 1.44.0 — 3 passed
- focused regression tests with the minimum supported OpenTelemetry SDK 1.39.0 — 2 passed, 1 expected skip
- `uv run pyright logfire/_internal/config.py tests/test_configure.py`
- Ruff formatting and lint checks

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -982,47 +982,30 @@ def _register_at_fork_resource_updates(
 
         pid_resource = Resource({'process.pid': os.getpid()})
 
-        # This callback is registered before the OpenTelemetry providers' callbacks, so reset their locks before
-        # calling their lock-taking resource updaters. Their later callbacks can safely repeat the reset.
+        # This callback runs before OpenTelemetry resets its locks, so update the same backing state as its
+        # lock-taking `_update_resource` methods without acquiring inherited locks. Tests snapshot those methods
+        # so that changes to these private implementation details are caught when OpenTelemetry is upgraded.
         tracer_provider = proxy_tracer_provider.provider
         if isinstance(tracer_provider, SDKTracerProvider):
-            handle_fork = getattr(tracer_provider, '_handle_fork', None)
-            update_resource = getattr(tracer_provider, '_update_resource', None)
-            if handle_fork and update_resource:
-                handle_fork()
-                update_resource(pid_resource)
-            else:
-                new_resource = tracer_provider.resource.merge(pid_resource)
-                tracer_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
-                for proxy_tracer in proxy_tracer_provider.tracers:
-                    if isinstance(proxy_tracer.tracer, SDKTracer):
-                        proxy_tracer.tracer.resource = new_resource
+            new_resource = tracer_provider.resource.merge(pid_resource)
+            tracer_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
+            for proxy_tracer in proxy_tracer_provider.tracers:
+                if isinstance(proxy_tracer.tracer, SDKTracer):
+                    proxy_tracer.tracer.resource = new_resource
 
         meter_provider = proxy_meter_provider.provider
         if isinstance(meter_provider, MeterProvider):
-            handle_fork = getattr(meter_provider, '_handle_fork', None)
-            update_resource = getattr(meter_provider, '_update_resource', None)
-            if handle_fork and update_resource:
-                handle_fork()
-                update_resource(pid_resource)
-            else:
-                meter_provider._sdk_config.resource = meter_provider._sdk_config.resource.merge(  # pyright: ignore[reportPrivateUsage]
-                    pid_resource
-                )
+            meter_provider._sdk_config.resource = meter_provider._sdk_config.resource.merge(  # pyright: ignore[reportPrivateUsage]
+                pid_resource
+            )
 
         logger_provider = proxy_logger_provider.provider
         if isinstance(logger_provider, SDKLoggerProvider):
-            handle_fork = getattr(logger_provider, '_handle_fork', None)
-            update_resource = getattr(logger_provider, '_update_resource', None)
-            if handle_fork and update_resource:
-                handle_fork()
-                update_resource(pid_resource)
-            else:
-                new_resource = logger_provider.resource.merge(pid_resource)
-                logger_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
-                for proxy_logger in proxy_logger_provider.loggers:
-                    if isinstance(proxy_logger.logger, SDKLogger):
-                        proxy_logger.logger._resource = new_resource  # pyright: ignore[reportPrivateUsage]
+            new_resource = logger_provider.resource.merge(pid_resource)
+            logger_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
+            for proxy_logger in proxy_logger_provider.loggers:
+                if isinstance(proxy_logger.logger, SDKLogger):
+                    proxy_logger.logger._resource = new_resource  # pyright: ignore[reportPrivateUsage]
 
     os.register_at_fork(after_in_child=fix_pid)
 
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -982,9 +982,14 @@ def _register_at_fork_resource_updates(
 
             pid_resource = Resource({'process.pid': os.getpid()})
 
+            # This callback is registered before the OpenTelemetry providers' callbacks, so reset their locks before
+            # calling their lock-taking resource updaters. Their later callbacks can safely repeat the reset.
             tracer_provider = proxy_tracer_provider.provider
             if isinstance(tracer_provider, SDKTracerProvider):
-                if update_resource := getattr(tracer_provider, '_update_resource', None):
+                handle_fork = getattr(tracer_provider, '_handle_fork', None)
+                update_resource = getattr(tracer_provider, '_update_resource', None)
+                if handle_fork and update_resource:
+                    handle_fork()
                     update_resource(pid_resource)
                 else:
                     new_resource = tracer_provider.resource.merge(pid_resource)
@@ -995,7 +1000,10 @@ def _register_at_fork_resource_updates(
 
             meter_provider = proxy_meter_provider.provider
             if isinstance(meter_provider, MeterProvider):
-                if update_resource := getattr(meter_provider, '_update_resource', None):
+                handle_fork = getattr(meter_provider, '_handle_fork', None)
+                update_resource = getattr(meter_provider, '_update_resource', None)
+                if handle_fork and update_resource:
+                    handle_fork()
                     update_resource(pid_resource)
                 else:
                     meter_provider._sdk_config.resource = meter_provider._sdk_config.resource.merge(  # pyright: ignore[reportPrivateUsage]
@@ -1004,7 +1012,10 @@ def _register_at_fork_resource_updates(
 
             logger_provider = proxy_logger_provider.provider
             if isinstance(logger_provider, SDKLoggerProvider):
-                if update_resource := getattr(logger_provider, '_update_resource', None):
+                handle_fork = getattr(logger_provider, '_handle_fork', None)
+                update_resource = getattr(logger_provider, '_update_resource', None)
+                if handle_fork and update_resource:
+                    handle_fork()
                     update_resource(pid_resource)
                 else:
                     new_resource = logger_provider.resource.merge(pid_resource)
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -972,57 +972,57 @@ def _register_at_fork_resource_updates(
     proxy_meter_provider_ref = weakref.ref(proxy_meter_provider)
     proxy_logger_provider_ref = weakref.ref(proxy_logger_provider)
 
+    @handle_internal_errors
     def fix_pid():
-        with handle_internal_errors:
-            proxy_tracer_provider = proxy_tracer_provider_ref()
-            proxy_meter_provider = proxy_meter_provider_ref()
-            proxy_logger_provider = proxy_logger_provider_ref()
-            if not proxy_tracer_provider or not proxy_meter_provider or not proxy_logger_provider:
-                return
+        proxy_tracer_provider = proxy_tracer_provider_ref()
+        proxy_meter_provider = proxy_meter_provider_ref()
+        proxy_logger_provider = proxy_logger_provider_ref()
+        if not proxy_tracer_provider or not proxy_meter_provider or not proxy_logger_provider:
+            return
 
-            pid_resource = Resource({'process.pid': os.getpid()})
+        pid_resource = Resource({'process.pid': os.getpid()})
 
-            # This callback is registered before the OpenTelemetry providers' callbacks, so reset their locks before
-            # calling their lock-taking resource updaters. Their later callbacks can safely repeat the reset.
-            tracer_provider = proxy_tracer_provider.provider
-            if isinstance(tracer_provider, SDKTracerProvider):
-                handle_fork = getattr(tracer_provider, '_handle_fork', None)
-                update_resource = getattr(tracer_provider, '_update_resource', None)
-                if handle_fork and update_resource:
-                    handle_fork()
-                    update_resource(pid_resource)
-                else:
-                    new_resource = tracer_provider.resource.merge(pid_resource)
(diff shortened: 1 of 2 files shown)


logfire/_internal/config.py
```diff
@@ -1452,9 +1452,13 @@ def check_tokens():
...
                     with handle_internal_errors:
-                        new_resource = resource.merge(Resource({'process.pid': os.getpid()}))
+                        pid_resource = Resource({'process.pid': os.getpid()})
+                        new_resource = resource.merge(pid_resource)
                         tracer_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
-                        meter_provider._resource = new_resource  # pyright: ignore[reportAttributeAccessIssue]
+                        if isinstance(meter_provider, MeterProvider):
+                            meter_provider._sdk_config.resource = meter_provider._sdk_config.resource.merge(  # pyright: ignore[reportPrivateUsage]
+                                pid_resource
+                            )
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 2b3ae42f. The fork test now covers metrics, spans, and logs, including the OpenTelemetry 1.39 compatibility path.

tests/test_configure.py
```diff
@@ -1225,6 +1226,61 @@ def test_host_and_os_resource_attributes_populated_by_default(
...
+    logger = get_logger('fork.logger')
+    logger.emit(LogRecord(body='parent'))
+    with logfire.span('parent'):
+        pass
+
+    metrics_data = metrics_reader.get_metrics_data()
+    assert metrics_data is not None
+    assert metrics_data.resource_metrics[0].resource.attributes['process.pid'] == os.getpid()
+    exporter.clear()
+    logs_exporter.clear()
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 497fab2e with direct raw-resource assertions for the parent span and log. Direct assertions preserve the exact PID comparison rather than the test exporter's normalized snapshot value.

logfire/_internal/config.py
```diff
@@ -1499,6 +1493,38 @@ def fix_pid():  # pragma: no cover
...
+                            update_resource(pid_resource)
+                        else:
+                            new_resource = tracer_provider.resource.merge(pid_resource)
+                            tracer_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
+                            for proxy_tracer in self._tracer_provider.tracers:
+                                if isinstance(proxy_tracer.tracer, SDKTracer):
+                                    proxy_tracer.tracer.resource = new_resource
+
+                        if isinstance(meter_provider, MeterProvider):
+                            meter_provider._sdk_config.resource = meter_provider._sdk_config.resource.merge(  # pyright: ignore[reportPrivateUsage]
```
alexmojaki: why not also use _update_resource here?

alexmojaki: (AI) Addressed in 015c70a8. Metrics now use `_update_resource()` when available, with the direct `_sdk_config.resource` update retained for OpenTelemetry 1.39–1.43 where that method does not exist.

logfire/_internal/config.py
```diff
@@ -1499,6 +1493,41 @@ def fix_pid():  # pragma: no cover
             )
             logger_provider = SDKLoggerProvider(resource)
             logger_provider.add_log_record_processor(root_log_processor)
+
+            if hasattr(os, 'register_at_fork'):  # pragma: no branch
```
alexmojaki: move this block outside this method

tests/test_configure.py
```diff
@@ -1225,6 +1226,63 @@ def test_host_and_os_resource_attributes_populated_by_default(
...
+    assert metrics_data.resource_metrics[0].resource.attributes['process.pid'] == os.getpid()
+    assert exporter.exported_spans[-1].resource.attributes['process.pid'] == os.getpid()
+    assert logs_exporter.get_finished_logs()[-1].resource.attributes['process.pid'] == os.getpid()
+    exporter.clear()
+    logs_exporter.clear()
+
+    read_fd, write_fd = os.pipe()
+    with warnings.catch_warnings():
+        warnings.filterwarnings('ignore', category=DeprecationWarning, message='.*fork.*')
+        child_pid = os.fork()
```
alexmojaki: is this safe in tests, especially with xdist?

logfire/_internal/config.py
```diff
@@ -1499,6 +1493,41 @@ def fix_pid():  # pragma: no cover
...
+                            tracer_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
+                            for proxy_tracer in self._tracer_provider.tracers:
+                                if isinstance(proxy_tracer.tracer, SDKTracer):
+                                    proxy_tracer.tracer.resource = new_resource
+
+                        if isinstance(meter_provider, MeterProvider):
+                            if update_resource := getattr(meter_provider, '_update_resource', None):
+                                update_resource(pid_resource)
+                            else:
+                                meter_provider._sdk_config.resource = meter_provider._sdk_config.resource.merge(  # pyright: ignore[reportPrivateUsage]
```
alexmojaki: why does this not need the same kind of proxy loop?

logfire/_internal/config.py
alexmojaki: (AI) Addressed in e6260679. Fork-resource registration and compatibility handling now live in the module-level `_register_at_fork_resource_updates()` helper.

tests/test_configure.py
alexmojaki: (AI) Verified with `uv run pytest -n 2 tests/test_configure.py -k resource_process_pid_updated_after_fork -q`; it passes under xdist. Each worker owns its pipe and forked child, the child exits via `os._exit()`, and the worker waits for that exact PID, so workers do not share this test state. The normal suite also runs with `-n logical`, so CI exercises this mode.

logfire/_internal/config.py
alexmojaki: (AI) The metric reader storage retains the same mutable `_sdk_config` object and reads `_sdk_config.resource` at collection time, so changing that field is visible to existing meters/readers. SDK tracer and logger objects instead cache the `Resource` object itself when created, which is why the pre-1.44 fallback has to update those existing proxy-backed objects explicitly.

logfire/_internal/config.py
```diff
@@ -955,6 +960,49 @@ def _load_configuration(
...
+    tracer_provider: SDKTracerProvider,
+    meter_provider: MeterProvider | NoOpMeterProvider,
+    logger_provider: SDKLoggerProvider,
+    proxy_tracer_provider: ProxyTracerProvider,
+    proxy_logger_provider: ProxyLoggerProvider,
+) -> None:
+    if not hasattr(os, 'register_at_fork'):  # pragma: no cover
+        return
+
+    def fix_pid():  # pragma: no cover
```
alexmojaki: does this still need the pragma

alexmojaki: (AI) Yes. The callback only runs in the forked child, and the regression test deliberately terminates that child with `os._exit()`, so its inherited coverage state is not flushed or merged into the parent coverage data. The end-to-end behavior is tested, but these lines still appear uncovered to coverage.py.

alexmojaki: add a test which patches os.register_at_fork and os.getpid so that this function gets called in the main process and can have its coverage tested. keep the test which actually forks.

logfire/_internal/config.py
```diff
@@ -955,6 +960,49 @@ def _load_configuration(
...
+        with handle_internal_errors:
+            pid_resource = Resource({'process.pid': os.getpid()})
+
+            if update_resource := getattr(tracer_provider, '_update_resource', None):
+                update_resource(pid_resource)
+            else:
+                new_resource = tracer_provider.resource.merge(pid_resource)
+                tracer_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
+                for proxy_tracer in proxy_tracer_provider.tracers:
+                    if isinstance(proxy_tracer.tracer, SDKTracer):
```
alexmojaki: why check this

logfire/_internal/config.py
alexmojaki: (AI) Addressed in 2b27096e. The new test captures `os.register_at_fork`, patches `os.getpid`, invokes the callback in the main process for coverage, and asserts updated metric/span/log resources. The separate real-fork test remains.

logfire/_internal/config.py
alexmojaki: (AI) `proxy_tracer.tracer` is typed as the API `Tracer` and may be an SDK tracer, `SuppressedTracer`, or another no-op implementation. Only `SDKTracer` has the cached mutable `resource` attribute that needs updating; the guard avoids assigning an SDK implementation detail onto suppressed/no-op tracers.

logfire/_internal/config.py
```diff
@@ -955,6 +960,54 @@ def _load_configuration(
         self.metrics = metrics
 
 
+def _register_at_fork_resource_updates(
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 568e11a5. Each `LogfireConfig` now registers one callback during initialization. The callback retains only weak references to the stable proxy providers and resolves their current SDK providers at fork time, so repeated reconfiguration no longer registers provider-capturing callbacks or retains old provider/exporter graphs.

logfire/_internal/config.py
```diff
@@ -955,6 +960,53 @@ def _load_configuration(
...
+            proxy_meter_provider = proxy_meter_provider_ref()
+            proxy_logger_provider = proxy_logger_provider_ref()
+            if not proxy_tracer_provider or not proxy_meter_provider or not proxy_logger_provider:
+                return
+
+            pid_resource = Resource({'process.pid': os.getpid()})
+
+            tracer_provider = proxy_tracer_provider.provider
+            if isinstance(tracer_provider, SDKTracerProvider):
+                new_resource = tracer_provider.resource.merge(pid_resource)
```
alexmojaki: what happened to the _update_resource calls

alexmojaki: (AI) `_update_resource()` was removed because it was only added in OpenTelemetry 1.44, while Logfire supports 1.39–1.44. The replacement updates the backing resource storage used across that full range (`_resource` for traces/logs and `_sdk_config.resource` for metrics), plus the already-cached tracer/logger resources that provider-only updates don't change.

alexmojaki: (AI) Addressed in 418774a9

logfire/_internal/config.py
```diff
@@ -955,6 +960,62 @@ def _load_configuration(
...
+            proxy_logger_provider = proxy_logger_provider_ref()
+            if not proxy_tracer_provider or not proxy_meter_provider or not proxy_logger_provider:
+                return
+
+            pid_resource = Resource({'process.pid': os.getpid()})
+
+            tracer_provider = proxy_tracer_provider.provider
+            if isinstance(tracer_provider, SDKTracerProvider):
+                if update_resource := getattr(tracer_provider, '_update_resource', None):
+                    update_resource(pid_resource)
```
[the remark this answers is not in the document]
alexmojaki: (AI) Tracking as `## avoid-pre-reinit-provider-locks`:

The callback ordering is as described. I'll keep OpenTelemetry 1.44's lock-taking update behind its own lock-reinitializing fork callback, preserve the lock-free 1.39–1.43 fallback, and add regression coverage for the ordering/version paths.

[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 8ce72fe3

logfire/_internal/config.py
```diff
@@ -955,6 +960,62 @@ def _load_configuration(
...
+                else:
+                    new_resource = tracer_provider.resource.merge(pid_resource)
+                    tracer_provider._resource = new_resource  # pyright: ignore[reportPrivateUsage]
+                    for proxy_tracer in proxy_tracer_provider.tracers:
+                        if isinstance(proxy_tracer.tracer, SDKTracer):
+                            proxy_tracer.tracer.resource = new_resource
+
+            meter_provider = proxy_meter_provider.provider
+            if isinstance(meter_provider, MeterProvider):
+                if update_resource := getattr(meter_provider, '_update_resource', None):
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 8ce72fe3

tests/test_configure.py
```diff
@@ -1225,6 +1227,195 @@ def test_host_and_os_resource_attributes_populated_by_default(
...
+        provider_locks = [
+            tracer_provider._tracers_lock,  # pyright: ignore[reportPrivateUsage]
+            meter_provider._meter_lock,  # pyright: ignore[reportPrivateUsage]
+            logger_provider._active_loggers_lock,  # pyright: ignore[reportPrivateUsage]
+        ]
+        for lock in provider_locks:
+            lock.acquire()
+
+    parent_pid = os.getpid()
+    signal.alarm(10)
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 2dea8857

logfire/_internal/config.py
```diff
@@ -955,6 +960,73 @@ def _load_configuration(
...
+                        pid_resource
+                    )
+
+            logger_provider = proxy_logger_provider.provider
+            if isinstance(logger_provider, SDKLoggerProvider):
+                handle_fork = getattr(logger_provider, '_handle_fork', None)
+                update_resource = getattr(logger_provider, '_update_resource', None)
+                if handle_fork and update_resource:
+                    handle_fork()
+                    update_resource(pid_resource)
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 9f236767. The fork callback now always performs lock-free backing-state updates and no longer calls either `_handle_fork()` or `_update_resource()`. The tests also snapshot the three upstream `_update_resource()` implementations so future drift in these private assumptions fails loudly.

logfire/_internal/config.py
```diff
@@ -955,6 +960,73 @@ def _load_configuration(
...
+
+            # This callback is registered before the OpenTelemetry providers' callbacks, so reset their locks before
+            # calling their lock-taking resource updaters. Their later callbacks can safely repeat the reset.
+            tracer_provider = proxy_tracer_provider.provider
+            if isinstance(tracer_provider, SDKTracerProvider):
+                handle_fork = getattr(tracer_provider, '_handle_fork', None)
+                update_resource = getattr(tracer_provider, '_update_resource', None)
+                if handle_fork and update_resource:
+                    handle_fork()
+                    update_resource(pid_resource)
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 9f236767. Logfire no longer calls `_handle_fork()`, so the callback no longer creates a `ThreadPoolExecutor` or relies on the concurrent-futures at-fork handler having run first.

tests/test_configure.py
```diff
@@ -1225,6 +1228,238 @@ def test_host_and_os_resource_attributes_populated_by_default(
...
+    callback()
+
+
+def test_otel_resource_updater_sources() -> None:
+    provider_classes = {
+        'traces': config_module.SDKTracerProvider,
+        'metrics': config_module.MeterProvider,
+        'logs': config_module.SDKLoggerProvider,
+    }
+    if not all(hasattr(provider_class, '_update_resource') for provider_class in provider_classes.values()):
```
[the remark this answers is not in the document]
alexmojaki: (AI) Thanks, no action needed. The exact-source brittleness is intentional here: this code manually mirrors private OpenTelemetry implementation details, so even a cosmetic upstream edit should force a human to re-check and explicitly refresh the snapshot rather than allowing drift silently. OpenTelemetry <1.44 has no `_update_resource()` implementation to inspect; those supported versions remain covered by both the callback test and the real-fork behavioral regression test (including the dedicated 1.39 CI job).

PR #2207, opened 2026-08-05
Add single source of truth for agent setup prompt and revise content [murmur:logfire-sdk/agent-setup-prompt-revamp]

### LINKED ISSUE #2201: Revise the agent setup prompt: auth flow, browser guidance, project creation, MCP-first verification, next steps
We QA-tested the agent setup prompt (the `<AgentSetup>` / `<CopyPrompt>` block in `docs/index.md` and `docs/first-trace.md`) by running it through coding agents (mainly Codex) against real projects. It works, but several observed behaviors suggest revisions:

**Observed problems**

- The agent starts by opening browser tooling to read the docs links in the prompt: slow and expensive. It should fetch them as text (or the prompt should inline the essential steps).
- The prompt has the agent run `logfire auth`. That opens the *user's* browser, which is good, but the agent does not understand this: it opened its own browser tool and asked the user to log in there instead. It needs explicit guidance about what happens and how to wait (poll for credentials to appear rather than driving a browser).
- The agent created a new Logfire project without asking. Probably not the default we want.
- The agent spent a long time trying to hand-roll trace queries to verify the setup (reading SDK internals to find a query client, considering read tokens). The MCP server makes this easy, and the prompt only mentions it as a later optional step.
- It committed directly to the user's current branch.
- Suggesting evals as a next step is odd in general, and especially when the project has no AI in it.

**Proposed revisions**

1. Open with a transparency preamble along the lines of "This prompt was copied from the Logfire docs; don't assume the user has read it", so the agent frames its interaction correctly.
2. Tell the agent to fetch referenced guides as plain text (e.g. `curl` a `.txt` URL) instead of using browser tools. (A stable text endpoint for the guide is being considered separately.)
3. Auth: prefer having the *user* run the first auth commands themselves (`uvx logfire auth` works regardless of the project's language and tooling); if the agent runs it, explain that a browser tab opens for the user, and tell the agent to wait/poll for `~/.logfire` credentials instead of opening its own browser.
4. Do not create a new project without asking the user first.
5. Move MCP setup earlier and make verification go through it; explicitly discourage hand-rolled trace querying without the MCP.
6. Have the agent create a new branch before committing.
7. Rework the suggested next steps: standard dashboards, auto-tracing (Python), system metrics instrumentation (with links to the corresponding UI pages), optionally manual tracing, and configuring a `service_name` plus a pointer to the Services page. Drop evals as a default suggestion (or condition it on the project actually containing AI).
8. Since we suggest the MCP anyway, also suggest installing the Logfire skills/plugins, and point the agent at the bundled `logfire-instrumentation` skill (shipped in the `logfire` package under `.agents/skills/`), which is much more detailed than the prompt itself.

The same prompt is duplicated on the marketing site; a companion issue covers deduplicating it, but any wording change here should be propagated to all copies in the meantime.

### LINKED ISSUE #2202: Single source of truth for the agent setup prompt (3 verbatim copies today)
The agent setup prompt exists verbatim in three places today:

1. `docs/index.md` (inside `<AgentSetup>`)
2. `docs/first-trace.md` (inside `<CopyPrompt>`)
3. The marketing site homepage (a copy of the same text in the site's repo)

They will drift. Proposal, in increasing order of effort:

1. **Minimum (this repo)**: a test asserting the prompt block in `docs/index.md` and `docs/first-trace.md` are byte-identical, so a change to one fails CI until the other is updated.
2. **Single source in this repo**: move the prompt text to one canonical file (e.g. `docs/agent-setup-prompt.txt`) and include it from both pages via the docs build (mkdocs snippets/include, or whatever mechanism the docs pipeline supports), so there is exactly one copy to edit.
3. **Stable text endpoint**: publish the canonical prompt at a stable `.txt` URL on the docs site. This has a second benefit: QA showed coding agents open browser tooling to read the docs links inside the prompt, and a plain-text URL lets the prompt say "fetch this with curl" instead. The marketing site can then either fetch it at build time or keep its copy with a CI drift check against the published URL.

Item 1 is trivial and worth doing immediately even if 2/3 follow.

dmontagu:
Fixes #2201, fixes #2202.

## Summary

- Adds `docs/agent-setup-prompt.txt` as the single source of truth for the Logfire agent setup prompt. The docs build expands `{{ agent_setup_prompt() }}` in both the home page and the first-trace guide, so the two copies cannot drift.

- Reworks the prompt into an end-to-end setup flow that tells coding agents to:
  - protect existing work and create a setup branch when the project uses version control;
  - detect every service and install the appropriate SDK for Python, Node.js, browsers, Cloudflare Workers, Rust, or another OpenTelemetry-supported language;
  - authenticate securely, using the local Logfire CLI flow for Python or `LOGFIRE_TOKEN` where appropriate, without exposing or committing credentials;
  - fetch the instrumentation skill as plain text, then configure Logfire and instrument supported frameworks, databases, large language model clients, HTTP clients, system metrics, and standard logging;
  - run the application, provide the relevant Live View URL, and ask the user to confirm traces rather than trying to operate the Logfire UI itself; and
  - offer the matching Logfire plugins, skills, and Model Context Protocol (MCP) server after the initial setup works.

- Adds tests that verify the canonical prompt exists, both public pages use the placeholder instead of hardcoded text, and the docs plugin expands the exact canonical prompt in both locations.

## Scope

This PR changes only the reusable prompt, its docs-build substitution, and the tests that keep the two public prompt locations in sync. The hand-written setup instructions elsewhere in `first-trace.md` are unchanged.

[Created via Murmur](https://cloud.murmur.dev/github_app/pydantic/task/github_app%2Fpydantic%2Fw%2Flogfire-sdk%2Fgithub_oauth%2Fdmontagu%2Fagent-setup-prompt-revamp)

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/plugins/main.py
```diff
@@ -175,6 +176,25 @@ def before_you_start(markdown: str, page: Page) -> str:
...
+def agent_setup_prompt(markdown: str, page: Page) -> str:
+    """Expand ``{{ agent_setup_prompt() }}`` with the canonical prompt text.
+
+    The canonical text lives in ``docs/agent-setup-prompt.txt`` so there is a single
+    source of truth for the agent setup prompt that appears in both ``index.md``
+    (inside ``<AgentSetup>``) and ``first-trace.md`` (inside ``<CopyPrompt>``).
+    """
+    if page.file.src_uri not in _AGENT_SETUP_PROMPT_PAGES:
+        return markdown
+    if '{{ agent_setup_prompt() }}' not in markdown:
```
alexmojaki: extract `'{{ agent_setup_prompt() }}'` to a constant

docs/plugins/main.py
```diff
@@ -175,6 +176,25 @@ def before_you_start(markdown: str, page: Page) -> str:
...
+
+def agent_setup_prompt(markdown: str, page: Page) -> str:
+    """Expand ``{{ agent_setup_prompt() }}`` with the canonical prompt text.
+
+    The canonical text lives in ``docs/agent-setup-prompt.txt`` so there is a single
+    source of truth for the agent setup prompt that appears in both ``index.md``
+    (inside ``<AgentSetup>``) and ``first-trace.md`` (inside ``<CopyPrompt>``).
+    """
+    if page.file.src_uri not in _AGENT_SETUP_PROMPT_PAGES:
+        return markdown
```
alexmojaki: drop this, the check below is fine

docs/agent-setup-prompt.txt
```diff
@@ -0,0 +1,34 @@
+Note for the agent: This prompt was copied from the Logfire docs by the user. The user may not have read it. Briefly tell the user what you plan to do before starting.
+
+Set up Pydantic Logfire in this project so it sends traces to Logfire.
+
+Before any code changes, create a new git branch.
+
+If the logfire-instrumentation skill is available in your agent environment, read it now for detailed per-language instrumentation steps. For a Python project, find it in the installed package with:
+  cat "$(python -c 'import pathlib, logfire; print(pathlib.Path(logfire.__file__).parent / ".agents/skills/logfire-instrumentation/SKILL.md")')"
```
alexmojaki: this needs to happen after installing logfire

alexmojaki: (AI) Addressed in 5f726d42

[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in e5901524

docs/plugins/main.py
alexmojaki: (AI) Addressed in 76845d71

docs/plugins/main.py
alexmojaki: (AI) Addressed in dfd96ec2

docs/plugins/main.py
```diff
@@ -175,6 +176,25 @@ def before_you_start(markdown: str, page: Page) -> str:
...
+    if '{{ agent_setup_prompt() }}' not in markdown:
+        return markdown
+    prompt_text = _AGENT_SETUP_PROMPT_FILE.read_text(encoding='utf-8')
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in f6ef5617

PR #2245, opened 2026-08-13
fix: validate extra scrubbing patterns

### LINKED ISSUE #2234: ScrubbingOptions: a bare string or empty-matching regex in extra_patterns silently redacts all telemetry
> This issue was posted by Claude Code using claude-opus-5 on behalf of David.

### Description

`ScrubbingOptions.extra_patterns` is typed `Sequence[str] | None`. Because `str` satisfies `Sequence[str]`, passing a bare string type-checks cleanly under pyright and is then iterated character by character, producing a single-character alternation that matches almost any text.

```python
from logfire._internal.scrubbing import Scrubber

# intended: extra_patterns=['password']
Scrubber('password').scrub_value(('attributes', 'x'), 'the quick brown fox')
# -> "[Scrubbed due to 'r']"
```

The compiled pattern becomes `p|a|s|s|w|o|r|d`, so essentially all telemetry for that process is replaced with `[Scrubbed due to '<letter>']`. There is no error and no warning, and `make typecheck` does not flag the call site.

A closely related case is a pattern that matches the empty string, which has the same "redact everything" outcome with an empty reason string:

```python
Scrubber([''])       .scrub_value(('attributes', 'x'), 'totally harmless log line')  # "[Scrubbed due to '']"
Scrubber(['[0-9]*']) .scrub_value(('attributes', 'x'), 'totally harmless log line')  # "[Scrubbed due to '']"
Scrubber([r'\b'])    .scrub_value(('attributes', 'x'), 'totally harmless log line')  # "[Scrubbed due to '']"
```

`'[0-9]*'` in particular looks like a reasonable thing to write when you mean `'[0-9]+'`.

### Suggested fix

Validate in `ScrubbingOptions.__post_init__` and raise `LogfireConfigError`:

- reject a bare `str` for `extra_patterns`, naming the `['...']` form in the message
- reject any pattern where `re.compile(p).match('') is not None`

Compiling each pattern individually before joining would also make invalid-regex errors attributable. Today an invalid entry surfaces as a `re.PatternError` whose reported position is an offset into the joined pattern, e.g. `extra_patterns=['foo(']` reports `missing ), unterminated subpattern at position 262`.

### Versions

Reproduced on `logfire==4.40.0`, Python 3.13.3, macOS.

sidsri14:
## Summary

- reject a bare string passed to `ScrubbingOptions.extra_patterns` with an actionable list-form error
- validate each custom regex independently and identify malformed entries
- reject custom patterns that can produce zero-length matches before they redact unrelated telemetry
- add regression coverage for bare strings, malformed regexes, quantifiers, boundaries, and empty-only patterns

A bare string currently satisfies `Sequence[str]` and is iterated character by character. Zero-length matches are also treated as a reason to replace the complete value. Both cases can silently redact harmless telemetry.

Fixes #2234

## Testing

- `uv run --no-sync pytest tests/test_secret_scrubbing.py -q` (25 passed)
- `uv run --no-sync ruff check`
- `uv run --no-sync ruff format --check --diff`
- `uv run --no-sync pyright logfire/_internal/scrubbing.py tests/test_secret_scrubbing.py` (0 errors)
- applicable pre-commit hooks passed; the repository-wide Pyright hook was skipped on Windows after its 28 POSIX API errors were reproduced on untouched `main`
- full pytest attempt: 2247 passed, 14 skipped, 2 xfailed; 51 failures and 8 errors remained in unrelated Windows/integration paths, including unavailable Docker services, external CLI fixtures, and platform-specific snapshots

## AI assistance

Implemented with OpenAI Codex assistance. I reviewed the issue, implementation, diff, and verification results.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2245?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/logfire/_internal/scrubbing.py
+++ b/logfire/_internal/scrubbing.py
@@ -14,6 +14,8 @@ from opentelemetry.attributes import BoundedAttributes
 from opentelemetry.sdk.trace import Event
 from opentelemetry.trace import Link
 
+from logfire.exceptions import LogfireConfigError
+
 from .constants import (
     ATTRIBUTES_CONFIG,
     ATTRIBUTES_JSON_SCHEMA_KEY,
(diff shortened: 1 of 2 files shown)


sidsri14:
Closing this in favor of #2236, which was opened earlier and already covers #2234, including the zero-width behavior raised in review. Thanks to the reviewers for catching the finite-probe gap.

PR #2256, opened 2026-08-13
Document native Elasticsearch tracing

### LINKED ISSUE #1538: Elasticsearch instrumentation integration
### Description

It would be great to have integration with `elasticsearch` for auto instrumentation.

dmontagu:
## What changed

- Add an Elasticsearch integration guide for the Python client's native OpenTelemetry tracing.
- Explain that `logfire.configure()` activates the global tracer provider used by Elasticsearch 8.15 and later, with no Logfire wrapper or contrib instrumentor.
- Document query-body capture at its privacy decision point and use the client's supported `raw` value.
- Remove Elasticsearch from the CLI's legacy contrib-instrumentor map.
- Add a real offline client test using a custom `BaseNode`, and add Elasticsearch only to development dependencies so CI can exercise that behavior.

## Why

Elasticsearch 8.15 and later emit native OpenTelemetry spans through the active global tracer provider. The existing CLI map could recommend the older `opentelemetry-instrumentation-elasticsearch` package, which can duplicate the client's native spans. The integration also had no first-party setup or troubleshooting guide.

## Impact

Users can configure Elasticsearch tracing with the supported native path and verify the resulting request span in Logfire. Existing runtime dependencies and the public Logfire API are unchanged. The new Elasticsearch dependency is test-only.

## Checks

- Offline Elasticsearch integration test: current 9.5 and documented minimum 8.15
- Focused integration and CLI checks: 26 passed
- Documentation tests: 123 passed
- Full CLI tests: 195 passed
- Ruff, Ruff format, direct Pyright, lock check, and changed-file pre-commit hooks
- `git diff --check`

The new documentation was AI-assisted and then edited and verified against the current Elasticsearch client documentation and real client behavior.

Closes #1538

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2256?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/integrations/databases/elasticsearch.md
+++ b/docs/integrations/databases/elasticsearch.md
@@ -26,17 +26,17 @@ pip install 'logfire' 'elasticsearch>=8.15'
 
 ## Trace Elasticsearch requests
 
-Call `logfire.configure()` before creating the Elasticsearch client. This activates Logfire's OTel
-tracer provider, which the client's native tracing uses automatically.
+Call `logfire.configure()` before making an Elasticsearch request. The client looks up the active
+OTel tracer provider when it starts a request, so you can create the client before or after this call.
 
 ```python skip-run="true" skip-reason="external-connection"
 from elasticsearch import Elasticsearch
 
 import logfire
 
-logfire.configure()
-
 client = Elasticsearch('http://localhost:9200')
+
+logfire.configure()
 client.search(index='products', query={'match': {'name': 'coffee'}})
 ```
 
@@ -51,10 +51,18 @@ export OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_CAPTURE_SEARCH_QUERY=raw
 ```
 
 !!! warning
-    Setting this option to `raw` can send query values to Logfire, where they are stored with the
-    span. Review the values your application puts in Elasticsearch queries before enabling it.
-    Logfire's [data scrubbing](../../how-to-guides/scrubbing.md) provides an additional safeguard,
-    but you should not rely on scrubbing to remove every sensitive value.
+    Setting this option to `raw` sends query values to Logfire, where they are stored with the span.
+    Logfire does not scrub the `db.query.text` attribute. Enable query capture only when the queries
+    do not contain sensitive values.
+
+## Stop tracing Elasticsearch requests
+
+The Elasticsearch client creates spans by default whenever an OTel tracer provider is active. To
+turn off its native tracing, set this environment variable before creating the client:
+
+```bash
+export OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_ENABLED=false
+```
 
 ## Verify the spans
 
@@ -65,8 +73,9 @@ body.
 
 ## Troubleshoot missing spans
 
-- **No Elasticsearch spans appear:** check that you use `elasticsearch` 8.15 or later, and call
-  `logfire.configure()` before creating the client.
+- **No Elasticsearch spans appear:** check that you use `elasticsearch` 8.15 or later, call
+  `logfire.configure()` before the first request, and have not set
+  `OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_ENABLED=false`.
 - **The query body is missing:** restart the application after setting
   `OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_CAPTURE_SEARCH_QUERY=raw`. Query capture is disabled by
   default.
--- a/docs/integrations/databases/elasticsearch.md
+++ b/docs/integrations/databases/elasticsearch.md
@@ -28,6 +28,8 @@ pip install 'logfire' 'elasticsearch>=8.15'
 
 Call `logfire.configure()` before making an Elasticsearch request. The client looks up the active
 OTel tracer provider when it starts a request, so you can create the client before or after this call.
+Requests made before configuration are not sent to Logfire, but requests made after configuration
+are traced normally.
 
 ```python skip-run="true" skip-reason="external-connection"
 from elasticsearch import Elasticsearch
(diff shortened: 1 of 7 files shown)


tests/otel_integrations/test_elasticsearch.py
```diff
@@ -0,0 +1,54 @@
+from elastic_transport import ApiResponseMeta, BaseNode, HttpHeaders
+from elastic_transport._node._base import NodeApiResponse
```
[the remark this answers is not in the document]
dmontagu: Fixed in d4539ed5. The offline node now implements the documented public tuple protocol with a local `NamedTuple`, so the test no longer imports `elastic_transport` private modules. The focused test passes with both Elasticsearch 8.15.1 and 9.5.0, and Ruff/Pyright are clean.

docs/integrations/databases/elasticsearch.md
```diff
@@ -0,0 +1,78 @@
...
+variable before starting your application:
+
+```bash
+export OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_CAPTURE_SEARCH_QUERY=raw
+```
+
+!!! warning
+    Setting this option to `raw` can send query values to Logfire, where they are stored with the
+    span. Review the values your application puts in Elasticsearch queries before enabling it.
+    Logfire's [data scrubbing](../../how-to-guides/scrubbing.md) provides an additional safeguard,
```
alexmojaki: scrubbing doesn't help here because the value gets stored under the attribute name db.query.text which is marked as safe

docs/integrations/databases/elasticsearch.md
```diff
@@ -0,0 +1,78 @@
...
+
+Install Logfire and Elasticsearch 8.15 or later:
+
+```bash
+pip install 'logfire' 'elasticsearch>=8.15'
+```
+
+## Trace Elasticsearch requests
+
+Call `logfire.configure()` before creating the Elasticsearch client. This activates Logfire's OTel
```
alexmojaki: calling configure after creating the client works fine

docs/integrations/databases/elasticsearch.md
```diff
@@ -0,0 +1,78 @@
...
+integration: otel
+---
+# Elasticsearch
+
+See every request your application makes to Elasticsearch, including its operation, duration,
+outcome, and target, as a **span** (one unit of work: a single operation, with a name, a start, and a
+duration) in Logfire.
+
+The Elasticsearch Python client has native **OpenTelemetry (OTel)** support from version 8.15. OTel
+is the open industry standard for collecting traces, metrics, and logs. The client creates its own
```
alexmojaki: since it just creates spans by default, we need to tell users how to opt out. if there's no native way, then we can recommend suppress_scopes like with bigquery

alexmojaki: https://www.elastic.co/docs/reference/elasticsearch/clients/python/opentelemetry#opentelemetry-config-enable

OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_ENABLED=false

docs/integrations/databases/elasticsearch.md
```diff
@@ -0,0 +1,78 @@
...
+description: "Trace requests made by the Elasticsearch Python client 8.15 and later with its native OpenTelemetry support."
+integration: otel
+---
+# Elasticsearch
+
+See every request your application makes to Elasticsearch, including its operation, duration,
+outcome, and target, as a **span** (one unit of work: a single operation, with a name, a start, and a
+duration) in Logfire.
+
+The Elasticsearch Python client has native **OpenTelemetry (OTel)** support from version 8.15. OTel
```
alexmojaki: the wording makes it sound like not all logfire is otel

dmontagu: Updated in 6f84774c. The introduction now states that Logfire is built on OpenTelemetry and that Elasticsearch uses OTel directly, avoiding the implication that OTel is only one part of Logfire.

docs/integrations/databases/elasticsearch.md
dmontagu: Updated in 68a29d98. I added the native `OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_ENABLED=false` opt-out and a regression test confirming that it suppresses the Elasticsearch span.

docs/integrations/databases/elasticsearch.md
dmontagu: Updated in 68a29d98. The warning now says explicitly that Logfire does not scrub `db.query.text` and that raw query capture should only be enabled when queries contain no sensitive values.

docs/integrations/databases/elasticsearch.md
dmontagu: Updated in 68a29d98. The docs now require `logfire.configure()` before the first request, not before client creation, and the example creates the client first to demonstrate that either order works.

docs/integrations/databases/elasticsearch.md
```diff
@@ -0,0 +1,87 @@
...
+turn off its native tracing, set this environment variable before creating the client:
+
+```bash
+export OTEL_PYTHON_INSTRUMENTATION_ELASTICSEARCH_ENABLED=false
+```
+
+## Verify the spans
+
+Run the application and open Logfire's [Live view](../../guides/web-ui/live.md). Search for
+`elasticsearch` and make a request. You should see an Elasticsearch span with the request method,
```
alexmojaki: "Search for `elasticsearch`" means nothing. i think just skip this section.

docs/integrations/databases/elasticsearch.md
```diff
@@ -0,0 +1,87 @@
...
+target, duration, and status. If query capture is enabled, the span also contains the search query
+body.
+
+## Troubleshoot missing spans
+
+- **No Elasticsearch spans appear:** check that you use `elasticsearch` 8.15 or later, call
```
alexmojaki, line 71: calling configure after the first request still works for requests after

dmontagu: Clarified this in ba664ae4: requests made before `logfire.configure()` are not sent to Logfire, while subsequent requests are traced normally. The troubleshooting text now says to configure before the specific request you want to trace.

docs/integrations/databases/elasticsearch.md
dmontagu: Removed the vague Verify section in ba664ae4.

Issue #2235, opened 2026-08-12
extra_patterns are matched on the application thread, so a nested-quantifier pattern stalls the caller


dsfaccini:
> This issue was posted by Claude Code using claude-opus-5 on behalf of David.

### Description

Patterns supplied via `ScrubbingOptions.extra_patterns` are compiled into the scrubbing regex and run over attribute values. `MessageValueCleaner.clean_value` is called during message formatting at span-creation time, so this runs **on the application's own thread**, not on the exporter thread.

A user pattern with nested quantifiers therefore turns an attribute value into a stall of the calling code:

```python
import time, logfire

logfire.configure(
    send_to_logfire=False,
    scrubbing=logfire.ScrubbingOptions(extra_patterns=[r'(a+)+$']),
)

payload = 'a' * 26 + '!'          # e.g. a field from an incoming request
t0 = time.perf_counter()
logfire.info('user said {msg}', msg=payload)
print(f'{(time.perf_counter() - t0) * 1000:.0f} ms')
```

Measured on `logfire==4.40.0`, Python 3.13.3, macOS:

| input length | time |
|---|---|
| 16 | 13 ms |
| 20 | 183 ms |
| 24 | 3.0 s |
| 26 | 17.8 s |
| 27 (through `logfire.info`) | 28.3 s |

Growth is exponential, so a slightly longer value stalls the thread indefinitely.

The 18 entries in `DEFAULT_PATTERNS` are all linear and are not affected — the exposure comes only from user-supplied patterns. But the value being matched is frequently attacker-influenced (a request field, a header, a form value), so a well-meaning pattern turns instrumentation into a denial-of-service vector against the instrumented application itself.

### Suggested fix

At minimum, document the risk on `ScrubbingOptions.extra_patterns` — nothing currently warns that user patterns run synchronously on the application thread against untrusted input.

Beyond that: rejecting obvious nested-quantifier shapes at configure time would catch the common cases cheaply. A full guarantee is not reachable with `re`.

### Versions

`logfire==4.40.0`, Python 3.13.3, macOS.

PR #2258, opened 2026-08-14
Check extra_patterns at configuration time

### LINKED ISSUE #2234: ScrubbingOptions: a bare string or empty-matching regex in extra_patterns silently redacts all telemetry

### LINKED ISSUE #2237: Numeric backreferences in extra_patterns refer to groups from other patterns, so the pattern silently never matches
> This issue was posted by Claude Code using claude-opus-5 on behalf of David.

### Description

`Scrubber.__init__` joins every pattern into a single alternation, so all patterns share one capture-group numbering space. A numeric backreference in one `extra_patterns` entry therefore refers to a group in an *earlier* entry rather than to its own:

```python
import re
from logfire._internal.scrubbing import Scrubber

Scrubber(['(a)', r'(b)\1']).scrub_value(('attributes', 'x'), 'bb')
# -> 'bb'   (not scrubbed)

re.compile(r'(b)\1').search('bb')
# -> <re.Match object; span=(0, 2), match='bb'>   (the pattern works on its own)
```

In the joined pattern, `(a)` is group 1 and `(b)` is group 2, so the `\1` in the second entry points at `(a)`, which never participated in the match. The pattern silently never matches, so data the user expected to be redacted is exported.

This is long-standing and not specific to any recent change — it reproduces identically on `main`. It only affects users with **two or more** `extra_patterns` where a later one uses a numeric backreference; a single pattern with a backreference works, because the default patterns contain no capture groups.

### Suggested fix

Neither option is free, which is why this is its own issue rather than a drive-by:

- **Match each pattern separately.** Fully isolates numbering, but replaces one `re` call per value with N, on a path that runs during span creation on the application's thread.
- **Rewrite numeric backreferences when joining**, offsetting each by the number of groups already emitted. Keeps the single-pass search, but hand-rewriting regex internals is its own source of bugs.
- **Reject numeric backreferences in `extra_patterns`** with a clear error, pointing at named groups (`(?P<x>...)` / `(?P=x)`), which are unaffected because names aren't renumbered.

The third is the cheapest and turns a silent failure into a loud one, which is the main harm here.

### Versions

Reproduced on `logfire==4.40.0`, Python 3.13.3, macOS. Found by the Macroscope reviewer on #2236.

aviseth:
`ScrubbingOptions.extra_patterns` entries are joined with the default patterns into one regex. A mistake in a single entry therefore changes the behaviour of the whole scrubber, and does so silently — either redacting every value or redacting none. This checks each pattern on its own before joining, so those cases raise `LogfireConfigError` at configuration time.

Fixes #2234. Fixes #2237.

### Cases now rejected

| Input | Current behaviour on `main` | Now |
|---|---|---|
| `extra_patterns='password'` | iterated per character into `p\|a\|s\|s\|w\|o\|r\|d`, redacting nearly every value | error naming the `['password']` form |
| `extra_patterns=['[0-9]*']` | matches every value, reason string empty | error |
| `extra_patterns=['(a)', r'(b)\1']` | `\1` refers to `(a)`, so the pattern never matches and the data is exported | error pointing at named groups |
| `extra_patterns=['foo(']` | `missing ), unterminated subpattern at position 281` | same message at position 3 |

The first three are silent, which is what makes them worth an error rather than a warning: the first two destroy all telemetry for the process, and the third exports data the user asked to have redacted.

Named groups (`(?P<name>...)` / `(?P=name)`) are unaffected by the joining because names aren't renumbered, so they're what the backreference message recommends, and there's a test covering that they still work.

### Notes for review

- Validation lives in `Scrubber.__init__` rather than a `ScrubbingOptions.__post_init__` so that it also covers direct construction; `config.py` is the only other caller.
- `_has_numeric_backreference` tracks escapes and character classes, since `\1` inside `[...]` is an octal escape and `\\1` is a literal backslash followed by a digit. Both are covered by tests.
- The empty-string check uses `compiled.search('')`, which catches `''`, `[0-9]*` and similar. It does not catch a pattern like `\b` that only ever matches zero-width in context — that still yields `[Scrubbed due to '']`. Happy to extend this if you'd prefer it handled here; I left it out because the alternatives I could see either probe with an arbitrary sample string or move the check to match time, and moving it to match time would turn over-redaction into under-redaction, which seems like the wrong direction for a scrubber.
- #2235 (nested-quantifier patterns running on the application thread) is the third of the trio and isn't addressed here — it's a documentation change rather than validation, so it seemed better kept separate. Happy to follow up with it.

### Verification

- `make test` on this branch: 2306 passed, 6 failed. The same failures occur on `main` (2291 passed, 7 failed) and are unrelated — missing local service dependencies for celery/mysql/redis, plus some order-dependent tests in the existing suite. The difference of 15 passing tests is exactly the tests added here.
- `make lint` and `make typecheck` are clean.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2258?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/logfire/_internal/scrubbing.py
+++ b/logfire/_internal/scrubbing.py
@@ -14,6 +14,8 @@ from opentelemetry.attributes import BoundedAttributes
 from opentelemetry.sdk.trace import Event
 from opentelemetry.trace import Link
 
+from logfire.exceptions import LogfireConfigError
+
 from .constants import (
     ATTRIBUTES_CONFIG,
     ATTRIBUTES_JSON_SCHEMA_KEY,
@@ -205,12 +207,83 @@ class NoopScrubber(BaseScrubber):
 NOOP_SCRUBBER = NoopScrubber()
 
 
+def _has_numeric_backreference(pattern: str) -> bool:
+    r"""Whether `pattern` contains a backreference to a group by number, e.g. the `\1` in `(a)\1`.
+
+    Escapes and character classes are tracked because `\1` inside `[...]` is an octal escape,
+    and `\\1` is a literal backslash followed by a digit. Neither is a backreference.
+    """
+    in_class = False
+    i = 0
+    while i < len(pattern):
+        char = pattern[i]
+        if char == '\\' and i + 1 < len(pattern):
+            if not in_class and pattern[i + 1] in '123456789':
+                return True
+            i += 2
+            continue
+        if char == '[':
+            in_class = True
+        elif char == ']':
+            in_class = False
+        i += 1
+    return False
+
+
+def _check_extra_patterns(patterns: Sequence[str] | None) -> list[str]:
+    """Check user-supplied scrubbing patterns, raising `LogfireConfigError` on ones that can't work.
+
+    The patterns are joined into a single regex, so a mistake in one of them changes the behaviour of
+    the whole scrubber. Checking each pattern on its own here means the error can name the pattern
+    that caused it, and means these mistakes fail at configuration time instead of silently either
+    redacting everything or redacting nothing.
+    """
+    if patterns is None:
+        return []
+
+    if isinstance(patterns, str):
+        # `str` satisfies `Sequence[str]`, so this type checks cleanly, and then iterating the string
+        # turns 'password' into the alternation `p|a|s|s|w|o|r|d`, which matches nearly all text.
+        raise LogfireConfigError(
+            f'`extra_patterns` must be a sequence of regular expressions, not a single string. '
+            f'Use `extra_patterns=[{patterns!r}]` to pass one pattern.'
+        )
+
+    checked: list[str] = []
+    for pattern in patterns:
+        try:
+            compiled = re.compile(pattern, re.IGNORECASE | re.DOTALL)
+        except re.error as e:
+            # Compiling here rather than after joining means the position in the message refers to
+            # this pattern, not to an offset within the combined pattern.
+            raise LogfireConfigError(f'Invalid regular expression in `extra_patterns`: {pattern!r}: {e}') from e
+
+        if compiled.search('') is not None:
+            raise LogfireConfigError(
+                f'The `extra_patterns` entry {pattern!r} matches the empty string, so it would match every '
+                f'value and redact all telemetry. Did you mean `+` where you wrote `*`?'
+            )
+
+        if _has_numeric_backreference(pattern):
+            # All the patterns share one group numbering space once they're joined, so a numeric
+            # backreference points at a group from another pattern and the pattern never matches.
+            raise LogfireConfigError(
+                f'The `extra_patterns` entry {pattern!r} contains a backreference to a group by number. '
+                f'Patterns are combined into a single regex, so numbered groups from different patterns '
+                f'would collide. Use a named group instead, e.g. `(?P<name>...)` and `(?P=name)`.'
+            )
+
+        checked.append(pattern)
+
+    return checked
+
+
 class Scrubber(BaseScrubber):
     """Redacts potentially sensitive data."""
 
     def __init__(self, patterns: Sequence[str] | None, callback: ScrubCallback | None = None):
         # See ScrubbingOptions for more info on these parameters.
-        patterns = [_DEFAULT_PATTERN, *(patterns or [])]
+        patterns = [_DEFAULT_PATTERN, *_check_extra_patterns(patterns)]
         self._pattern = re.compile('|'.join(patterns), re.IGNORECASE | re.DOTALL)
         self._callback = callback
 
--- a/logfire/_internal/scrubbing.py
+++ b/logfire/_internal/scrubbing.py
@@ -227,9 +227,17 @@ def _has_numeric_backreference(pattern: str) -> bool:
                     return True
             i += 2
             continue
-        if char == '[':
+        if char == '[' and not in_class:
             in_class = True
-        elif char == ']':
+            i += 1
+            # A `]` at the very start of a class is a member of it rather than the end of it,
+            # as in `[]a]`, and it may follow a negating `^`.
+            if i < len(pattern) and pattern[i] == '^':
+                i += 1
+            if i < len(pattern) and pattern[i] == ']':
+                i += 1
+            continue
+        if char == ']' and in_class:
             in_class = False
         i += 1
     return False
--- a/logfire/_internal/scrubbing.py
+++ b/logfire/_internal/scrubbing.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import copy
 import json
 import re
+import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


aviseth:
Pushed cd5be4f, which takes all four review points.

**Three-digit octal escapes** — correct, and it was a real false positive: `re.compile(r'\123')` matches `S`, and the detector was rejecting it on the leading `\1`. It now treats three octal digits as an escape and only flags the rest. Added to the detector's table alongside `\\1` and `[\1]`, plus an end-to-end test that `sea\123hell` scrubs `seaShell`.

**Non-string entries** — taken. `re.compile` was raising `TypeError`, which broke the contract the rest of this function establishes. Now a `LogfireConfigError` naming the type.

**Zero-length matches in non-empty values** — this is the gap I described in the PR body as unresolved, and `(?=.)` is a better example than the `\b` I had. `search('')` only catches patterns that match the empty string outright. The check now also runs the pattern over a short probe of ordinary characters (`aA0_ -.`) and rejects it if any match is empty, which covers `\b`, `\B`, `(?=.)` and similar. Both are in the parametrized test.

The probe is a heuristic and worth a maintainer's opinion: it cannot catch a pattern that only matches an empty span against characters outside the probe, e.g. `(?=é)`. I chose it over the alternatives because handling zero-length matches at match time would silently stop scrubbing instead of over-scrubbing, and under-redaction seems the worse failure for this feature. Happy to drop it back to the `search('')` check alone if you'd rather not carry a heuristic here.

**Patterns quoted in error messages** — taken. Someone scrubbing a known literal value would put it in `extra_patterns`, and a configuration error tends to end up in a traceback or a log, which is the thing this feature exists to prevent. Messages now identify the entry by index and don't reproduce it. The regex-compilation errors still carry `re`'s own message, including the position within that pattern, which was the point of compiling them individually.

`make lint`, `make typecheck` and the test suite are clean; the suite failures on this branch are the same ones present on `main` (missing local service dependencies, plus some order-dependent tests).

aviseth:
Pushed a97a3fd.

**Zero-width patterns** — you're right that the probe was the wrong mechanism; `(?=z)` defeats it, and so would any pattern written against characters I didn't think to include. Replaced it with the minimum match width from the regex parser, which answers the question directly rather than approximately: 0 for `[0-9]*`, `\b`, `(?=.)`, `(?=z)`, `(?<=a)` and `(?:password)?`, and at least 1 for anything that has to consume a character. All of those are now in the parametrized test, along with `(?<=a)` for the lookbehind case.

The catch is that the parser is private — `re._parser` on 3.11+, `sre_parse` before that. It's looked up by name through `import_module`, and the probe stays as a fallback if that ever stops resolving, so the check degrades rather than breaking. If you'd rather not depend on a private module at all, the alternative I see is rejecting patterns whose width can't be proven positive, which would mean dropping the fallback and erroring instead. Your call.

**"is a int"** — fixed, phrased as "is of type int" so the article isn't a problem for any type name.

`make lint` and `make typecheck` pass. Test suite: 2314 passed, 6 failed, all of which fail on `main` too (celery/mysql/redis need local services, and a few CLI and instrumentation tests are order-dependent — the set rotates between runs on an unmodified checkout).

aviseth:
Pushed fbfa465, which takes three of the four remaining points and pushes back on one.

**Numeric backreferences were rejected too broadly.** Correct, and this was the worst of the four: groups are numbered in order across the joined pattern, so only groups from *earlier* entries move a backreference off its target. The default patterns capture nothing, so `Scrubber([r'(.)\1'])` kept its own numbering and worked — and my check was turning it away. The repeated-character idiom is a reasonable thing to write, so that was a regression against `main` for anyone using it.

Now it counts capturing groups as it goes and only rejects when there are some ahead of the backreference, with the count in the message. `[r'(.)\1']` and `['(?:foo)', r'(.)\1']` are allowed; `['(a)', r'(b)\1']` — the case from #2237 — still errors. Both directions are tested.

**Entries invalid only once combined.** Also correct, and it defeated the purpose of the PR: two entries using the same group name gives `redefinition of group name 'c' as group 2 ... at position 291`, a raw `re.error` carrying an offset into the combined pattern. That is precisely the failure mode this change set out to remove, and worse, using a named group is what the backreference error recommends, so following the advice could land you there. The joined compile is now wrapped and reports as a `LogfireConfigError` naming the two causes I could find: a shared group name, and an inline flag such as `(?i)` (valid alone, invalid once it isn't at the start).

**Zero-width message overstated the effect.** Fair. `(?<=-)` matches an empty span only at `-`, so it redacts a subset rather than everything, and the `+`/`*` hint doesn't apply to look-arounds. Reworded to say it reports a match on values containing nothing sensitive, which are then redacted with an empty reason, and the hint now mentions both causes.

**Python 3.10 `re.error` wording** — I don't think this one holds. The claim is that 3.10 raises "unbalanced parenthesis" for `re.compile('foo(')`, which would break the snapshot. Checked on 3.10, 3.12 and 3.13:

```
py3.10: missing ), unterminated subpattern at position 3
py3.12: missing ), unterminated subpattern at position 3
py3.13: missing ), unterminated subpattern at position 3
```

Same message on all three, and the 3.10 CI job has passed on every push here. "unbalanced parenthesis" is what `re` raises for an unmatched *closing* paren, e.g. `foo)`. Leaving the assertion as it is, but happy to loosen it to the stable prefix if you'd rather not have a stdlib message in a snapshot at all.

`make lint`, `make typecheck` and coverage pass. Suite: 2317 passed, 6 failed, all of which fail on `main` too.

aviseth:
Two more, pushed in b70d423 and the commit above.

**Python 3.10.** You were right that 3.10 behaves differently, though not in the way the earlier note said — the `re.error` text for `foo(` is identical on 3.10, 3.12 and 3.13, and that assertion passes. What actually differs is the inline-flag case I had just added: before 3.11 a global flag away from the start is a `DeprecationWarning` rather than an error, so `extra_patterns=['(?i)foo']` compiled on 3.10 and my new test failed there. That was a genuine break and the 3.10 job caught it.

Rather than make the test version-conditional I made the behaviour consistent, because the 3.10 outcome is the worse one: the pattern compiles and the flag silently applies to the whole combined expression, defaults included. The warning is now turned into an error for the duration of that compile, so the same configuration is refused on every supported version. Verified by running the scrubbing tests under both 3.10 and 3.12.

**Leading `]` in a character class.** Correct, and a nice catch. Python treats a `]` directly after `[` (or `[^`) as a member rather than the terminator, so `[]a\1]` is a single class and the `\1` in it is an octal escape. The scanner took the leading `]` for the end of the class and then read `\1` as a backreference, so `Scrubber(['(a)', r'[]a\1]'])` was rejected despite working. It now consumes an optional `^` and an optional leading `]` when a class opens. `[]a\1]` and `[^]a\1]` are in the detector's table, along with `[a][b]\1` to confirm two complete classes followed by a real backreference is still caught.

All checks green, including the 3.10 job. Suite: 2319 passed, 6 failed, same set as `main`.

PR #2260, opened 2026-08-14
Default the CLI base URL to LOGFIRE_BASE_URL

### LINKED ISSUE #2257: CLI: honor `LOGFIRE_BASE_URL` env var for the base URL (consistency with the SDK)
The SDK reads `LOGFIRE_BASE_URL` to target a self-hosted instance, but the CLI (`logfire auth`, `logfire projects`, …) only takes the endpoint via `--base-url`/`--region` — it ignores `LOGFIRE_BASE_URL`. So you can configure the SDK by env but must pass `--base-url` to every CLI call, and `LOGFIRE_BASE_URL=… logfire auth` drops into the interactive region picker instead of using the URL.

### Request

When neither `--base-url` nor `--region` is passed, default the CLI's base URL to `LOGFIRE_BASE_URL` if it is set (before falling back to the region prompt). This matches the SDK's behavior and improves CI/scripting/self-hosted ergonomics.

### Repro

```console
$ LOGFIRE_BASE_URL=http://localhost:8080 logfire auth
# prompts for a region (us/eu) instead of using the URL

$ logfire --base-url http://localhost:8080 auth
# works
```

(`logfire` 4.40.0)

jgjoe:
Closes #2257.

### The problem

`_get_logfire_url` only looks at the arguments it is given, so the CLI never consults `LOGFIRE_BASE_URL`, while the SDK reads it through `BASE_URL` in `config_params.py`. The visible effect is that `LOGFIRE_BASE_URL=http://localhost:8080 logfire auth` drops into the region picker, because `parse_auth` sees `logfire_url` as `None` and falls into its `if not logfire_url:` branch. Configuring a self-hosted instance by environment therefore works for the SDK but has to be repeated as `--base-url` on every CLI call.

### The change

One line in `_get_logfire_url`, so the resolution order becomes:

`--base-url` / `--logfire-url` → `--region` → `LOGFIRE_BASE_URL` → the interactive region prompt

An empty `LOGFIRE_BASE_URL` is treated as unset and still reaches the prompt.

Two notes on scope:

- The value is read with `os.environ` rather than `ParamManager`. `--data-dir` is a subcommand argument rather than a global one, so it is not reliably available where the URL is resolved, and reading the environment is what the issue asks for. This deliberately does not extend the CLI to file-based config.
- `parse_whoami` already resolves this parameter through `param_manager.load_param('base_url', ...)` and passes the CLI argument as the override, so its behaviour is unchanged.

`docs/reference/cli.md` gets the environment variable next to the existing `--base-url` note. I can't preview the docs site as an external contributor, so the rendering of that note block is worth a glance.

### Tests

In `tests/test_cli.py`:

- `LOGFIRE_BASE_URL` set: `logfire auth` runs the device flow against that host and never prompts for a region, asserted on the request history
- `--base-url` and `--region` each override the environment variable (parametrized)
- `test_auth_no_region_specified` is parametrized over unset and empty, so the prompt path stays covered in both cases

Full suite locally: no new failures. On Windows this checkout has 50 pre-existing failures on `main` (path separators, box-drawing characters, unavailable services); on this branch the set is 48 — a strict subset, with `test_dspy` and `test_litellm` passing on the rerun rather than anything this change touches.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crushonyou2:
Checked this one. `ParamManager.load_param` resolves `runtime > env vars > config file`, and the value being passed as `runtime` here *is* `LOGFIRE_BASE_URL`, so `whoami` ends up with the same URL it had before in every combination: env set (with or without a config file) already resolved to the env value, since env outranks file config; env unset passes `None` and leaves resolution untouched; an empty value is treated as unset on both sides.

So there's no override to restrict — but if you'd like the equivalence pinned down, I'm happy to add a `whoami` regression test covering env and config file together.

crushonyou2:
Rebased onto the latest main — the conflict came from the upstream CLI/auth refactoring, not from this change itself.

The production change is unchanged: `_get_logfire_url` falls back to `LOGFIRE_BASE_URL` only after `--base-url` and `--region`. The `import os` line I had added is no longer needed, since main already imports it.

One test assertion changed to match upstream behavior: `_read_line()` now forwards its default empty prompt directly, so the "press Enter" call is `input('')` rather than `input()`.

crushonyou2:
Hi, quick follow-up on this PR. GitHub reports that it is mergeable, with no conflicts and all checks passing. I saw that review was requested from @strawgate on Aug 27. Is there anything else you'd like me to adjust before review? Thanks!

tests/test_cli.py
```diff
@@ -1343,7 +1396,12 @@ def test_auth_logout_wrong_region(default_credentials: Path, capsys: pytest.Capt
     assert 'No user token was found matching' in capsys.readouterr().err
 
 
-def test_auth_no_region_specified(tmp_path: Path) -> None:
+@pytest.mark.parametrize('base_url_env', [None, ''], ids=['unset', 'empty'])
+def test_auth_no_region_specified(tmp_path: Path, monkeypatch: pytest.MonkeyPatch, base_url_env: str | None) -> None:
+    if base_url_env is None:
+        monkeypatch.delenv('LOGFIRE_BASE_URL', raising=False)
+    else:
+        monkeypatch.setenv('LOGFIRE_BASE_URL', base_url_env)
```
Kludex: Non-blocking: I'd prefer to avoid conditionals in tests. Could we parametrize the environment overrides as `{}` and `{'LOGFIRE_BASE_URL': ''}`, clear the variable unconditionally with `monkeypatch.delenv()`, and apply the overrides with `patch.dict(os.environ, env)` inside the existing `ExitStack`? That keeps both unset and empty cases without branching or duplicating the test.

Kludex: Updated in 705237dd: both environment states now use dictionary parameters and `patch.dict()` without a conditional. All 288 CLI tests, Ruff, and the targeted Pyright check pass.

PR #2262, opened 2026-08-14
Document ingest limits

### LINKED ISSUE #2244: Document ingest time skew and payload size limits
We have limits on time skew (24 hours), span attribute size (10MB), string length limits (on `span_name` and others) as well as a total ingest payload size limit (100MB IIRC).

We should clearly document these in our docs.

adriangb:
Trimmed the page after Bill's feedback that it shares too much, using how other vendors document limits as the yardstick.

**What the survey found** (Sentry, Datadog, Grafana, Elastic, New Relic):

| | Where limits live | Size | Format |
|---|---|---|---|
| Sentry | Dedicated "Size Limits" page | ~225 words | Bullets; delegates exact numbers to a linked reference and notes they may change |
| Datadog | No limits page — a "Limitations" subsection on the Log Collection page | ~150 words | Short bullet list |
| Grafana | Split: a "Usage limits" reference plus a separate "Reduce trace size" how-to | ~1,300 each | Tables with a one-line description column |
| Elastic | Limits are config options (`max_event_size`, default 300 KB) in the config reference | n/a | Config docs |
| New Relic | "View system limits" table, 20-25 limits | Table | One line each |

Two patterns: none of them explain mechanism, they state the number and the consequence and stop; and Grafana keeps the reference table separate from the how-to.

**The page went from 2,015 words to 755.** What came out:

- **The truncation cascade** (string truncation, then container caps, then leaf pruning) and its 32,000 / 64 / 1024 constants. Those were never limits on what you may send: they are the order an over-budget record is cut down in, which is why the page needed a callout saying "those three numbers are not limits on what you can send". A number that needs a disclaimer explaining it is not a limit does not belong on a limits page.
- **The 14-item inventory** of 512-byte and 32,000-byte fields, now one table row naming the common ones.
- **The per-signal timestamp table** and the verbatim rejection message, keeping the `logfire ingest error` pointer that makes the problem findable.
- **`aggregation_temporality` and int32 count rejections**, too rare to earn space.

What stays is the four things a reader acts on: request size, the timestamp window (the one that looks like silent data loss), truncation-not-rejection with `logfire.truncated` to see what was cut, and Summary metrics. The troubleshooting table now carries the actionable material, Grafana-style.

There is a second reason to cut, beyond length: **each published number becomes a contract.** "100 MB per request" and "24 hours" are stable product promises. `MAX_STRING_BYTES = 32_000` is an implementation detail we would want freedom to move, and Sentry hedges its own numbers for exactly this reason.

Closes #2244.

Substantially AI-drafted (Claude Opus 5), written against `dev-docs/documentation-style-guide.md`.


adriangb:
@coderabbitai review

docs/guides/web-ui/metrics-explorer.md
```diff
@@ -94,6 +96,7 @@ Refresh the Metrics view. `hello` appears in **Recently active** and as its own
 |---------|--------------|
 | Metric appears in the catalog but the chart is empty | The metric stopped reporting within the current window. The wizard hides metrics with no recent data, but the catalog entry remains. Widen the time picker. |
 | Custom metric lands in **Everything else** instead of its own namespace | The metric name has no dot (e.g. `requests_total` instead of `app.requests.total`). The grouping is structural: give the name a dotted prefix to create a namespace. |
+| One metric never appears in the catalog while others from the same source do | It is an OTLP `Summary`, which Logfire does not store. Check for a `logfire ingest error` record naming the metric, and see [Summary metrics are not supported](../../reference/limits.md#summary-metrics-are-not-supported). |
```
strawgate: unrelated change?

docs/guides/onboarding-checklist/add-metrics.md
```diff
@@ -88,6 +88,8 @@ for duration in [10, 20, 30, 40, 50]:
 
 You can read more about the Histogram metric in the [OpenTelemetry documentation][histogram-metric].
 
+A histogram is also what to reach for if you are used to Prometheus summaries. Logfire does not store OTLP `Summary` metrics, which report quantiles the sender has already computed: see [Summary metrics are not supported](../../reference/limits.md#summary-metrics-are-not-supported). Record the raw values in a histogram and compute percentiles at query time instead.
```
strawgate, line 91: unrelated change?

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+
+Logfire does not store OTLP `Summary` metrics, a legacy type that reports quantiles the sender has already computed. Quantiles that arrive pre-computed cannot be re-aggregated: averaging two p95 values from two hosts does not give the p95 across both.
+
+A `Summary` is dropped and never reaches the metrics catalog, while the other metrics in the same request are stored. They usually come from a Prometheus scrape forwarded through an OpenTelemetry Collector, whose `prometheus` receiver turns every Prometheus summary into an OTLP `Summary`. Send a histogram instead and compute percentiles at query time.
+
+## Troubleshooting
+
+| Symptom | Cause | Fix |
+| --- | --- | --- |
+| The exporter reports the payload is too large and a batch never arrives | The request was over 100 MB | Lower the export batch size (`OTEL_BSP_MAX_EXPORT_BATCH_SIZE`), or stop attaching very large values to every record |
```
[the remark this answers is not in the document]
adriangb: Valid, and the reasoning is right: `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` is the Batch **Span** Processor's, so it does nothing for logs or metrics, and no batch size helps when a single record is itself oversized.

Checked the OTel SDK env var spec before fixing: logs have `OTEL_BLRP_MAX_EXPORT_BATCH_SIZE`, and metrics have no batch-size variable at all (the periodic reader takes `OTEL_METRIC_EXPORT_INTERVAL`). So the row now names the per-signal knob and says plainly that shrinking the batch does not help an oversized record, which covers the metrics case without inventing a variable. Fixed in 85c02d40.

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+
+This is the limit that most often looks like data silently going missing, because the request itself succeeds. Records outside the window are dropped, the rest of the request is stored, and Logfire writes an explanation into your project as a span named `logfire ingest error`. Search for that if data you expected never arrived.
+
+The usual causes are a wrong clock on the sending host, data buffered offline and exported much later, and attempts to load historical data. Backfilling records older than 24 hours is not supported.
+
+## Truncation
+
+Attributes and long text fields are never rejected for being too big, only shortened. A record whose attributes exceed 10 MB has its largest values cut down until it fits.
+
+You do not have to guess whether this happened. Every shortened value is listed in the record's `logfire.truncated` attribute, and the record's detail panel in the [Live view](../guides/web-ui/live.md) shows a **Truncation** section naming exactly what was cut.
```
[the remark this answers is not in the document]
adriangb: Valid, and it's a regression I introduced: the pre-trim version said explicitly that fixed-length fields are "truncated silently: these fields are not listed in \`logfire.truncated\`", and cutting the page dropped that caveat while keeping the broader claim.

Restored as a distinction rather than a repeat of the old wording: values cut to fit the 10 MB attribute budget are listed in `logfire.truncated` and shown in the Truncation panel, while a span name or message cut at 512 bytes is not flagged anywhere. Fixed in f3c58156.

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+# Ingest limits
+
+Logfire applies a few limits to the data you send. Going past one either **drops** the data or **shortens** a value, and this page covers which, and what to do about it.
+
+| Limit | Value | What happens past it |
+| --- | --- | --- |
+| Request size | 100 MB | The request is rejected and none of it is stored |
+| Timestamp range | 24 hours in the past to 1 hour in the future | Records outside the window are dropped; the rest of the request is stored |
+| Attributes per record | 10 MB | Oversized values are shortened; the record is stored |
+| Long text fields | 512 bytes for span names, messages, and service names; 32,000 bytes for exception messages and stack traces | The value is shortened; the record is stored |
```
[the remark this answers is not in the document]
adriangb: Half right, and the half that lands is worth fixing. The 10 MB log-body limit was in the pre-trim page and came out with everything else; grouping it under "messages ... 512 bytes" does leave a log author able to read the wrong number.

Rather than restore the row, the Truncation section now says a log's `body` gets its own 10 MB budget, separate from the much shorter message. That resolves the ambiguity without re-expanding the table, which is the direction this PR is deliberately going after review feedback that the page shared too much. Fixed in f3c58156.

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Truncation
+
+Attributes and long text fields are never rejected for being too big, only shortened. A record whose attributes exceed 10 MB has its largest values cut down until it fits.
+
+You do not have to guess whether this happened. Every shortened value is listed in the record's `logfire.truncated` attribute, and the record's detail panel in the [Live view](../guides/web-ui/live.md) shows a **Truncation** section naming exactly what was cut.
+
+## Summary metrics are not supported
+
+Logfire does not store OTLP `Summary` metrics, a legacy type that reports quantiles the sender has already computed. Quantiles that arrive pre-computed cannot be re-aggregated: averaging two p95 values from two hosts does not give the p95 across both.
```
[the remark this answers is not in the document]
adriangb: Correct, and it is our own rule: the style guide requires spelling out field-specific acronyms at first use on each page, and the trim removed the OTLP gloss along with the intro that carried it. Expanded on both pages. Fixed in f3c58156.

docs/reference/limits.md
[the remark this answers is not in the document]
adriangb: This is the same finding as the inline comment on the troubleshooting row, and your web query confirms the fix I pushed: `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` is trace-only, logs use `OTEL_BLRP_MAX_EXPORT_BATCH_SIZE`, and metrics have no batch-size variable. The row now names the per-signal knob and says a smaller batch does not help when one record is itself oversized.

docs/reference/limits.md
```diff
@@ -0,0 +1,154 @@
...
+
+```text
+service_name=my-api -> scope_name=my.tracer -> metric_name=http_request_duration_seconds: dropped 12 data points because Summary metrics are a legacy data type and are not supported by Logfire. Please use a Histogram instead.
+```
+
+In practice `Summary` metrics reach Logfire from one of two places:
+
+- **A Prometheus scrape forwarded through an OpenTelemetry Collector.** The Collector's `prometheus` receiver turns every Prometheus summary into an OTLP `Summary`. Change the application to expose a Prometheus histogram instead, which the receiver maps to an OTLP histogram that Logfire stores.
+- **An old OpenTelemetry SDK or a hand-built exporter.** Emit a histogram: `logfire.metric_histogram()` in the Logfire SDK, or your SDK's histogram instrument.
+
```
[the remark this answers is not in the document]
adriangb: Stale: this was against the pre-trim version of the page, which has since been cut from 2015 words to ~760. `logfire.metric_histogram()` no longer appears in `limits.md` at all, so there is nothing left to fix here.

docs/guides/web-ui/metrics-explorer.md
adriangb: Fair question. They're cross-links rather than independent changes: this PR adds a "Summary metrics are not supported" section to the new limits page, and these two lines point at it from the places someone actually hits the problem. The Metrics explorer lists the four supported metric kinds, so `Summary`'s absence belongs next to that list, and this row is the troubleshooting entry for "my metric never showed up".

Happy to drop both if you'd rather keep the PR to the new page alone. The cost is that the limits page becomes the only place that says `Summary` is unsupported, and nobody debugging a missing metric starts there.

docs/guides/onboarding-checklist/add-metrics.md
adriangb: Same reasoning as the Metrics explorer comment: this is the histogram section of the metrics onboarding page, which is where someone migrating from Prometheus summaries lands. One sentence pointing at the new section, rather than restating it.

Same offer: I'll drop it if you'd prefer this PR touch only `limits.md` and the nav.

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+Logfire accepts timestamps from **24 hours in the past** to **1 hour in the future**, measured against Logfire's clock when the request arrives, not the clock of the machine that sent it. Every timestamp is checked: a span's start and end, each span event, a log's timestamp, and each metric data point.
+
+This is the limit that most often looks like data silently going missing, because the request itself succeeds. Records outside the window are dropped, the rest of the request is stored, and Logfire writes an explanation into your project as a span named `logfire ingest error`. Search for that if data you expected never arrived.
+
+The usual causes are a wrong clock on the sending host, data buffered offline and exported much later, and attempts to load historical data. Backfilling records older than 24 hours is not supported.
+
+## Truncation
+
+Attributes and long text fields are never rejected for being too big, only shortened. A record whose attributes exceed 10 MB has its largest values cut down until it fits, and a log's `body` gets its own 10 MB budget separate from the much shorter message.
```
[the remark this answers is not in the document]
adriangb: Valid, and it was self-inflicted: the previous commit added the log `body` budget to fix a different reviewer's point and left the page asserting both "a log's message stores 512 bytes" and "a log's body gets 10 MB" with nothing connecting them.

Checked how they actually relate in the ingest path: a log's text is stored **twice**, in full in `body` (capped at 10 MB) and again as the short display message (capped at 512 bytes), with `logfire.msg` overriding only the latter. So both statements were true, just unexplained.

Rather than document that, the page now states **one 10 MB limit per span, log, or metric point**. Enforcement really is per mechanism (attributes and log body each get their own 10 MB), which makes the single number *stricter* than what we accept: anyone who keeps a record under 10 MB is guaranteed never to be truncated, while going over may still be fine. That is the right direction for a simplification, and it removes the body-versus-message distinction from the page entirely instead of explaining it. Fixed in cbf3a853.

docs/guides/web-ui/metrics-explorer.md
```diff
@@ -40,6 +40,8 @@ Aggregations default sensibly by metric kind, with the rest available from the d
 | Histogram | `avg` | `avg`, `sum`, `min`, `max`, `count` |
 | Exponential histogram | `avg` | `avg`, `sum`, `min`, `max`, `count` |
 
+Those four kinds are the whole list. Logfire does not store OpenTelemetry Protocol (OTLP) `Summary` metrics, a legacy type that carries pre-computed quantiles instead of buckets: a `Summary` is dropped at ingest and never reaches the catalog. See [Summary metrics are not supported](../../reference/limits.md#summary-metrics-are-not-supported).
```
adriangb:
```suggestion
Logfire does not store OpenTelemetry Protocol (OTLP) `Summary` metrics, a legacy type that carries pre-computed quantiles instead of buckets: a `Summary` is dropped at ingest, emitting an error into your project. See [Summary metrics are not supported](../../reference/limits.md#summary-metrics-are-not-supported).
```

docs/guides/web-ui/metrics-explorer.md
adriangb:
```suggestion
| One metric never appears in the catalog while others from the same source do | It is a`Summary` metric, which is deprecated in OTLP and which Logfire does not support. Check for a `logfire ingest error` record naming the metric, and see [Summary metrics are not supported](../../reference/limits.md#summary-metrics-are-not-supported). |
```

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+| Size of one span, log, or metric point | 10 MB | Oversized values are shortened; the record is stored |
+| Long text fields | 512 bytes for span names, service names, and the message shown in the UI; 32,000 bytes for exception messages and stack traces | The value is shortened; the record is stored |
+
+The limits are the same in the [US and EU regions](data-regions.md) and are not configurable per project. Separately, [Summary metrics](#summary-metrics-are-not-supported) are not stored at all.
+
+## Timestamps
+
+Logfire accepts timestamps from **24 hours in the past** to **1 hour in the future**, measured against Logfire's clock when the request arrives, not the clock of the machine that sent it. Every timestamp is checked: a span's start and end, each span event, a log's timestamp, and each metric data point.
+
+This is the limit that most often looks like data silently going missing, because the request itself succeeds. Records outside the window are dropped, the rest of the request is stored, and Logfire writes an explanation into your project as a span named `logfire ingest error`. Search for that if data you expected never arrived.
```
adriangb:
```suggestion
This limit can look like data silently going missing, because the request returns a partial success message if some of the spans do pass the timestamp range checks. In this case records outside the window are dropped, the rest of the request is stored, and Logfire writes an explanation into your project as a span named `logfire ingest error`. Search for that if data you expected never arrived. If all records in a request are outside the time range (e.g. a backfill) an error response will be returned.
```

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+
+Values cut to fit that limit are listed in the record's `logfire.truncated` attribute, and the record's detail panel in the [Live view](../guides/web-ui/live.md) shows a **Truncation** section naming them. The fixed-length fields in the table above are a different case: they are shortened silently, so a span name or message cut at 512 bytes is not flagged anywhere. Keeping a record under 10 MB avoids truncation entirely.
+
+## Summary metrics are not supported
+
+Logfire does not store OpenTelemetry Protocol (OTLP) `Summary` metrics, a legacy type that reports quantiles the sender has already computed. Quantiles that arrive pre-computed cannot be re-aggregated: averaging two p95 values from two hosts does not give the p95 across both.
+
+A `Summary` is dropped and never reaches the metrics catalog, while the other metrics in the same request are stored. They usually come from a Prometheus scrape forwarded through an OpenTelemetry Collector, whose `prometheus` receiver turns every Prometheus summary into an OTLP `Summary`. Send a histogram instead and compute percentiles at query time.
```
adriangb:
```suggestion
A `Summary` is dropped on arrival. If a payload contains `Summary` metrics alongside supported metrics the supported metrics will be ingested and the summary metrics will be dropped; a partial success status will be returned. They usually come from a Prometheus scrape forwarded through an OpenTelemetry Collector, whose `prometheus` receiver turns every Prometheus summary into an OTLP `Summary`. Send a histogram instead and compute percentiles at query time.
```

adriangb (commented):
some nits

docs/reference/limits.md
adriangb: Applied in 962eb14e. Verified the two-outcome behaviour in the ingest path first: rejection is per record, so a partly-out-of-window request answers `200` with an OTLP partial success, and `build_response` returns `422 Unprocessable` once nothing valid is left. Your point lands: the page described only the first case, which made a fully rejected backfill look like it should have half-worked.

Kept your substance, tightened the wording and put it in active voice to match the rest of the page.

docs/reference/limits.md
adriangb: Applied in 962eb14e, and confirmed metrics behave the same way: mixed payloads store the supported metrics and return a partial success, while an all-`Summary` payload hits `num_valid_metrics == 0` and fails outright. Added that second case too, so this section and the timestamp one describe the same two outcomes.

In active voice rather than "will be ingested"/"will be returned", per the style guide's rule on passive by default. Say the word if you prefer your phrasing verbatim.

docs/guides/web-ui/metrics-explorer.md
adriangb: Applied verbatim in 962eb14e. Better than what I had: "never reaches the catalog" told the reader where the metric is absent, yours tells them where to look.

docs/guides/web-ui/metrics-explorer.md
adriangb: Applied the substance in 962eb14e, with one change: I did not write "deprecated in OTLP", because it is not.

I checked `opentelemetry/proto/metrics/v1/metrics.proto` on main before publishing it. `Summary` carries no deprecation annotation. Its comment reads: quantile summaries "cannot always be merged in a meaningful way. While they can be useful in some applications, histogram data points are recommended for new applications." Discouraged for new use, not deprecated, and our own admission code calls it "a legacy data type" rather than deprecated.

The row now reads "a legacy type that OTLP itself discourages and Logfire does not store", which keeps your point that this is not just a Logfire quirk without claiming a status the spec has not given it. Happy to go stronger if you know of a deprecation I could not find. Also fixed the missing space in `a`Summary``.

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+
+This is the limit that most often looks like data silently going missing, because the request itself succeeds. Records outside the window are dropped, the rest of the request is stored, and Logfire writes an explanation into your project as a span named `logfire ingest error`. Search for that if data you expected never arrived.
+
+The usual causes are a wrong clock on the sending host, data buffered offline and exported much later, and attempts to load historical data. Backfilling records older than 24 hours is not supported.
+
+## Truncation
+
+Attributes and long text fields are never rejected for being too big, only shortened. A record over 10 MB has its largest values cut down until it fits.
+
+Values cut to fit that limit are listed in the record's `logfire.truncated` attribute, and the record's detail panel in the [Live view](../guides/web-ui/live.md) shows a **Truncation** section naming them. The fixed-length fields in the table above are a different case: they are shortened silently, so a span name or message cut at 512 bytes is not flagged anywhere. Keeping a record under 10 MB avoids truncation entirely.
```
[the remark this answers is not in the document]
adriangb: Valid, and it was mine from the simplification commit: the sentence claimed staying under 10 MB avoids truncation entirely, immediately after the sentence saying fixed-length fields are shortened silently. A 600-byte span name is trimmed in a 2 KB record, so the two lines contradicted each other.

Tied the advice to the truncation it actually prevents ("A record over 10 MB has its largest values cut down until it fits, so keeping records under 10 MB avoids that") and made the fixed fields explicit that they are capped whatever the record's size. Fixed in 39a602fc.

docs/guides/web-ui/metrics-explorer.md
adriangb: That's the source I should have found; I only checked `metrics.proto` for a deprecation annotation and stopped there, when the data model spec is where the status is actually stated.

So "legacy" is not a hedge, it is the spec's own word: the section is titled **Summary (Legacy)** and says the point type "is not recommended for new applications and exists for compatibility with other formats". Updated both pages to mirror that rather than my "discourages" paraphrase, and linked the section so the claim is checkable:

> Logfire does not store OpenTelemetry Protocol (OTLP) `Summary` metrics, a type the OpenTelemetry spec itself marks [legacy](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#summary-legacy) and does not recommend for new applications.

In b4cfad87.

adriangb:
@coderabbitai approve

docs/reference/limits.md
```diff
@@ -0,0 +1,54 @@
...
+| Request size | 100 MB | The request is rejected and none of it is stored |
+| Timestamp range | 24 hours in the past to 1 hour in the future | Records outside the window are dropped; the rest of the request is stored |
+| Size of one span, log, or metric point | 10 MB | Oversized values are shortened; the record is stored |
+| Long text fields | 512 bytes for span names, service names, and the message shown in the UI; 32,000 bytes for exception messages and stack traces | The value is shortened; the record is stored |
+
+The limits are the same in the [US and EU regions](data-regions.md) and are not configurable per project. Separately, [Summary metrics](#summary-metrics-are-not-supported) are not stored at all.
+
+## Timestamps
+
+Logfire accepts timestamps from **24 hours in the past** to **1 hour in the future**, measured against Logfire's clock when the request arrives, not the clock of the machine that sent it. Every timestamp is checked: a span's start and end, each span event, a log's timestamp, and each metric data point.
```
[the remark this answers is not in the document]
adriangb: Valid. Confirmed in `retain_admissible_data_points`: a point's own timestamp goes through `window.admits(recorded)` (both bounds), while its start timestamp goes through `window.admits_future(start)` only. The admission code says why: "Metric \`start_time_unix_nano\` is legitimately old (it is often when the process booted), and we don't use it for retention or partitioning, so only the impossible direction is checked."

Worth saying why I fixed this rather than leaving it simplified. This page deliberately rounds toward the stricter rule in places (it states one 10 MB limit per record where enforcement is actually per mechanism), because a reader who follows a stricter rule is safe. That reasoning does not hold here: a cumulative counter's start timestamp is process boot time, which can be weeks old and is not something the reader controls. Rounding strict there does not give them a safe rule to follow, it tells them correct data is being dropped and sends them chasing a problem that does not exist.

Fixed in b8678b6d, in one sentence rather than restructuring the paragraph:

> A metric point's *start* timestamp is checked only against the future bound, because for a cumulative metric it is often when the process booted and legitimately old.

PR #2275, opened 2026-08-18
Let `logfire auth` complete without a TTY

### LINKED ISSUE #2274: `logfire auth` cannot complete without a TTY: it blocks on input() twice
### Description

`logfire auth` cannot be completed by any caller without a terminal — CI, a container, or a coding agent — because it blocks on `input()` twice. Both raise `EOFError` and print a traceback.

```console
$ logfire auth < /dev/null
Welcome to Logfire! 🔥
Before you can send data to Logfire, we need to authenticate you.

Logfire is available in multiple data regions. Please select one:
1. US (GCP region: us-east4)
2. EU (GCP region: europe-west4)
Selected region [1/2]: Traceback (most recent call last):
  ...
  File ".../logfire/_internal/cli/auth.py", line 50, in parse_auth
    input(f'Selected region [{"/".join(str(i) for i in range(1, len(REGIONS) + 1))}]: ')
EOFError: EOF when reading a line
```

`--region` answers the first prompt, and then it stops at the second:

```console
$ logfire --region us auth < /dev/null
  ...
  File ".../logfire/_internal/cli/auth.py", line 61, in parse_auth
    input()
EOFError: EOF when reading a line
```

That second prompt is `Press Enter to open <host> in your browser...`. Nothing is done with the result — it exists to give a person a beat before a browser window appears.

**Nothing else in the flow needs a terminal.** With a single newline piped in, the whole thing completes:

```console
$ printf '\n' | logfire --region us auth
Press Enter to open logfire-us.pydantic.dev in your browser...
Please open https://logfire-us.pydantic.dev/auth/device/dc-0001... in your browser to authenticate if it hasn't already.
Waiting for you to authenticate with Logfire...
Successfully authenticated!
```

The URL is printed, and the device-code poll simply waits. A non-interactive caller can surface that link and the login completes when the user opens it.

### Why it matters

This is the first command in our own setup instructions, and the docs' "copy AI prompt" is designed to be pasted into a coding agent. Agent shells do not allocate a TTY — Claude Code's Bash tool, for one — so the agent gets a traceback on the very first step.

The prompt works around it by telling the agent to hand the command to the user and wait ("Then tell me when you're done"), which costs a round trip on every first-time setup. Observed agent behaviour when it tries anyway:

> *"I tried running `logfire auth` myself, but it hard-stops needing an interactive browser sign-in and a region prompt from a real TTY — that's account-linking only you can do."*

They are reading the situation correctly. The point is that they should not have to.

### Suggested behaviour

When `sys.stdin` is not a TTY:

- **skip the "Press Enter" prompt** — there is no beat to give and no browser to open, and the URL is already printed;
- **for the region, fail with a message that says what to do** (`pass --region us|eu`) rather than an `EOFError` traceback. Which region holds your data is a real choice and shouldn't be guessed.

Interactive behaviour is unchanged.

strawgate:
Fixes #2274.

`logfire auth` blocks on `input()` twice, so any caller without a terminal — CI, a container, a coding agent — gets an `EOFError` traceback on the first command in our own setup instructions.

```console
$ logfire --region us auth < /dev/null
  ...
  File ".../logfire/_internal/cli/auth.py", line 61, in parse_auth
    input()
EOFError: EOF when reading a line
```

## The change

**Read stdin and handle `EOFError`, rather than checking `isatty()`.** That distinction turned out to matter — see below. It covers all three cases with no special-casing: a terminal, a pipe with answers in it, and nothing at all.

**`Press Enter to open <host> in your browser...` no longer blocks when there is nothing to read.** Nothing is done with the result — it exists to give a person a beat before a browser window appears, and with no one there, there is no beat to give and no browser to open (so none is opened). Everything else in the flow already works headless: the URL is printed and the device-code poll simply waits, so a caller can surface the link and the login completes when the user opens it.

**The region stays a real choice.** Which region holds your data is not ours to guess, so with no answer available the caller is told what to pass and the command exits 1 — no traceback, and each suggestion is a line you can paste:

```console
$ logfire auth < /dev/null
Logfire is available in multiple data regions and no region was selected.
Re-run in an interactive terminal to choose, or pass one:
  logfire --region us auth
  logfire --region eu auth
```

Interactive behaviour is unchanged.

## A correction, and why the first version was wrong

The first version of this PR gated the prompts on `sys.stdin.isatty()`. That is a different question — *"is a terminal attached"* rather than *"can I read an answer"* — and it broke a case that worked before:

```console
$ printf '1\n\n' | logfire auth      # worked on main; hard-failed on the first version
LogfireConfigError: ... no region was selected ...
```

A pipe is not a tty, but the answers are sitting right there. Piping them in is how scripts have always driven this command, so gating on `isatty()` would have turned working setup scripts into a hard error — trading one broken caller for another.

I found this by running the real CLI in a container with its stdin on a pipe, which is how our own tooling drives it. Reading and handling `EOFError` is both simpler and correct for all three shapes.

## Verified

In a container against a real backend:

| stdin | before | after |
| --- | --- | --- |
| `< /dev/null`, `--region us` | `EOFError` traceback | authenticated, credentials written |
| `< /dev/null`, no region | `EOFError` traceback | clean guidance on stderr, exit 1 |
| `printf '1\n\n' \|` (a pipe) | works | works |

## Tests

Two new tests for the headless cases, and one that pins the pipe case above so the `isatty()` mistake cannot come back — I confirmed it goes red when the gate is reinstated. The tests also assert the login URL is printed (with no browser to open, it is the only way to surface the login to a person) and that the region suggestions never regress into the un-pasteable `us|eu` form.

`tests/test_cli.py tests/test_auth.py`: **222 passed**.

Two pre-existing tests exercise the interactive path by patching `input` directly, and are unchanged.

### DIFF
--- a/logfire/_internal/cli/auth.py
+++ b/logfire/_internal/cli/auth.py
@@ -18,12 +18,13 @@ def _read_line(prompt: str = '') -> str | None:
     piping the answers in is how scripts have always driven this command. Gating on
     isatty() would turn that into a hard failure.
 
-    `sys.stdin` can also be None (pythonw, some embedded runtimes) and reading it can
-    raise on a closed stream, so both are treated as "no answer available".
+    `sys.stdin` can also be None (pythonw, some embedded runtimes), where `input()` raises
+    `RuntimeError('lost sys.stdin')`, and reading a closed stream raises `ValueError`.
+    All of them mean the same thing here: no answer is available.
     """
     try:
         return input(prompt)
-    except (EOFError, AttributeError, ValueError):
+    except (EOFError, RuntimeError, AttributeError, ValueError):
         return None
 
 
--- a/logfire/_internal/cli/auth.py
+++ b/logfire/_internal/cli/auth.py
@@ -67,11 +67,19 @@ def parse_auth(args: argparse.Namespace) -> None:
                 # Nothing to read and nothing to guess from: which region holds your data
                 # is not ours to pick. It is answerable ahead of time, and saying so beats
                 # looping forever or raising EOFError from a prompt nobody can reply to.
-                raise LogfireConfigError(
-                    'Logfire is available in multiple data regions and no region was selected. '
-                    f'Pass one with `logfire --region {"|".join(REGIONS)} auth`, or run this in '
-                    'an interactive terminal to choose.'
+                #
+                # Written and exited rather than raised: an exception escapes `main()`,
+                # which catches only KeyboardInterrupt, so the caller gets a traceback --
+                # which is the thing this change set out to remove. Each region is spelled
+                # out as its own runnable line, because `--region us|eu` is a shell
+                # pipeline if you paste it.
+                sys.stderr.write(
+                    'Logfire is available in multiple data regions and no region was selected.\n'
+                    'Re-run in an interactive terminal to choose, or pass one:\n'
                 )
+                for region_id in REGIONS:
+                    sys.stderr.write(f'  logfire --region {region_id} auth\n')
+                sys.exit(1)
             try:
                 selected_region = int(answer)
             except ValueError:
--- a/logfire/_internal/cli/auth.py
+++ b/logfire/_internal/cli/auth.py
@@ -10,16 +10,21 @@ from ..auth import DEFAULT_FILE, UserTokenCollection, poll_for_token, request_de
 from ..config import REGIONS
 
 
-def _stdin_is_interactive() -> bool:
-    """Is there a person at a terminal who can answer a prompt?
+def _read_line(prompt: str = '') -> str | None:
+    """Read one line of input, or `None` if there is nothing there to read.
 
-    `sys.stdin` can be None (pythonw, some embedded runtimes) and `isatty()` can raise on
-    a closed stream, so neither is assumed.
+    Deliberately not `sys.stdin.isatty()`. That answers "is a terminal attached", which is
+    a different question: a pipe is not a tty and is perfectly answerable, and
+    piping the answers in is how scripts have always driven this command. Gating on
+    isatty() would turn that into a hard failure.
+
+    `sys.stdin` can also be None (pythonw, some embedded runtimes) and reading it can
+    raise on a closed stream, so both are treated as "no answer available".
     """
     try:
-        return bool(sys.stdin) and sys.stdin.isatty()
-    except (AttributeError, ValueError):  # pragma: no cover - defensive
-        return False
+        return input(prompt)
+    except (EOFError, AttributeError, ValueError):
+        return None
 
 
 def parse_auth(args: argparse.Namespace) -> None:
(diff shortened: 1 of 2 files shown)


strawgate:
All three addressed in 48d7090 — the first two were right and the fix was worse than I thought.

### P2 — `LogfireConfigError` reaches the user as a traceback

Correct, and it undercut the whole point of the PR: `main()` catches only `KeyboardInterrupt`, so I had replaced an `EOFError` traceback with a `LogfireConfigError` traceback. Now writes to stderr and `sys.exit(1)`, matching how the rest of the CLI reports errors (e.g. the `Not logged in.` / `No Logfire credentials found` paths).

Verified in a container — before, this ended in `Traceback (most recent call last): ... LogfireConfigError`; now:

```console
$ logfire auth < /dev/null
Logfire is available in multiple data regions and no region was selected.
Re-run in an interactive terminal to choose, or pass one:
  logfire --region us auth
  logfire --region eu auth
$ echo $?
1
```

### P2 — `--region us|eu` is a shell pipeline if you paste it

Also correct, and the more embarrassing one given the audience for this fix is people (and agents) copying the suggested command. Each region is now printed as its own runnable line, generated from `REGIONS` so it stays correct if a region is added. The test asserts both lines are present *and* that `us|eu` does not appear, so it cannot regress into the pipeline form.

Confirmed each suggested line runs as printed.

### P3 — the test does not assert the URL is printed

Fair, and it is the feature rather than incidental output: with no browser to open, that printed URL is the only way a caller can surface the login to a person. Now asserted via `capsys`.

`tests/test_cli.py tests/test_auth.py`: 222 passed.

strawgate:
Three findings on the latest commits. One fixed here, two answered.

### 🟠 `RuntimeError: input(): lost sys.stdin` escapes `_read_line` — **fixed** (1c6090d)

Valid, and the docstring was already claiming otherwise, which is the worse part. `RuntimeError` is now in the tuple alongside `EOFError` / `ValueError` / `AttributeError` — every one of them means "no answer is available".

Tested against both a mocked `input` and the real thing (`sys.stdin = None`), so it cannot pass on a mock that does not match CPython's actual behaviour.

### 🟠 The unconditional read still blocks on an open, silent stdin — **valid, and out of scope here**

Correct, and worth being explicit that this PR does not fix it. Confirmed:

```console
$ sleep 20 | timeout 10 logfire --region us auth
# killed, still waiting
```

Handling `EOFError` only helps when a read *returns*. An idle pipe never sends EOF, so `input()` waits forever — there is nothing to detect.

The suggested alternative, a non-blocking availability check, would break the case this PR exists to protect: `printf '1\n' | logfire auth` is a race against whether the writer has flushed yet, so a `select`-style poll would sometimes decide "no input" while the answer was in flight. That trades a hang for a wrong answer.

The fix is to let the caller declare it up front, which is #2283 (`--non-interactive`), stacked on this branch:

```console
$ sleep 20 | timeout 10 logfire --non-interactive --region us auth
Successfully authenticated!    # exit 0
```

### 🔵 `Press Enter to open …` is printed before it is known to be readable — **keeping as is**

The suggested reordering — read first, print the message inside the `is not None` branch — makes the interactive case worse: a person would get a blocking cursor with no explanation of what is being waited on, because the prompt only appears *after* they press the key.

Printing it and then discovering there is nobody there costs one stray line in a log. Blocking a human with an unlabelled prompt costs them the ability to guess what to do. I would rather have the stray line.

With #2283 both the message and the read are skipped together, which is the clean version of this — but that needs the flag, since it is the same "cannot know without reading" problem as above.

PR #2283, opened 2026-08-18
Add `logfire --non-interactive`

strawgate:
Adds a global `--non-interactive` flag to the CLI. It never assumes an answer — it **fails with guidance** naming the argument that would have supplied it.

> **Stacked on #2275.** That PR is the first commit on this branch; review/merge it first. This one touches the same region block, so branching off `main` would just conflict.

## Why, when #2275 already handles no-stdin

Because handling `EOFError` only helps when a read **returns**. Stdin can be open and silent — a CI runner, a supervisor, or an agent harness holding an idle pipe — and then `input()` waits forever, with no traceback and no output. The job just stops.

```console
$ sleep 20 | timeout 10 logfire auth                      # killed, still waiting
$ sleep 20 | timeout 10 logfire --non-interactive auth    # exit 1, with guidance
```

Nothing can *detect* that case, because there is nothing to detect. Only declaring intent up front avoids it, since then nothing has to be read.

So the two are complements rather than alternatives:

- **EOF handling** covers callers that cannot be changed — a command copied out of the docs into an agent, which will never know to pass a flag.
- **`--non-interactive`** covers callers that *can* say so, and is the only thing that catches the hang.

Closest precedents: `ssh -o BatchMode=yes`, `pip --no-input`, `terraform -input=false`.

## Behaviour

| Prompt | With `--non-interactive` |
| --- | --- |
| `Press Enter to open <host> in your browser…` | skipped — nothing is being asked |
| `Press Enter to continue` after a project is made | skipped; the message still prints |
| region | error naming `logfire --region us\|eu auth`, one line each |
| organization on `projects new` | error naming `--org` / `--default-org` |
| project name | error naming the positional |
| project selection on `projects use` | error naming the positional |

```console
$ logfire --non-interactive projects new orders
This would create a project in the organization "agent-eval".
Cannot prompt because --non-interactive was passed.
Supply it instead with:
  logfire projects new <name> --org agent-eval
  logfire projects new <name> --default-org
```

**Creating a project is refused, not assumed**, even here. It is a side effect, and `--default-org` is how a caller says yes ahead of time. The flag means "don't ask me", not "do whatever you like".

Without the flag, every prompt behaves exactly as before.

## Verified

In a container against a real backend:

- the hang above, fixed (measured by the CLI's own exit code — wall-clock is the wrong instrument, since `sleep 20 | …` makes the *shell* wait regardless of when the CLI exits);
- each blocking command now exits 1 with guidance and no traceback;
- the already-working paths (`projects use <name> --org <org>`, `projects list`) are unaffected and still exit 0.

Tests: 3 new (refusal without reading stdin, the interactive path unchanged, and that the switch does not leak between invocations — it is process state, and a leak would be invisible). `tests/test_cli.py tests/test_configure.py`: **315 passed**.

One pre-existing failure elsewhere in the suite, `test_log_non_scalar_args[pydantic_file_path]`, reproduces on unmodified `main` and is unrelated.

## Scope

CLI only. `logfire.configure()` — the second half of what our own setup docs tell people to run — is a Python API, so a CLI flag cannot reach it, and it still prompts. Worth saying plainly: this does not finish the job on its own.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2283?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/logfire/_internal/cli/__init__.py
+++ b/logfire/_internal/cli/__init__.py
@@ -26,7 +26,7 @@ from ..auth import HOME_LOGFIRE
 from ..client import LogfireClient
 from ..config import REGIONS, LogfireCredentials, get_base_url_from_token
 from ..config_params import ParamManager
-from ..interactive import NonInteractiveError, set_non_interactive
+from ..interactive import NonInteractiveError, is_non_interactive, require_answer, set_non_interactive
 from ..server_response import install_logfire_response_hook
 from ..tracer import SDKTracerProvider
 from .auth import parse_auth, parse_logout
@@ -114,7 +114,14 @@ def parse_clean(args: argparse.Namespace) -> None:
     files_to_delete.append(data_dir / 'logfire_credentials.json')
 
     files_to_display = '\n'.join([str(file) for file in files_to_delete if file.exists()])
-    confirm = input(f'The following files will be deleted:\n{files_to_display}\nAre you sure? [N/y]')
+    if not args.yes:
+        require_answer(
+            f'This would delete:\n{files_to_display}',
+            'logfire clean --yes',
+        )
+    confirm = (
+        'y' if args.yes else input(f'The following files will be deleted:\n{files_to_display}\nAre you sure? [N/y]')
+    )
     if confirm.lower() in ('yes', 'y'):
         for file in files_to_delete:
             file.unlink(missing_ok=True)
@@ -362,6 +369,7 @@ def _main(args: list[str] | None = None) -> None:
     cmd_clean.set_defaults(func=parse_clean)
     cmd_clean.add_argument('--data-dir', default='.logfire')
     cmd_clean.add_argument('--logs', action='store_true', default=False, help='remove the Logfire logs')
+    cmd_clean.add_argument('--yes', '-y', action='store_true', default=False, help='do not ask for confirmation')
 
     cmd_gateway = subparsers.add_parser('gateway', help=_parse_gateway.__doc__)
     cmd_gateway.add_argument('gateway_args', nargs=argparse.REMAINDER)
@@ -480,6 +488,7 @@ def main(args: list[str] | None = None) -> None:
     file_handler.setLevel(logging.DEBUG)
     logging.basicConfig(handlers=[file_handler], level=logging.DEBUG)
 
+    previous_non_interactive = is_non_interactive()
     try:
         _main(args)
     except KeyboardInterrupt:
@@ -490,4 +499,8 @@ def main(args: list[str] | None = None) -> None:
         sys.stderr.write(f'{e}\n')
         sys.exit(1)
     finally:
+        # Restore rather than clear: `main()` is importable and an application may call it
+        # in-process, where leaving the switch set would silently stop every later prompt
+        # -- including ones in `logfire.configure()`, which this flag does not govern.
+        set_non_interactive(previous_non_interactive)
         file_handler.close()
--- a/logfire/_internal/cli/__init__.py
+++ b/logfire/_internal/cli/__init__.py
@@ -26,6 +26,7 @@ from ..auth import HOME_LOGFIRE
 from ..client import LogfireClient
 from ..config import REGIONS, LogfireCredentials, get_base_url_from_token
 from ..config_params import ParamManager
+from ..interactive import NonInteractiveError, set_non_interactive
 from ..server_response import install_logfire_response_hook
 from ..tracer import SDKTracerProvider
 from .auth import parse_auth, parse_logout
@@ -335,6 +336,11 @@ def _main(args: list[str] | None = None) -> None:
 
     parser.add_argument('--version', action='store_true', help='show the version and exit')
     global_opts = parser.add_argument_group(title='global options')
+    global_opts.add_argument(
+        '--non-interactive',
+        action='store_true',
+        help='never prompt; fail with guidance if an answer would be required',
+    )
     url_or_region_grp = global_opts.add_mutually_exclusive_group()
     url_or_region_grp.add_argument('--logfire-url', help=argparse.SUPPRESS)
     url_or_region_grp.add_argument(
@@ -434,6 +440,7 @@ def _main(args: list[str] | None = None) -> None:
         namespace.script_and_args = unknown_args + (namespace.script_and_args or [])
     else:
         namespace = parser.parse_args(args)
+    set_non_interactive(namespace.non_interactive)
 
     if namespace.logfire_url:
         warnings.warn(
@@ -477,6 +484,10 @@ def main(args: list[str] | None = None) -> None:
         _main(args)
     except KeyboardInterrupt:
         sys.stderr.write('User cancelled.\n')
+        sys.exit(1)
+    except NonInteractiveError as e:
+        # The whole point is guidance instead of a traceback, so it must not escape.
+        sys.stderr.write(f'{e}\n')
         sys.exit(1)
     finally:
         file_handler.close()
--- a/logfire/_internal/cli/auth.py
+++ b/logfire/_internal/cli/auth.py
@@ -8,6 +8,7 @@ from urllib.parse import urlparse
 from ...exceptions import LogfireConfigError
 from ..auth import DEFAULT_FILE, UserTokenCollection, poll_for_token, request_device_code
 from ..config import REGIONS
+from ..interactive import require_answer
 
 
 def _read_line(prompt: str = '') -> str | None:
@@ -57,6 +58,10 @@ def parse_auth(args: argparse.Namespace) -> None:
         )
     )
     if not logfire_url:
+        require_answer(
+            'Logfire is available in multiple data regions and no region was selected.',
+            *(f'logfire --region {region_id} auth' for region_id in REGIONS),
+        )
         selected_region = -1
         while not (1 <= selected_region <= len(REGIONS)):
             sys.stderr.write('Logfire is available in multiple data regions. Please select one:\n')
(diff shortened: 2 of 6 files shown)


strawgate:
Both findings fixed, and the first was a real hang in the case this PR is most likely to be used in. Rebased on the updated #2275.

### 🟠 `--non-interactive --region <r> auth` still blocked at the browser prompt — **fixed**

Valid and embarrassing: with `--region` supplied the region prompt is skipped, so `Press Enter to open … in your browser` is the *next* stop, and it read unconditionally. That is precisely the hang this PR exists to prevent, in the invocation most likely to be used.

Measured by the CLI's own exit code — wall-clock cannot tell a clean failure from a hang here, because `sleep 20 | …` makes the shell wait regardless of when the CLI exits:

| command, stdin = open and silent | before | after |
| --- | --- | --- |
| `logfire auth` | hangs | hangs *(unchanged — no flag)* |
| `logfire --region us auth` | hangs | hangs *(unchanged — no flag)* |
| `logfire --non-interactive auth` | exit 1, guidance | exit 1, guidance |
| `logfire --non-interactive --region us auth` | **hung** | **exit 0, authenticated** |

**Why my own tests missed it:** they simulate no-stdin with `side_effect=EOFError`, and an EOF-based test passes whether or not the flag is honoured — the read returns either way. The new test asserts `input` is *never called*, which is the only assertion that can catch a blocking read.

### 🟡 `sys.stdin is None` raises `RuntimeError` — **fixed in #2275**

Valid. It belongs on the base branch, since that is where `_read_line` is introduced, so it landed there (1c6090d) and this branch was rebased onto it.

### Also in this push

- **Dropped two guards in `initialize_project`.** That is the `logfire.configure()` path; the CLI calls `create_new_project` / `use_existing_project` directly, so the switch — which only CLI argument parsing sets — could never be on when they ran. Unreachable code in a CLI-scoped PR, and untestable.
- **`require_answer` now takes at least one remedy positionally.** Refusing to prompt while offering no way forward just moves the dead end, and the signature now says so. This also removed the only uncovered branch.

### Coverage

| file | coverage |
| --- | --- |
| `logfire/_internal/interactive.py` | **100%** |
| `logfire/_internal/cli/auth.py` | **100%** |
| `logfire/_internal/cli/__init__.py` | **100%** |

Branches included. The new `config.py` guards are covered by the existing `projects new` / `projects use` tests.

`tests/test_cli.py tests/test_configure.py`: **320 passed**. One pre-existing failure elsewhere in the suite, `test_log_non_scalar_args[pydantic_file_path]`, reproduces on unmodified `main`.

strawgate:
All six addressed in c0490d1, threads resolved. The reach findings were the important ones — the flag only guarded the paths I had happened to touch, which is not what it promises.

### 🟠 P1 — `clean` and `gateway` still prompt

Right, and it made the flag's contract false. With an open, silent stdin, judged by the CLI's own exit code:

| | before | after |
| --- | --- | --- |
| `logfire --non-interactive clean` | hangs | exit 1 |
| `logfire --non-interactive projects new 'bad name' --default-org` | hangs | exit 1 |
| `logfire --non-interactive gateway` | hangs | exit 1 |

`clean` needed a way through, so it gains `--yes`. Refusing is the only safe answer without it: the prompt defaults to **N**, so assuming the default would do nothing while reporting success, and assuming yes would delete credentials nobody confirmed. Verified the file survives the refusal and is gone with `--yes`.

### 🟠 P1 — an invalid supplied name still prompts

Correct: my guard only covered a *missing* name, so `projects new 'bad name'` fell through to the recovery prompt and read stdin. Now refused before it, with the reason:

```
The project name 'Bad Name' is invalid: it may contain lowercase alphanumeric
characters and single hyphens, and may not start or end with a hyphen.
```

### 🟠 Major — the remedy dropped `--org` / `--default-org`

Also right — suggesting a bare `projects new <name>` to someone who passed `--org` hands them a command that stops at the prompt they had already answered. Built from the active organization mode now:

```console
$ logfire --non-interactive projects new --org agent-eval
  logfire projects new app --org agent-eval

$ logfire --non-interactive projects new --default-org
  logfire projects new app --default-org
```

### 🟡 P2 — `logfire projects new Ellipsis`

Good catch, and worse than it reads: that command *runs*, and creates a project called `Ellipsis`. After a rejection `project_name_default` is set to `...` despite its `str` annotation. The refusal now repeats the real rejection reason and suggests `<name>`:

```console
$ logfire --non-interactive projects new taken-name --default-org
A project with the name 'taken-name' already exists. Please enter a different project name
Cannot prompt because --non-interactive was passed.
Supply it instead with:
  logfire projects new <name> --default-org
```

### 🟡 P2 — the switch outlived `main()`

Valid. `main()` is importable, so an application calling it in-process lost prompting everywhere afterwards — including in `logfire.configure()`, which this flag does not govern at all. Restored on the way out.

**The test I had written for this asserted the buggy behaviour** (`assert is_non_interactive() is True` after `main()` returned) — it pinned the defect as correct, so it is deleted rather than adjusted, and replaced by one that asserts the switch does not outlive the call.

### 🔵 P3 — duplicated test

Correct, byte-for-byte the same experiment as `test_auth_reads_piped_answers_even_though_a_pipe_is_not_a_tty`. Dropped.

---

`make lint` and `make typecheck` clean across the repo. `tests/test_cli.py tests/test_configure.py tests/test_auth.py`: **339 passed**.

Coverage of the changed files, branches included: `interactive.py` **100%**, `cli/auth.py` **100%**, `cli/__init__.py` **100%**, `cli/gateway.py` 99% (the two partial branches are a pre-existing `contextlib.suppress` at line 376, ~75 lines from anything this PR touches).

strawgate:
All five fixed in abcbc8e, threads resolved. Two were the same class of mistake I had already made once in this PR.

### 🟠 High — `--org` + `--default-org` together suggests the wrong organization

Valid. The remedy preferred `--default-org`, so following it would create the project in the user's *default* organization rather than the one they asked for.

Fixed by always naming the settled organization outright rather than choosing between the flags:

```console
$ logfire --non-interactive projects new --org agent-eval --default-org
  logfire projects new app --org agent-eval
```

That reproduces the same result whichever flag chose the organization — including when `--default-org` resolved it — so there is no conflict left to get wrong. I did not make the flags mutually exclusive: that is a pre-existing behaviour and a separate decision, though it would be a reasonable follow-up.

### 🟠 High — the suggestions are not pasteable

Correct, and this is the second time in this PR: `--region us|eu` was a pipeline, and now `<name>` / `<organization>` are input redirection. `logfire projects new <name>` in a shell tries to read a file called `name`.

Now `PROJECT_NAME` and `ORGANIZATION`, at **every** site rather than the two flagged — the `projects use` remedies had the same problem. And since I have now made this mistake twice, there is a test that rejects any suggestion containing `<>|&;$\`()`, so it cannot come back a third time.

Verified in a container that a suggested command runs as printed:

```console
$ logfire projects new app --org agent-eval
Project created successfully. You will be able to view it at: ...
```

### 🟡 P2 — the gateway refusal has no test

Right, and the interesting part is *why* it looked fine: `require_answer(...)` is a single statement that executes whether or not the flag is set — the refusal happens inside the helper. So the line showed as covered at 100% while nothing had ever driven the refusal from that call site. **Coverage could not distinguish them**; only a test with the flag set can.

Added, asserting `Prompt.ask` is never reached.

### 🔵 P3 — "several are installed" when one is

Correct: with a single integration the prompt is still reached, because it has a default. Wording now follows the count — `claude is installed.` vs `Installed: claude, codex.`

---

`make lint` and `make typecheck` clean across the repo. **342 passed**. Coverage of the changed files, branches included: `interactive.py`, `cli/auth.py`, `cli/__init__.py` at 100%; `cli/gateway.py` 99%, both partial branches a pre-existing `contextlib.suppress` ~75 lines away.

strawgate:
Rebased onto `main` now that #2275 has merged. This PR was stacked on that branch, so it went conflicting the moment #2275 landed as a squash — `main` had the content under different SHAs while the branch still carried the originals.

Replayed only the five `--non-interactive` commits onto current `main`, dropping the merged ones. The diff is now exactly this PR's own work:

```
 logfire/_internal/cli/__init__.py |  26 ++++-
 logfire/_internal/cli/auth.py     |  24 +++--
 logfire/_internal/cli/gateway.py  |   6 ++
 logfire/_internal/config.py       |  72 ++++++++++++--
 logfire/_internal/interactive.py  |  67 +++++++++++++
 tests/test_cli.py                 | 200 +++++++++++++++++++++++++++++++++++++-
```

`make lint` and `make typecheck` clean against the merged `main`; **345 passed**.

---

@coderabbitai review — the `CHANGES_REQUESTED` above is from 14:52 and its three findings were fixed later the same day:

- **skip the browser keypress in non-interactive mode** (`auth.py` ~107) — fixed in `86ab1dc`. The flag is now checked *before* reading, which is the point: with `--region` supplied that prompt is the next stop, and an open, silent stdin hung there. The test asserts `input` is never called, since an EOF-based test passes either way.
- **`sys.stdin is None` raises `RuntimeError`** — fixed in #2275 (`1c6090d`), where `_read_line` lives.
- the remaining inline notes were addressed in `abcbc8e`: pasteable placeholders, the `--org` / `--default-org` conflict, gateway wording, and the gateway refusal's missing test.

All review threads on this PR are resolved and CI was green before the rebase.

PR #2293, opened 2026-08-20
Describe what the Pydantic plugin's include and exclude actually do

### LINKED ISSUE #2053: Pydantic plugin: docs say third-party modules are not instrumented by default, but empty include instruments everything

aviseth:
The docs for `instrument_pydantic` say that third-party modules are not instrumented by default. They are. With `include` empty, which is the default, `_include_model` returns `True` for every model, so every Pydantic class in the process is instrumented, including every model belonging to a dependency. There is no third-party detection anywhere in the decision — only a fixed list of three ignored modules (`fastapi`, `fastui`, `logfire_backend`).

Fixes #2053, which reports 2,221 model classes instrumented at startup in a Django app, the majority from dependencies.

Checked against `main`:

```python
>>> check(include=set(), module='openai._models')      # third-party, no include
True
>>> check(include={r'apps\..*'}, module='django.apps.config')
True                                                   # anchored at the end only
>>> check(include={r'^apps\..*'}, module='django.apps.config')
False
```

This is documentation only. The behaviour may well be the intended one — `include`/`exclude` work fine as an opt-in filter — but the description is the opposite of it, and someone reading it has no reason to pass `include` at all. Changing the default would be a breaking change for anyone relying on models being instrumented without configuration, so this describes what happens instead.

Three things are now stated:

- `include` empty means everything, dependencies included.
- Patterns are matched with `re.search(f'{pattern}$', ...)`, so they are anchored at the end but not at the start. `apps\..*` also matches `django.apps.config`, which is easy to hit and hard to spot. The examples use `^`.
- `exclude` is checked before `include`, so a model matching both is not instrumented.

The docs section also mentions why this is worth configuring rather than leaving alone: with `record='failure'`, libraries that use a failed validation as control flow and fall back to another schema get their expected failures recorded as warnings. The issue gives two concrete cases, in the OpenAI SDK and in Opik.

Every claim added here was verified by calling `_include_model` directly against the current implementation. `make lint`, `make typecheck` and the full suite pass.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/integrations/pydantic.md
+++ b/docs/integrations/pydantic.md
@@ -73,24 +73,37 @@ imported *after* that call.
 
 ## Advanced
 
-### Third-party modules
+### Choosing which models to instrument
 
-By default the plugin does not instrument third-party modules, to avoid noise. Opt specific ones in
-with the [`include`][logfire.PydanticPlugin.include] setting:
+The plugin instruments **every** Pydantic model in the process, including models defined by the
+libraries you depend on. In an application with a few large dependencies that can be thousands of
+model classes, and with `record='failure'` it also records validations those libraries perform
+deliberately — several use a failed validation as control flow and fall back to another schema, so
+the failure is expected rather than a problem you can act on.
 
-```py skip-run="true" skip-reason="global-instrumentation"
-import logfire
-
-logfire.instrument_pydantic(include={'openai'})
-```
-
-Opt your own modules out with the [`exclude`][logfire.PydanticPlugin.exclude] setting:
+Narrow it to your own code with the [`include`][logfire.PydanticPlugin.include] setting:
 
 ```py skip-run="true" skip-reason="global-instrumentation"
 import logfire
 
-logfire.instrument_pydantic(exclude={'app.api.v1'})
+logfire.instrument_pydantic(include={r'^app\..*'})
 ```
+
+Patterns are regular expressions matched against `<module>::<class name>`, and they are anchored at
+the end but not at the start. `app\..*` on its own therefore also matches `django.apps.config`,
+which is why the example begins with `^`.
+
+Use [`exclude`][logfire.PydanticPlugin.exclude] to leave particular models out. It is checked before
+`include`, so a model matching both is not instrumented:
+
+```py skip-run="true" skip-reason="global-instrumentation"
+import logfire
+
+logfire.instrument_pydantic(exclude={r'^app\.api\.v1\..*'})
+```
+
+Models in `fastapi`, `fastui` and `logfire_backend` are never instrumented, regardless of these
+settings.
 
 ### Per-model configuration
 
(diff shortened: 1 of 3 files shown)


aviseth:
Good catch, taken. The docs page listed `fastapi`, `fastui` and `logfire_backend` but the two docstrings said "every model" without the qualification, so they disagreed with each other. Both now name the exception.

aviseth:
Right, and the inconsistency was mine: the warning was about `apps\..*` (from the issue) but the example above it used `app\..*`, which needs a literal `app.` and so doesn't match `django.apps.config` at all. Verified both ways:

```
'app\..*'   vs django.apps.config -> False
'apps\..*'  vs django.apps.config -> True
```

The whole example now uses `apps`, so the warning describes the pattern that's actually on the page.

alexmojaki:
This is being handled in https://github.com/pydantic/logfire/pull/2054

PR #2341, opened 2026-09-01
chore: Update dependencies

alexmojaki:


tests/otel_integrations/test_httpx.py
```diff
@@ -1117,7 +1117,7 @@ async def test_httpx_client_capture_all_environment_variable(exporter: TestExpor
                     'logfire.msg': 'GET example.org/foo',
                     'http.request.header.host': ('example.org:8080',),
                     'http.request.header.accept': ('*/*',),
-                    'http.request.header.accept-encoding': ('gzip, deflate, zstd',),
+                    'http.request.header.accept-encoding': ('gzip, deflate',),
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in a2b584bf

logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -275,7 +275,7 @@ def record_chunk(self, chunk: anthropic.types.MessageStreamEvent) -> None:
 
         if type(chunk).__module__.startswith('anthropic.types.beta'):
             self._message = beta_accumulate_event(
-                event=cast(Any, chunk), current_snapshot=self._message, request_headers=httpx.Headers()
+                event=cast(Any, chunk), current_snapshot=self._message, request_headers=cast(Any, httpx.Headers())
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 5c458ae8

logfire/_internal/integrations/llm_providers/anthropic.py
```diff
@@ -275,7 +274,7 @@ def record_chunk(self, chunk: anthropic.types.MessageStreamEvent) -> None:
 
         if type(chunk).__module__.startswith('anthropic.types.beta'):
             self._message = beta_accumulate_event(
-                event=cast(Any, chunk), current_snapshot=self._message, request_headers=httpx.Headers()
```
alexmojaki, line 278: why not just switch to httpx2?

tests/otel_integrations/test_anthropic.py
```diff
@@ -1034,8 +1031,15 @@ def test_messages_without_stop_reason(instrumented_client: anthropic.Anthropic,
 
 
 def test_unknown_method(instrumented_client: anthropic.Anthropic, exporter: TestExporter) -> None:
-    response = instrumented_client.completions.create(max_tokens_to_sample=1000, model='claude-2.1', prompt='prompt')
-    assert response.completion == 'completion'
+    response = cast(
```
alexmojaki, line 1034: why the change

logfire/_internal/integrations/llm_providers/anthropic.py
alexmojaki: (AI) We still support Anthropic 0.27+, whose older releases install httpx rather than httpx2. Switching unconditionally would invert the clean-install failure for those users. The accumulator only calls .get() on this empty value, so an empty mapping is version-neutral and avoids requiring either HTTP transport package directly.

tests/otel_integrations/test_anthropic.py
alexmojaki: (AI) Anthropic 1.2 removed both the legacy .completions resource and the Completion response model. The low-level post keeps this test exercising the same unknown-endpoint fallback (/v1/complete). cast_to=object is needed because cast_to=dict fails in Anthropic 1.2, while dict[str, Any] triggers an issubclass(GenericAlias, ...) TypeError on Python 3.10; the returned object is then narrowed to the expected test dictionary.

tests/otel_integrations/test_anthropic.py
```diff
@@ -49,9 +48,7 @@ def request_handler(request: httpx.Request) -> httpx.Response:
     if request.url == 'https://api.anthropic.com/v1/complete':
         return httpx.Response(
             200,
-            json=Completion(id='test_id', completion='completion', model='claude-2.1', type='completion').model_dump(
-                mode='json'
-            ),
+            json={'id': 'test_id', 'completion': 'completion', 'model': 'claude-2.1', 'type': 'completion'},
```
alexmojaki, line 51: why the change

tests/otel_integrations/test_anthropic.py
```diff
@@ -1089,16 +1093,20 @@ def test_request_parameters(instrumented_client: anthropic.Anthropic, exporter:
...
-        max_tokens=1000,
-        model='claude-3-haiku-20240307',
-        system='You are a helpful assistant.',
-        messages=[{'role': 'user', 'content': 'What is four plus five?'}],
-        temperature=0.7,
-        top_p=0.9,
-        top_k=40,
-        stop_sequences=['END', 'STOP'],
-        tools=cast(Any, tools),
+    response = instrumented_client.post(
```
alexmojaki, line 1096: why the change

tests/otel_integrations/test_anthropic.py
```diff
@@ -1121,8 +1129,10 @@ def test_request_parameters(instrumented_client: anthropic.Anthropic, exporter:
                         'messages': [{'role': 'user', 'content': 'What is four plus five?'}],
                         'model': 'claude-3-haiku-20240307',
                         'stop_sequences': ['END', 'STOP'],
-                        'system': 'You are a helpful assistant.',
                         'temperature': 0.7,
+                        'top_p': 0.9,
```
alexmojaki: revert non-changes

tests/otel_integrations/test_anthropic.py
alexmojaki: (AI) Anthropic 1.2 removed the Completion model, so the mock can no longer construct that SDK type. This literal is the same serialized /v1/complete response shape the test used before, without depending on a type that no longer exists.

tests/otel_integrations/test_anthropic.py
alexmojaki: (AI) Anthropic 1.2 removed temperature, top_p, and top_k from the typed messages.create signature, but this test specifically verifies that our endpoint instrumentation extracts those fields from request bodies. Calling the same /v1/messages endpoint through the low-level API preserves that coverage and still parses the response as Message.

tests/otel_integrations/test_anthropic.py
alexmojaki: (AI) Addressed in bcccb428

pyproject.toml
```diff
@@ -207,7 +207,10 @@ dev = [
     "google-genai >= 0",
     "openinference-instrumentation-litellm >= 0",
     "litellm != 1.92.0; python_version >= '3.14'",
-    "litellm >= 1.92.0; python_version < '3.14'",
+    "litellm >= 1.92.0; python_version >= '3.11' and python_version < '3.14'",
+    # LiteLLM 1.97 no longer matches our Python 3.10 VCR recordings, while 1.98 imports
```
alexmojaki: are you saying litellm no longer supports 3.10?

alexmojaki: (AI) No. LiteLLM 1.98 still advertises Python 3.10 support, but it currently has a 3.10 import regression because one module imports NotRequired from typing instead of typing_extensions. LiteLLM 1.97 imports on 3.10, but its requests no longer replay against these existing VCR cassettes. The 1.96.2 cap is a compatibility workaround for this matrix, not a claim that upstream dropped 3.10.

tests/otel_integrations/test_anthropic.py
alexmojaki: ensure there's still a test calling instrumented_client.messages.create without those fields

alexmojaki: (AI) Yes—test_sync_messages at line 157 still calls instrumented_client.messages.create(...) with the normal typed arguments and none of the removed fields. The same path is also exercised by the stream, empty-chunk, tool-message, and no-stop-reason tests at lines 370, 483, 623, 828, and 937.

pyproject.toml
alexmojaki: how is it that the cassettes work on only some python versions?

alexmojaki: (AI) Addressed in 1559c242. The cassettes are not Python-version-specific; I misdiagnosed a non-fatal VCR warning. The actual LiteLLM 1.97 failures on Python 3.10 were `Message` Pydantic model forward references left unresolved. LiteLLM 1.98 avoids that failure on newer Python versions, but cannot import on 3.10 because it imports `NotRequired` from `typing`. I corrected the dependency comment to state the real reason for the cap.

PR #2345, opened 2026-09-02
Skip the Logfire token check inside AWS Lambda

reidemeister94:
## Problem

`logfire.configure()` validates the token in a background thread (`check_logfire_token`, added in #274 so that `configure()` does not block).

In AWS Lambda, `configure()` is usually called at import time, i.e. during the **init phase**. Lambda may pre-initialize an execution environment and then **freeze** it until the first invocation arrives ("proactive initialization" — see [Understanding the Lambda execution environment lifecycle](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html) and [AJ Stuyvenberg's write-up](https://aaronstuyvenberg.com/posts/understanding-proactive-initialization)). That can be minutes later. A thread frozen in the middle of the TLS handshake finds the connection closed by the server once thawed and warns:

```
UserWarning: Logfire API is unreachable, you may have trouble sending data. Error: HTTPSConnectionPool(host='logfire-eu.pydantic.dev', port=443): Max retries exceeded with url: /v1/info (Caused by SSLError(SSLEOFError(8, '[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1032)')))
```

Nothing is actually wrong: the exporters work and no data is lost. But the warning lands in CloudWatch Logs on every affected cold start (≈23 % of cold starts on a production function with ~14k invocations/day) and any log-based alarm that matches "Error" fires on it.

Evidence from that function (Logs Insights, 88 cold starts over 3 days, logfire 4.32.1):

| cold starts | gap between end of init and first `START` | count |
| --- | --- | --- |
| with the warning | 23 s min, 144 s median, 316 s max — every one > 1 s | 14 / 14 |
| without the warning | 0.00 s median; 2 of 74 above 1 s | 74 |

The warning is printed 0.07–0.21 s after the first `START` in every failing case — exactly the thaw.

## Fix

When `AWS_LAMBDA_FUNCTION_NAME` is set (the runtime sets it in every Lambda execution environment), skip the token check. As suggested in review, running it synchronously would slow down every cold start, and the check is not essential: it only prints the project link and warns early about a rejected token, which the exporters report on the first export anyway. `platform_is_aws_lambda()` lives next to `platform_is_emscripten()` in `_internal/utils.py`.

Users on older versions can work around it by joining the thread after `configure()`:

```python
for t in threading.enumerate():
    if t.name == 'check_logfire_token':
        t.join(3.0)
```

## Checks

- New test `test_token_check_is_skipped_on_aws_lambda`: the mocked `/v1/info` response is held back on an event, so a background thread started by mistake would still be alive and visible; asserts no `check_logfire_token` thread and no request. Fails on the previous threaded code, passes with this change.
- `pytest tests/test_configure.py -k "aws_lambda or send_to_logfire_if_token_present"`: 9 passed.
- `ruff check`, `ruff format --check`, `pyright` on the touched files: clean.

### DIFF
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -126,6 +126,7 @@ from .utils import (
     SeededRandomIdGenerator,
     ensure_data_dir_exists,
     handle_internal_errors,
+    platform_is_aws_lambda,
     platform_is_emscripten,
     suppress_instrumentation,
 )
@@ -1370,6 +1371,14 @@ class LogfireConfig(_LogfireConfigData):
                                         validated_credentials.print_token_summary()
 
                     if emscripten:  # pragma: no cover
+                        check_tokens()
+                    elif platform_is_aws_lambda():
+                        # Run the check synchronously: `configure()` is typically called at import
+                        # time, i.e. in the Lambda init phase, and Lambda may freeze the environment
+                        # right after init (proactive initialization) until the first invocation
+                        # arrives, sometimes minutes later. A thread frozen mid-TLS-handshake reads an
+                        # EOF once thawed and warns "Logfire API is unreachable" although nothing is
+                        # wrong. The synchronous check costs one HTTP round trip during init.
                         check_tokens()
                     else:
                         thread = Thread(target=check_tokens, name='check_logfire_token')
--- a/logfire/_internal/config.py
+++ b/logfire/_internal/config.py
@@ -1373,13 +1373,14 @@ class LogfireConfig(_LogfireConfigData):
                     if emscripten:  # pragma: no cover
                         check_tokens()
                     elif platform_is_aws_lambda():
-                        # Run the check synchronously: `configure()` is typically called at import
-                        # time, i.e. in the Lambda init phase, and Lambda may freeze the environment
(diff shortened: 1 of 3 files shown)


reidemeister94:
Addressed the review note on the test: the mocked `/v1/info` response is now held back on an event and `configure()` runs in a helper thread, so the test asserts that `configure()` is still waiting while the response is blocked. Checked that this version fails against the previous threaded implementation and passes with this change.

reidemeister94:
@coderabbitai review

alexmojaki:
I imagine users would generally prefer not to slow down the cold start for this. It might be better to just disable the check in this case, it's not essential.

reidemeister94:
Agreed, thanks. Changed the PR to skip the check inside Lambda instead of running it synchronously (commit 69c859a); title, description and test updated accordingly. The exporters still report a rejected token on the first export, so nothing is lost besides the project link in the init logs.

alexmojaki (approved):
Thanks!

PR #2348, opened 2026-09-02
Add Snowflake integration (instrument_snowflake)

duncankmckinnon:
## Summary
- Adds `logfire.instrument_snowflake()`, instrumenting `snowflake-connector-python`'s `connect()`/`execute()`/`executemany()` with a span per query (SQL text, Snowflake's own query ID `sfqid`, row count, connection identity), following the `surrealdb.py` hand-rolled-patch pattern since no upstream `opentelemetry-instrumentation-snowflake` package exists.
- Connection identity (account/warehouse/database/schema/role) is read only from the connection object's own public properties, never from `connect()` kwargs, so credentials can never reach a span attribute.
- Supports instrumenting the whole module (default) or a single connection instance, with idempotency guards covering both orderings when combined.
- Compute/warehouse-usage telemetry (credits, bytes scanned, etc.) is explicitly out of scope for this integration; the docs page points at the OpenTelemetry Collector's `snowflakereceiver` for that instead.

## Test plan
- [x] `uv run pytest tests/otel_integrations/test_snowflake.py -v` — 10/10 passing
- [x] `make test` (full suite) — 2475 passed, 9 skipped, 2 xfailed, no failures
- [x] `make lint` / `make typecheck` clean
- [x] `docs/integrations/databases/snowflake.md` added with a nav entry; code examples tagged `skip-run` since they require a live Snowflake connection

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


duncankmckinnon:
Addressed the review in b96d768c.

Changes made:
- Query parameters are no longer captured by default. `capture_parameters=True` provides an explicit opt-in for both `execute()` and `executemany()`. Instrumentation remains idempotent, so the first call for a given target determines the capture setting; this is documented in the public API and integration guide.
- Missing Snowflake itself still produces the installation guidance, while unrelated transitive `ModuleNotFoundError`s now propagate unchanged.
- Added contract coverage for passing the `snowflake.connector` module and for `executemany()` exception propagation.
- Clarified the troubleshooting order for module-wide versus single-connection instrumentation.

Two suggestions were intentionally not applied:
- Reversible/context-manager instrumentation is a broader lifecycle/API change. Making only the cursor-class patch reversible would leave `connect()` and per-connection patches in place, and safe restoration also needs defined behavior for nested instrumentation and third-party patches. The current permanent, idempotent behavior remains in scope for this PR; reversible instrumentation can be considered separately if needed.
- The `[logfire.Logfire.instrument_snowflake]` API reference is intentionally unchanged. This is the unified-docs API-reference syntax used throughout the repository and does not require a local Markdown reference definition; replacing it with a hard-coded generated anchor would be inconsistent.

Verification:
- `uv run pytest tests/otel_integrations/test_snowflake.py tests/test_logfire_api.py -q`: 20 passed, 1 skipped
- `make typecheck`: 0 errors
- `make lint`: passed

duncankmckinnon:
@coderabbitai review

Kludex:
Thanks for this! A few things before merging:

1. `wrapped_connect(**kwargs)` breaks positional args. The real signature is `connect(connection_name=None, connections_file_path=None, **kwargs)`, so `connect('my_named_connection')` from `connections.toml` raises `TypeError: SnowflakeConnection.__init__() takes 0 positional arguments but 1 was given` (checked against 4.7.3). Use `*args, **kwargs` and add a test that goes through the real `connect` with a positional arg - the autouse fixture currently hides this.
2. Please add `db.system='snowflake'` and `db.statement=<command>` to the execute spans. The live view's DB details panel keys on those (and already knows the snowflake dialect for SQL formatting), so without them these render as generic spans.
3. The `(Task 5)` reference in the `_instrument_module` comment leaked from a plan - please drop it and trim the surrounding comments.
4. Minor: unset connection attributes end up as the string `"null"` (see the error test snapshot). Skipping `None` values would be cleaner.

Also worth a decision on the first-call-wins `capture_parameters` behaviour cubic flagged: a connection instrumented with `capture_parameters=True` silently stops capturing once the module is instrumented with the default.

logfire/_internal/integrations/snowflake.py
```diff
@@ -0,0 +1,150 @@
...
+    def wrapped_cursor_factory(*args: Any, **kwargs: Any) -> SnowflakeCursor:
+        cursor = original_cursor_factory(*args, **kwargs)
+        # Always wrap this connection's cursors with this call's capture_parameters,
+        # using the unpatched methods so a later module-level patch cannot override
+        # them or double-wrap.
+        execute = _unpatched(SnowflakeCursor.execute)
+        cursor.execute = types.MethodType(_wrap_execute(logfire_instance, execute, capture_parameters), cursor)
+        executemany = _unpatched(SnowflakeCursor.executemany)
+        cursor.executemany = types.MethodType(
+            _wrap_executemany(logfire_instance, executemany, capture_parameters), cursor
```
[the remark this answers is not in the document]
Kludex: Addressed in c70b49a4. The connection wrapper now unwraps `type(cursor).execute` / `executemany`, so `conn.cursor(CustomCursor)` keeps subclass overrides. Covered by `test_instrument_single_connection_custom_cursor_class`.

logfire/_internal/integrations/snowflake.py
```diff
@@ -0,0 +1,159 @@
...
+from logfire._internal.utils import handle_internal_errors
+
+try:
+    import snowflake.connector as sf_connector
+    from snowflake.connector.connection import SnowflakeConnection
+    from snowflake.connector.cursor import SnowflakeCursor
+except ModuleNotFoundError as e:
+    if e.name not in {'snowflake', 'snowflake.connector'}:
+        raise
+    raise ImportError('Run `pip install snowflake-connector-python` to use `logfire.instrument_snowflake()`.') from e
```
Kludex, line 19: Shouldn't we have a pyproject extras for this?

logfire/_internal/integrations/snowflake.py
```diff
@@ -0,0 +1,159 @@
...
+
+CONNECTION_ATTRS = ('account', 'warehouse', 'database', 'schema', 'role')
```
Kludex: This doesn't need to be a constant, let's just inline it.

docs/integrations/databases/snowflake.md
```diff
@@ -0,0 +1,138 @@
+---
+title: "See Snowflake queries in Logfire"
```
Kludex: Does this matches how other titles are in this docs section?

logfire/_internal/integrations/snowflake.py
```diff
@@ -0,0 +1,159 @@
...
+    original_cursor_factory = conn.cursor
+    if getattr(original_cursor_factory, '_logfire_patched', False):
+        return
+
+    def wrapped_cursor_factory(*args: Any, **kwargs: Any) -> SnowflakeCursor:
+        cursor: SnowflakeCursor = original_cursor_factory(*args, **kwargs)
+        # Wrap the unpatched methods of the cursor's own class so a module-level patch
+        # cannot double-wrap and cursor subclass overrides are preserved.
+        cursor_class = type(cursor)
+        execute = _unpatched(cursor_class.execute)
```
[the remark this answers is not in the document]
Kludex: Addressed in 908fb457 by removing the second wrapper layer entirely. Connection instrumentation now only records the connection in a `WeakKeyDictionary`; the single class-level wrapper looks up `cursor.connection` there and falls back to the module setting. With one layer, `super().execute()` in a cursor subclass cannot produce a second span. Covered by `test_custom_cursor_super_call_with_module_and_connection_instrumented`.

logfire/_internal/main.py
```diff
@@ -1011,6 +1012,28 @@ def instrument_surrealdb(
...
+    ) -> None:
+        """Instrument the [Snowflake Connector for Python](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector) so that a span is created for each query.
+
+        Args:
+            conn_or_module: Pass a single connection instance to instrument only that connection.
+                By default (`None`), all connections are instrumented, including ones created later.
+            capture_parameters: Set to `True` to capture query parameters as span attributes.
+                Be cautious when enabling this, as it may lead to sensitive data being captured in traces.
+                Instrumenting the same target again has no effect; the first call determines this setting,
+                and a later call with a different value emits a warning.
```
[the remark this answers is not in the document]
Kludex: Addressed in 908fb457. Re-instrumenting a connection with a different `capture_parameters` now warns the same way the module path does. Covered by `test_instrument_connection_capture_parameters_change_warns`.

logfire/_internal/integrations/snowflake.py
Kludex: Inlined in 524a37c1.

docs/integrations/databases/snowflake.md
Kludex: Aligned in 524a37c1 to the "Instrument X: see every query your app runs" pattern used by 9 of the other 10 database pages, with the matching description. SurrealDB is the only other page using the "See X in Logfire" form.

logfire/_internal/integrations/snowflake.py
Kludex: I would say no, following the existing precedent. Every entry in `[project.optional-dependencies]` and in `INSTRUMENTATION_TO_EXTRA` in `cli/run.py` installs an `opentelemetry-instrumentation-*` package the user would not otherwise know about. There is no such package for Snowflake, so the extra would only pin `snowflake-connector-python`, which anyone using this integration already has. SurrealDB, the other hand-rolled integration, has no extra either, and the skill reference lists it as "no extra". I added a matching Snowflake row there. Happy to add an extra if you want the `logfire[snowflake]` spelling anyway.

logfire/_internal/integrations/snowflake.py
```diff
@@ -0,0 +1,150 @@
...
+            _warn_capture_parameters_ignored(_module_capture_parameters)
+    elif isinstance(conn_or_module, SnowflakeConnection):
+        existing = _connection_capture_parameters.get(conn_or_module)
+        if existing is None:
+            _connection_capture_parameters[conn_or_module] = capture_parameters
+        elif existing != capture_parameters:
+            _warn_capture_parameters_ignored(existing)
+    else:
+        raise ValueError(f"Don't know how to instrument {conn_or_module!r}")
+    _patch_cursor_class(logfire_instance)
```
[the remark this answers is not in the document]
Kludex: Addressed in 3d7871db. The module and per-connection registrations now store `(logfire_instance, capture_parameters)`, and the class wrapper selects both when the cursor method runs. Covered by `test_instrument_connections_with_different_logfire_instances`.

logfire/_internal/integrations/snowflake.py
```diff
@@ -0,0 +1,150 @@
...
+                    span.set_attribute(key, value)
+            return conn
+
+    sf_connector.connect = wrapped_connect
+
+
+def _patch_cursor_class(logfire_instance: Logfire) -> None:
+    original_execute = SnowflakeCursor.__dict__.get('execute', SnowflakeCursor.execute)
+    if not getattr(original_execute, '_logfire_patched', False):
+        SnowflakeCursor.execute = _wrap_execute(logfire_instance, original_execute)
```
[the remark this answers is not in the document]
Kludex: Addressed in 3d7871db. Each connection registration keeps its own `Logfire` instance and the wrapper selects it at execution time, with the module-wide pair as the fallback. Covered by `test_instrument_connections_with_different_logfire_instances`.

PR #2354, opened 2026-09-04
Document API keys that can send telemetry

strawgate:
## Summary

Correct the API-key guide, which currently says API keys cannot send telemetry. A project-scoped key with **Send telemetry** (`project:write_otlp`) can send logs, traces, and metrics. The Assistant was repeating the outdated distinction when answering setup questions.

- Explain the required permission and project scope, including using the key as `LOGFIRE_TOKEN` with the Python SDK.
- Add send/query telemetry permissions to the scope table.
- Keep write tokens documented as supported; no migration is required.

## Validation

- Checked scope and ingestion behavior against current Platform code and Python SDK token handling.
- Documentation pre-commit checks and `git diff --check` passed.
- Local unified-docs preview starts, but its rendered Markdown still comes from the sibling checkout despite the explicit worktree source; that preview is not proof of this patch. Hosted docs preview requested.

Documentation only; no runtime or API changes. This touches the same guide as #2353 but changes the telemetry introduction and scope rows, not SCIM setup.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2354?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/how-to-guides/create-write-tokens.md
+++ b/docs/how-to-guides/create-write-tokens.md
@@ -2,7 +2,8 @@
 title: "Logfire Write Tokens: Setup Guide"
 description: "Step-by-step guide for creating a Logfire write token in the web UI by injecting LOGFIRE_TOKEN in apps & for optionally suppressing it in local development."
 ---
-To send data to **Logfire**, you need to create a <OpenInLogfire path="settings/write-tokens" variant="inline" label="write token" />.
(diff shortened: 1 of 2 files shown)


strawgate:
Addressed the conditional-send wording in 611e9dbb: it now names project API keys with Send telemetry alongside write tokens. Checked against SDK token handling; normal pre-commit hooks passed.

PR #2357, opened 2026-09-04
Move query client out of experimental

alexmojaki:
## Summary

- move the query client implementation to `logfire.query_client`
- remove the legacy `logfire.experimental.query_client` module
- update the DB-API integration and tests to use the canonical module

This is a behavior-preserving module relocation following the v5 query API cleanup.

## Tests

- `uv run pytest tests/test_query_client.py tests/test_query_client_errors.py tests/test_db_api.py`
- `uv run pytest tests/test_logfire_api.py`
- `uv run pyright logfire/query_client.py logfire/db_api.py tests/test_db_api.py`
- `uv run ruff check logfire/query_client.py logfire/db_api.py tests/test_db_api.py`
- `uv run ruff format --check logfire/query_client.py logfire/db_api.py tests/test_db_api.py`


logfire/query_client.py
```diff
@@ -22,3 +36,563 @@
...
+_ACCEPT = Literal['application/json', 'application/vnd.apache.arrow.stream', 'text/csv']
+_USER_AGENT = f'logfire-sdk-python/{VERSION} (Python {platform.python_version()}, os {platform.platform()}, arch {platform.machine()})'
+
+
+class _BaseLogfireQueryClient(Generic[T]):
+    def __init__(self, base_url: str, read_token: str, timeout: Timeout, client: type[T], **client_kwargs: Any):
+        self.base_url = base_url
+        self.read_token = read_token
+        self.timeout = timeout
+        headers = client_kwargs.pop('headers', {})
```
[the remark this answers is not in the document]
alexmojaki: (AI) Valid, but this header handling was moved unchanged from `logfire.experimental.query_client` as part of this behavior-preserving relocation. Normalizing the accepted header inputs should be a separate behavior-fix PR.

logfire/query_client.py
```diff
@@ -1,15 +1,29 @@
...
+
+if sys.version_info >= (3, 11):
+    from datetime import UTC
+else:
+    UTC = timezone.utc
+
+try:
+    from httpx import AsyncClient, Client, Response, Timeout
+    from httpx._client import BaseClient
+except ImportError as e:  # pragma: no cover
```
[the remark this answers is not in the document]
alexmojaki: (AI) Valid, but this handler was moved unchanged from `logfire.experimental.query_client` as part of this behavior-preserving relocation. Changing its exception semantics should be handled separately rather than mixed into this mechanical PR.

logfire/query_client.py
```diff
@@ -1,15 +1,29 @@
-# For git history and backwards compatibility, everything is kept in the experimental module.
```
alexmojaki, line 1: is there a way to do this move that keeps git history on the new file?

logfire/db_api.py
alexmojaki, line 1: remove the stubs as per agents.md

alexmojaki: (AI) Addressed in 29589ace

logfire/query_client.py
alexmojaki: (AI) Not without rewriting the existing branch history. Because `logfire/query_client.py` already exists on the base branch as a wrapper, Git compares the final path as a modification rather than a rename from `logfire/experimental/query_client.py`. A fresh sequence that removes the wrapper first and then renames the implementation would let `git log --follow` cross the move, but this repository explicitly prohibits rebasing or otherwise rewriting history. Keeping the implementation in the experimental module would preserve its history, but would defeat this PR's relocation.

PR #2374, opened 2026-09-05
Don't warn when inspect_arguments has no caller source

### LINKED ISSUE #2223: InspectArgumentsFailedWarning repeats for each pytest span when source path differs from executed pyc
### Description

`logfire.span()` emits `InspectArgumentsFailedWarning` once per test when pytest is executed from a read-only bind-mounted checkout but the imported `conftest.py` code object points to a source path that is unavailable in the container. The warning says to report the issue and repeats for every span, obscuring otherwise clean bounded-test output.

### Environment

- Python 3.11.15
- logfire 4.38.0
- pytest executed in a Linux container with a read-only repository bind mount
- Span is created by a `pytest_runtest_protocol` hook wrapper

### Span call

```python
with logfire.span(
    "wc-test {test_name}",
    test_name=item.name,
    test_class=item.cls.__name__ if item.cls else "",
):
    yield
```

### Observed warning

```text
Failed to introspect calling code. Please report this issue to Logfire.
Falling back to normal message formatting which may result in loss of information
if using an f-string. Set inspect_arguments=False in logfire.configure() to suppress
this warning. The problem was: No source code available. This happens when running
in an interactive shell, using exec(), or running .pyc files without the source .py files.
```

The message repeats once per test. The message template is a literal, not an f-string, and all interpolation values are provided explicitly.

### Expected behavior / suggestion

When the message template is a literal with explicit keyword values, inability to recover the caller source should not need a warning per span. Possible improvements:

1. recognize that no f-string/source inspection is needed for this call shape, or
2. deduplicate the warning per call site/process, while retaining the documented `inspect_arguments=False` escape hatch.

Implementation encountered during a resource-bounded deployment verification workflow. Reported by Codex in collaboration with @dirkenglund.

gyanu2507:
## Summary
`inspect_arguments` warned on every span when `executing` could not read source (exec, REPL, or .pyc without the .py). A literal template with explicit kwargs already falls back correctly, so the warning just repeated in pytest output, including bind-mounted checkouts where the code object's path is unreadable.

Skip the warning when there is no source. Real f-string inspection failures still warn.

## Test plan
- [x] `tests/test_source_code_extraction.py` (6 passed)

Fixes #2223

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2374?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/logfire/_internal/ast_utils.py
+++ b/logfire/_internal/ast_utils.py
@@ -169,14 +169,10 @@ class CallNodeFinder(ABC):
         # This shouldn't happen in most cases, but it's best not to rely on it always working.
 
         if not self.source.text:
-            # This is a very likely cause.
-            # There's nothing we could possibly do to make magic work here,
-            # and it's a clear case where the user should turn the magic off.
-            self.warn_inspect_arguments(
-                'No source code available. '
-                'This happens when running in an interactive shell, '
-                'using exec(), or running .pyc files without the source .py files.',
-            )
+            # Nothing we can inspect: exec(), a REPL, or a .pyc without sources.
+            # The formatter already falls back to the provided template and kwargs,
+            # which is correct for a literal message. Warning on every span made
+            # that expected fallback look like a defect (#2223).
             return None
 
         msg = '`executing` failed to find a node.'
--- a/tests/test_source_code_extraction.py
+++ b/tests/test_source_code_extraction.py
@@ -9,7 +9,7 @@ from inline_snapshot import snapshot
 
 import logfire
 from logfire._internal import ast_utils
-from logfire._internal.ast_utils import InspectArgumentsFailedWarning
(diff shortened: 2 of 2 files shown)


gyanu2507:
The exec regression now uses a literal template plus kwargs (`logfire.span('from {name}', name='module')`) and asserts there is no `InspectArgumentsFailedWarning`.

alexmojaki:
When there's no source, inspect_arguments should be set to False because the magic can't work. The warning is valid. I'd prefer https://github.com/pydantic/logfire/pull/2225

PR #2375, opened 2026-09-05
fix: require path boundaries for non-user code

yhz5613813:
## Description

`is_non_user_path` currently uses string prefix matching. Registering `/path/library` therefore also classifies `/path/library-extra/app.py` as non-user code.

Use `Path.is_relative_to` for component-aware containment checks.

## Testing

- Added coverage for a file inside a registered prefix and a sibling directory with the same textual prefix.
- `pytest -c /dev/null -p no:cacheprovider --noconftest tests/test_stack_info.py -q`
- Ruff check, formatting check, and `compileall` on both changed files.


alexmojaki:
Thanks!

PR #2451, opened 2026-09-22
fix(metrics): sanitize metric attributes to avoid unrecoverable exporter/aggregation failures

### LINKED ISSUE #782: Metric attributes with wrong value types fail badly
```python
import logfire

logfire.configure()

counter = logfire.metric_counter('my_counter')
counter.add(1, {'k': object()})
```

logs:

```
Failed to encode key k: Invalid type <class 'object'> of value <object object at 0x10864d320>
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/__init__.py", line 113, in _encode_attributes
    pb2_attributes.append(_encode_key_value(key, value))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/__init__.py", line 94, in _encode_key_value
    return PB2KeyValue(key=key, value=_encode_value(value))
                                      ^^^^^^^^^^^^^^^^^^^^
  File "/Users/alex/work/logfire/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/__init__.py", line 90, in _encode_value
    raise Exception(f"Invalid type {type(value)} of value {value}")
Exception: Invalid type <class 'object'> of value <object object at 0x10864d320>
```

This happens in the `PeriodicExportingMetricReader` thread and gives no way for the user to know the original offending code.

We should report this to OTEL if it isn't already and also decide what we want to do about it ourselves. I'm leaning towards warning the user and dropping the attribute.

RichardoMrMu:
Fixes #782.

## Problem

Passing a metric attribute whose value can't be sent to the backend blows up in a place the user can't connect to their own code. From the issue:

```python
counter = logfire.metric_counter('my_counter')
counter.add(1, {'k': object()})
```

Nothing goes wrong at the `add()` call. The failure happens later, inside the `PeriodicExportingMetricReader` export thread, as:

```
Failed to encode key k: Invalid type <class 'object'> of value <object object at ...>
```

There's no stack back to the `add()` that produced it, and the whole export batch is affected. Unhashable values (e.g. a `list`) are worse: they raise a bare `TypeError` from the metrics SDK's aggregation (`frozenset(attributes.items())`).

Span attributes don't have this problem because they go through `prepare_otlp_attribute`. Metric attributes are passed straight through, and neither the metrics SDK nor the exporter cleans them.

## Change

Sanitize metric attributes at the point they're recorded, in the proxy instruments (`add`/`record`/`set`). A value is kept if it's a primitive (`bool`/`str`/`bytes`/`int`/`float`) or a tuple of those. Anything else is dropped with a `UserWarning` naming the attribute, and the rest of the attributes and the measurement are kept.

Tuples only (not lists): a list is a valid OTLP attribute value in general, but it's unhashable and the metrics SDK keys aggregation on `frozenset(attributes.items())`, so a list would fail there regardless.

This matches the direction suggested in the issue (warn and drop the attribute rather than let it break export), and keeps the warning attributable to the user's call via `stacklevel`.

## Tests

Added to `tests/test_metrics.py`:

- counter / histogram / up-down-counter / gauge each drop an un-encodable value and warn, while keeping the valid attributes;
- a tuple containing a bad element is dropped;
- a `list` value is dropped (it would otherwise crash aggregation);
- fully valid attributes (including a tuple) produce no warning and are recorded unchanged.

<a href="https://cubic.dev/pr/pydantic/logfire/pull/2451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


RichardoMrMu:
Thanks @coderabbitai — good catch, fixed. Dropped the `element is None` branch in `_metric_attribute_value_is_valid`, so a tuple containing `None` (e.g. `(1, None)`) now fails validation and is warned-on + dropped like any other invalid metric attribute, instead of being forwarded with version-dependent exporter behavior.

Added `test_metric_attributes_with_none_sequence_element_are_dropped` covering `{'seq': (1, None), 'good_seq': ('a', 'b')}`, asserting the `None`-bearing tuple triggers the "invalid type tuple" warning and only `good_seq` survives on the data point. Also updated the comment to note that `None` is not a valid metric attribute sequence element.

